PR #22, opened 2016-01-26
Fix typo in `linkState()` example

darrenscerri:


developit:
Awesome, thanks! I had fixed it in the [Wiki](https://github.com/developit/preact/wiki/Linked-State) but not the README.

PR #35, opened 2016-02-01
Updated code to call the 'empty' function instead of doing direct comparison

ericperez:
Found three cases where the exact logic for the 'empty' function was being used directly in the code instead of a call to the 'empty' function and updated them.


developit:
Great catch! I don't know how I missed those, thanks. Will merge tonight.

Issue #59, opened 2016-02-19
defaultProps and getDefaultProps


c58:
Hi!
Currently defaultProps and getDefaultProps is not supported. It is a big problem with making some react components to work with preact. Any plans to implement it?

developit:
Yup, coming to Preact itself. Will update here as it does.

c58:
Awesome, thanks!

developit:
Fixed on master and passing your lovely tests. Will release as `4.0` under a beta npm dist-tag for a short time and then we'll be live.

mrbar42:
Sorry to comment on closed issue, but the non-es6 class construction still doesn't ever invoke getDefaultProps (I'm on 4.2.0).
apparently in the non es6 way all of the life cycle methods are under nodeName.prototype.
though I did no see *any* reference to `getDefaultProps` at all.
as seen here `preact/src/vdom/index.js:32`
```javascript
let defaultProps = vnode.nodeName.defaultProps;
```
perhaps something like this would solve it.
```javascript
let nodeName = vnode.nodeName;
let defaultProps = nodeName.defaultProps;
if (!defaultProps && nodeName.prototype.getDefaultProps) {
    defaultProps = nodeName.prototype.getDefaultProps();
}
```

developit:
@mrbar42 no problem!

`getDefaultProps()` is not part of the ES class API, so it is actually handled in [preact-compat](https://github.com/developit/preact-compat) rather than Preact itself.  The [implementation](https://github.com/developit/preact-compat/blob/master/src/index.js#L189-L214) in `preact-compat` basically what you described.

I think we should at least document this somewhere (in the Wiki?), since I can see it causing some confusion.  Also, perhaps we have a use-case here for something lighter than preact-compat that adds support for the `createClass()` API?  A separate module would make it nice and optional, but easy to integrate without relying on all of preact-compat (not that it's particularly large).

Interested to hear your thoughts. Also any info about your use case would be helpful too!

mrbar42:
@developit 

My use case is as follows, I hate fat bundles.. that's for one.
Secondly, I dislike es6 classes (as classes in JS are virtual)
as well as dislike jsx.
but that's really a personal preferences (I seem to be in minority ).

Coming from react I'm looking for the minimum to continue more or less the same.
For example I like the `DOM` shorthand
this for example will add ~800 bytes to the bundle with this minimal hack.
I don't need the whole compat package - just these couple of things.
If the case for preact is to only support es6 than so be it.
[...]

PR #62, opened 2016-02-20
fixed #60 componentWillReceiveProps to be called on rerender

### LINKED ISSUE #60: componentWillReceiveProps called on initial render
In react `componentWillReceiveProps` called only on re-render from parent. In Preact it is called on initial render too. It is a problem for some React components (react-router for example)

c58:
Also i added initial tests for lifecycle methods

### DIFF
--- a/src/vdom/component.js
+++ b/src/vdom/component.js
@@ -39,7 +39,9 @@ export function setComponentProps(component, props, opts, context) {
 		component.context = context;
 	}
 
-	hook(component, 'componentWillReceiveProps', props, component.context);
+	if (component.base) {
+		hook(component, 'componentWillReceiveProps', props, component.context);
+	}
 
 	if (!component.prevProps) component.prevProps = clone(component.props);
 	component.props = props;
--- a/test/browser/lifecycle.js
+++ b/test/browser/lifecycle.js
@@ -1,0 +1,188 @@
+import { h, render, rerender, Component } from '../../src/preact';
+/** @jsx h */
+
+describe('Lifecycle methods', () => {
+	let scratch;
+
+	before( () => {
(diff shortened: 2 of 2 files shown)


developit:
Awesome! So happy to have more tests for this. I'll merge, is it alright if I bundle this with the 4.0 release? I had some stuff depending on the old behavior so I think we can put this in with the other breaking changes (refs, etc). Let me know :)

c58:
4.0 is ok :) I'm working on #59 now
Also i have some suggestions about code style. I dislike comments like this:
```
/** Some comment about something
 *	@private
 */
```

Can we change it to something more general, like this?
```
/**
 *  Some comment about something
 *  @private
 **/

developit:
@c58 I don't see much of a difference, but I don't mind changing them.  I have some fairly drastic changes on a local branch for 4.0 though, maybe it's better for you to work off that branch?  Will push.

c58:
It would be grate to work on latest version to make easier future merging :)

c58:
Hm, linting isn't passing successfully. All because i'm using spaces instead tabs. Is there any reason to use tabs instead of spaces for indention? There is many conversations about the problem, but in my opinion spaces is better :)

developit:
I just prefer tabs. If you're using one of the popular editors like atom, the is an editor config file that should automatically switch to tabs. I'm definitely a tabs person, so I'd like to keep this using them :P

c58:
Ok, tab person, it is not a problem :) I have installed editorconfig plugin, but it does not translate existing code for a style :(

developit:
I think running `eslint --fix` would do it. Or I use an atom plugin to automatically convert between (I think it's called "convert whitespace").

PR #63, opened 2016-02-20
added support for getDefaultProps and default props #59

c58:


developit:
@c58 Looks good. I might play with this a bit to try to get the filesize down.

developit:
@c58 - looking at this now and I'm confused - preact was already invoking `getDefaultProps()` in the component constructor. Not sure how to resolve the two.

c58:
```javascript
this.props = props || hook(this, 'getDefaultProps') || {};
```
if i understand correctly, it does not merge props, but it should. Also, as mentioned in a React docs, it must be invoked before any instances are created. Invoking it in a constructor is not a good idea.

developit:
lol - you work late! Agreed about avoiding in constructor, so I'm just moving your `setDefaultProps()` call into `setComponentProps()` (it just gets called less often).

c58:
Ok :)

developit:
@c58 Actually, on second look I was confusing the semantics of `getNodeProps()` - the place you had it is definitely the best.  So really we're just moving the cache generation into preact-compat. I shouldn't code before my morning coffee, haha.

developit:
Weird, github never marked this as merged but it's definitely merged...

PR #77, opened 2016-03-07
Git ignore src directory

### LINKED ISSUE #76: Exclude src directory from npm package
To do that we need to add this to `.npmignore`
```
src
```

steelbrain:
The main benefit is an extremely lightweight npm package

Fixes #76


developit:
@steelbrain commented on the issue.

PR #91, opened 2016-03-14
Add safe guard to dangerouslySetInnerHTML

steelbrain:
@developit Any ideas how to add tests for this one? You don't seem to have functional tests and this is a deep down change. Also I am not sure how to repro this manually

Fixes https://github.com/developit/preact/issues/90

### DIFF
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -57,7 +57,7 @@ export function setAccessor(node, name, value) {
 		node.style.cssText = value || '';
 	}
 	else if (name==='dangerouslySetInnerHTML') {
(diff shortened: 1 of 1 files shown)


developit:
@steelbrain Awesome, will merge. I think it needs to be slightly tweaked so that unsetting it doesn't overwrite newly created children, something like:

```js
	if (value && value.__html) node.innerHTML = value.__html;
}
```

developit:
Released in [4.3.2](https://github.com/developit/preact/releases/tag/4.3.2)

PR #95, opened 2016-03-16
Allow on* handlers defined as strings to be handled by the EventProxy

AkeemMcLennon:
This change allows on* handlers defined as strings to be correctly registered as listeners and handled by preact's EventProxy. 

For example, `render(<div onClick="console.log(arguments[0]);" />);` should allow the user to click on an element and see a console log of the MouseEvent that triggered it. 

This is particularly useful for parsing html that has been generated server side.

### DIFF
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -79,9 +79,17 @@ function setComplexAccessor(node, name, value) {
 		let type = normalizeEventName(name),
 			l = node._listeners || (node._listeners = {}),
 			fn = !l[type] ? 'add' : !value ? 'remove' : null;
-		if (fn) node[fn+'EventListener'](type, eventProxy);
 		l[type] = value;
-		return;
+		if (typeof(value) === "string"){
+			node.setAttribute(name, value);
+			value = node['on'+type];
+			l[type] = value;
+		}
+		if (fn && typeof(value) !== "string"){
+			node[fn+'EventListener'](type, eventProxy);
+			node.removeAttribute(name);
+			return;
+		}
 	}
 
 	let type = typeof value;
--- a/test/browser/render.js
+++ b/test/browser/render.js
@@ -154,6 +154,21 @@ describe('render()', () => {
 		proto.addEventListener.restore();
 	});
 
+	it('should correctly proxy on* handlers defined as strings', () => {
+		window.stub = sinon.stub();
+		let click = () => {},
+			onclick = "stub(arguments[0]);";
+		render(<div click={ click } onClick={ onclick } />, scratch);
+		let element = scratch.childNodes[0];
+		let ev = document.createEvent("MouseEvent");
+		ev.initMouseEvent("click");
+		expect(window.stub.called,"stub should not be called before click handler").to.equal(false);
+		element.dispatchEvent(ev);
+		expect(window.stub.called,"stub should be called from click handler").to.equal(true);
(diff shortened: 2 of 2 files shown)


developit:
@AkeemMcLennon While I agree that this could be useful for parsing HTML, it's also a script injection vulnerability.  React blocks `String` event handlers and even throws an error in the console:
https://jsfiddle.net/developit/6aL9829j/

Below is an example that enables Preact to handle those String event handlers exactly the way you want, without modifying Preact itself. I would be extremely careful given that any approach that allows this behavior is creating implicit trust in the contents of these attributes.

[JSFiddle Example](https://jsfiddle.net/developit/v4pwunc1/)

```js
import { h, render, options } from 'preact';

let oldVnodeHook = options.vnode;
options.vnode = vnode => {
  let props = vnode.attributes;
  if (typeof vnode.nodeName==='string' && props) {
    for (let key in props) {
      if (props.hasOwnProperty(key) && key.match(/^on.+/gi) && typeof props[key]==='string') {
        props[key] = new Function(props[key]);
      }
    }
  }
  return oldVnodeHook(vnode);
};
```

I'm open to feedback on this because I think it's an interesting discussion, but I am not sure it makes sense to specifically add a feature that the React community has determined is unsafe.

AkeemMcLennon:
Hi @developit, thanks for the quick reply and additional insight!

You make a good point about the potential for script injection through this method while parsing untrusted HTML. The tradeoffs and use cases seem analogous to dangerouslySetInnerHTM. Perhaps an option for a "dangerous" event handler might be appropriate? 

In my use case, I am using preact / preact-markup in conjunction with a 3rd party templating library (nunjucks/ jinja2). This is somewhat similar to the portion of the React tutorial where dangerouslySetInnerHTML is used to render Markdown. Ultimately, the same security concerns exist since you could of course include any arbitrary html / javascript. However, with this method, you lose some of the advantages of parsing the html into a VirtualDom node (such as efficient diffing / handling components). 

On a side note, while playing around with this just now, I discovered tit's actually possible to render script nodes. If we're making the assumption that VirtualNodes are untrusted by default, then it could be filed as a separate issue. For example,

```
render(h('script',null,'console.log("hello world")'), document.body);
```
works in https://jsfiddle.net/syygabsf/1/

developit:
Definitely an interesting case, I'm not sure where to go given the two extremes.  It seems like the "safe" bet would be to disable scripts within script tags - that would leave only `dangerouslySetInnerHTML` as a vector for HTML/script injection, which seems like a reasonably well-documented escape hatch for that sort of thing.

As for your use-case, I think this is the kind of thing that could be added as an option in `preact-markup` itself. It would make sense in terms of where that code should live (semantically speaking), because preact-markup is where the HTML is parsed and converted to VDOM ([attribute conversion here](https://github.com/developit/preact-markup/blob/master/src/to-vdom.js#L20-L30)).  Perhaps an option that does something similar to my vnode hook suggestion would work?  The API could just be:

```js
<Markup type="html" markup={html} allow-scripts allow-events />
```

...where `allow-scripts` would turn on `<script>` support and `allow-events` would turn on the conversion from String `on*` attributes to handler functions?

AkeemMcLennon:
Sounds like a good idea. I'll create a separate PR tomorrow.

AkeemMcLennon:
Haven't had time for the new PR yet, but closing in the meantime due to above discussion.

PR #99, opened 2016-03-21
Update diff.js maybe faster

aimwhy:


developit:
@aimwhy This looks like how diff.js used to work, and I switched to sparse Arrays for performance. Are you getting faster diffs with this approach?

developit:
[Some quick testing](http://esbench.com/bench/56ef4a03db965b9a009659cf) seems to show there is little difference, so maybe my manual tracking of `min` is unnecessary here. Need more testing though.

aimwhy:
I think it reduces the number of cycles.  
No experiment

developit:
Yes, it would definitely reduce the number of iterations from within `removeOrphanedChildren`, though my thought was that would be more than offset by not having to mutate the Object and Array shapes.  Might try to get some others to chime in.

aimwhy:
also reduce the number of iterations for next child 

> 5 minutes ago

developit:
@aimwhy Right, yes, in the case where the `min` lower bound does not get incremented because the matching child is out of order.

aimwhy:
this is my first use github ,  and my english is poor 
With the aid of translation software
Hope useful

developit:
Very useful! I enjoy discussing performance any day :)

kruczy:
IMHO, splice will be much slower then sparse array especially you might need to call it a lot

if you want to reduce the amount of iterations down the chain you could collapse all the nulls using `.filter,  though `removeOrphanedChildren` does `if(child)` which is likely to be faster than making a new array anyway

developit:
I thought so too, though `filter()` seems to end up being slower than just looping (🐼).  I _think_ the current sparse implementation ends up being the fastest, but it's not something that comes out in microbenchmarks very well because it's more about memory complexity than time complexity.

The strange thing is - I ran into a case a while ago where I had assumed looping over an Array and then setting `.length=0` would be the fastest way to perform a queue-flush type operation, since it resulted in only a single mutation.  However, the tests showed that `while(pop())` is actually the fastest by quite a bit:
https://esbench.com/bench/56bd078245df6895002e03c0

Sometimes I find my own conception of what will be fast ends up being predicated on assumptions I've made about the JavaScript engine that are untrue, at least in V8. It's humbling 😊

developit:
Going to close this one out - I've done enough testing locally to know that the sparse Array performs better (also see benchmark above), as noted by @kruczy.  I appreciate all your hard work on this, it's been a good exercise in performance analysis/tuning over the past few months. Thanks!!

kruczy:
Cool, thanks alot :)

PR #101, opened 2016-03-21
no need "call" for context

aimwhy:
var x = {call:{}};
It is not accurate.


developit:
Great, thanks! This shaved 8 bytes off the output. :100:

PR #102, opened 2016-03-22
Easy to understand

aimwhy:


developit:
Ah, yes I do like this better. Originally innerDiffNode worked a but differently, but now it seems nice to flatten things out as you have. Might even help with performance :)

Issue #65, opened 2016-02-24
componentDidMount fires before child component is attached to DOM


unindented:
I have an easy repro here: http://jsfiddle.net/pvbvj23o/

developit:
Interesting. Debugger shows it's attached to the parentNode from `<App>`, but `<App>` hasn't yet been appended to the DOM itself.  I'd love to avoid deferring `componentDidMount()`, since that would cost a traversal during render, but that might be the only way.  Are you sure React only calls this hook once the whole tree is mounted?

unindented:
Looks like React only calls the hook once the element is mounted: https://jsfiddle.net/x6xpdkjL/

The docs seem to confirm: https://facebook.github.io/react/docs/component-specs.html#mounting-componentdidmount

developit:
Awesome, thanks for investigating. I'll try to look into this tonight.

c2h5oh:
Still present in 4.5.

developit:
Yup, on my to-do list. thx.

developit:
Hooray!! 👏 🎉 🍻

developit:
Released in `5.2.0-beta.0`.

PR #128, opened 2016-04-18
added tests for issues #65 to verify component(Did|Will)(Mount|UnMoun…

billneff79:
added tests for issues #65 to verify component(Did|Will)(Mount|UnMount) fires correctly relative to when element is loaded in the dom


developit:
It only took me `86.3155245949074` days to merge this.

PR #130, opened 2016-04-19
added test coverage with isparta-loader

billneff79:
Added test coverage with isparta-loader

Needed to remove `"loose": "all"` config from .babelrc or isparta blew up.  If that is needed, will need to reconcile.


developit:
@billneff79 Definitely still need `loose` in there, but it seems like isparta is using Babel 6, whereas Preact is transpiled via Babel 5 (for performance and filesize reasons).

It might be time to reinvestigate moving to Babel 6.

I'll merge and move the `loose` option into the rollup config for now, that way isparta will happily ignore it.

PR #142, opened 2016-05-10
Add type check for eventProxy function

AkeemMcLennon:
Very that event proxy callback function is actually a function

### DIFF
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -106,7 +106,7 @@ function setComplexAccessor(node, name, value) {
  */
 function eventProxy(e) {
 	let fn = this._listeners[normalizeEventName(e.type)];
-	if (fn) return fn.call(this, optionsHook('event', e) || e);
+	if (fn && typeof(fn) === "function") return fn.call(this, optionsHook('event', e) || e);
 }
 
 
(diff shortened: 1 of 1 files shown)


developit:
Hi Akeem,

Any context for this PR? I had left out the type check in order to show an error when the handler was the wrong type. Though I guess the error would be a bit useless ("cannot call method call of undefined"). Seems like it would be worth showing some form of error though?  Interested to hear your thoughts.

AkeemMcLennon:
Hi @developit, this really goes in conjunction with https://github.com/developit/preact-markup/pull/3

When I wrote a test for the negative case where the `allow-events` weren't enabled, I got the undefined error. 

However, I'm fine with throwing an Error / Exception instead if that's what you'd prefer.

developit:
Ah, excellent, that's what I was looking for.  I'm going to check if there is a way to throw a more useful error here, will let you know.  Part of an ongoing effort to provide a better debugging experience, but retain the dev <-> prod parity, instead of requiring that people switch builds or apply globals to get suitably production-worthy builds.

AkeemMcLennon:
Updated to throw an Error now

developit:
I think we might be able to combine this with #153. If it can cover all the cases for event handlers, need add more tests and look at compat, then I believe the TypeError would be similar to this one. Will keep you posted as I merge these on Thursday/Friday (sorry for vacation delays!).

developit:
Hey Akeem - I'm going to close this one out since the work for #153 ended up throwing a `TypeError` here for non-function handlers.  Thanks so much for your work!!

PR #150, opened 2016-05-17
Fixed named export for using with rollup and meet the js modules spec

impronunciable:


developit:
Tweaking a bit after merge, but this is great thanks!

PR #160, opened 2016-05-23
Update index.js

aeosynth:
a function which doesn't explicitly return a value returns `undefined`, so this check doesn't do anything

it's the same as:

```
if (node[ATTR_KEY]!==undefined) return node[ATTR_KEY]
else return node[ATTR_KEY]
```

### DIFF
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -9,7 +9,7 @@ export function ensureNodeData(node, data) {
 
 
 export function getNodeData(node) {
-	if (node[ATTR_KEY]!==undefined) return node[ATTR_KEY];
+	return node[ATTR_KEY];
 }
 
 


developit:
Hmm - this is a V8 optimization. The existence check prevents the function from being megamorphic.

aeosynth:
then it should have a comment to prevent people like me from getting confused, and a link to a benchmark

developit:
That's a good idea :)

developit:
@aeosynth Turns out inlining this ended up being faster + smaller!

aeosynth:
neat. you can get rid of the function entirely now.

developit:
Released in `5.0.0-beta2`

PR #161, opened 2016-05-24
Add default export for compatibility

le0nik:
For example, many people have `preact.h` as their pragma. They also configure their `eslint` config to ignore `preact` variable, because it's required for transforming `jsx`.

This PR makes it possible to import stuff like this(if you use `webpack-2`+ or `rollup`):

```js
import preact, { Component } from 'preact';
```

Which is exactly how you import `react`. Haven't found example of this in official react docs, but here it is in `react-native` docs(first lines of first code example): https://facebook.github.io/react-native/.

So tree-shaking will work for people who want it. They just won't use the default import. But it's still going to be possible to use for people, who need the `preact` variable if they use `webpack-2`+ or `rollup`.

### DIFF
--- a/src/preact.js
+++ b/src/preact.js
@@ -1,5 +1,21 @@
-export { h } from './h';
-export { Component } from './component';
-export { render } from './render';
-export { rerender } from './render-queue';
-export { default as options } from './options';
+import { h } from './h';
+import { Component } from './component';
+import { render } from './render';
+import { rerender } from './render-queue';
+import options from './options';
+
+export default {
+	h,
+	Component,
+	render,
+	rerender,
+	options
+};
+
+export {
+	h,
+	Component,
+	render,
+	rerender,
+	options
+};


developit:
Thanks @le0nik - I figured this would end up needing to be re-revisited.  I had tried to simplify #150, but it would indeed break the default export ES Modules (such as via `jsnext:main`).

I was curious if it would be possible to drop down to CommonJS here and export an object literal, but your point about tree shaking makes it worthwhile to keep the solution proposed here and by @impronunciable in his similar PR.

developit:
Just a little update:  Adding the reflected default exports bumped the filesize up quite a bit, so I have [devised a way](https://github.com/developit/preact/blob/master/rollup.config.js#L8-L11) to keep it in the ES Modules source (jsnext:main), but strip it out of the CommonJS bundle at build time. This means the CommonJS source simply exports everything as named, which works with Babel's cjs interop mode since it emulates the reflected default export we're doing in `src/preact.js`.

I'm hoping this covers both the ES Modules and CJS use cases nicely, let me know if you run into issues!  The revised exports are published as `preact@5.0.0-beta3`.

impronunciable:
Awesome! I'll try it and fill an issue if it doesn't work

PR #171, opened 2016-05-30
Added child validation using strict render check

mrbar42:
When working on my new module [trixion](https://github.com/mrbar42/trixion), I saw this:
On consecutive renders, children are validated with a strict comparison to the child's constructor reference.

I try a method where the components can be object literals and if so they are converted to component class on the go whenever they are rendered.

Because the component constructor gets re-created - it breaks the comparison mechanism thus re-rendering the children every them.

This PR adds another mechanism where the render method is a validation options as well.
I figure that render strict ref is as good as constructor ref since they are both almost always composed solely for a given component.

### DIFF
--- a/src/vdom/component.js
+++ b/src/vdom/component.js
@@ -117,7 +117,8 @@ export function renderComponent(component, opts, mountAll) {
 			// set up high order component link
 
 			let inst = component._component;
-			if (inst && inst.constructor!==childComponent) {
+			let childRender = childComponent.prototype.render;
+			if (inst && !(inst.constructor===childComponent || inst.render===childRender)) {
 				toUnmount = inst;
 				inst = null;
 			}


developit:
Hmm - interesting use-case and I think I follow what you're going for with the shared constructor, but using `render()` breaks component inheritance:

```js
class Base {
  render() {
    return <div>{this.name}</div>;
  }
}
class A extends Base {
  name = 'A';
}
class B extends Base {
  name = 'B';
}
```

Just so I'm on the same page, are you referring to [this function](https://github.com/mrbar42/trixion/blob/master/src/lib/preact.js#L28-L44)?  I know @ld0rman put together a version of it that was working, maybe compare the two?  His is [here](https://github.com/ld0rman/preact-classless-component/blob/master/index.js).

One thought I had for your case specifically would be to use a map to cache conversion of object literal component to a classful component:

```js
let cache = [];
export function createClass(obj) {
  // try to find a cached class
  for (let i=cache.length; i--; ) {
    if (cache[i].obj===obj) return cache[i].component;
  }

  function F() {
    Component.apply(this, arguments);  // important to still pass (props, context)
  }

  let p = F.prototype = Object.create(Component.prototype);
  for (let i in obj) {
    if ('getDefaultProps' == i) {
      //noinspection JSUnfilteredForInLoop
      F['defaultProps'] = obj[i]();
    }
    //noinspection JSUnfilteredForInLoop
    p[i] = obj[i];
  }

  // cache it
  cache[i] = { obj, component:F };
  return p.constructor = F;
}
```

developit:
@mrbar42 Just going to drop a hint here for those who come across this later:

trixion provides a `createClass()` mechanism to construct stateful components from plain objects, but doing so made it possible to construct `N` preact Component variants from the same source object.  @mrbar42's solution was to store a reference to the generated Component (constructor) on the source object, and simply return that cached value for subsequent calls to `createClass()` given the same object as input.

Here is [his revised version](https://github.com/mrbar42/trixion/blob/cc3351380a540e7336e9e7a66c5113c9a850310f/src/lib/preact.js#L30-L62) (inlined for clarity):

```js
export function createClass(obj) {

  if (obj.__CREF__) return obj.__CREF__;

  function F() {
    Component.apply(this, arguments);
  }

  let p = F.prototype = Object.create(componentPrototype);
  F.prototype.getDOMNode = function () {
    return this.base;
  };
  F.prototype.isMounted = function () {
    return !!this.base;
  };
  for (let i in obj) {
    if ('getDefaultProps' == i) {
      //noinspection JSUnfilteredForInLoop
      F['defaultProps'] = obj[i]();
    }
    //noinspection JSUnfilteredForInLoop
    p[i] = obj[i];
  }

  p.constructor = F;

  Object.defineProperty(obj, '__CREF__', {
    value: F,
    writable: true
  });

  return F;
}
```

PR #202, opened 2016-06-21
🎨 Add built-in support for inline SVG

developit:
It looks like around 3/4 of people wanted to merge SVG support into Preact itself.  I did a poor job of conveying the fact that there would actually be a massive size _savings_ for anyone using not only `preact-svg`, but actually also `preact-compat`, since it in turn pulls in `preact-svg` (to normalize for the fact that React ships SVG support by default).

Total added weight is `114 bytes`.

### DIFF
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -30,7 +30,7 @@ export function removeNode(node) {
  *	@param {any} previousValue	The last value that was set for this name/node pair
  *	@private
  */
-export function setAccessor(node, name, value) {
+export function setAccessor(node, name, value, isSvg) {
 	ensureNodeData(node)[name] = value;
 
 	if (name==='key' || name==='children') return;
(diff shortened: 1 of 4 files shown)


zhenkunou:
hi @developit, 

In this [codepen](http://codepen.io/ryan_ou/pen/VjmvQy?editors=0110) demo, the `className` is lost after render. (`preact@5.1.0-beta.16`)

It is so weird that, i edit the [test case call 'should render SVG to DOM'](https://github.com/developit/preact/pull/202/commits/e6a7bd7c6e8b60273da13a15203a138f10b6644d#diff-941ea4c3b24fc6d29b0f3a5cce3aba5bR37) to re-appear the bug:
```javascript
const Deom = () => (
    <svg viewBox="0 0 360 360" className="svg-class">
        ...
    </svg>
);
render(<Deom />, scratch);

expect(scratch.innerHTML).to.equal('<svg viewBox="0 0 360 360" class="svg-class" ...')
```
but it is passing.

I hope this information could help. thx for your great work again.

developit:
Ahh @zhenkunou good catch, I will fix this ASAP!

developit:
@zhenkunou fixed and released, here's the codepen against `5.1.0-beta.17`: http://codepen.io/developit/pen/beBBJM?editors=0110

PR #206, opened 2016-06-22
Added start coffeescript project

crisward:
For those who want to use preact with CJSX. It may not be to everyones taste, but it's good to see it's possible. This helloworld is 10kb gzip/min which is pretty good. Plan to extend this to add a example test setup and possible css integration.


developit:
Awesome work, this looks useful for coffeescripters!

PR #220, opened 2016-06-30
Add preact flow definition

### LINKED ISSUE #219: Flow support
Hi,

I was wondering if you have looked into supporting [Flow](https://flowtype.org), it has built in support for JSX type checking, but I'm unsure whether it would be possible to make it work with Preact.

bouk:
Because flow has special built-in support for React/JSX, we can piggy back on that by re-exporting everything from react in the preact flow definition (and also adding h = React.createElement).

I tested this out on my project and get correct errors for JSX property type mismatches and such!

for review @developit 

Fixes #219

### DIFF
--- a/src/preact.js.flow
+++ b/src/preact.js.flow
@@ -1,0 +1,9 @@
+/* @flow */
+
+import { createElement as h, cloneElement, Component, render } from 'react';
+
+export { h, cloneElement, Component, render };
+export default { h, cloneElement, Component, render };
+
+declare export function rerender(): void;
+declare export var options: Object;
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
   "minified:main": "dist/preact.min.js",
   "scripts": {
     "clean": "rimraf dist/",
-    "build": "npm-run-all clean transpile strip optimize minify size",
+    "copy-flow-definition": "cp src/preact.js.flow dist/preact.js.flow",
+    "build": "npm-run-all clean transpile copy-flow-definition strip optimize minify size",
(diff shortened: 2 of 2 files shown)


developit:
Looks good, though I wonder if it'd be cleaner to fork React's definition?  If you have a link to it handy that'd be great, I haven't done much with Flow quite yet.

src/preact.js.flow
```diff
@@ -0,0 +1,11 @@
+/* @flow */
+
+import * as React from 'react';
+
+export * from 'react';
+
+const Preact = Object.assign({}, React, {
```
developit: Usually we lowercase, but I don't imagine it actually matters here since the export is a default.

bouk:
The react definition is actually hardcoded inside Flow: https://github.com/facebook/flow/blob/master/lib/react.js

developit:
Oh wow there's quite a lot in there.  I have no problem merging this as-is, but how hard do you think it would be to only (re-)export the core stuff preact provides?

In my head I'm thinking something like this:

```js
import { createElement as h, cloneElement, Component, render } from 'react';

export { h, cloneElement, Component, render };
export default { h, cloneElement, Component, render };

// can these just be added here? (not super familiar with Flow definitions)
declare function rerender(): void;
declare var options: any;
```

bouk:
@developit I've updated the definition

developit:
@bouk Perfect, happy to merge this!

developit:
🎉 Released as [preact@5.1.0-beta.22](https://github.com/developit/preact/releases/tag/5.1.0-beta.22)! If you are able to test it out, let me know if it worked 👍

PR #224, opened 2016-07-07
Stop event listener from being added when event handler is null

HenningM:
Hi,

I've discovered a minor problem with the removal of event listeners. More specifically this problem is triggered by replacing a component that has an event handler attribute with another component having no such handler. It turns out that in these cases `!l[name] == true` and `value == null` in setAccessor(), meaning that an event listener is added in this case. This leads to an error in eventProxy() since null is called as a function.

I've set up a simple example reproducing the behavior here: http://jsfiddle.net/jsLhf3Lz/. Hovering up & down the list of items should produce the error I've described here. Please excuse me if I simply misunderstood how this is supposed to work.

### DIFF
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -51,8 +51,8 @@ export function setAccessor(node, name, value, isSvg) {
 	else if (name[0]==='o' && name[1]==='n') {
 		let l = node._listeners || (node._listeners = {});
 		name = toLowerCase(name.substring(2));
-		if (!l[name]) node.addEventListener(name, eventProxy);
-		else if (!value) node.removeEventListener(name, eventProxy);
+		if (!value && l[name]) node.removeEventListener(name, eventProxy);
+		else if (!l[name]) node.addEventListener(name, eventProxy);
 		l[name] = value;
 	}
 	else {


developit:
@HenningM Looks like a good fix for a valid bug! I might re-reverse these post-merge, but I think we can merge this 👍

slaskis:
I'd love if this could get merged, we're using the latest beta for some sweet bug fixes to the component lifecycle and now when some buttons are being clicked we get these beauties:

```
preact.js:188 Uncaught TypeError: this._listeners[e.type] is not a function
    eventProxy @ preact.js:188
```

developit:
@slaskis I'll publish a release this morning, sorry!

slaskis:
Yaaay, Thanks a bunch for the speedy turnaround!

developit:
Released as `preact@5.1.0-beta.24` 🎈

slaskis:
Tested it out a little bit and it seems to work wonderfully, although the exception was only thrown sometimes, but so far so good!

slaskis:
Oh wow I posted too soon. It failed just as I clicked the Comment button here...

Sorry for the false alarm then, I must be bumping in to some other issue which seemed related in my app. My event listener is assigned to `null` somehow.

screenshot 2016-07-08 14 09 16

I'll post a new issue if I can figure out why...

HenningM:
Yeah, actually this "fix" of mine doesn't really do anything to remedy the issue since it the input combination listed in the PR description still ends up adding the listener. So `value` also needs to be part of condition for adding the listener. Sorry about not testing this properly before submitting the PR!

developit:
@HenningM Indeed, just noticed that and [a fix](https://github.com/developit/preact/commit/dfb4872683411f90f22665728d0e87df15cf2007) is now published as `preact@5.1.0-beta.26`.

HenningM:
@developit Great. Thanks a lot! And again: really sorry for not consulting with a computer before sending the PR.

developit:
No problem. The tests should have caught it, send they're missing an edge case.

slaskis:
Wow, thanks for fixing it so quickly, will try this version out too now

slaskis:
So far so good now! Thanks again guys!

PR #229, opened 2016-07-11
Support for Buble and Rollup

povrazor:
Hello. I had to make some very minor tweaks to get Preact working with Buble (Bublé?) and Rollup JS.

Most changes are tweaks to the imports. Rollup JS (without plugins) doesn't support NPM style implicit `index.js` files, so anywhere an `import blah from '.'` or `import blah from '../dom'` was used had to be tweaked. According to the Rollup JS author (and several bug reports suggesting this behaviour), implicit `index.js` files aren't part of the module import spec. Best I can tell he's right.

https://developer.mozilla.org/en/docs/web/javascript/reference/statements/import

As for Buble, there is one tiny change I made in `vdom/diff.js`. The variable `child` is an uninitialised variable inside a loop. It's possible for the variable never to be set before it's used, and sometimes carries the value from the prior iteration. This was causing JSX code to only ever output the last child node. i.e:

```jsx
render((
    <div id="foo">
        <div>Hello, world!</div>
        <div>How are you?</div>
        <div>Good</div>
    </div>
), document.body);
```

Would output this:

```html
<div id="foo">
    <div>Good</div>
</div>
```

### DIFF
--- a/src/dom/recycler.js
+++ b/src/dom/recycler.js
@@ -1,5 +1,5 @@
 import { toLowerCase } from '../util';
-import { ensureNodeData, getNodeType, getRawNodeAttributes, removeNode } from '.';
+import { ensureNodeData, getNodeType, getRawNodeAttributes, removeNode } from './index';
 
 /** DOM node pool, keyed on nodeName. */
 
--- a/src/vdom/component.js
+++ b/src/vdom/component.js
@@ -3,11 +3,11 @@ import options from '../options';
 import { isFunction, clone, extend, empty } from '../util';
 import { hook, deepHook } from '../hooks';
 import { enqueueRender } from '../render-queue';
-import { getNodeProps } from '.';
+import { getNodeProps } from './index';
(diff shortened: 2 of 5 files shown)


developit:
Hi @povrazor - this looks great, and I totally agree with the principle you alluded to: if Preact is going to ship ES2015 source as `jsnext:main`, that source should be proper ES2015.

I just wanted to clarify though that the `child` issue you noticed is actually a Bublé edge case, and isn't a bug when using Babel or the transpiled `preact.js` output.  That being said, it's such a small fix I think it's worth doing.  I actually went to fairly absurd lengths to account for that very edge case in [a codemod](https://github.com/developit/preact/blob/master/config/codemod-strip-tdz.js) that strips Babel's `void 0` TDZ initialization.  Your fix is simple, and removes the edge case entirely, so I'm happy to merge.  I'll likely just hoist those declarations to the top of `innerDiffNode` once merged (if it helps the file size).

PR #232, opened 2016-07-16
Fix `ref` becoming null before `componentWillUnmount`.

slmgc:
Hi @developit!

While checking `preact` with my big app I found this issue:

```es6
class TestUnmount extends React.Component {
	componentWillUnmount() {
		console.assert(
			this.refs.works,
			'`works` points to a DOM-element as it should'
		)

		console.assert(
			this.refs.breaks,
			'`breaks` should point to a DOM-element but contains null'
		)
	}

	render() {
		return (
			<div ref="breaks">
				<div ref="works" />
			</div>
		)
	}
}
```

EDIT: ~~I moved the hook which nullifies `ref` to be executed just before the `componentDidUnmount` hook to solve this issue.~~

Sorry, looks like I've been checking the fix with ```React``` instead of ```preact```. It doesn't resolve the issue.

P.S. nevertheless, by looking at the code I found a function call responsible for this issue:
https://github.com/developit/preact/blob/master/src/vdom/component.js#L265

Looks like `removeOrphanedChildren` nullifies nodes before the `componentWillUnmount`. By disabling this call I was able to evade the issue but I don't know all ins and outs, your assistance would be very much appreciated.

### DIFF
--- a/src/vdom/component.js
+++ b/src/vdom/component.js
@@ -247,7 +247,6 @@ export function buildComponentFromVNode(dom, vnode, context, mountAll) {
 export function unmountComponent(component, remove) {
 	// console.log(`${remove?'Removing':'Unmounting'} component: ${component.constructor.name}`);
 
-	hook(component, '__ref', null);
 	hook(component, 'componentWillUnmount');
 
 	// recursively tear down & recollect high-order component children:
@@ -267,5 +266,6 @@ export function unmountComponent(component, remove) {
 
 	}
 
+	hook(component, '__ref', null);
 	hook(component, 'componentDidUnmount');
 }


developit:
Interesting bug! I'll need a bit too look into this, but it seems like your change is the right spot.

slmgc:
@developit I've checked the app with ```preact@5.3.0```  & ```preact-compat@2.0.0``` and it still fails in some cases with the same error. I'm still looking into making a minimal reproducible example.

developit:
Odd :(  Might be worth double-checking against `5.3.1` (just pushed), since that was addressing a silly mistake for unmount events.  Repro is for sure the way to go here. If you can find a way, I'll happily port it to be a unit test and then we'll be guaranteed a fix 👍

slmgc:
@developit ```5.3.1``` fixed the issue, thanks! I have a few more different errors, but I think they aren't related to this one :)

PR #236, opened 2016-07-18
Remove preact-svg, add utility libraries

queckezz:
This PR removes `preact-svg` (since it's deprecated) from the readme as well as adding `preact-classless-component`, `preact-hyperscript`.

Do you prefer to keep a maintainable, high-quality list of modules or is it possible to move the tools section to a wiki so anybody could add their library?

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,6 @@ As one would expect coming from React, Components are simple building blocks for
 ## Libraries & Add-ons
 
 - :earth_americas: [**preact-router**](https://git.io/preact-router): URL routing for your components.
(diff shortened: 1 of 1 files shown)


developit:
Hi @queckezz - I'd actually like to get these into the Preact website itself, since the intent is to remove the wiki entirely.  The website was originally forked from the wiki content, and has been kept up to date better.

If you're willing, I'd love a PR for these changes to the website:
https://github.com/developit/preact-www/blob/master/content/about/libraries-addons.md

PR #237, opened 2016-07-19
Improving performance by making direct calls to hook objects

trashhalo:
roadhog_paris___overwatch_by_plank_69-d9apg7f

So I was profiling preact in uibench and one of the top hot areas is calling the **hook** function.

## [Inferno](https://github.com/trueadm/inferno/blob/61ab85c780b0e4938f8f8166bdbeb9f918cde5e4/src/DOM/utils.js#L247)
```javascript
if (!isNullOrUndefined(hooks.created)) {
  hooks.created(dom);
}
```
## [Preact](https://github.com/developit/preact/blob/master/src/hooks.js#L14)
```javascript
export function hook(obj, name, a, b, c) {
	if (obj[name]) return obj[name](a, b, c);
}
// ...
if (empty(b) || mountAll) {
	hook(component, 'componentWillMount');
}
```

In inferno each place that calls a hook does its own if check and calls the hook via a dot call. Its unclear to me if its the extra function call or its the dot notation vs bracket lookup that's the issue here but I swapped out preact's implementation for the style inferno's doing and we seem to be getting a 5-10% speed bump. 

Other small note I left the hook calls that were weirder to me. Like the options hook stuff and the ATTR hook stuff. Wasnt clear if I could cleanly swap them out for dot calls.

# Bench: Left Column my branch. Right is master
ui benchmark

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -1,5 +1,4 @@
 import { FORCE_RENDER } from './constants';
-import { hook } from './hooks';
 import { extend, clone, isFunction } from './util';
 import { createLinkedState } from './linked-state';
 import { triggerComponentRender, renderComponent } from './vdom/component';
@@ -26,7 +25,7 @@ export function Component(props, context) {
 	/** @type {object} */
 	this.props = props;
 	/** @type {object} */
-	this.state = hook(this, 'getInitialState') || {};
+	this.state = this.getInitialState && this.getInitialState() || {};
 }
 
 
--- a/src/hooks.js
+++ b/src/hooks.js
@@ -3,13 +3,5 @@ import options from './options';
 
 /** Invoke a hook on the `options` export. */
 export function optionsHook(name, a) {
-	return hook(options, name, a);
-}
-
-
-/** Invoke a "hook" method with arguments if it exists.
- *	@private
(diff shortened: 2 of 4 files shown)


developit:
This awesome.

developit:
Looks like the size bumped up by about 20 bytes. Maybe if there is a way to nix the rest of the hooks so we can drop both `hook()` and `optionsHook()` it'll be a net size loss 💃

trashhalo:
whats the deal with this guy? https://github.com/developit/preact/blob/master/src/constants.js#L11 If I can drop the Symbol stuff I can remove another hook call.

developit:
It's used for hanging properties off of DOM nodes. In my testing (and this is something I'm not 100% sure of) using a symbol sped things up, I assume because it avoided triggering repaints?  It's probably worth looking into.  Right now I'm not aware of anything using this (rather private/undocumented) "api", so it's possible we can change it to drop some bytes.  That said, it should be possible to still remove the rest of the hooks without removing the Symbol.

developit:
This has been released in `5.4.0`! Thanks again! 🎉 🎈

PR #251, opened 2016-08-01
add TypeScript typings

kruczy:


developit:
Build error is just a saucelabs issue, ignore it. Thanks for this!

jmarsh999:
Just a quick issue with the typings definition for linkState.  Currently it is defined to return a void, but in reality it should return a function:  
linkState:(name:string) => (e: Event)=>void;

developit:
Correct! Want to PR?

jmarsh999:
Sure, the PR has been submitted

PR #253, opened 2016-08-02
improve typings, install via typings cli

kruczy:
this will allow preact typings to be installed globally via typings cli
`typings install github:developit/preact --global`

the reason to do it is that JSX typings will not work if they are not global

the next step would be to add a reference in typings/registry

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,6 @@
   "version": "5.6.0",
   "description": "Tiny & fast Component-based virtual DOM framework.",
   "main": "dist/preact.js",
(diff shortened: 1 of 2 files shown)


developit:
@kruczy in order to make `typings install preact` work, should we be including `typings.json` in the npm release? Or is that what their registry does?

kruczy:
@developit there will need to be an entry in the registry pointing at preact github repo, I don't think it needs to be included in the dist npm package, will make a pr against the registry to include preact there

developit:
Perfect!

PR #264, opened 2016-08-10
Fix to allow props & context in componentWillMount

mdierolf:
I found that I was unable to access `this.context` from within `componentWillMount()`, and made the patch below.

I am not sure if this introduces differences vs React but I was unable to build a contextual component that functions like the following without the patch:

 ```
componentWillMount() {
 this.context.form.registerField(this);
}
componentWillUnmount() {
 this.context.form.unregisterField(this);
}

```

### DIFF
--- a/src/vdom/component.js
+++ b/src/vdom/component.js
@@ -37,22 +37,22 @@ export function setComponentProps(component, props, opts, context, mountAll) {
 	if ((component.__key = props.key)) delete props.key;
 
 	if (empty(b) || mountAll) {
-  component.context = context;
-  component.props = props;
+		component.context = context;
+		component.props = props;
 		if (component.componentWillMount) component.componentWillMount();
 	} else {
-	 if (component.componentWillReceiveProps) {
-		 component.componentWillReceiveProps(props, context);
-	 }
+		if (component.componentWillReceiveProps) {
+			component.componentWillReceiveProps(props, context);
+		}
 
- 	if (context && context!==component.context) {
-	 	if (!component.prevContext) component.prevContext = component.context;
-		 component.context = context;
- 	}
+		if (context && context!==component.context) {
+			if (!component.prevContext) component.prevContext = component.context;
+			component.context = context;
+		}
 
-	 if (!component.prevProps) component.prevProps = component.props;
- 	component.props = props;
- }
+		if (!component.prevProps) component.prevProps = component.props;
+		component.props = props;
+	}
 	component._disableRendering = false;
 
 	if (opts!==NO_RENDER) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by developit** [comment]
Hey Mark - thanks for the PR! I'll double check React's internals but it seems like context should be available in that method.  One thing to watch out for though - `componentWillMount` is invoked during server rendering, whereas `componentDidMount` is only invoked on the client.

**Comment by developit** [comment]
Don't worry about the build failure by the way, that's just a saucelabs config issue I haven't been able to sort out yet (tests are passing).

**Comment by mdierolf** [comment]
Thanks Jason - it seemed like an omission but I see no mention of it in the react docs, so i'm not sure what the expected behavior should be.

**Comment by developit** [comment]
@mdierolf quick question - is there any chance the class you noticed this in was not inheriting from `Component`?

**Comment by mdierolf** [comment]
Jason,
I figured out where the problem lies.
We are inheriting from Component, but we do so through a subclass.

I did some more testing, and it looks like the issues is this code:

```
class FormFieldComponent extends Component {
  constructor() {
    super();
     ...
  }
}

class Input extends FormFieldComponent {
  componentWillMount() {
    console.log(this.props); // shows no props
    console.log(this.context); // shows no context
  }
  render() {
    console.log(this.props); // now we have props
    console.log(this.context); // now we have context
}
```

It seems like the props & context are passed through the constructor, so by subclassing and using super() we are dropping them, yet they appear again later during render()

This code solves the problem:

```
class FormFieldComponent extends Component {
  constructor(props,context) {
    super(props,context);
    ...
  }
}
```

I'm not sure if this is simply undefined behavior or would be considered a regression from React. It appears React sets props & context from outside the component, and some people expect that:

> Call super(props) only if you want to access this.props inside the constructor. React automatically set it for you if you want to access it anywhere else.

Links:
https://discuss.reactjs.org/t/should-we-include-the-props-parameter-to-class-constructors-when-declaring-components-using-es6-classes/2781

http://stackoverflow.com/questions/30571875/whats-the-difference-between-super-and-superprops-in-react-when-using-e

developit:
Ahh okay that makes a lot more sense!  So `this.props` and `this.context` are indeed set prior to `componentWillMount()`, it was argument forwarding to the super that was causing the issue.

This is actually a known issue in Preact, and something that has varied in previous react versions.  I'd like to get it fixed, but I'm thinking that happens in a separate issue.  I do have some stuff locally that might work, just needs proper tests.

developit:
@mdierolf Preact 5.7 now enforces `this.props` and `this.context` even if components don't inherit from `Component` or don't forward arguments on to the Component constructor.  Are you able to check if this resolves the underlying issue?

developit:
I've confirmed that this is fixed to more closely follow React's behavior (enforcing `this.props` and `this.context` regardless of constructor/super calls), so I'm going to close this out.  Let me know if there's more to do and we can re-open!

PR #269, opened 2016-08-15
Correct jsnext:main

piuccio:
According to jsforum/jsforum#5 jsnext:main should contain the transpiled version with the exception of ES6 modules.

Similar to developit/preact-compat#144 however the build process here is slightly different.

### DIFF
--- a/config/rollup.config.js
+++ b/config/rollup.config.js
@@ -1,8 +1,13 @@
+import fs from 'fs';
 import nodeResolve from 'rollup-plugin-node-resolve';
 import babel from 'rollup-plugin-babel';
 import memory from 'rollup-plugin-memory';
 
+const pkg = JSON.parse(fs.readFileSync('./package.json'));
+
 export default {
+	entry: 'src/preact',
+	sourceMap: true,
 	exports: 'named',
 	useStrict: false,
 	plugins: [
@@ -19,5 +24,9 @@ export default {
 			blacklist: ['es6.tailCall'],
 			exclude: 'node_modules/**'
 		})
+	],
+	targets: [
+		{ dest: pkg['dev:main'], format: 'umd', moduleName: pkg.amdName },
+		{ dest: pkg.module, format: 'es' }
 	]
 };
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "version": "5.6.0",
   "description": "Tiny & fast Component-based virtual DOM framework.",
   "main": "dist/preact.js",
-  "jsnext:main": "src/preact.js",
+  "jsnext:main": "dist/preact.es.js",
+  "module": "dist/preact.es.js",
   "dev:main": "dist/preact.dev.js",
   "minified:main": "dist/preact.min.js",
   "scripts": {
@@ -12,7 +13,7 @@
     "copy-flow-definition": "cp src/preact.js.flow dist/preact.js.flow",
     "copy-typescript-definition": "cp src/preact.d.ts dist/preact.d.ts",
     "build": "npm-run-all --silent clean transpile copy-flow-definition copy-typescript-definition strip optimize minify size",
-    "transpile": "rollup -c config/rollup.config.js -m ${npm_package_dev_main}.map -f umd -n $npm_package_amdName $npm_package_jsnext_main -o $npm_package_dev_main",
+    "transpile": "rollup -c config/rollup.config.js",
     "optimize": "uglifyjs $npm_package_dev_main -c conditionals=false,sequences=false,loops=false,join_vars=false,collapse_vars=false --pure-funcs=Object.defineProperty -b width=120,quote_style=3 -o $npm_package_main -p relative --in-source-map ${npm_package_dev_main}.map --source-map ${npm_package_main}.map",
     "minify": "uglifyjs $npm_package_main -c collapse_vars,evaluate,screw_ie8,unsafe,loops=false,keep_fargs=false,pure_getters,unused,dead_code -m -o $npm_package_minified_main -p relative --in-source-map ${npm_package_main}.map --source-map ${npm_package_minified_main}.map",
     "strip": "jscodeshift --run-in-band -s -t config/codemod-strip-tdz.js dist/preact.dev.js && jscodeshift --run-in-band -s -t config/codemod-const.js dist/preact.dev.js",


developit:
@piuccio I know you're on vacation, just wanted to drop a note here mentioning that I'm going to try out your `preact-compat` fix first and if that is working nicely then merge this PR.

config/rollup.config.js
```diff
@@ -1,8 +1,13 @@
+import fs from 'fs';
 import nodeResolve from 'rollup-plugin-node-resolve';
 import babel from 'rollup-plugin-babel';
 import memory from 'rollup-plugin-memory';
 
+const pkg = JSON.parse(fs.readFileSync('./package.json'));
```
KenanY: How about?:

``` js
const pkg = require('./package.json');
```

developit: @KenanY If I recall correctly, that doesn't work because rollup preprocesses this file for imports/requires.

KenanY: @developit Oh, I've seen it [used](https://github.com/rollup/rollup-starter-project/blob/f076671cf11f02408487b9d4bc52458301e49605/rollup.config.js#L5) [elsewhere](https://github.com/rollup/rollup-plugin-babel/blob/8c1bc1c7db6c3acc06755e7a5d411cfb86c3e19c/rollup.config.js#L3) but if it doesn't work here then I'll take your word for it :)

piuccio: I think it works if you load rollup-plugin-json but I mostly copied from preact-compat that uses this line

developit: Ah I'm probably just crazy and it works fine. I can change it haha.

mkautzmann:
How is this going? This is important for people using Webpack 2 and similar tools.

People usually do not transpile `node_modules` content. They expect there to have transpiled code already.

Without transpilation to ES5, browsers that do not support basic ES6 stuff (let, for example) may break when importing preact via Webpack 2, Rollup, etc because it's grabbing the `src` directory.

developit:
@mkautzmann Could use some extra eyes on this PR actually. It looks good to me, and is similar to what was done for other repos (I believe preact-redux?).  I like the idea of shipping a bundled `modules:main` if that solves Webpack 2. Someone raised that exact concern yesterday in Gitter.

mkautzmann:
Yeah, the main problem is that we don't have a universal pattern for ES modules yet. Things can change.

Rollup, Webpack 2 and other tools began rolling out their own `package.json` properties to handle ES modules in an effort to set a pattern.

I think we can, at least for now, export `jsnext:main` and `module`. I know Webpack 2 and Rollup handle these.

developit:
Yup. I think there is already a `jsnext:main` in the package.json, but it points to the `src/` directory. This technically works, but the bundle you get back is generally quite a bit larger than just using `dist/preact.min.js` (since it doesn't have the codemods and stuff run on it).

mkautzmann:
You are right, it works in most cases but I can see it breaking if the user runs the code directly in an old browser (with no transformation). That's why I think @piuccio point is valid. We should export a new file that contains the transpiled code + ES module syntax.

I think I will clone this branch and compare to `6.4.0` in a Webpack 2 scenario and report my results.

developit:
Sound good to me!

NicolasParada:
I don't agree. The jsnext:main points a to an ES6 module, so it can have the complete ES6 syntax.
The person who is gonna use the jsnext:main is using rollup, so it's up to him transpile to old code or not.

I would prefer my bundle with `const`, `arrow functions` and all those stuff, we're almos two years from ES2015 now.

piuccio:
I'm not sure anymore what's the solution, the only thing for sure is that the jsnext version should have modules. 

For advanced use cases it's actually handy to point to source code, ES6, because if you run your app in modern browsers only you might not need any transpilation at all. 
However most people transpile back to ES5, and they prefer not having to tinker with node_modules and webpack configurations. 

So it's a choice between convenience for the majority or flexibility. 

It feels like jsnext is too simplistic, it doesn't specify the target environment. 
Something like this would have been better

```
"jsnext:main": {
   "es2015" : "blah", 
   "es2016" : "bleh" 
} 
```

Anyway, as far as I'm concerned, this can be closed by simply updating the readme with examples on how to use babel on node_modules with wp/rollup

developit:
Interesting note: I tested bundle size via Rollup and Webpack 2 (both recursive transpiled w/ jsnext:main). Rollup was trivially larger than using preact's dist, Webpack 2 nontrivially larger.

piuccio:
@developit same version of babel, same configuration, same uglify? If I remember correctly I also had larger file size with rollup, haven't tried webpack, nothing big, few bytes

mkautzmann:
@developit Transpiling to ES5 on both? That's odd, do you know why the bundle size is larger? As you are not defining an `umd`, but just an ES module it should be smaller actually...

developit:
The benefit of UMD is that it wraps the entire codebase in a function, which means Uglify can mangle what were previously top-level variable names. Normally it can't do this. That means all the internals of Preact get mangled, keeping things super small.  Also there are a few codemods applied to Preact during its own build that bring the size down.

NicolasParada:
With rollup and babily to minify, I got for a simple app:
* jsnext:main: 10.8kB
* without jsnext:main: 10.4kB

mkautzmann:
I understand. Well, the way I see `jsnext:main` or `module` today is that it indicates that the file exports ES modules.

That bundle could be written in any JS version + exporting ES modules.

The JS version you are going to transpile to goes down to browser support. As preact intends to support IE9, it's mostly ES5.

If we are going to keep exporting `src` we should inform the user that transpilation may be necessary to get support for old browsers.

btd:
Just want to bump this historic issue, because i spend around 3 hours today when tried to get work rollup and preact. In Rollup docs clearly said that jsnext:main **must** be es5 + modules and nothing else. 
There is could be point to store es6 sources but it could be used in only fresh browsers and more or less last node. If you support ie9 i do not see any reason for this.

Please include huge red notice for rollup users that this module must be transpilled!

To anyone googling solution i come to putting my files to one folder (in my case `./js`) and relevant part of rollup config:
```
rollupBabel({
        plugins: ["external-helpers", ["transform-react-jsx", { pragma: "h" }]],
        presets: [
          [
            "es2015",
            {
              loose: true,
              modules: false
            }
          ]
        ],
        babelrc: false,
        include: ["./js/**", "node_modules/preact/**"]
})
```

@developit can you please do something with this issue please? For newcomers (react -> preact + rollup) it is big pain point.

developit:
@btd I'd like to update Preact's build so that we produce an ES Modules + ES5 bundle, which is how `preact-compat` works.  It's similar to this PR, I just haven't found the time yet.

btd:
@developit great to hear. do you need help with that PR if you do not have time?

developit:
That might be the best way forward, yeah.  I had been trying to find time because I'm thinking it'll require upgrading to Babel 6 (we're still on 5 here haha), and that is going to mean a lot of size tweaking.  Right now there are codemods that expect Babel 5 -style output.  I'd be up for PRs though, even this one is quite a good start.

config/rollup.config.js
lukeed: Reviving, but strolling thru 🚶... 

```js
const pkg = require('../package');
```

will work perfectly fine -- I do it in all my modules. 😄

PR #274, opened 2016-08-17
Setup preact aliases bundle

ngasull:
In response to https://github.com/developit/preact/issues/168
These changes set up a new transpiling target for aliases.

The goal isn't to provide more features but just direct compatibility with a supposedly minimalistic React. For instance at the moment, `preact/aliases` export the same stuff as `preact` with `h` as `createElement` on top of that.

When using `preact` as a node module, one can use `preact/aliases` as a drop in replacement for a minimalistic use of `react` module (without `preact-compat` :raised_hands: features)

If anything isn't properly done/named in the configuration, please tell me, thanks! :pray:


config/rollup.config.aliases.js
```diff
@@ -0,0 +1,12 @@
+import memory from 'rollup-plugin-memory';
+import rollupConfig from './rollup.config';
+
+export default Object.assign({}, rollupConfig, {
+	plugins: [
+		memory({
+			path: 'src/preact',
+			contents: `import { h } from './preact';export * from './preact';export { h as createElement };`
```
developit: Wondering if it's worth pulling this into a file (`src/aliases.js`?) for visibility.  Or if there's anything else we might add in the future, it'd be easier in a file. Thoughts?

developit:
@ngasull Let me know if you want me to move the file after merge :)

config/rollup.config.aliases.js
ngasull: Wondered about that too but I preferred sticking to the default rollup config in order to minimize the complexity added by this PR. Moreover I think there shouldn't be anything more from basic React's API that isn't in `preact`: we only hit an issue with `createElement` because it's React's implementation of jsx's `h`, which in the end isn't standard.

However still this code as a literal is pretty ugly :mask: and I would be ok with isolating it. You've got the last word: should I do it? :athletic_shoe:

ngasull: Forgot to highlight you @developit :smile:

developit: I'm fine leaving it in memory for now, just was riffing.
I'll merge this, but I'm currently _slaving_ over a really annoying bug that'll likely block further releases until it's fixed :( :( :(

Issue #260, opened 2016-08-08
unexpect `class` rendered with merge


csbun:
I used [preact-render-to-string](https://github.com/developit/preact-render-to-string) as a server-side rendering tool and rerender on the client-side with merge.

```jsx
import { h, render, Component } from 'preact';
/** @jsx h */

class Cmp extends Component {
  render() {
    return <p id="cmp" className="my-cmp">Cmp</p>;
  }
}

class App extends Component {
  render() {
    return (<div id="my-app">
      <p id="before">Before</p>
      <Cmp />
      <p id="after-1">After1</p>
      <p id="after-2">After2</p>
      <p id="after-3">After3</p>
    </div>);
  }
}
```

- server-side rendering
```jsx
import { h } from 'preact';
import render from 'preact-render-to-string';
const ssrHtml = render(<App />);
res.send(`
  <div>Preact</div>
  <div id="container">${ssrHtml}</div>
  <div id="log"></div>
`);
```

- client-side rendering
```jsx
import { h, render } from 'preact';
render(
  <App />,
  document.getElementById('container'),
  document.getElementById('my-app')
);
```

Here is the [full example](http://codepen.io/csbun/pen/RRqKGZ) on codepen.

As the result comes out, there is an unexpect `class` on `p#after-1`:
Server-side Rendering:
```html
<div id="my-app"><p id="before">Before</p><p id="cmp">Cmp</p><p id="after-1">After1</p><p id="after-2">After2</p><p id="after-3">After3</p></div>
```

Meanwhile Client-side Rendering:
```html
<div id="my-app"><p id="before">Before</p><p id="cmp">Cmp</p><p id="after-1" class="">After1</p><p id="after-2">After2</p><p id="after-3">After3</p></div>
```

And when I debug on Chrome DevTool, it seems `p#my-cmp`, `p#after-1`, `p#after-2`, `p#after-3` are all replaced. `removeOrphanedChildren` was call twice.

developit:
Thanks for the great detailed report @csbun :)
I'm wondering if the is the same root cause as https://github.com/developit/preact/issues/257

developit:
@csbun hmm - I'm seeing the same output in both cases. Was this with a specific browser?  I checked in Chrome.

csbun:
Sorry I made an mistake on CodePen by missing `className="my-cmp"` in `class Cmp`. [Now](http://codepen.io/csbun/pen/RRqKGZ) it works out:

- Server-side Rendering

```html
[...]

PR #276, opened 2016-08-19
fix: isSameNodeType between Component and real DOM

csbun:
This might be an solution of #260 fixing rendering Component with merge.

`isSameNodeType` in `innerDiffNode` (`src/vdom/diff.js` [line 170](https://github.com/developit/preact/blob/master/src/vdom/diff.js#L170)) return `false` in this use case:

```html
<div id="my-app">
  <p id="before">Before</p>
  <p id="cmp" className="my-cmp">Cmp</p>
  <p id="after">After</p>
</div>
```

```jsx
class Cmp extends Component {
  render() {
    return <p id="cmp" className="my-cmp">Cmp</p>;
  }
}

class App extends Component {
  render() {
    return (<div id="my-app">
      <p id="before">Before</p>
      <Cmp />
      <p id="after">After</p>
    </div>);
  }
}
render(
  <App />,
  document.body,
  document.getElementById('my-app')
);
```

`isSameNodeType` between `p#cmp` and `<Cmp>` SHOULD be `TRUE`.

### DIFF
--- a/src/vdom/index.js
+++ b/src/vdom/index.js
@@ -1,5 +1,6 @@
 import { clone, extend, isString, isFunction, toLowerCase } from '../util';
 import { isFunctionalComponent } from './functional-component';
+import { buildComponentFromVNode } from './component';
 import { getNodeType } from '../dom/index';
 
 
@@ -16,7 +17,9 @@ export function isSameNodeType(node, vnode) {
 		return isNamedNode(node, vnode.nodeName);
 	}
 	if (isFunction(vnode.nodeName)) {
-		return node._componentConstructor===vnode.nodeName || isFunctionalComponent(vnode);
+		return node._componentConstructor===vnode.nodeName
+			|| isNamedNode(node, buildComponentFromVNode(null, vnode).nodeName)
+			|| isFunctionalComponent(vnode);
 	}
 }
 


csbun:
Or maybe we can generate a key to each Component automatically like `react-id`.

developit:
@csbun Hmm - it seems like this is already returning `true` for that case?

```js
// <p id="cmp">._componentConstructor === Cmp
// vnode.nodeName === Cmp
node._componentConstructor===vnode.nodeName
```

Pre-resolving during diff would be nice, but that would trigger mount events and all sorts of shenanigans.

csbun:
It won't be `true` cause `<p id="cmp">` is rendered from server-side and `<p id="cmp">._componentConstructor` should be `undefined`.

developit:
@csbun Let me know if this is solved in 6.0.0 - I've marked the other issue as closed, but want to make sure I've covered your usecase before closing this out.  Thanks for your hard work!!

csbun:
Thanks! I'll try it in my new project these days.

developit:
Going to close this out since I'm fairly confident 6.0.1 being released fixes the issue. Let me know if I'm crazy!

PR #279, opened 2016-08-20
Added support for children that may be arrays.

aweber1:
I encountered this scenario when trying to use preact and preact-compat with radium (https://github.com/FormidableLabs/radium). For whatever reason, when rendering radium-enhanced components, sometimes the child components aren't `simple` or `VNode` or `function`, but instead are 1-element arrays with the first element being `simple` or `VNode`. 

When that is the case, the array is evaluated as a string and rendered as [Object Object] where you would normally expect the component markup to be. The code in this PR attempts to detect the array, then effectively flatten it and add the elements to the children collection.

While the code I've added in this PR "works", it's terribly ugly and may have unintended side effects. So while I'm submitting a PR, I'm mostly trying to bring awareness to the issue and hoping that a more elegant solution can be implemented by someone with better skills/understanding of preact.

### DIFF
--- a/src/h.js
+++ b/src/h.js
@@ -16,7 +16,7 @@ const SHARED_TEMP_ARRAY = [];
  */
 export function h(nodeName, attributes, firstChild) {
 	let len = arguments.length,
-		children, arr, lastSimple;
+		children, arr, lastSimple, lastSimple2;
 
 
 	if (len>2) {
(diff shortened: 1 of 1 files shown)


developit:
@aweber1 thanks for opening this up, definitely an issue we need to resolve. 

I'll take your implementation and a few alternatives I vome up with, and benchmark them to figure out what the best overall technique is.

slaskis:
I think it might be related. But I was trying out using hyperx instead of jsx with preact, to avoid a build step when using Electron.

And after some questioning why preact kept rendering `[Object object]` I figured it was because preacts `h()` expects the children to be more arguments while hyperx passes an array.

So my workaround for now is using a module like this:

```
const {h} = require('preact')
const hyperx = require('hyperx')

module.exports = hyperx(hr)

function hr(tagName, attrs, children) {
  if (Array.isArray(children)) {
    return h.apply(null, [tagName, attrs].concat(children))
  } else {
    return h(tagName, attrs, children)
  }
}
```

KrofDrakula:
I guess this was also resolved by #313.

developit:
Yup! Super appreciative of everyone's effort and contributions here. I ended up doing a bunch of performance testing and used @mhink's implementation and tests with some additional crazy optimizations.

developit:
So just to clarify - @aweber1 @slaskis both of your issues should be fixed in 6.1 :)

aweber1:
Many thanks! Glad a better solution was found than the hack I originally submitted :)

PR #280, opened 2016-08-20
Added check for component.__ref as function.

aweber1:
I'm trying to migrate a code base from react to preact and encountered instances where `component.__ref` was a string value and not a function, thereby resulting in `component.__ref is not a function` errors. 

The fix in this PR seemed to fix the problem, but hopefully there aren't unintended consequences of `.__ref` not being a function.

### DIFF
--- a/src/vdom/component.js
+++ b/src/vdom/component.js
@@ -62,7 +62,7 @@ export function setComponentProps(component, props, opts, context, mountAll) {
 		}
 	}
 
-	if (component.__ref) component.__ref(component);
+	if (component.__ref && isFunction(component.__ref)) component.__ref(component);
 }
 
 
@@ -274,6 +274,6 @@ export function unmountComponent(component, remove) {
 		removeOrphanedChildren(base.childNodes, !remove);
 	}
 
-	if (component.__ref) component.__ref(null);
+	if (component.__ref && isFunction(component.__ref)) component.__ref(null);
 	if (component.componentDidUnmount) component.componentDidUnmount();
 }


developit:
Hi there! This was actually intentional si that an error is thrown for non-functions. For string refs use preact-compat

aweber1:
Thanks for the quick reply! I am actually using preact-compat already. Is there something specific that needs to be enabled/configured in order to use string refs?

developit:
Make sure you are inheriting from `preact-compat`'s Component and using its createElement for JSX.

aweber1:
Ok, we're just using the preact-compat webpack aliases for react and react-dom, so everything "should" be using preact-compat Component and render. However, I think the components that were throwing ref errors are also products of `cloneElement` and `.Children` operations, so maybe an edge case. I'll dig around a bit more now that I know string refs should be handled by react-compat.

Thanks!

And thanks for preact in general - it's a great project and I really like the goals behind it.

developit:
Thanks! I appreciate your effort in tracking this down!

robbiewxyz:
@aweber1 I'm trying to get Auth0 working with preact-compat and I'm hitting this same error, so I'm on with you tracking down this string ref problem.

developit:
I wonder what the commonality is?  @robbiewxyz which Auth0 integration are you using? There seem to be a few.

robbiewxyz:
I'm working with [auth0/lock](https://github.com/auth0/lock).  Like @aweber1, I've aliased `react` and `react-dom` to `preact-compat`, but somehow a string ref is sneaking through and tripping up preact.

developit:
Are you using `babel-preset-react`? I'm wondering if this could be related to `babel-plugin-transform-react-inline-elements` or `babel-plugin-transform-react-constant-elements`.

robbiewxyz:
I'm not using any of those packages.  This is my `.babelrc`:
```javascript
{
	presets: [ 'stage-0', 'es2016' ],
	plugins: [
		[ 'transform-react-jsx', { pragma: 'h' } ]
	]
}
```

developit:
Hmmm - is the ref in your own codebase?  String refs in `preact-compat` are only applied when the JSX is transformed to `createElement()` calls (`import { createElement } from 'preact-compat'`).

robbiewxyz:
No, I'm certain it's not coming from my code.  I'm pretty sure that `cloneElement` is the common denominator with our errors.  My error traces to where the Auth0 codebase calls something like this:

```
return cloneElement (createElement (...), { ref: "something" })
```

And because that new `ref` is going through `cloneElement`, isn't ever getting transformed through `createStringRefProxy`.  So then we have a string ref getting through `preact-compat` and killing `preact`.

developit:
Ah! You're right - the original VNode would no longer be associated because it's created via `cloneElement()`.  I assume `cloneElement()` is being invoked from within a component's `render()`?

If that's the caveat, it should be a pretty easy fix - just have to have cloneElement() apply the same transformations as createElement :)

robbiewxyz:
Yep, that's the case.  Thanks for the support!  I've got a version that's working for me with Auth0, I'd be happy to write a test and open a PR to [preact-compat](https://github.com/developit/preact-compat).

developit:
@robbiewxyz I had been playing with it locally and ended up with a fix and test as well, want to take a look?  I want to be sure it covers your case.

developit:
I believe we can close this out since `preact-compat` now prevents String refs from erroneously falling through cloneElement!

ianlamb:
Hey there, I'm running into this exact same issue trying to use https://github.com/reactjs/react-tabs with preact/preact-compat

They seem to be using cloneElement as well. Maybe this is an edge case that was missed in the original fix, or possibly a regression?

developit:
Hmm, not sure. I haven't seen anything about this in a long long time. Maybe worth a new issue on the compat repo?

PR #289, opened 2016-08-29
Specify which version of node to use

ivantm:
If I try to build this with an older version of node (4.x), I get an error around a new ES6 feature that isn't supported:

```
Unexpected token ...
SyntaxError: Unexpected token ...
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:373:25)
```

I thought it would be useful to have a .nvmrc file (for [NVM](https://github.com/creationix/nvm) users) and also use the [engines property of package.json](https://docs.npmjs.com/files/package.json#engines)


developit:
Hmm - @ivantm what did you run to get that error?  The build and tests should run on Node 4+.

developit:
ahhh - I think it's one of the devDependencies.

developit:
@ivantm I'm stalled on this because you're absolutely right, developing preact requires Node 5+ (or even 6+), but specifying `node>=6` in the `engines` field will actually show a warning to everyone who installs preact using node 4. That is a little awkward, since the built library doesn't care about node versions at all.  I might be wrong about how it works though, just hard to verify without actually publishing the package since `npm link` doesn't work between nvm envs.

robertknight:
https://github.com/npm/npm/issues/9765#issuecomment-200430838 describes how Facebook handle this in the React repo:

1. Add a `devEngines` field to package.json
2. Add a post-install hook which checks the Node version in development environments

developit:
@robertknight awesome, that seems to be what we want here!

robertknight:
I can implement the suggestion above and it doesn't sound like it would be a whole lot of work to do so, but given that Node v6 is now the LTS release, how important is it that we continue to support older versions?

developit:
Personally I've stopped assuming anything works with node 4. It ships with npm 2, which now send to be less and less supported too.

.nvmrc
```diff
@@ -0,0 +1 @@
+6
```
robertknight: Does this mean "use at least Node v6" or "use exactly Node v6"?

robertknight:
Given that Node v6 has been the LTS release for some time and we don't do any regular testing with Node v4 or npm v2 I would suggest that we just make it a requirement for preact by merging this PR. I had a question over the behavior of the nvmrc file though.

.nvmrc
developit: Not sure.

ivantm: It means any version of Node 6 (e.g. 6.1.0 or 6.4.0, but not 7.0.0). Unfortunately nvm does not appear to support a value like `>= 6.0.0` 😢 

I'm happy to amend this PR to remove the .nvmrc file, since not everyone uses nvm, and the `engines` field of package.json is more robust.

developit:
@robertknight wouldn't that make installation fail for users running Node 4?  I'm happy to set Node 6+ as a requirement for development in the preact repo, I just thought `engines.node` applied even when doing `npm install preact`. Since that's really just putting a few files into `node_modules`, it doesn't depend on any version of node.

ivantm:
> wouldn't that make installation fail for users running Node 4

if people install preact with node 4 and the engines field set to ">= 6.0.0" then npm will just warn [as per the package.json docs](https://docs.npmjs.com/files/package.json#engines)

> this field is advisory only

developit:
Right, just it seems odd since anyone installing the package gets the warning even though it isn't a requirement for Preact itself, but for those developing the package `git clone && npm i` doesn't warn, right?

kumarabhishek:
I would vote for moving to node 6.x and soon to 8.x as in October node LTS is scheduled and in April 2018 Node 4.x is hitting end-of-life. At least, after October we should stop dead trails of node 4.x. Refer: https://github.com/nodejs/LTS#lts-schedule1

PR #291, opened 2016-09-01
Install missing dependency for npm t

ivantm:
When running `npm t`, I got the following error:

```
Error loading module mocha!
You have enabled diff output. That only works with karma-mocha and mocha installed!
Run the following command in your command line:
  npm install karma-mocha mocha diff
```

So I ran 

`npm install --save-dev karma-mocha mocha diff`

and it turns out `diff` was missing from package.json.


developit:
Great - thanks!

PR #292, opened 2016-09-01
Ensure node data for refs on pre-rendered nodes with no attributes

ivantm:
I've come across an exception in diff.js where the node does not have the `[ATTR_KEY]` property, leading it to throw on this block of code near the end of the `idiff` function

```js
if (originalAttributes && originalAttributes.ref) {
    (out[ATTR_KEY].ref = originalAttributes.ref)(out);
}
```

The node may not have had `ensureNodeData` called on it because:
a) it has no attributes
b) the DOM node it's diffing against has been pre-rendered (and thus doesn't have an existing `[ATTR_KEY]` property)
c) there is an attempt to get a ref on the component for whom the node represents the root node

Here is the failing test case (a part of this PR):

```js
it('should add refs to components representing DOM nodes with no attributes if they have been pre-rendered', () => {
	// Simulate pre-render
	const parent = document.createElement('div');
	const child = document.createElement('div');
	parent.appendChild(child);
	scratch.appendChild(parent); // scratch contains: <div><div></div></div>

	let ref = spy('ref');

	function Wrapper() {
		return <div></div>;
	}

	render(<div><Wrapper ref={ref} /></div>, scratch, scratch.firstChild);
	expect(ref).to.have.been.calledOnce.and.calledWith(scratch.firstChild.firstChild);
});
```

But it passes if I make a small change by adding at least one attribute to the element:

```js
function Wrapper() {
	return <div class=""></div>;
}
```

The test case is slightly convoluted because it does not work if the node with the ref is the root node in the `render` function (hence the artificial component structure)

I'm not 100% sure the solution proposed in this PR is the best one, or if there is a more elegant way of describing the test case, but this PR fixes the issue.

### DIFF
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -3,8 +3,8 @@ import options from '../options';
 import { toLowerCase, empty, falsey, isString, isFunction } from '../util';
 
 
-export function ensureNodeData(node, data) {
-	return node[ATTR_KEY] || (node[ATTR_KEY] = (data || {}));
+export function ensureNodeData(node, data={}) {
+	return node[ATTR_KEY] || (node[ATTR_KEY] = data);
 }
 
 
@@ -31,7 +31,7 @@ export function removeNode(node) {
  *	@private
  */
 export function setAccessor(node, name, value, old, isSvg) {
-	ensureNodeData(node)[name] = value;
+	node[ATTR_KEY][name] = value;
 
 	if (name==='key' || name==='children' || name==='innerHTML') return;
 
--- a/src/vdom/diff.js
+++ b/src/vdom/diff.js
@@ -237,9 +237,8 @@ export function recollectNodeTree(node, unmountOnly) {
 
 /** Apply differences in attributes from a VNode to the given DOM Node. */
 function diffAttributes(dom, attrs) {
-	ensureNodeData(dom);
-
 	const old = dom[ATTR_KEY] || getRawNodeAttributes(dom);
+	ensureNodeData(dom);
 
 	// removeAttributes(dom, old, attrs || EMPTY);
 	for (let name in old) {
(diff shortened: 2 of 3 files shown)


src/dom/index.js
```diff
@@ -31,7 +31,7 @@ export function removeNode(node) {
  *	@private
  */
 export function setAccessor(node, name, value, old, isSvg) {
-	ensureNodeData(node)[name] = value;
+	node[ATTR_KEY][name] = value;
```
ivantm: Only `diffAttributes` calls `setAccessor` (in 2 different cases) so I moved the `ensureNodeData` call into `diffAttributes` so that it only gets called once.

`idiff` always calls `diffAttributes`, which means `ensureNodeData` will always be called.

src/vdom/diff.js
```diff
@@ -237,7 +237,9 @@ export function recollectNodeTree(node, unmountOnly) {
 
 /** Apply differences in attributes from a VNode to the given DOM Node. */
 function diffAttributes(dom, attrs) {
-	let old = dom[ATTR_KEY] || getRawNodeAttributes(dom);
+	ensureNodeData(dom);
```
developit: I think this will cause `setAttribute` to be carried on already-correct attributes.

ivantm: Good point - missed that. I've swapped the order of those 2 lines.

developit:
Will take another look at this tomorrow!

developit:
Wondering if we can further simplify this by removing the return value of `ensureNodeData()` and placing a call to it at the top of `idiff()`.  I'll probably merge this and play with it post-merge if that's alright.

developit:
Hi @ivantm - I ended up taking some of your suggestions here as part of the 6.2 release. However, I'd still like to merge your tests. I'll close this PR out once I've done so. Thanks for the help!

developit:
Merged! I changed the author in my merge commit so the tests will get properly attributed to you. Thanks!!

PR #294, opened 2016-09-04
allow input's list to pass through as an attribute

### LINKED ISSUE #287: input + datalist don't work
This is rather obscure, but I was trying to get the new `<datalist>` element working with `<input type="range">` today. For some reason, Preact isn't picking up the `list="..."` element on the input.

Here's a working (non-working) example: http://jsfiddle.net/4apL8a8t/

Here's what it should look like: http://stackoverflow.com/questions/26612700/ticks-for-type-range-html-input

FWIW, it looks like it doesn't work in React either: https://github.com/facebook/react/issues/6717

vutran:
This PR should fix #287 so `list` can be passed through as an attribute since this specific DOM attribute becomes normalized as a ref to the HTMLElement itself.

### DIFF
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -65,7 +65,7 @@ export function setAccessor(node, name, value, old, isSvg) {
 		}
 		l[name] = value;
 	}
-	else if (name!=='type' && !isSvg && name in node) {
+	else if (name!=='list' && name!=='type' && !isSvg && name in node) {
 		setProperty(node, name, empty(value) ? '' : value);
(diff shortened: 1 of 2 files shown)


developit:
@vutran ahh that's sweet, I guess using an attribute makes the insertion order not matter!

developit:
@vutran Hmm - I'm not sure this fixes the issue, at least in PhantomJS.  If I add an assertion to check that `input.list` points to the `<datalist>` element, it fails.  Also it seems like the `.list` DOM property already does that normalization?  Maybe earlier though.

PR #313, opened 2016-09-13
Add support for single level of array children in h()

mhink:
Should be pretty straightforward: if any of the "...children" arguments passed to h() is an Array, flatten them once and iterate over the resulting array, rather than the raw arguments list. The 'core' algorithm doesn't change, other than referencing the new array.

Note that this commit _does_ introduce two extra Array allocations: once for `toArray` and again in `flattenOnce` (which returns a new Array which is the 'flattened' version of its parameter). If this is unacceptable, let me know and I'll keep tinkering with it.

### DIFF
--- a/src/h.js
+++ b/src/h.js
@@ -1,6 +1,6 @@
 import { VNode } from './vnode';
 import options from './options';
-import { falsey, isFunction, isString, hashToClassName } from './util';
+import { falsey, isFunction, isString, hashToClassName, toArray, flattenOnce } from './util';
 
 
 const SHARED_TEMP_ARRAY = [];
@@ -18,8 +18,10 @@ export function h(nodeName, attributes, firstChild) {
 	let len = arguments.length,
 		children, arr, lastSimple;
 
+	const flatChildren = flattenOnce(toArray(arguments, 2));
 
 	if (len>2) {
+		let firstChild = flatChildren[0];
 		let type = typeof firstChild;
 		if (len===3 && type!=='object' && type!=='function') {
 			if (!falsey(firstChild)) {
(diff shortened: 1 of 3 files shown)


developit:
Looks good, just a few things we'll need to tweak.  Generally for single-use functions it's both smaller and faster (in terms of runtime performance) to inline.  Also as you mentioned, it'd be awesome to avoid the additional Array allocation.  Last one is a bit weird, but Preact's build process actually inlines constants, so variables are better declared using `let` when they are locals.

I am always happy to do this stuff post-merge if you want. I tend to tweak for size and perf endlessly anyway. The tests really help, makes it easier to tweak things without breaking the use-case :)

developit:
@mhink So this prompted me to do some performance analysis, just because I'm a huge nerd for this kind of stuff.  Here's what I found:

[results](https://esbench.com/bench/57d85f78330ab09900a1a07e)

https://esbench.com/bench/57d85f78330ab09900a1a07e

Using a loop + stack is super fast and recursion-free. "TIL" I guess!

mhink:
> Looks good, just a few things we'll need to tweak. Generally for single-use functions it's both smaller and faster (in terms of runtime performance) to inline. Also as you mentioned, it'd be awesome to avoid the additional Array allocation. Last one is a bit weird, but Preact's build process actually inlines constants, so variables are better declared using let when they are locals.

> I am always happy to do this stuff post-merge if you want. I tend to tweak for size and perf endlessly anyway. The tests really help, makes it easier to tweak things without breaking the use-case :)

I'm fine doing it now. It's not terribly much work. Thanks for the feedback!

developit:
I added you to the gitter room btw.

PR #316, opened 2016-09-15
Add missing annotations to Typescript definition file

jrf0110:
I noticed that Typescript was complaining about my use of object-style `className`

```javascript
{ className: { foo: true } }
```

This adds annotations for:

* `className: string | Object`
* `ref: (el?: Element) => void)`
* `dangerouslySetInnerHTML?:Object;`

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -4,8 +4,14 @@ declare namespace preact {
 		key?:string;
 	}
 
+	interface DangerouslySetInnerHTML {
+	 __html: string;
+	}
+
 	interface PreactHTMLAttributes {
+		dangerouslySetInnerHTML?:DangerouslySetInnerHTML;
 		key?:string;
+		ref?:(el?: Element) => void;
 	}
 
 	interface VNode {
@@ -51,8 +57,8 @@ declare namespace preact {
 		abstract render(props:PropsType & ComponentProps, state:any):JSX.Element;
 	}
 
-	function h<PropsType>(node:ComponentConstructor<PropsType, any>, params:PropsType, ...children:(JSX.Element|string)[]):JSX.Element;
-	function h(node:string, params:JSX.HTMLAttributes&JSX.SVGAttributes, ...children:(JSX.Element|string)[]):JSX.Element;
+	function h<PropsType>(node:ComponentConstructor<PropsType, any>, params:PropsType, ...children:(JSX.Element|JSX.Element[]|string)[]):JSX.Element;
+	function h(node:string, params:JSX.HTMLAttributes&JSX.SVGAttributes&{[propName: string]: any}, ...children:(JSX.Element|JSX.Element[]|string)[]):JSX.Element;
 
 	function render(node:JSX.Element, parent:Element, merge?:boolean):Element;
 
@@ -277,8 +283,8 @@ declare namespace JSX {
 		charSet?:string;
 		challenge?:string;
 		checked?:boolean;
-		class?:string;
-		className?:string;
+		class?:string | { [key:string]: boolean };
+		className?:string | { [key:string]: boolean };
 		cols?:number;
 		colSpan?:number;
 		content?:string;


src/preact.d.ts
```diff
@@ -5,7 +5,9 @@ declare namespace preact {
 	}
 
 	interface PreactHTMLAttributes {
+		dangerouslySetInnerHTML?:Object;
```
jrf0110: Actually, not sure if this should be in PreactHTMLAttributes or just HTMLAttributes

developit: I can't help you there! haha
If it's just a naming thing, doesn't matter to me.

developit:
Pinging @kruczy since he's super knowledgeable on this stuff.

src/preact.d.ts
```diff
@@ -277,8 +279,8 @@ declare namespace JSX {
 		charSet?:string;
 		challenge?:string;
 		checked?:boolean;
-		class?:string;
-		className?:string;
+		class?:string | Object;
+		className?:string | Object;
```
kruczy: would this be better `string | {[key:string]:string}`?

all values and keys will then be forced as string

src/preact.d.ts
kruczy: The effect will be the same, tho for organisational purposes it should be here.

maybe also better to define the type specifically with an interface? instead of allowing any object to be passed
this could be something like:
```
interface DangerouslySetInnerHTML {
 __html: string;
}
```

src/preact.d.ts
kruczy: read your example again, this should be then: `string | {[key:string]:string|boolean}`

developit: Not sure how the TS annotations work, but values can also be numbers.

src/preact.d.ts
developit: @kruczy That seems nice, since anything else in that object would be ignored anyway.

jrf0110:
Thanks for the feedback @kruczy. I'm pretty new to TypeScript so the help is much appreciated.

I'll get those suggestions in in a bit!

jrf0110:
Sorry it took so long to get back this. Should be good to go 👍

src/preact.d.ts
```diff
@@ -141,6 +147,7 @@ declare namespace JSX {
 		y1?:number | string;
 		y2?:number | string;
 		y?:number | string;
+		[propName: string]: any;
```
jrf0110: One more thing I added was this catchall so we can have `data-*` attributes on HTML elements

kruczy: this should not be needed
from https://www.typescriptlang.org/docs/handbook/jsx.html

```
Note: If an attribute name is not a valid JS identifier (like a data-* attribute), it is not considered to be an error if it is not found in the element attributes type.
```

src/preact.d.ts
```diff
@@ -277,8 +284,8 @@ declare namespace JSX {
 		charSet?:string;
 		challenge?:string;
 		checked?:boolean;
-		class?:string;
-		className?:string;
+		class?:string | Object;
+		className?:string | { [key:string]: string | boolean | number };
```
kruczy: now it occurred to me that you want an object for className, does that actually work?

maybe you wanted to use `style`?

jrf0110: Preact allows classnames to specified as

```javascript
className: {
  'some-class-that-will-be-showed': true,
  'some-class-that-will-NOT-be-showed': false,
}
```

src/preact.d.ts
jrf0110: Hrmm yeah, I added it because of this error I saw:

> Object literal may only specify known properties, and ''data-canonical'' does not exist in type 'HTMLAttributes & SVGAttributes'

src/preact.d.ts
kruczy: ah, then it should only allow boolean right? string and number are unnecessary
and it should be the same for `class` and `className` right?

jrf0110: oh yeah. dang. That's my bad. That should be:

```javascript
className?: string | number | { [key:string]: boolean };
```

And I guess `class?` should be the same

jrf0110: To be honest, I've never used `className: number`, but @developit says it can be number so it can be number :)

src/preact.d.ts
kruczy: I would guess this is because you are calling `h` directly and not using jsx? if that is the case it would be better to add it in definition of `h` than here

jrf0110: Good lord. I just realized I put that under SVGAttributes. I dunno what happened there. I guess I was just yolo-pasting.

It is probably better suited in `HTMLAttributes`

src/preact.d.ts
developit: FWIW any value in an attribute or classname object value that preact treats as Boolean will be implicitly cast, so numbers and strings are fair game there. That said, Boolean is the meaning, regardless of the type given.

developit:
You guys ready for me to merge this? @kruczy @jrf0110 
Looks good to me but I'm not much help for TS.

src/preact.d.ts
kruczy: then I think it should be only boolean, there is no reason to pass a string or number there

developit: Agreed, as far as typings are concerned anything else may just be misleading.

jrf0110: Okay final type for `className?` and `class?` will be

```typescript
// Supports className: 'foo bar'
// Supports className: { foo: true, bar: true }
className?: string | { [key:string]: boolean };
```

developit: @jrf0110 that looks good to me.

jrf0110:
👍  Everything is solid for me.

kruczy:
i would still move `[propName: string]: any;` out into a separate interface and only use id in the definition of `h` as it actually is not really valid for `HTMLAttributes`

jrf0110:
@kruczy I'm not sure I totally follow. Do you mean adding `[propName: string]: any;` to `ComponentProps`?

src/preact.d.ts
```diff
@@ -51,8 +57,8 @@ declare namespace preact {
 		abstract render(props:PropsType & ComponentProps, state:any):JSX.Element;
 	}
 
-	function h<PropsType>(node:ComponentConstructor<PropsType, any>, params:PropsType, ...children:(JSX.Element|string)[]):JSX.Element;
-	function h(node:string, params:JSX.HTMLAttributes&JSX.SVGAttributes, ...children:(JSX.Element|string)[]):JSX.Element;
+	function h<PropsType>(node:ComponentConstructor<PropsType, any>, params:PropsType, ...children:(JSX.Element|JSX.Element[]|string)[]):JSX.Element;
+	function h(node:string, params:JSX.HTMLAttributes&JSX.SVGAttributes, ...children:(JSX.Element|JSX.Element[]|string)[]):JSX.Element;
```
kruczy: cool, this makes sense :+1:

maybe add `&{[propName: string]: any}` after `params:JSX.HTMLAttributes&JSX.SVGAttributes`? And remove it from `HtmlAttributes`?

jrf0110: Gotcha lemme try that :)

jrf0110: Awesome. Worked for me. I wasn't super familiar with intersection types when I first started, but that totally makes sense

developit:
Is this ready to drop into `6.1.1`? :)

jrf0110:
I think so. I'm ready if @kruczy is ready

kruczy:
sorry for delay, looks good now :+1:

PR #329, opened 2016-09-22
fix instanceof not pointing to the right function

matthewmueller:
This was pretty brutal to track down, but I was running into a situation where the dev version works, but the production version doesn't (vnode's get stringified into `[Object object]`).

While I'm not sure exactly the state of my build to produce that, I'm assuming that the VNode was created inside another instance of Preact so `child instanceof VNode` is false. Here's the build if you'd like to have a look: https://gist.github.com/matthewmueller/bca6d225cd15c95a61939c77f1bb8648

This is an example I used to recreate this issue:

```js
const { HTML, render } = require('vcom')
const { h } = require('preact')

function app () {
  // return another
  return HTML.div(h('div', {}, ['hi there!']))
}

render(app(), document.body)
```

### DIFF
--- a/src/h.js
+++ b/src/h.js
@@ -32,7 +32,7 @@ export default function h(nodeName, attributes) {
 			}
 			for (let j=0; j<arr.length; j++) {
 				let child = arr[j],
-					simple = !empty(child) && !(child instanceof VNode);
+					simple = !empty(child) && !(child.nodeName);
 				if (simple) child = String(child);
 				if (simple && lastSimple) {
 					children[children.length-1] += child;


developit:
This happens when you have a build containing two copies of Preact.

matthewmueller:
ah yah the browserify build included another one. yah, i think this is a little too easy to accidentally do when you have multiple dependencies that require preact.

developit:
For context, React uses a Symbol (falling back to a string) to identify, which lets them use POJO's instead of classes. I'm actually already looking into that, so I might go that route rather than duck-typing.  Though honestly yours might be faster, so I'll be sure to check first.

matthewmueller:
ah okay – i think the check is fine unless there could be some other objects passed through with a nodeName property, but that's totally your call.

i'm also totally fine with it saying: "you have more than 1 preact, please remove one and try again."

but i actually just ran into this again with this case: running npm@2 will likely run into this problem unless all the preact versions in all dependencies are the same. i had to use the step debugger again to realize what was happening because it's non-obvious from the error.

developit:
For sure, and then it gets weirder when you throw a module bundler in there that may or may not de-dupe deps, or might look at their version and selectively de-dupe, etc etc.  Definitely a problem worth solving.  I might even just use your check and add in the other two properties, will depend on the performance implications.

matthewmueller:
awesome, yah i think that's the right call. thinking about it some more there are actually two different stages:

1) dev time: just work, i don't care about build size. preact should allow multiple instances of itself to work
2) release time: figure out why there are 2+ preact in my project and optimize my modules at that point.

developit:
Agreed. I believe there may be other issues with running 2 copies of preact and swapping nodes between the two though, since both would be fighting to recycle DOM elements and components separately.

matthewmueller:
Ah okay that makes sense, so this could just be the tip of the iceberg. One other thing you could do is break out the `{ h }` and perhaps the `render()` function, seems like most plugins use `h` and sometimes `render()`.

developit:
Indeed. Also the current solution is more difficult for bundlers to prune when performing tree shaking.  Duck typing like this PR implements keeps things more separable.

developit:
@matthewmueller I figured out a way to do this by inverting that check, entirely removing the need for typing of any kind :)

matthewmueller:
woah awesome! can this be closed then?

developit:
I was going to mark it as closed from the commit but forgot to, so sure. Usually I go around and close things out when I cut a release just to avoid duplicate issues coming in.

developit:
Fixed in 6.2!

matthewmueller:
thanks! ✨

PR #334, opened 2016-09-29
regression:  serializing class={{ }} objects for SVG tags

robbiewxyz:
e7c6ab6 broke serializing `class={{ }}` for SVG elements.  This fixes
that regression and does a little cleans up the logic a little bit.

I've also added a test verifying the proper behavior for SVG.

### DIFF
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -28,11 +28,15 @@ export function setAccessor(node, name, value, old, isSvg) {
 
 	if (name==='className') name = 'class';
 
+	if (name==='class') {
+		value = value && typeof value==='object' ? hashToClassName(value) : value;
+	}
+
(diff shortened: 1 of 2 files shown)


developit:
I believe there is already a commit on master for this: 2de530b9d3329fe3fcd32816b143a3f8fff1d43f
I might keep that commit but merge the test - is that alright?  The fix on master accounts for `className` being an object.

robbiewxyz:
The test currently fails on master because `hashToClassName()` is only run if `isSvg == false` (See [`dom/index` lines 34-36](https://github.com/developit/preact/blob/master/src/dom/index.js#L34)).

So `<circle class="red"/>` works, but `<circle class={{ red: true }}/>` fails.  This commit makes both notations work with HTML or SVG elements.

developit:
oooh dang, good catch.

PR #339, opened 2016-10-02
Add React Developer Tools integration

### LINKED ISSUE #320: compatibility with React developer tools
React provides a convenient [chrome extension](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi/related?hl=en) as a debugger tool.

Preact doesn't seem to work with it. So, adding compatibility - is it in the road map?

robertknight:
This PR adds a module which provides support for using Preact with the React Developer Tools by emulating the renderer interface that React DOM exposes to the developer tools via `__REACT_DEVTOOLS_GLOBAL_HOOK__.inject(...)`

I have done a moderate amount of testing with the Chrome extension and the `preact-www` project, along with a few others. It could use additional testing with more complex projects.

Enabling the devtools integration is currently done by invoking the `initDevTools()` function:

```js
import { initDevTools } from 'preact';

...

// Call this before first render() call
initDevTools();
```

preact-devtools-2

Fixes #320

The implementation is in a separate module and only a few small hooks needed to be added to the Preact core, but the functionality is not yet exposed as a separate entry point. I could use a couple of pointers on how best to do that.


PaulBGD:
You should add a TypeScript definition, otherwise +1 from me.

developit:
@robertknight Sorry for just sitting on this for so long! I've been pondering how to get this in without bundling the shim into the prod build of Preact.  Might be worth using dev/prod builds, or providing the hooks but then splitting devtools into its own submodule:

```js
import { initDevTools } from 'preact/devtools';

initDevTools();
```

PaulBGD:
@developit for anyone who uses rollup or the new version of webpack, the devtools code will be removed anyways.

robertknight:
> @developit for anyone who uses rollup or the new version of webpack, the devtools
> code will be removed anyways.

I think it would be best not to rely on tree shaking for this. Anyone using Browserify or older versions of Webpack would end up with unnecessary code in their bundles.

> Might be worth using dev/prod builds, or providing the hooks but then splitting devtools into its own submodule:

Providing the hooks but moving the devtools into a submodule which is not included when importing preact's main exports via `require("preact")` (or the equivalent ES6 import) sounds good to me. So how about I:

1. Include the `afterMount`, `afterUpdate`, `beforeUnmount` hooks in the prod build?
2. Move `devtools.js` to `devtools/index.js` and remove the `initDevTools` re-export from `src/preact.js`?

developit:
Yup, that seems very reasonable. I'm super curious to see what the file size hit is for those three hooks. If it's too much, there might be a way to use a single VNode Hook to inject the others 🕵

robertknight:
I've moved the devtools integration/shim to a separate bundle, required via `preact/devtools`, so it can be loaded in a similar way to the `aliases` bundle.

The gzipped-size hit for the extra hooks is 0.04KB:

master: gzip size: 3843 / 3.84 kB
devtools branch: gzip size: 3880 / 3.88 kB

developit:
Awesome, that much less of a hit than I thought we'd be taking!

src/devtools.js
```diff
@@ -0,0 +1,434 @@
...
+
+/**
+ * Find all root component instances rendered by preact in `node`'s children
+ * and add them to the `roots` map.
+ *
+ * @param {DOMElement} node
+ * @param {[key: string] => ReactDOMComponent|ReactCompositeComponent}
+ */
+function findRoots(node, roots) {
+	Array.from(node.childNodes).forEach(child => {
```
robertknight: Ah, cool. I didn't know that :) . This is _super_-new in Firefox though (>=50) according to https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach so perhaps best not to rely on it for the moment?

src/devtools.js
```diff
@@ -0,0 +1,434 @@
...
+	const name = vnode.nodeName.name || '(Function.name missing)';
+	const wrappers = functionalComponentWrappers;
+	if (!wrappers.has(originalRender)) {
+		const wrapper = class extends Component {
+			render(props, state, context) {
+				return originalRender(props, context);
+			}
+		};
+		// Function.name is non-writable, but we can replace it using
+		// defineProperty() instead in most browsers.
```
robertknight: 👍 - Just testing that out

src/devtools.js
```diff
@@ -0,0 +1,434 @@
...
+ *
+ * See https://github.com/facebook/react-devtools/blob/e31ec5825342eda570acfc9bcb43a44258fceb28/backend/getData.js
+ */
+function createReactCompositeComponent(component) {
+	const _currentElement = createReactElement(component);
+	const node = component.base;
+
+	let instance = {
+		// --- ReactDOMComponent properties
+		getName() {
```
robertknight: Do you mean mixing in these getters to the Component prototype? That might work. You'd still need to create `ReactDOMComponent`-like instances for DOM elements which don't have a preact component though.

src/devtools.js
```diff
@@ -0,0 +1,434 @@
...
+function createReactDOMComponent(node) {
+	const childNodes = node.nodeType === Node.ELEMENT_NODE ?
+		Array.from(node.childNodes) : [];
+
+	const isText = node.nodeType === Node.TEXT_NODE;
+
+	return {
+		// --- ReactDOMComponent interface
+		_currentElement: isText ? node.textContent : {
+			type: node.nodeName.toLowerCase(),
```
robertknight: Ah, I didn't know about [.localName](https://developer.mozilla.org/en-US/docs/Web/API/Element/localName). That's only been introduced in Firefox very recently, so I'd be inclined to stick with `toLowerCase()` for the moment and avoid the hassle of polyfilling this. Preserving the case is not that important, it just looks nicer in the devtools if the tag names don't get upper-cased.

src/devtools.js
robertknight: I've done a little experimentation and it looks like this would be possible and might remove the need for the internal map between Preact Component instances / DOM Nodes and their corresponding React*Component-like objects. It'll need some more thought and checking that none of the other properties of Preact components are going to cause problems in the dev tools. I would be inclined to get the current implementation merged first and then revisit this afterwards.

src/devtools.js
```diff
@@ -0,0 +1,434 @@
...
+		forceUpdate: component.forceUpdate.bind(component),
+		setState: component.setState.bind(component),
+
+		// --- Additional properties used by preact devtools
+		node
+	};
+
+	// React DevTools exposes the `_instance` field of the selected
+	// item in the component tree as `$r` in the console
+	instance._instance = instance;
```
robertknight: Done. As long as `_instance` is compatible with the React `Component` API (props, state, setState, forceUpdate), this works.

robertknight:
OK, status update:

- The developer tools are now built as a separate UMD bundle, similar to `aliases.js` except that the core Preact code is imported rather than duplicated.
- In a CJS/ES6 environment, you can do this:

```js
import { initDevTools } from 'preact/devtools'
initDevTools();
```

In a tooling-free environment, you should be able to do this:

```html
<script src="https://unpkg.com/preact/dist/preact.min.js"></script>
<script src="https://unpkg.com/preact/devtools.js"></script>
<script>
preactDevTools.initDevTools();
</script>
```

PaulBGD:
This is such an awesome release to celebrate 4,000 stars! Great job on it @robertknight!
P.S. would it be better if I opened a PR for a TypeScript definition?

Issue #343, opened 2016-10-08
Build fails on Windows


Download:
``` sh
c:\ws\preact> npm install

...

> preact@6.2.1 prepublish c:\ws\preact
> npm run build


> preact@6.2.1 build c:\ws\preact
> npm-run-all --silent clean transpile copy-flow-definition copy-typescript-definition strip optimize minify size

'cp' is not recognized as an internal or external command,
operable program or batch file.
ERROR: "copy-flow-definition" exited with 1.

npm ERR! Windows_NT 10.0.14393
npm ERR! argv "C:\\nodejs\\v6.3.1\\node.exe" "C:\\Users\\...\\AppData\\Roaming\\npm\\node_modules\\npm\\bin\\npm-cli.js" "run" "build"
npm ERR! node v6.3.1
npm ERR! npm  v3.10.7
npm ERR! code ELIFECYCLE
npm ERR! preact@6.2.1 build: `npm-run-all --silent clean transpile copy-flow-definition copy-typescript-definition strip optimize minify size`
npm ERR! Exit status 1
```

Also, I see files getting created named `$npm_package_aliases_main.js` etc and I guess that's wrong  :)

I will create a PR to see if we can make this build cross-platform together. I will need your help though as I don't yet fully understand how the build works.

developit:
Ack! Sounds good, I think it will just be the `npm-scripts` that need to be adjusted.  Might just be a matter of adding [cross-env](https://npm.im/cross-env) as a prefix?

Download:
I found this lib that will fix the cp issue I think:
https://www.npmjs.com/package/copy

But I'm really fuzzy on the `$npm_package_*` identifiers you are using. Do you have any pointers to info on these?

Download:
I ended up using [copyfiles](https://npmjs.com/package/copyfiles) instead. It seems to work. The build is progressing further now, but I'm stuck on those `$npm_package_*` expressions. They don't work under Windows... 

```sh
c:\ws\preact> npm install

> preact@6.2.1 prepublish c:\ws\preact
> npm run build


> preact@6.2.1 build c:\ws\preact
> npm-run-all --silent clean transpile copy-flow-definition copy-typescript-definition strip optimize minify size

Skipping path dist/preact.dev.js which does not exist.
No files selected, nothing to do.
Skipping path dist/preact.dev.js which does not exist.
No files selected, nothing to do.
fs.js:640
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
[...]

PR #344, opened 2016-10-08
Fixes build for Windows #343

Download:
- Added dev dependency on [copyfiles](https://npmjs.com/package/copyfiles)
- Added dev dependency on [pkgcfg](https://npmjs.com/package/pkgcfg)
- Changed `size` script to use pipes
- Replaced all occurrences of `$npm_package_*` with the literal value, except for `$npm_package_version` in the `release` script
- Moved script `release` to `smart-release`
- Added script `release`, which only calls `run smart-release`, invoking `run` from `pkgcfg` so it can substitute expressions

I replaced all occurances of `$npm_package_*` with the literal value, because I don't know how (if at all possible) to get them to work on Windows. These are pretty static anyway. It's not DRY but I hope we can live with it. 

There was one instance that I could not replace though; the `release` script uses the package version and hardcoding that would be very error-prone. So for that one, I included `pkgcfg` which can replace variables much like those dollar expressions, but does work on Windows. Disclosure: I wrote that package. I hope you can live with that. Or we have to find a way to make the dollar vars work on Windows somehow.

Let me know what you think (and whether this still works correctly for you as well).


Download:
@developit  Are there things that need to be changed in order to get this merged?

developit:
Nope, it's on my todo list. I wish there was a way to have `npm run` inject the variables manually :(

Download:
@developit  Is it the dependency on pkgcfg? I've been using it myself for a while now and it does seem to be stable. Another option could be to externalize the work of the `release` script to an external .js file that has easy access to the environment.

developit:
@Download I was thinking of maybe switching to [np](https://github.com/sindresorhus/np) instead of the `release` npm script. It looks really nice and does more or less the same thing.

Download:
`np` looks nice indeed!

> My hesitation was just that I don't have any way to test on Windows 

Hah! I know the feeling, but consider this: I *am* on Windows. The build is currently broken. This fixes it on the Windows end (and looking at the Travis results, still builds on *nix as well). So AFAIC (and most other Windows users I guess), there isn't much to loose on our end.  😃 

Maybe you can implement only those parts of the fixes you are comfortable with, making sure everything still works on your end, and I will verify the build on Windows and suggest additional fixes if needed.

Alternatively, I could take a stab at implementing `np`. I could practice the publish stuff on my own fork and let you know once I think I have got it working. Let me know.

Together we can make this work!

developit:
Agreed - I'm good to just merge this since you're able to get a build working with these changes. Will try to find some time tonight (hopefully!).

Download:
Cool! Let me know if you need help. I'll be online for a few more hours.

developit:
@Download ended up having to swap pkgcfg for cross-env to do the release script (but I kept your release/smart-release idea) because on OSX it seems to inject multiple values (wasn't able to see the exact issue, but I kept getting errors when it was invoking `git commit -am`).

PR #350, opened 2016-10-10
Added more missing annotations to Typescript definition file

arusanov:
1. Changed `ref` argument to `Component`
2. Added `defaultValue` for inputs
3. Added `context` property to `Component`

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -12,7 +12,7 @@ declare namespace preact {
 		dangerouslySetInnerHTML?:DangerouslySetInnerHTML;
 		key?:string;
 		defaultValue?:string;
-		ref?:<T,S>(component:Component<T,S>) => void;
+		ref?:<T,S>(el:Component<T,S>|Element)=>void;
 	}
 
 	interface VNode {


src/preact.d.ts
```diff
@@ -11,7 +11,8 @@ declare namespace preact {
 	interface PreactHTMLAttributes {
 		dangerouslySetInnerHTML?:DangerouslySetInnerHTML;
 		key?:string;
-		ref?:(el?: Element) => void;
+		defaultValue?:string;
+		ref?:<T,S>(component:Component<T,S>) => void;
```
developit: `ref` can receive both components and elements.

arusanov: fixed in https://github.com/developit/preact/pull/350/commits/791cfd4c0195c5cce872fa235da8386054b0db7e

developit:
LGTM but I've basically never used TypeScript.
Maybe @jrf0110 or @kruczy can confirm 😀

src/preact.d.ts
```diff
@@ -11,7 +11,8 @@ declare namespace preact {
 	interface PreactHTMLAttributes {
 		dangerouslySetInnerHTML?:DangerouslySetInnerHTML;
 		key?:string;
-		ref?:(el?: Element) => void;
+		defaultValue?:string;
```
kruczy: should this be in `DOMAttributed`?

src/preact.d.ts
```diff
@@ -11,7 +11,8 @@ declare namespace preact {
...
+		ref?:<T,S>(el:Component<T,S>|Element)=>void;
```
developit: Ah, no you're right - sorry I missed that! Indeed, an Element VNode will only ever invoke its ref with either an `Element` or `null`.

src/preact.d.ts
nojvek: Could someone answer why its called DomAttributed rather than DomAttributes? Its kind of confusing on the first look.

geppy: This typo was resolved in #402 (correctly, I think, based on usage).

Issue #351, opened 2016-10-10
clear `src` attribute on recycled `<img>` nodes


robbiewxyz:
Here's [a codepen](http://codepen.io/robbiewxyz/pen/xEAzLB) with the problem:  when the `src` attribute of an image is changed, the old picture will keep showing while the new picture is loading.  This behavior persists even if the image is removed from the DOM.  When combined with Preact's node-caching, this can get crazy, with old pictures showing up again in unrelated places.

Clearing the image with `.removeAttribute('src')` should solve the issue.  I'll plan on putting together a pull request here shorty.

developit:
Seems like there are two identical issues. Want to delete one of them? 😛

robbiewxyz:
Oh, oops.  I was on a bad internet connection.  Can I actually delete an issue or just close it?

mikestead:
@developit I'm seeing something similar with the latest Preact (8.1.0). If I comment out the contents of Preact's `collectComponent` function it goes away.

The scenario we have is a list of products, you click one and go to the detail page with an image header (`picture` element), click back, then click a new product. When you hit the product page the second time it shows the old product image for a split second. If we switch back to React all is ok again.

I've been trying for the past hour to create a smaller codepen example but no luck recreating it, and ran out of time for now I'm afraid.

developit:
Preact still recycles components even after we removed DOM recycling. #664 proposes removing this, and I have [a branch](https://github.com/developit/preact/compare/scratch/perf) that tries to regain the optimization lost by removing the component recycler.

That said, if you're seeing the image for a really short time, that might be related to the async change in Preact 8, which has been [reverted to the 7.x behavior](https://github.com/developit/preact/commit/576dda310a8e31e524fa2ecc803d6d406597c41e) and will be in the next release.

mikestead:
Thanks for the details, and all the hard work on Preact.

I tried reverting to pre 8.x versions and the issue still seems to be present. I then tried resetting the src element on img elements in `collectComponent` and this seemed to mostly work so I assume it's the recycling issue. I'll subscribe to #664 and keep an eye on progress 👍

jackmoore:
I've also been running into this problem with old images showing while new images load, which is not a trivial amount of time. The old images may be visible for seconds.

I tried using `.removeAttribute('src')` on `componentWillUnmount`, which did work to start each image on a clean slate but there is some sort of race condition there because sometimes the new image would end up without an src attribute. It was unworkable.

[...]

PR #353, opened 2016-10-10
clear src= on recycled <img>s

robbiewxyz:
Solves #351

### DIFF
--- a/src/dom/recycler.js
+++ b/src/dom/recycler.js
@@ -1,3 +1,4 @@
+import { ATTR_KEY } from '../constants';
 import { toLowerCase } from '../util';
 import { removeNode } from './index';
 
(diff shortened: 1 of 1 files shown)


developit:
Wondering if there is a way to make this target Firefox since that's the only browser affected by the src reuse issue.

robbiewxyz:
It looks like it affects my Chrome as well.  You can try it out with [my old codepen](http://codepen.io/robbiewxyz/pen/xEAzLB).

developit:
Ah okay, didn't realize it was in both.

robbiewxyz:
Hey, just bumping this again:  is there a conflict with merging in this bugfix?

developit:
Nope, just on vacation. On my flight home actually!

PR #355, opened 2016-10-11
Correct "preact/devtools" type definitions

robertknight:
The `initDevTools()` function was removed in favor of just initializing
the devtools integration when the "preact/devtools" module is imported.


developit:
Good catch.

PR #359, opened 2016-10-13
Rethrow errors in a separate tick

### LINKED ISSUE #357: Errors thrown in `render()` swallowed
Since `Promise.resolve().then(...)` is used for deferring to another tick, any errors thrown inside `render () { ... }` gets silently thrown away. At least in Safari TP 15, I think that some other browsers reports unhandled rejections.

This is very unintuitive when developing since there is no way to see what went wrong.

LinusU:
Fixes #357

I have tested this in Safari TP 15 and it works great. Open to any other strategy for solving it though :)


LinusU:
Actually, it seems like Safari isn't keep the stack trace, but instead generating a new one pointing to the `throw` 😞

PR #362, opened 2016-10-14
Disable React Developer Tools integration tests under IE

robertknight:
Disable React Developer Tools integration tests under IE

Developer tools integration is currently not supported in Internet
Explorer. The immediate issue is that Function.name is not supported,
though there may be other issues as well.


developit:
Saucelabs just being flakey (environment var fix worked though)

PR #375, opened 2016-10-28
clear src= on recycled <img>s

robbiewxyz:
This solves #351.  This will resets the `src` attribute to blank (`''`).  This is because most browsers will keeps displaying old image until the new one has loaded.

I've tested with a few other media-related attributes:  `<video src=>`, `<audio src=>`, and `<object data=>`.  None of them appear to care about previous attribute values.  I have not tested the SVG `<image xlink:href=>` attribute, it might also need a similar workaround.

I've closed my old PR, #353, in order to clean things up a bit with git.

### DIFF
--- a/src/dom/recycler.js
+++ b/src/dom/recycler.js
@@ -1,3 +1,4 @@
+import { ATTR_KEY } from '../constants';
 import { toLowerCase } from '../util';
 import { removeNode } from './index';
 
@@ -9,6 +10,10 @@ export function collectNode(node) {
 	removeNode(node);
 
 	if (node instanceof Element) {
+		if (node.nodeName === 'IMG') {
+			node.src = node[ATTR_KEY].src = '';
+		}
+
 		node._component = node._componentConstructor = null;
 
 		let name = node.normalizedNodeName || toLowerCase(node.nodeName);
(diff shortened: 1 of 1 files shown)


developit:
@robbiewxyz thanks for rebasing, this is next on my list 😇

robbiewxyz:
Okay, so `<image xlink:href=""/>` also needs this fix.  See [codepen.io/robbiewxyz/pen/ObNbVp](http://codepen.io/robbiewxyz/pen/ObNbVp?editors=0010).  I'll be pushing a fix for this shortly.

developit:
Good to know. I'm actually investigating dropping element recycling, will let you know if that looks likely since it would circumvent these issues.

robbiewxyz:
Hmm... I'd be interested to know what recycling is doing for the DOM performance.  Have you played around with any benchmarks yet?

developit:
Only minimally, not sure if I will try to pull thst into v7 or if its too high risk.

robbiewxyz:
Could we merge this fix as a patch release in the meantime?

developit:
I can definitely merge this into 7.x

developit:
Reason I was waiting was because this is going to end up being whack-a-mole with stateful DOM properties. Ideally we'd just drop recycling if the performance and memory consumption is good enough, then all these issues go away.

developit:
@robbiewxyz think it'd be alright to just pull in the `img` case for the patch release? (in 7.x)  Or are you needing the svg one too

robertknight:
Thank-you for the fix, but DOM Element recycling is being removed entirely in the next release of Preact (v8), which should fix this problem.

PR #386, opened 2016-11-04
Fix CI build

mkautzmann:
Hello.

I noticed that preact's Travis build is failing so I went on to investigate.

I discovered that the Sauce Labs integration is the problem.

The iPhone/iOS version that's been currently used is deprecated in the Sauce Labs platform. Besides that, the mobile browsers seems to take a lot longer to "boot" on Sauce Labs, which causes another problem:
 - The default timeout [karma-sauce-launcher](https://www.npmjs.com/package/karma-sauce-launcher) is providing is too low, causing the Sauce Labs tunnel to close before the mobile browser gets up and running.

I've seeing the same problem in my personal projects. In fact, there is an open [PR](https://github.com/karma-runner/karma-sauce-launcher/pull/99) suggesting the project to enable developers to set their own timeout of preference.

So I'm suggesting that we remove remote mobile testing for a while, until `karma-sauce-launcher` is OK to use for that purpose.

Also, I'm adding tests for two commonly used browsers: Desktop Safari and Edge.

The plan is to add iOS/Android as soon as it is possible with Karma.

I tested the changes of this PR and if you merge it, it should green the badge on README.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "karma-mocha": "^1.1.1",
     "karma-mocha-reporter": "^2.0.4",
     "karma-phantomjs-launcher": "^1.0.1",
-    "karma-sauce-launcher": "^1.0.0",
+    "karma-sauce-launcher": "^1.1.0",
     "karma-source-map-support": "^1.1.0",
     "karma-sourcemap-loader": "^0.3.6",
     "karma-webpack": "^1.7.0",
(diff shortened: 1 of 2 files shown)


developit:
Wow - thanks for digging into this!  I was getting fed up with the iphone failures, had no idea they deprecated that for OpenSauce.

Issue #387, opened 2016-11-04
Travis leaking env variables


mkautzmann:
While researching for #386 I've found out that Travis is leaking the SAUCE_USERNAME and SAUCE_ACCESS_KEY environment variables.

I figure you don't want to share the access key. So I recommend you to change the access key and follow [this guide](https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml) to secure it.

developit:
It was actually previously obscured as per that article, but it meant Pull Requests failed to connect to saucelabs. I'm thinking maybe we turn off saucelabs entirely for PRs though.

mkautzmann:
If I'm not mistaken it is already off for PRs. I'm talking about the Travis build page, where it shows the env variables in the log.

See the below screenshot:
![travis_preact_screen](https://cloud.githubusercontent.com/assets/1051644/20025814/6839dcd6-a2d9-11e6-8ca5-23e2bb0925f9.jpg)


The article I linked contains a way to secure the env variables in Travis.

developit:
In the article you linked, it says "Encrypted environment variables are not available to pull requests from forks due to the security risk of exposing such information to unknown code.".  This is what I ran into, and why I disabled the securing technique.

I think the best thing would be to re-enable the obscured access_key, and turn off saucelabs-based testing for PRs.  This would still leave phantomjs tests running, and perhaps then it'd be reasonable to look into running chrome/firefox headless tests in addition to saucelabs.

mkautzmann:
You are right, the way Travis CI handles env variables makes it hard to use them safely on PRs.

I got confused, I was focusing on commits/merges to master because the following line should already be blocking Sauce when testing PRs (assuming TRAVIS_PULL_REQUEST is passed correctly by Travis): https://github.com/developit/preact/blob/master/test/karma.conf.js#L4

In reality it is not, as we can see from [Travis' logs](https://travis-ci.org/developit/preact/builds/173047962).

I agree that Sauce should not execute on PRs, we can use secure env variables and test with Sauce only on commits to master.

robertknight:
Enable Sauce labs for PRs was something I suggested after my devtools PR broke the tests for IE.

I assume that we're using a Sauce Labs open source account for Preact? In that case, is it really a problem if the account key is exposed to PRs?

developit:
[...]

PR #390, opened 2016-11-07
Run only local tests for pull requests

mkautzmann:
This PR is regarding the discussion #387.

From my understanding of [this article](https://docs.travis-ci.com/user/environment-variables/#Convenience-Variables) from Travis Docs the TRAVIS_PULL_REQUEST env variable gets set to the current PR **number** if it is a build from a PR.

Considering we are using a RegExp expecting to match '1' or 'true' values in the TRAVIS_PULL_REQUEST variable the match will always be false.

That said lets simply check for falses, if it is 'false' (not a PR) we use PhantomJS else we use Sauce Labs browsers.

### DIFF
--- a/test/karma.conf.js
+++ b/test/karma.conf.js
@@ -1,8 +1,10 @@
 /*eslint no-var:0, object-shorthand:0 */
 
 var coverage = String(process.env.COVERAGE)!=='false',
-	sauceLabs = String(process.env.SAUCELABS).match(/^(1|true)$/gi) && !String(process.env.TRAVIS_PULL_REQUEST).match(/^(1|true)$/gi),
-	performance = !coverage && !sauceLabs && String(process.env.PERFORMANCE)!=='false',
+	pullRequest = !String(process.env.TRAVIS_PULL_REQUEST).match(/^(0|false|undefined)$/gi),
+	realBrowser = String(process.env.BROWSER).match(/^(1|true)$/gi),
+	sauceLabs = realBrowser && !pullRequest,
+	performance = !coverage && !realBrowser && String(process.env.PERFORMANCE)!=='false',
 	webpack = require('webpack');
 
 var sauceLabsLaunchers = {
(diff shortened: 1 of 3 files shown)


mkautzmann:
OK. It is working, but thinking of it I find strange that we run tests twice on PhantomJS. Maybe we could avoid running the last script entirely when testing PRs.

What do you think?

developit:
Agreed for sure - maybe that's something that makes its way into the `npm-scripts`? That said, another option would be to let phantomjs be the default as it is now, and if we detect that karma is running from a travis PR, use Chromedriver?

developit:
This is fantastic btw, thanks for looking into it

test/karma.conf.js
```diff
@@ -1,7 +1,7 @@
 /*eslint no-var:0, object-shorthand:0 */
 
 var coverage = String(process.env.COVERAGE)!=='false',
-	sauceLabs = String(process.env.SAUCELABS).match(/^(1|true)$/gi) && !String(process.env.TRAVIS_PULL_REQUEST).match(/^(1|true)$/gi),
+	sauceLabs = String(process.env.SAUCELABS).match(/^(1|true)$/gi) && String(process.env.TRAVIS_PULL_REQUEST).match(/^(0|false)$/gi),
```
developit: Might be worth also allowing TRAVIS_PULL_REQUEST to be undefined here? (`/^(0|false|undefined)$/gi`)

mkautzmann:
@developit Yeah it makes sense to use Chrome running on Travis directly with Karma. That way I see the testing scenario as follows:
  1. Unit tests first.
  2. Karma integration on PhantomJS (always runs with npm test).
  3. Karma integration locally on Chrome if from PR else remotely using Sauce Labs.

I can try an approach with Karma ChromeDriver like you said and add in this PR. Is that OK for you?

test/karma.conf.js
mkautzmann: Sure! I will add it.

mkautzmann:
OK, Chrome is working on Travis. Now I will re-enable PhantomJS for simple `npm test`.

mkautzmann:
I have changed the SAUCELABS env variable to BROWSER. This way the developer can specify that real browser testing is required, rather than PhantomJS.

Sauce Labs is implicitly enabled when running on CI, disabling it when testing PRs as we want it to do.

I think this is a good approach because people won't be testing on Sauce Labs locally anyways.

Also, the first commit message now seems a bit misleading. I could change that.

Please let me know what you think.

developit:
This looks really good, and once I re-read your comment I understand the decision to rename that env var.  It's also nice that we don't have to set the BROWSER environment variable in Travis.

Think we're good to merge this then?  I'm excited to have properly repeatable testing for PRs haha.

mkautzmann:
Yes, I think we are good to merge 🚀

developit:
Awesome, giving it a whirl.

PR #395, opened 2016-11-11
Reset style props set to null, undefined or false

linde12:
To be compatible with React we have to empty style props
when the value is null, undefined or false.

### DIFF
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -41,7 +41,7 @@ export function setAccessor(node, name, old, value, isSvg) {
 		if (value && typeof value==='object') {
 			if (!isString(old)) {
 				for (let i in old)
-					if (!(i in value) || value[i] == null || value[i] === false) node.style[i] = '';
+					if (value[i] == null || value[i] === false) node.style[i] = '';
 			}
 			for (let i in value) {
 				node.style[i] = typeof value[i]==='number' && !NON_DIMENSION_PROPS[i] ? (value[i]+'px') : value[i];
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -40,8 +40,8 @@ export function setAccessor(node, name, old, value, isSvg) {
 		}
 		if (value && typeof value==='object') {
 			if (!isString(old)) {
-        for (let i in old)
-          if (!(i in value) || value[i] == null || value[i] === false) node.style[i] = '';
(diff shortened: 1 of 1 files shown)


linde12:
@developit I noticed that, in MaterialUI, after i hovered buttons the "hovered" styles were still applied. After some research i noticed that the style-props weren't cleared when setting a property to false, undefined or null like this:

`this.setState({backgroundColor: this.state.isHovering && 'gray'});`

This would become falsy when isHovering is false and so we have to remove the already applied `backgroundColor`

linde12:
I'll fix the spaces -> tabs tomorrow :-)

src/dom/index.js
```diff
@@ -40,7 +40,8 @@ export function setAccessor(node, name, old, value, isSvg) {
 		}
 		if (value && typeof value==='object') {
 			if (!isString(old)) {
-				for (let i in old) if (!(i in value)) node.style[i] = '';
+				for (let i in old)
+					if (!(i in value) || value[i] == null || value[i] === false) node.style[i] = '';
```
developit: Can skip the `in` check now I think

linde12: Done! :smile:

developit:
@linde12 sitting on this for a bit because I think there might be a way to roll the empty check into the second iteration pass. I'll let you know if I think of it.

linde12:
@developit Okay, great! :-)

linde12:
@developit Should we resolve the conflicts? It looks like the diff is only the way we check if `old` is a string or not

developit:
I think we actually need to move the `false` handling to L48, since the check should be on the new value ("if the value is now false, remove the property"):
https://github.com/developit/preact/pull/395/files#diff-4935bb00e75b7854383877cbfd9d770fR48

```js
node.style[i] = typeof value[i]==='number' && !NON_DIMENSION_PROPS[i] ? (value[i]+'px') : value[i]===false ? '' : value[i];
```

linde12:
Ah, i think you're correct! If a value is falsy during the first render the style would, with the current code, be set to false. But if we always check the new properties we will avoid the problem i think.

paranoidjk:
👍  +1 for this, when will this get merged?

linde12:
Looks like a similar issue was merged on master. Can we close this?

PR #396, opened 2016-11-11
Fix usage of copyfiles

tbekolay:
The last argument to `copyfiles` is always interpreted as a directory, which is created if it does not exist. This means that previously, `preact.d.ts` and `preact.js.flow` directories were being made in `dist`.

This PR fixes the usage of `copyfiles` to copy the files to the `dist` directory directly, and point to them in package.json where appropriate.

---

Before this PR:

```bash
$ tree dist
dist
├── preact.dev.js
├── preact.dev.js.map
├── preact.d.ts
│   └── src
│       └── preact.d.ts
├── preact.js
├── preact.js.flow
│   └── src
│       └── preact.js.flow
├── preact.js.map
├── preact.min.js
└── preact.min.js.map

4 directories, 8 files
```

After this PR:

```bash
$ tree dist
dist
├── preact.dev.js
├── preact.dev.js.map
├── preact.d.ts
├── preact.js
├── preact.js.flow
├── preact.js.map
├── preact.min.js
└── preact.min.js.map

0 directories, 8 files
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,8 +10,8 @@
   "minified:main": "dist/preact.min.js",
   "scripts": {
     "clean": "rimraf dist/ aliases.js aliases.js.map  devtools.js devtools.js.map",
-    "copy-flow-definition": "copyfiles src/preact.js.flow dist/preact.js.flow",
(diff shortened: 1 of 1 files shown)


developit:
!

PR #398, opened 2016-11-12
fix render function's signature

tusharmath:


developit:
Awesome, thanks for fixing that 😊

tusharmath:
@developit Is this released? What's the version?

developit:
Not yet. It will be a minor release so likely 7.1 (beta)

PR #402, opened 2016-11-14
Rename DomAttributed typo to DomAttributes.

nojvek:
Removing extra lines for nicely sectionized .d.ts.


developit:
This seems fine to me, but I don't really have the context to know why it was called that in the first place. @kruczy might.

robertknight:
Given the names of the other interfaces in the file and that s/d are right next to each other on many keyboards, it looks clear to me that this is a typo.

PR #409, opened 2016-11-18
Adds requestIdleCallback to vdom/component

paullewis:
So from our offline conversation, I've been working on adding `requestIdleCallback` to `Component`. This is a straw man approach, so I'm not going to be all hurt if you're like "ick, no, don't do THAT!" or anything 😄 

What this does is wraps the `renderComponent` function inners in `requestIdleCallback` if the component in question is set to render as async:

```javascript
class MyComponent extends Component {
  get renderAsync () {
   return true;
  }

  render () {
    // usual gubbins...
  }
}
```

If the component is pending reconciliation and a second call to `renderComponent` is made for it, it cancels the first and replaces it with the most recent.

When faced with a decent number of components to boot from SSR, I'm seeing the main thread blocking time on my Nexus 5X go from ~1 second:

before

All the way down to ~160ms:

after

Obviously this could be a breaking change, though given it's an opt-in behaviour I hope it wouldn't really sting anyone!

I mean, who doesn't enjoy a 6x reduction in blocking main thread time? 🎉 🎉 🎉 🎉

### DIFF
--- a/src/vdom/component.js
+++ b/src/vdom/component.js
@@ -17,39 +17,67 @@ import { removeNode } from '../dom/index';
  *	@param {boolean} [opts.render=true]			If `false`, no render will be triggered.
  */
 export function setComponentProps(component, props, opts, context, mountAll) {
-	if (component._disable) return;
-	component._disable = true;
 
-	if ((component.__ref = props.ref)) delete props.ref;
-	if ((component.__key = props.key)) delete props.key;
+	const requestIdleCallback =
+      typeof window !== 'undefined' &&
+      typeof window.requestIdleCallback !== 'undefined' ?
+         window.requestIdleCallback :
+         null;
 
-	if (!component.base || mountAll) {
-		if (component.componentWillMount) component.componentWillMount();
-	}
-	else if (component.componentWillReceiveProps) {
-		component.componentWillReceiveProps(props, context);
-	}
+	const cancelIdleCallback =
+      requestIdleCallback !== 'null' ?
+      window.cancelIdleCallback :
+      null;
 
-	if (context && context!==component.context) {
-		if (!component.prevContext) component.prevContext = component.context;
-		component.context = context;
-	}
+	let handler = cb => cb();
 
-	if (!component.prevProps) component.prevProps = component.props;
-	component.props = props;
+  // Only support async if rIC is supported and the component requests it.
+	if (requestIdleCallback && component.renderAsync) {
+		handler = requestIdleCallback;
 
-	component._disable = false;
-
-	if (opts!==NO_RENDER) {
-		if (opts===SYNC_RENDER || options.syncComponentUpdates!==false || !component.base) {
-			renderComponent(component, SYNC_RENDER, mountAll);
-		}
-		else {
-			enqueueRender(component);
+    // Remove any pending requests for this component.
+		if (component.renderId) {
+			cancelIdleCallback(component.renderId);
 		}
 	}
 
-	if (component.__ref) component.__ref(component);
+	component.renderId = handler(() => {
+		component.renderId = null;
+
+		if (component._disable) return;
+		component._disable = true;
+
+		if ((component.__ref = props.ref)) delete props.ref;
+		if ((component.__key = props.key)) delete props.key;
+
+		if (!component.base || mountAll) {
+			if (component.componentWillMount) component.componentWillMount();
+		}
+		else if (component.componentWillReceiveProps) {
+			component.componentWillReceiveProps(props, context);
+		}
+
+		if (context && context!==component.context) {
+			if (!component.prevContext) component.prevContext = component.context;
+			component.context = context;
+		}
+
+		if (!component.prevProps) component.prevProps = component.props;
+		component.props = props;
+
+		component._disable = false;
+
+		if (opts!==NO_RENDER) {
+			if (opts===SYNC_RENDER || options.syncComponentUpdates!==false || !component.base) {
+				renderComponent(component, SYNC_RENDER, mountAll);
+			}
+			else {
+				enqueueRender(component);
+			}
+		}
+
+		if (component.__ref) component.__ref(component);
+	});
 }
 
 
--- a/src/component.js
+++ b/src/component.js
@@ -50,17 +50,17 @@ extend(Component.prototype, {
    *
    * @private
    */
-  set renderId (id) {
-    this._id = id;
-  },
+	set renderId (id) {
+		this._id = id;
+	},
 
   /** Gets the stored `requestIdleCallback` id.
    *
    * @private
    */
-  get renderId () {
-    return this._id || null;
-  },
+	get renderId () {
+		return this._id || null;
+	},
 
 	/** Returns a function that sets a state property when called.
 	 *	Calling linkState() repeatedly with the same arguments returns a cached link function.
--- a/src/component.js
+++ b/src/component.js
@@ -44,6 +44,23 @@ extend(Component.prototype, {
 	// 	return true;
 	// },
 
+	/** Sets a `requestIdleCallback` id on the component. If there are
+   * repeated calls to render the component (and the component is set to render
+   * asynchronously), the existing rIC callback will be cancelled using this id.
+   *
+   * @private
+   */
+  set renderId (id) {
+    this._id = id;
+  },
+
+  /** Gets the stored `requestIdleCallback` id.
+   *
+   * @private
+   */
+  get renderId () {
+    return this._id || null;
+  },
 
 	/** Returns a function that sets a state property when called.
 	 *	Calling linkState() repeatedly with the same arguments returns a cached link function.


addyosmani:
6x is such a nice win. Great work, @paullewis 🎉  Shipping it as an opt-in (at least for now) should give folks a chance to try it out for a while in the wild and let you know if anything breaks.

src/vdom/component.js
```diff
@@ -17,39 +17,67 @@ import { removeNode } from '../dom/index';
...
+         null;
 
-	if (!component.base || mountAll) {
-		if (component.componentWillMount) component.componentWillMount();
-	}
-	else if (component.componentWillReceiveProps) {
-		component.componentWillReceiveProps(props, context);
-	}
+	const cancelIdleCallback =
+      requestIdleCallback !== 'null' ?
```
tony19: `'null'`? Did you mean `null`?

    requestIdleCallback !== null

developit: whoops!

paullewis: Nah, I didn't. I did a typeof test earlier on, and that carried through. Awwwwkward.

makingoff: @paullewis you did return `window.requestIdleCallback` to `const requestIdleCallback` if `typeof window !== 'undefined' && typeof window.requestIdleCallback !== 'undefined'`. So, now `requestIdleCallback` contains a function or `null`. Not a `typeof` from earlier text.

developit:
Looking into some breakages from this, sorry for the merge wait!

src/vdom/component.js
Aweary: @makingoff is right, `requestIdleCallback` will be `null` not `'null'` when `window.requestIdleCallback` is not defined, so `cancelIdleCallback` will be `undefined` instead of `null` in that case.

screen shot 2016-12-02 at 3 47 44 pm


It doesn't actually cause an error since `cancelIdleCallback` is never called when `requestIdleCallback` is not defined, but still worth fixing 😄

Aweary: @makingoff is right, `requestIdleCallback` will be `null` not `'null'` when `window.requestIdleCallback` is not defined, so `cancelIdleCallback` will be `undefined` instead of `null` in that case.

screen shot 2016-12-02 at 3 47 44 pm


It doesn't actually cause an error since `cancelIdleCallback` is never called when `requestIdleCallback` is not defined, but still worth fixing 😄

ainth:
I like the idea but wonder if this should just be a component and not part of the framework. Roughly:

Instead of:
```
  get renderAsync () {
   return true;
  }
```
it would be:
```
render() {
  return (
    <AsyncRenderer>
      ...
    </AsyncRenderer>
  );
}
```

Rough implementation:
```
class AsyncRenderer extends Component {
  constructor(props) {
    super(props);
    this.state = { readyToRender: false };
  }

  componentWillMount() {
    requestIdleCallback(() => {
      this.setState({ readyToRender: true });
    })
  }

  render() {
    if (this.state.readyToRender) {
      return this.props.children;
    }
    return (<span />);
  }
}
```
(this would need a lot more to work in real life but hopefully conveys the idea). This feels more idiomatic and wouldn't require breaking changes or an expansion of the API.

matthewp:
One thing I like about @ainth's idea is that it allows for the async rendering to be contextual. I could see having different variations of AsyncRenderer for different strategies.

For example, I could see an InViewRenderer that waits until the component is within some threshold of being in the viewport before rendering. How would you do stuff like that using the `renderAsync` approach?

developit:
We have a `<RenderOnVisible />` that does what you described @matthewp:

example

Not compatible with SSR though, since it uses `getBoundingClientRect()`. That said, for lazy-loaded items it might be better to skip SSR.

I think that's the most interesting part of Paul's implementation here: you can render on the server, but then just slowly hydrate as needed on the client.  I think this is a really interesting space.

developit:
Also: for people who are as interested as I am with this - here's a really interesting demo of Paul's amended version of Preact asynchronously rendering a pythagoras tree. You can see the effect of the fully async rendering as the tree depth grows beyond what your CPU can render in in a single pass - it starts to render progressively based on available time.

[d52c0411766954026ad68af32d1ae253](http://jsfiddle.net/developit/n733adf9/)

Demo: http://jsfiddle.net/developit/n733adf9/

marvinhagemeister:
@developit That tree demo is mesmerising! I love where this is going 👍

sebmarkbage:
This demo looks very similar to the [Sierpinski Triangle Demo](https://github.com/facebook/react/pull/7180) that we've been using to test React Fiber's scheduling. The difference there is that we can hold the rendering to the screen until everything is complete.

We do have a mode that demos the progressive work it does but we only use that to demo the work that is done behind the scenes - not user visible.

In most cases you don't actually want to show the progression of too many incomplete states to the user. For example on Facebook we've actively been trying to get rid of that since the load experience is jarring when there are too many small visible updates to the screen. So instead of showing them immediately we delay them for later.

developit:
@sebmarkbage I believe this demo code was actually @Swizec's that I punked to test out this rendering mode. I do have a preact version of that triangle demo floating around somewhere though, it shows how the current async rendering works prior to Paul's PR here - not supporting async sibling rendering.

Your point about perceived lag and meaningful scheduling is the thing that's making me think this PR through perhaps a little more thoroughly than I ought to.  I don't think there's a one-size-fits-all scheduler, so I want to provide a clean extension point so that people can build their own schedulers and publish them.  There are many use-cases, so I foresee many scheduling techniques to match.

sebmarkbage:
@developit Certain scheduling techniques are anti-modular which are the ones that need built-in support. Anything else can be user space though.

pemrouz:
> I don't think there's a one-size-fits-all scheduler, so I want to provide a clean extension point so that people can build their own schedulers and publish them. There are many use-cases, so I foresee many scheduling techniques to match.

This is very important: How would this for example work with the microtask or rAF batching? It would be very hard for users to compose/control strategies with a `true`/`false` API or some other config.

Instead, [I think](https://github.com/rijs/fullstack/issues/31#issuecomment-264626124) it would be better for the framework to expose primitives to the user, for example like `pauseMeIfNeccessary` that can be naturally awaited:

```js
async render(node, data){
  const { pause } = framework
  // render some things
  await pause()
  // render some things
  await pause()
  // render some things
}
```

Just awaiting on a noop would push the component to the back of the render queue. You could have another one (deopt?) that would use rIC that would essentially make it low priority by only continuing rendering when other components have rendered (even if they joined the queue after the component was paused). You could potentially allow passing a value to indicate different priority lanes for rendering components. And a timeout. You could pause until visible, but perhaps first just fixing width/height though so the container has the right scrollbar length etc etc. Furthermore, these common cases could also be used by render middleware by decorating the component rather than from inside. This extension point allows more innovation to take place outside the framework core.

The key change to allow progressive rendering is that the render function needs to be `async`. This allows the component to stream/pipe/flush partial updates to the DOM early, rather than having to do it all at once in the final `return`.

The main objection I've heard against this is that it's no longer "pure". What's important with that is that components are still **declarative** and a **function of data**, it just doesn't have to resolve _immediately_ but can over time. 

The overall render process is fundamentally async. In most cases you have to wait for data to load. With the above, you can just do `await import(dep)` too. 

Forcing the tree of all components to be sync causes a lot of awkwardness atm for UI developers: it either pushes developers to try ideologically load everything at the root (doesn't scale), or splitting into an arbitrary async phase and sync phase with more lifecycle (i.e. `getInitialProps`, which is again only called at top-level), or causes performance problems (like having to double render) or a huge spike in complexity with other third-party solutions (sagas, thunks, etc). 

Related:
[1] [Need to manually re-call getInitialProps of sub-components](https://github.com/zeit/next.js/issues/192#issuecomment-258897709)
[2] [Support asynchronous server rendering (waiting for data before rendering)](https://github.com/facebook/react/issues/1739#issuecomment-267032331)
[3] [proposal: await dependencies inside component function](https://github.com/rijs/fullstack/issues/31#issuecomment-264626124)

developit:
Wouldn't a generator be more appropriate than an async function for modelling this?

For what it's worth, making render asynchronous does not solve Paul's issue on its own.  If you render 500 components in a common parent (as this demo does), each is not sufficiently aware of the other or its place in the try that it could determine an appropriate priority for itself.  Centralizing that decision into a hook makes this possible, since all component render requests flow through a single point that can use as much state as it needs to implement an appropriate algorithm for prioritizing updates.

swernerx:
@developit what is this `<RenderOnVisible />` thingy? Do you have a pointer to source / docs for this... sounds super useful!

luisvinicius167:
@swernerx I like this so much! https://github.com/developit/preact-virtual-list
s2

developit:
Yeah @luisvinicius167's reference is the closest I have publicly available. The other one is internal. I'll ask about open-sourcing it

jide:
Wow ! You may also want to have a look at https://github.com/Synchronized-TV/react-async-child (the react-router 4 example there is interesting)

developit:
@jide That's similar to what I had originally suggested ([AsyncComponent](http://codepen.io/developit/pen/jrOyPO?editors=0010)).  However, Paul pointed out that it breaks when using Server-Side Rendering.  The statically rendered HTML element children of the lazily-rendered component get removed.  The hope here is to work around that by not only lazily rendering children, but the component itself.

jide:
Ah ok, I better understand the issue !

kurtextrem:
@developit Any new plans yet regarding this PR? :)

kurtextrem:
By the way, as React Suspense is on the way. Might this PR be the beginning of Preact Suspense? 🤔

paullewis:
> Hey is this still on the table? 🙂 Are you guys considering it?

I'm always considering it... /me stares at @developit 😁

132yse:
Will this PR be merged in preact X? This seems to be another way to implement time slicing!

marvinhagemeister:
@132yse Sorry for the late response. I missed it in the pool of notifications :/

Time-Slicing refers to the ability to schedule updates with different priorities. This is a huge change and requires a 2-phase commit. It's something we may set our sights on for 2019, but nothing concrete has been decided yet.

The changes in this PR debounce updates via `requestIdleCallback` instead of `setTimeout` or a `Promise`-queue. It's something that we've supported for a while now via `options.debounceRendering`.

kurtextrem:
Friendly ping for this PR, it's been a year. Any plans, anything non-maintainers could help with getting this PR merged or improved?

JoviDeCroock:
hey @kurtextrem 

This is a PR for Preact 8, if you're using Preact X you can use this in user-land. We don't want to impose this increase in size for everyone who potentially won't need it.

```

options.debounceRendering = requestIdleCallback
```

paullewis:
Thanks for the update! 👍 

409 will always live on in my heart.

cristianbote:
As it will in our hearts as well, @paullewis. 🙇

PR #412, opened 2016-11-22
refactor:code style about indent

qiuyuntao:


developit:
I appreciate the effort, but these lead to bikeshedding without adding much value. Feel free to discuss on gitter though!

developit:
Actually this might be fine, but there are more of them around the codebase I'd want to change of so. Open to it.

PR #423, opened 2016-11-30
Use SVG namespace comparison instead of instanceof

nhunzaker:
I'm having trouble testing with Jest because JSDOM does not implement SVGElement. This PR changes the SVG check to compare the namespace URI.

### DIFF
--- a/src/vdom/diff.js
+++ b/src/vdom/diff.js
@@ -15,8 +15,9 @@ export const mounts = [];
 /** Diff recursion count, used to track the end of the diff cycle. */
 export let diffLevel = 0;
 
-let isSvgMode = false;
+const svgNamespace = "http://www.w3.org/2000/svg";
 
+let isSvgMode = false;
 
 export function flushMounts() {
 	let c;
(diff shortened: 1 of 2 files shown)


developit:
Looks good! I'll need to look into filesize and performance stuff since `instanceof` is faster than a property access, but I definitely see the point!

nhunzaker:
Sounds good!

nhunzaker:
FWIW, this is what I'm doing in the meantime for a project that doesn't use SVG:

https://github.com/vigetlabs/microcosm-preact/blob/master/test/helper.js:

```
class SVGElement extends HTMLElement {}

global.SVGElement = SVGElement
```

Which is not awesome, but unblocked me!

developit:
@nhunzaker ah yes, that's a good workaround.  Will try to get to this issue ASAP!

nhunzaker:
@developit Of course, an alternative is that we simply fix this on JSDOM's end. That's a bit of a longer turn around time, but I can look into that.

developit:
Yeah, I'm surprised they don't at least do what your workaround does. On preact's end, SVGElement just has to exist, it's actually better if it never matches the constructor when doing server-rendering.

robertknight:
I created an ESBench for this - https://esbench.com/bench/587110f7d5686f990078bc62 . Both methods of checking are very cheap (tens of millions of ops/sec on my laptop), with the string check being faster in Firefox on my system and instanceof being faster in Chrome.

So if the compatibility of doing it this way is better, I think we can merge. OK with this @developit ?

developit:
wow, @robertknight @nhunzaker - I'm silly, the string check is much faster. @robertknight I modified your benchmark to capture values (to avoid DCE) and iterate over a few possibilities for the values of `element` so that it's not as deterministic (more like what we see in this function).  Turns out to be like 3x faster with the string namespaceURI check!

screen shot 2017-01-09 at 11 51 38 pm

nhunzaker:
Hurrah!

developit:
Released as [7.2.0](https://github.com/developit/preact/releases/tag/7.2.0) (beta).

nhunzaker:
🌮 🎉 🎈

Issue #424, opened 2016-11-30
Remote testing branches


mkautzmann:
Currently we are running Sauce Labs on **pushes** to all branches. PRs are not included as of #390.

That is OK and working fine for PRs and pushes to `master`.

After checking the Greenkeeper behavior, I noticed that it pushes it's changes to a newly created local branch and then it creates a PR to `master`.

The problem is that it ends up starting two Travis jobs, one using Sauce Labs (push to branch) and one using local Chrome (PR). Since Sauce Labs connection can be flakey, it may not be ideal to use it to check dependency updates or even simple branch experiments.

I believe that we should only keep Sauce on for pushes to the `master` branch.

This seems right because the widget on README is related to `master` as well.

The question now is: how do we want to handle `beta`? Should we test it on Sauce Labs too or wait until it arrives on `master`?

developit:
That would be ideal. I think we can ignore beta - I've just been using it to keep my nearly-merged work public so that people don't submit PRs against 6.x.  I'll merge it into master now :)

developit:
Running sauce only on master was for some reason not working for me - I was trying to do it from within Karma, maybe it'd be simpler to just do that conditional in `npm run test`?

mkautzmann:
OK, I'll see what I can do.

mkautzmann:
OK, created branch `remote-testing` to test the new behavior.

Works fine so far (only Chrome running): https://travis-ci.org/developit/preact/builds/180537936

Now let's do a PR to `master` to complete walking through @greenkeeperio-bot's steps.

mkautzmann:
OK, #426 got merged and Sauce Labs tests [ran successfully](https://travis-ci.org/developit/preact/builds/180693798).

I think this ends this remote testing quest 😄.

developit:
Yay! 👍

PR #426, opened 2016-12-01
Run remote tests only when pushing to master

mkautzmann:
Related to #424.

We now handle `TRAVIS_BRANCH` to decide whether we use Sauce Labs (remote) testing or not.

With this change, we will only use Sauce Labs if it is pushing to `master`.

We can only close the issue #424 once we know it is actually running Sauce Labs once we merge this PR.

Please let me know it looks good before we proceed.

Thanks!

### DIFF
--- a/test/karma.conf.js
+++ b/test/karma.conf.js
@@ -3,8 +3,9 @@
 var coverage = String(process.env.COVERAGE)!=='false',
 	ci = String(process.env.CI).match(/^(1|true)$/gi),
 	pullRequest = !String(process.env.TRAVIS_PULL_REQUEST).match(/^(0|false|undefined)$/gi),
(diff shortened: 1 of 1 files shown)


developit:
Ahhh this is awesome, so it still runs browser tests, just in Chrome on Travis instead of via SauceLabs. This is freaking perfect!

developit:
Are we good to merge then since it seems to work?  It looks like master will still run sauce so we're good to go..

mkautzmann:
Yes, I think you can merge it now or review it so I can merge it.

After the merge I'll check if everything is OK in Travis and close the issue #424.

Then I will delete the `remote-testing` branch.

developit:
Perfect, here it goes 👯

PR #441, opened 2016-12-11
Remove an errant "to".

braddunbar:


developit:
Thanks for all these :)

PR #443, opened 2016-12-11
Update setAccessor argument documentation.

braddunbar:


developit:
Goo catch! Thanks @braddunbar

PR #450, opened 2016-12-13
Only save ref from original attributes.

braddunbar:
Since we only use the ref property of `originalAttributes`, we don't need a reference to the whole object.


developit:
Nice! Looks like this saves a byte :)

I'm actually wondering if we could drop the function type check there, too. That way if someone passes a String ref it'll throw a fairly reasonable error: "ref is not a function"

braddunbar:
> I'm actually wondering if we could drop the function type check there, too.

Seems reasonable to me!

braddunbar:
Updated in ef2eeb90d87b5465a002fdcc32351894d64c869d.

developit:
Woot! Another 2 bytes gone, and now we show a nice error message :)

developit:
Ended up being 3 bytes! 👍

PR #452, opened 2016-12-14
Do not inner diff dangerouslySetInnerHTML.

braddunbar:
The children of a node with dangerouslySetInnerHTML should not be diffed because this will always cause all children to be removed.

I'm not sure if this is a good solution or not. Let me know what you think! 😺

### DIFF
--- a/src/vdom/diff.js
+++ b/src/vdom/diff.js
@@ -152,17 +152,15 @@ function idiff(dom, vnode, context, mountAll) {
 	diffAttributes(out, vnode.attributes, props);
 
 
-	if (!props.dangerouslySetInnerHTML) {
-		// Optimization: fast-path for elements containing a single TextNode:
-		if (!hydrating && vchildren && vchildren.length===1 && typeof vchildren[0]==='string' && fc && fc instanceof Text && !fc.nextSibling) {
-			if (fc.nodeValue!=vchildren[0]) {
-				fc.nodeValue = vchildren[0];
-			}
+	// Optimization: fast-path for elements containing a single TextNode:
+	if (!hydrating && vchildren && vchildren.length===1 && typeof vchildren[0]==='string' && fc && fc instanceof Text && !fc.nextSibling) {
+		if (fc.nodeValue!=vchildren[0]) {
+			fc.nodeValue = vchildren[0];
 		}
-		// otherwise, if there are existing or new children, diff them:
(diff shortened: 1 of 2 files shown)


developit:
Were you seeing a bug with this one?  Currently they'll all get skipped because they weren't created by preact's recycler (specifically, because they would fail [this check](https://github.com/developit/preact/blob/75acb15ff7f9439ed89b8e1fd82d39961fdfa84d/src/vdom/diff.js#L204)).
Let me know!

src/vdom/diff.js
```diff
@@ -152,15 +152,17 @@ function idiff(dom, vnode, context, mountAll) {
...
-			fc.nodeValue = vchildren[0];
+	if (!props.dangerouslySetInnerHTML) {
+		// Optimization: fast-path for elements containing a single TextNode:
+		if (!hydrating && vchildren && vchildren.length===1 && typeof vchildren[0]==='string' && fc && fc instanceof Text && !fc.nextSibling) {
+			if (fc.nodeValue!=vchildren[0]) {
+				fc.nodeValue = vchildren[0];
+			}
+		}
+		// otherwise, if there are existing or new children, diff them:
+		else if (vchildren && vchildren.length || fc) {
```
developit: thinking since the first `if` block will never be true during hydration, this could actually just go into the second block:

```js
else if (!props.dangerouslySetInnerHTML && vchildren && vchildren.length || fc) {
```

braddunbar:
> is this only happening during hydration?

Yes! I should've mentioned that in the description.

> Were you seeing a bug with this one?

Yep. I put some rendered markdown into `dangerouslySetInnerHTML` and it's removed upon hydration.

developit:
Alright, I'll.prioritize this. I actually have to reverse the order of attribute/children diffing for another bug (initial select value) and it'll likely affect how this works. I'll try to merge first though :)

developit:
Wondering if this is the same bug as #454?

developit:
@braddunbar @gdub22 I merged this and it looks like your test is passing against master since [this change](https://github.com/developit/preact/commit/c9126bc7a2a2c45556e58ea949a1681b69807dcf#diff-a0896fa5d65b1d80ca4d496badbae50f). Rather than skipping the children diff, the order of attribute and children diffing was reversed (which brings us back to the order used in Preact 6 and prior).  Very happy to have your test though, and thanks for pointing in the right direction!

Issue #446, opened 2016-12-12
<select> elements aren't selecting a matching value after update


philipwalton:
preact version: `7.1.0`

**Update:** I created a [Codepen demo](http://codepen.io/anon/pen/YpOaPe?editors=0010) showing the behavior.

I have a component that renders three `<select>` elements representing a users Google Analytics account tree structure (accounts > properties > views). When the selected account is updated, the property and view `<select>`s should update accordingly, but in many cases an empty select is displayed. (Note the empty select is never displayed on the initial render, only after user updates).

Steps to reproduce:

1) Include the `ViewSelector` component code at the bottom of this issue in a Preact app.
2) Once the page loads, the `ViewSelector` renders with "Account A" selected
3) Select "Account C" then select "Account B"
4) Notice how once "Account B" is selected, the property and view `<select>`s aren't being populated.

**Note: the code works in React as well as Preact-compat.**

```js
import {h, Component} from 'preact';

const accountSummaries = {
  "kind": "analytics#accountSummaries",
  "username": "joe@gmail.com",
  "totalResults": 5,
  "startIndex": 1,
  "itemsPerPage": 1000,
  "items": [
    {
      "id": "1001",
      "kind": "analytics#accountSummary",
      "name": "Account A",
      "webProperties": [
        {
          "kind": "analytics#webPropertySummary",
          "id": "UA-1001-1",
          "name": "WebProperty A.A",
          "internalWebPropertyId": "3001",
          "level": "STANDARD",
          "websiteUrl": "http://example.com",
          "profiles": [
            {
              "kind": "analytics#profileSummary",
              "id": "2001",
              "name": "Profile A.A.A",
              "type": "WEB"
            },
            {
              "kind": "analytics#profileSummary",
              "id": "2002",
              "name": "Profile A.A.B",
              "type": "WEB"
            },
            {
              "kind": "analytics#profileSummary",
              "id": "2003",
              "name": "Profile A.A.C",
              "type": "WEB"
            }
          ]
        },
        {
          "kind": "analytics#webPropertySummary",
          "id": "UA-1001-2",
          "name": "WebProperty A.B",
          "internalWebPropertyId": "3002",
          "level": "STANDARD",
          "websiteUrl": "http://example.com",
          "profiles": [
            {
              "kind": "analytics#profileSummary",
              "id": "2004",
              "name": "Profile A.B.A",
              "type": "WEB"
            },
            {
              "kind": "analytics#profileSummary",
              "id": "2005",
[...]

PR #454, opened 2016-12-15
Dom mutation was not happening for elements with dangerouslySetInnerHTML.

harish2704:
For elements with dangerouslySetInnerHTML attr, 

initial rendering will take place correctly. ( only Test case for initial render is present in the unit tests ).

after that, after some updating, if we are again rendering the same element, It was not working. 

This pull request fixes that issue. ( Unit test also included )

### DIFF
--- a/src/vdom/diff.js
+++ b/src/vdom/diff.js
@@ -306,7 +306,8 @@ export function recollectNodeTree(node, unmountOnly) {
  */
 function diffAttributes(dom, attrs, old) {
 	// remove attributes no longer present on the vnode by setting them to undefined
-	for (let name in old) {
+	let name;
+	for (name in old) {
 		if (!(attrs && name in attrs) && old[name]!=null) {
 			setAccessor(dom, name, old[name], old[name] = undefined, isSvgMode);
 		}
@@ -314,7 +315,7 @@ function diffAttributes(dom, attrs, old) {
 
 	// add new & update changed attributes
 	if (attrs) {
-		for (let name in attrs) {
+		for (name in attrs) {
 			if (
 				name!=='children' &&  // break if attributeName is children
 				name!=='innerHTML' &&  // break if attributeName is innerHTML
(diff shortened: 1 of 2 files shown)


harish2704:
[INFO] This bug can be visualized by simply running the project.
https://github.com/harish2704/preact-router5-boilerplate

If we repeat the same with react library, ( just comment resolve part in webpack config ) it will work properly

-Thanks

developit:
Hi @harish2704! Thanks for the bug report, repro and PR (all in one)!
I need to check if this will cause excessive recycling, and if so might have to change the technique a little bit.  Regardless of that though, the tests and reproduction will help make sure any tweaks I have to make keep the functionality working :)

FWIW, I was hoping this would be fixed by the fix for #446.  I'll try to pull your tests in to see if that's the case, since that would be a super quick solution :)

src/vdom/diff.js
```diff
@@ -315,7 +315,17 @@ function diffAttributes(dom, attrs, old) {
 	// add new & update changed attributes
 	if (attrs) {
 		for (let name in attrs) {
-			if (name!=='children' && name!=='innerHTML' && (!(name in old) || attrs[name]!==(name==='value' || name==='checked' ? dom[name] : old[name]))) {
+			if (
+				name!=='children' &&  // break if attributeName is children
+				name!=='innerHTML' &&  // break if attributeName is innerHTML
+				(
+					( name=== 'dangerouslySetInnerHTML' ) || // continue if attributeName is dangerouslySetInnerHTML
```
harish2704: This line is the essence of PR #454

harish2704:
Hi,
In a quick test, the issue didn't seems to be resolved even after the merge.

I tested using this sample code against current master branch (bc84e2a6a41e4cfd033f5497a7d7fa9d99c9e486).

-info

developit:
@harish2704 what should I be looking for when I run the example?

developit:
I was able to get your repo running, just not really sure what I'm looking for.

harish2704:
Hi,
In the demo , the home route rendering using  **dangerouslySetInnerHTML**

If you try to goback to home route after visiting any other routes, You can see the problem.

It was not rendering

4preact

PR #388, opened 2016-11-04
test coverage for #254

sskoopa:
This slightly increments coverage, and tests for the re-emergence of issue #254

PR #455, opened 2016-12-16
Fix test case in IE

mkautzmann:
We have a problem with test case introduced in #388 when running in IE.

I'm heading to `bug/ie`, so we can put together all of the IE fixes and then merge it to master.

The line feeds won't match with what is being expected.

I switched to `textContent`.

I don't know if this is OK. At least it fixes the test case and don't break others.

Please review this when possible.

Thanks.

### DIFF
--- a/test/browser/components.js
+++ b/test/browser/components.js
@@ -214,7 +214,7 @@ describe('Components', () => {
 
 		let good, bad;
 		let root = render(<GoodContainer ref={c=>good=c} />, scratch);
-		expect(scratch.innerText, 'new component with key present').to.equal('A\nB');
+		expect(scratch.textContent, 'new component with key present').to.equal('AB');
 		expect(Comp.prototype.componentWillMount).to.have.been.calledTwice;
 		expect(sideEffect).to.have.been.calledTwice;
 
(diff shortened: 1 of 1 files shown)


developit:
Ahhh perfect! Definitely should have been `.textContent` all along. Since this passes I think we should merge immediately.

test/browser/components.js
```diff
@@ -214,31 +214,31 @@ describe('Components', () => {
 
 		let good, bad;
 		let root = render(<GoodContainer ref={c=>good=c} />, scratch);
-		expect(scratch.innerText, 'new component with key present').to.equal('A\nB');
+		expect(scratch.textContent, 'new component with key present').to.equal('AB');
```
developit: I wonder why the newline disappeared here?  I guess we'll merge and see how SauceLabs fares across the supported browsers.

NekR: Because `innerText` returns content as is (as it was entered in markup). `textContent` is more "strict" and behaves as specced (result depends on nodes' positions).

developit: Awesome.

Issue #430, opened 2016-12-05
Text node sometimes duplicated (after initial render) if its preceded by an empty node in IE9-11


rmales:
This issue seems to have been introduced since 7.0.1. I guess this may be what is causing the tests to fail on Saucelabs (although I can't figure out what test it is from that output).

Example: http://jsfiddle.net/28ckgyzj/13/

mkautzmann:
@rmales Yes, good catch. We are investigating this behavior.

developit:
Ahh perfect, now we have an issue to track it under. Thanks 👍

rmales:
If it helps at all reinstating the `instanceof` clause removed from this conditional seem to fix the bug (although I don't know if that affects the aim of the rest of the commit):
https://github.com/developit/preact/commit/c5553cefd67c4dc2ad262df2cfcb986114ef15a9#diff-a0896fa5d65b1d80ca4d496badbae50fL149

developit:
Ahh - awesome!

developit:
It's possible this is a Text node normalization issue in IE.  I have a potential fix for that, will test it out ASAP.

mkautzmann:
OK, Windows test setup working fine with preact. The problem seems to be related to `__preactattr_` not always being correctly set in Text nodes on IE. Then it ends up using `insertBefore` creating the duplicated nodes because the node reference check doesn't match. Couldn't figure out where is the origin of this behavior yet. Guess I need to get familiar with the codebase :)

developit:
Hmm.  I wonder if Text nodes can't accept properties in IE?

developit:
I seem to have found the fix. For some reason this bit was removed:

<img width="496" alt="screen shot 2016-12-15 at 6 14 44 pm" src="https://cloud.githubusercontent.com/assets/105127/21245738/64d218f4-c2f2-11e6-86a2-6c521655d9c0.png">

mkautzmann:
Humm, so in this patch you never reach the part of the code I was talking about. Seems fixed. Can you upload it to another branch so I can play with it in my local env?

developit:
yup!

developit:
Alright, it's up as [bug/ie](https://github.com/developit/preact/tree/bug/ie)

rmales:
I could be missing something, but having done a (quick) test that change doesn't seem to fix the issue in my example case (in IE11).

developit:
@rmales how did you test the change?

rmales:
[...]

PR #462, opened 2016-12-19
Fix duplicated text nodes issue in IE

mkxml:
Regarding #430.

### DIFF
--- a/src/vdom/diff.js
+++ b/src/vdom/diff.js
@@ -92,8 +92,6 @@ function idiff(dom, vnode, context, mountAll) {
 			dom = document.createTextNode(vnode);
 		}
 
-		// Mark for non-hydration updates
-		dom[ATTR_KEY] = true;
 		return dom;
 	}
 
--- a/src/vdom/diff.js
+++ b/src/vdom/diff.js
@@ -201,7 +201,7 @@ function innerDiffNode(dom, vchildren, context, mountAll) {
 				keyedLen++;
 				keyed[key] = child;
 			}
-			else if (hydrating || props) {
+			else if (hydrating || props || child instanceof Text) {
 				children[childrenLen++] = child;
 			}
 		}


developit:
Just some context:  we'll want to make sure we account for the issue solved by the commit that removed this. I can do some digging to figure out what it was:

https://github.com/developit/preact/commit/c5553cefd67c4dc2ad262df2cfcb986114ef15a9

mkautzmann:
Sure, I'm looking for potential breakages from this change in 7.1.0. Simulating SSR and seeing if it breaks the Text node behavior.

mkautzmann:
Specially in IE.

mkautzmann:
OK, I couldn't find problems by manually testing some scenarios in my local env.

Namely root Text Nodes and inner text nodes with hydration. Everything works fine.

I'm not fully aware of the problem that originated the commit you posted though. I know it has something to do with SSR and hydration.

Do you have a specific repro you would like to try to make sure it's working?

I can add a test to stop future regressions to this PR if applicable.

developit:
I'll try to get a repro together tonight!

developit:
Been thinking about this and I can't remember what the original issue was. The only difference between this and the property check would be this will remove externally created text nodes - but I don't really see much of a use case for them anyway.

developit:
@mkautzmann think we should drop the [`[ATTR_KEY]` prop marker](https://github.com/developit/preact/blob/123978c108579f147c49906c31301ad1712926be/src/vdom/diff.js#L95-L96) from Text nodes in this case since the `instanceof` check will always be true?

mkautzmann:
@developit Done, I've been thinking about the hydration lately and it will perform oddly for root text nodes. I hope other developers will understand that behavior.

I tried to make it work with ATTR_KEY on text nodes. No success.

I think we should ship this quick fix and keep an eye on this, fix IE for now and see if we can improve it in the future.

mkautzmann:
I can add a quick test case for root text nodes hydration. Would it help keep track of this?

developit:
Sure. If it's failing on IE we can mark it as flakey and it'll get skipped for the saucelabs tests.

mkautzmann:
@developit Yeah, spent some time trying to make a quick test case but it is flakey in IE, since it is precisely what I want to test I see no point ignoring it.

I guess I don't have a proper test case for this now. Feel free to merge it or not. 

Then `bug/ie` should be green to merge to `master`.

I'm open to any discussion.

Thanks!

PR #463, opened 2016-12-19
Throw when passing undefined component to h()

NekR:
This so much helps in development, but probably don't worth it to ship that to production. Should we utilize `process.env.NODE_ENV === 'development` in Preact in this case?

Error message is not final too, of course :-)

### DIFF
--- a/src/h.js
+++ b/src/h.js
@@ -15,6 +15,10 @@ const stack = [];
  *  render(<span>foo</span>, document.body);
  */
 export function h(nodeName, attributes) {
+	if (!nodeName) {
+		throw new Error('Component is missing');
+	}
+
 	let children = [],
 		lastSimple, child, simple, i;
 	for (i=arguments.length; i-- > 2; ) {


NekR:
And I'll test cases for this once we have agreement :-)

developit:
I don't think anything can depend on `process.env` in Preact, since it's often loaded off a CDN. However, we can emulate this exact behavior:

https://github.com/developit/preact/blob/75acb15ff7f9439ed89b8e1fd82d39961fdfa84d/src/vdom/diff.js#L108

Instead of invoking `String(vnode.nodeName)` there, we could just let it error out.  That'd be a free error (actually saves a few bytes) and would produce a pretty similar result.

NekR:
@developit The reason I put that into `h()` is because it's closest place in stack trace to broken code. i.e. I can just click to the next call in the stack trace and see the problem.

But yeah, anything that makes it throw is better than nothing, so I'm fine with simply removing `String()` call too.

developit:
Agreed, might just come down to size/perf

NekR:
@developit So I'm going to remove `String()` call, right? (+tests of course)

developit:
I think we can yeah, though that definitely counts as a breaking change.

NekR:
@developit yeah, I agree. Should we start 8.0 branch already then? Or just keep PRs open for those changes?

developit:
Definitely down to start a branch, but I'm thinking we need to merge some stuff first or things will get a little out of hand.

btw another thought: this awful line at the beginning of h:

```js
h(nodeName, ...etc) {
  nodeName.nodeName;
}
```

It'd spit out errors like this:
screen shot 2016-12-19 at 8 12 16 pm

developit:
I [renamed "The Wreckoning" project to `8.0`](https://github.com/developit/preact/projects/3) for marketing reasons

NekR:
> btw another thought: this awful line at the beginning of h:

It can be anything then, like: `nodeName.h`

NekR:
Crazy idea: still have `process.env.NODE_ENV` but compile to CDN with it set to `'production'`. i.e. remove all those conditions when compiling for CDN.

developit:
That's work with the existing `dist/preact.min.js`, but then people have to know which to use. There are so many cases where people use the development version and unknowingly bloat or even de-opt their builds, it would make me sad to think of that happening to people who use preact 😢

developit:
ah! I found the plugin I wrote for this - it hooks into h() and throws when `vnode.nodeName` is empty, giving you the more convenient stack trace:
https://gist.github.com/developit/7fda739f5dea675386d6c48ae96af667

Maybe we need something like a debug plugin?

```js
require('preact/debug');
```

NekR:
@developit sounds great actually and most importantly it's opt-in!

NekR:
I'll handle that of course, a bit later though. Probably makes sense to open an issue/new-PR about it and close this PR.

robertknight:
Closing this as suggested in the last comment.

PR #467, opened 2016-12-21
Fix #466: Type of setState argument should be Partial<StateType>

### LINKED ISSUE #466: setState typescript signature is too restrictive.
The typescript signature of `setState` is too restrictive, since it assumes it's parameter must be exactly the type of the component's `StateType` type parameter.

TypeScript has a builtin type `Partial<T>` that has the same type as `T`, but with all properties forced to be optional defined as:

```typescript
type Partial<T> = {
    [P in keyof T]?: T[P];
}
```

And define the signature of `setState` to be `setState(state: Partial<StateType>, opts?: any)`.

I'll PR this in when I have a moment!

mseddon:


src/preact.d.ts
```diff
@@ -52,7 +52,7 @@ declare namespace preact {
 
 		linkState:(name:string) => (event: Event) => void;
 
-		setState(state:StateType, opts?:any):void;
+		setState(state:Partial<StateType>, opts?:any):void;
```
jrf0110: This is cool. I've been making most of my State prop declarations optional, which is exactly what that generic type does https://github.com/Microsoft/TypeScript/pull/12276/files#diff-a6b488d9bd802977827b535a3011c1f3R1350

My only issue with using this is that it requires `typescript@next` if I'm not mistaken, and this could break some folks projects if released. I'm personally using `typescript@next`, so I'm good :)

mseddon: typescript@2.1.x supports this, but yeah, it's a bit bleeding edge. :)

jrf0110: Awesome. Using Partial is the way to go here. Just wish there was a way for libraries to specify what version of TypeScript their definition files require to compile.

robertknight:
I've just done some testing with TypeScript 2.0 and it looks like it is quite easy for users of earlier versions of TypeScript to add a "polyfill" for `Partial<T>` which would allow them to use these latest definitions:

```ts
type Partial<T> = T;
```

With this, any reference to `Partial<T>` is treated as if it were just `T`. On that basis, one option is that we could make this change but perhaps add a comment in the type definitions for the benefit of anyone using earlier TS versions, eg:

```ts
// Note: Partial<T> is a TS 2.1 feature. See <link to this PR>
setState(state: Partial<State>)
```
Thoughts?

SleeplessByte:
@robertknight The problem with that "polyfill" is that it's not. Nothing would be optional. Instead aliasing as any would be closer even :(

I think saying that you need TS 2.1+ is fine. TS 2.0 to 2.1 does not introduce breaking changes, even we upgraded from 1.3 to 2.1 and it was fine. Everything was stashed away with flags anyway.

robertknight:
> @robertknight The problem with that "polyfill" is that it's not.
> Nothing would be optional. Instead aliasing as any would be closer even :(

That's why I put "polyfill" in quotes. It would then work the same as it does now using the current `setState` typings, so there would be no regression. Once users update to TS 2.1 they will be able to take advantage of the proper feature.

So to update the status here, the current recommendation from the TypeScript team based on the discussion for the React typings is that there should be a grace period after a new TS release before releasing typings updates that depend on the new release.

I would suggest that we track https://github.com/DefinitelyTyped/DefinitelyTyped/pull/13155 and once that is merged then we can do the same here. That way Preact and React typings will be in sync.

SleeplessByte:
Then yup. I completely agree.

robertknight:
The corresponding PR for React has been merged upstream. `setState` in the React definitions uses `Pick` rather than `Partial`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/99d41f818c2de87790ce1f02956cbc39bce40f40/react/index.d.ts#L169

SleeplessByte:
@robertknight yes. The dropped "optional key"  support in state and chose pick.

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/13155 based of https://github.com/Microsoft/TypeScript/issues/12793#issuecomment-266055843

Preact could follow this.

src/preact.d.ts
```diff
@@ -45,7 +45,7 @@ declare namespace preact {
 
 		linkState:(name:string) => (event: Event) => void;
 
-		setState(state:StateType, callback?:() => void):void;
+		setState(state:Partial<StateType>, callback?:() => void):void;
 		setState(fn:(prevState:StateType, props:PropsType) => StateType, callback?:() => void):void;
```
vutran: Should be able to return a partial state here as well.

mseddon:
Updated to use `Pick`, and added the other signature.

mseddon:
🎉

developit:
Awesome, so happy to have this merged :D

PR #468, opened 2016-12-22
Stop stringifying true values in h

jxodwyer:
# Fix Issue https://github.com/developit/preact/issues/433
It has been requested that we would no longer want to render true bool values. Similar to that of functionality seen in React.

##### Input:
`<div>{true}</div>`
##### Output:
`<div>true</div>`

##### Input:
`<div>{true}</div>`
##### Output:
`<div></div>`

### DIFF
--- a/src/h.js
+++ b/src/h.js
@@ -28,8 +28,8 @@ export function h(nodeName, attributes) {
 		if ((child = stack.pop()) instanceof Array) {
 			for (i=child.length; i--; ) stack.push(child[i]);
 		}
-		else if (child!=null && child!==false) {
-			if (typeof child=='number' || child===true) child = String(child);
+		else if (child!=null && child!==true && child!==false) {
+			if (typeof child=='number') child = String(child);
 			simple = typeof child=='string';
 			if (simple && lastSimple) {
 				children[children.length-1] += child;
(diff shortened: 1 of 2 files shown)


developit:
Nice! Was toying with this, and I like your solution.

jxodwyer:
Whoa that was fast! Thanks a lot @developit.

src/h.js
```diff
@@ -28,8 +28,8 @@ export function h(nodeName, attributes) {
 		if ((child = stack.pop()) instanceof Array) {
 			for (i=child.length; i--; ) stack.push(child[i]);
 		}
-		else if (child!=null && child!==false) {
-			if (typeof child=='number' || child===true) child = String(child);
+		else if (child!=null && child!==true && child!==false) {
```
NekR: `typeof child != 'boolean'` maybe?

jxodwyer:
@NekR nice catch. Just updated

developit:
@jxodwyer @NekR strangely, the `!='boolean'` check is 4 bytes bigger after gzip. Odd! Maybe do the verbose one?

jxodwyer:
@developit that's super odd. Just reverted it back

src/h.js
noomorph: @NekR , @developit, isn't it slower? There's a small difference between these two ways in Chrome (in favor of typeof), but Firefox, for example, is 2x faster with plain comparison (`=== true || === false`). 

https://esbench.com/bench/58de1a6299634800a0347ba9

developit: I ended up opting to side with gzipped byte size here, which favors the explicit `true` / `false` check.

PR #471, opened 2016-12-22
Reuse type checking functions

mkxml:
I found out there are some bytes to be shaved by reusing `util`'s type checking functions.

These changes provide a very subtle bundle size reduction while improving code readability and flexibility.

Added an `isObject` function for free because this type was being checked many times.

Didn't add other `is***` functions yet because doing that would actually increase the current bundle size.

Please tell me if this is OK or not.


developit:
Looks good, happy to shave those bytes.  I'll need to double check the performance implications, but it shouldn't be too different.

andrewiggins:
Closing this PR since the `master` branch dramatically changed between Preact 8 and Preact X and I think this PR is open against Preact 8. Do re-open if you think this an error! :)

Issue #282, opened 2016-08-21
View not updated with preact-redux and immutable


delta-9:
I use Preact and preact-redux, with an immutable state.

I have a PictureList component connected to the state, and the props are updated. (I log nextProps in componentShouldUpdate() and in render())

When I start the app PictureList is empty, then I receive a list of ID and I print one component (Picture) by id (with map)..

The render() function of my PictureList component is called after the state update, but.. the DOM is not updated, and the components of the list are not mounted (Picture componentDidMount is never called)

I have no errors, I use ES6 Component, and preact-compat. When I switch to react everything work as expected.

I think there is a link with the immutable state because when I pass an array instead of an immutable List to the props, the DOM and the Picture's component are mounted.

I didn't found any discussion on the use of immutable with Preact, so I'm not sure if someone already had this problem and can point me to a solution..

developit:
Hi @delta-9 - you're right, there is likely a compatibility fix needed here for Immutable.js.  I have seen this issue come up previously but nobody has yet asked for a solution. Right now Preact doesn't do anything special to account for immutable's objects.  I've never actually used it, so I'll probably have to do some digging to figure out what the most logical approach would be to handle this transparently.  It might be as simple as automatically unboxing (`.toArray()`?) `children` in preact's `h()` method.  If the size is large enough, I would imagine this feature would end up in `preact-compat`.

delta-9:
To push this forward, I made a repo with a simple step to reproduce the issue. The main component print two component connected to redux. One get an Immutable List, the other get the List as array(). (the computation happened in the mapping in connect()).

Also I use babel aliasing to use preact instead of react.. (so the import look like it's react but not)

to use it simply do 
```npm i``` then ```npm run start```

There it is : https://github.com/delta-9/testPreact

delta-9:
I added two test case to the repo, it seems that simply if you pass an Immutable object to a component, this one doesn't render... (render() is called however render() is called..). 

However @developit  thanks for your help on this

delta-9:
Here : https://57ba0c9271e20a478e8c64e0--plumber-monkey-66073.netlify.com/

developit:
Awesome work @delta-9. Looks like the child iteration is failing, which is expected. 

For the array casting, is there a generic method that is generally used to get the raw JS value for an Immutable type?

delta-9:
I commit a new branch that use react with the same code

Here with preact : http://plumber-monkey-66073.netlify.com/
[...]

PR #476, opened 2016-12-28
[WIP] Add support for all iterable children, not just Arrays

neeharv:
This is still WIP, as this fix obviously changes behaviour wrt the children being added to the stack in reverse order.

I wanted to have a discussion on the best way to push the children on to the stack, considering there is no reverse iterable protocol (yet). I could push them all to an array and then call `.reverse()` on it, but I'm unsure of the perf implications if there are a large number of children.


developit:
Ah! I should have looked before I responded in 282! I'll take a look at this as soon as I can, happy to see the PR!

developit:
Initial thought: it would be good to find a way around relying on Symbol here.

NekR:
Also, Babel's `for..of` generated a lot of code because it relies on `Symbol` and then fallbacks to `Array`. So check for `Symbol` might not be needed there if you are going that way.

developit:
Ah yeah I want to avoid anything transpiled/polyfilled entirely if we can. Perhaps convert iterables to Arrays? We only ever care about the values.

NekR:
TBH I just don't see why support iterables right now. I mean, they aren't everywhere supported natively so typically you will need to convert, say `HTMLCollection` to `Array` anyway in your code. 

The only possible way here when it might be needed is when developers provide fully supported environment on their own, i.e. run polyfill before preact. Sounds like an edge case to me (though, of course, I can wrong).

neeharv:
@NekR well, my personal use case stems from developit/preact-compat#262. Since React allows iterable structures as children, it would be reasonable to assume that other developers coming to Preact would also expect similar behaviour. Additionally, not restricting iterable children to Arrays opens up use cases for libs like `Immutable` or newer data structures like `Map` and `Set`, heck even generators to be used. Constraining userland code to convert these data structures to Arrays may be impractical and a big loss for usability IMO.

neeharv:
@NekR @developit So if I understand correctly, there are 3 choices in front of us 

1. Polyfill `for...of` and `Symbol`, and go with the approach in this PR
   - Pros - Idiomatic way to do it (IMO), plus easier to grok
   - Cons - Increase code size, perf hit (?), need to reverse iteration order to preserve current stack order semantics
2. Convert iterables to an array using [`Array.from`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/from)
    - Pros - Keeps existing way of iterating and pushing to stack the same
    - Cons - Needs polyfill, unsure of perf hit
3. Don't support iterables as children, and add a nice-ish error when non array-like iterables are passed down, letting users know they need to convert to Arrays before calling it down
    - Pros - No changes
    - Cons - Not very intuitive when coming over from React, having this allows more idiomatic patterns of constructing children, doesn't allow use with black box data structure libraries like `Immutable`

NekR:
First 2 ways do 2 iterations (reverse + iterate; from + iterate). If we are
going to support this, then we should write iterater on our own and avoid
traspiled code and double iterations.

robertknight:
> Not very intuitive when coming over from React, having this allows more idiomatic patterns
> of constructing children, doesn't allow use with black box data structure libraries like Immutable

Hmm. If you are using Preact in order to reduce the script size of your application, are you likely to be using a library like Immutable which is much larger than Preact itself as opposed to something like [seamless-immutable](https://github.com/rtfeldman/seamless-immutable)?

I'll throw a variant of option (2) (Convert iterables to an array using Array.from) into the mix. Convert iterables to an array using `Array.from`, but only if the environment already provides `Array.from` (either natively or via a polyfill).

Pros: Less code in applications that don't need the feature or already have the polyfill
Cons: Consumers of the library may depend on this feature without realizing that a polyfill is required

developit:
Ahh I like that option @robertknight. If `Array.from` provides a tiny way to create an Array from an Iterable, it seems like a reasonably direct solution. For performance, maybe we recommend a well-suited polyfill or allow defining an iterator-to-array function on `options`?

neeharv:
@developit @robertknight That seems like a reasonable solution. The only caveat I have is that if we do something along the lines of detecting if `Array.prototype.from` exists, and only then allow this behaviour, consumers might use it without realising that they need to polyfill it.

Should we then do the following - 

1. If child is an Array, continue existing behaviour
2. If child is an iterable, then convert it to an array using `Array.prototype.from`, but also notify the user in dev mode (?) that if you're using iterables, you need to add an `Array.from` polyfill (I like the idea of a userland polyfill more than passing down a iterable->array function down as options, since `Array.from` is the semantic way of converting iterable->array)
3. If an iterable is used, and `Array.from` isn't available, warn the user and discard those children

neeharv:
>Hmm. If you are using Preact in order to reduce the script size of your application, are you likely to >be using a library like Immutable which is much larger than Preact itself as opposed to something >like seamless-immutable
@robertknight `Immutable` is just an example of a userland data structure lib. With the introduction of the `iterable` protocol, there exists an interoperable way to use older data structures (`Array`), ES2015 ones (`Map`, `Set`) as well as third party libs. Going forward, I'd expect lots of people to use whatever data structure fits there need, knowing that libraries have a standardised way of iterating over them. 

As for my personal use case with preact, I'm switching to it not just because of size, but because it has the best bits of React without any of the cruft. I don't need a synthetic event system, I don't need propTypes validation, so why pay for that cost. With preact + immutable, my js is still well under 100kb, and with SSR that feels blazing fast on 3G!

NekR:
@neeharv unfortunately there is no dev mode in Preact yet.

Just some code to consider:
image

developit:
Hmm. If is there a way to manually pull an Iterable into an array? We only care about values. I have done the reverse (feign Iterable API using an Array to avoid relying on Iterables themselves).

robertknight:
> Hmm. If is there a way to manually pull an Iterable into an array?

Using Babel's transpilation of `for ... of` to handle the iteration protocol:

```js
function from(iterable) {
  // TODO: Check for `Symbol` and `iterable[Symbol.iterator]` to test whether this
  // object is actually an iterable.
  let ary = [];
  for (let item of iterable) {
    ary.push(item);
  }
  return ary;
}
```

neeharv:
> Using Babel's transpilation of for ... of to handle the iteration protocol:

This is exactly what I have done in this PR 😄 @robertknight 

>In addition, since true iterables can not be polyfilled, this implementation does not support generic iterables as defined in the 6th edition of ECMA-262.

From [here](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/from)

@developit even if one were to use a "dumb" polyfill for `Array.from`, we can't support generic iterables. 

The only way to do that would be to use `Symbol.iterator`. I see no way around adding support for iterables without some sort pf polyfill at the very least. Even to detect that a child passed in is an iterable would need a polyfill for `Symbol.iterator`

I could do some benchmarking in terms of build size if `Symbol` and `Array.from` are polyfilled. Let me know your thoughts on adding this to preact. Personally I feel it'd be a hugely limiting not to have it going forward, but I understand `preact` prioritises size and speed!

NekR:
@developit can we just have `option.toArray` hook of something like that? Then those who need iterables work could just hook in that functionality. We can even publish it to npm as separate package, I guess.

robertknight:
> This is exactly what I have done in this PR 😄 @robertknight

Ah, sorry. I wasn't paying enough attention 🤦‍♂️ 

> Even to detect that a child passed in is an iterable would need a polyfill for Symbol.iterator

If the child passed in is an iterable presumably that means that `Symbol.iterator` must already have been defined or polyfilled in order to implement the interface. So Preact can just check for `Symbol` being an object and having an `iterator` property - it doesn't even need to provide the implementation itself.

neeharv:
> If the child passed in is an iterable presumably that means that Symbol.iterator must already have been defined or polyfilled in order to implement the interface. So Preact can just check for Symbol being an object and having an iterator property - it doesn't even need to provide the implementation itself.

That makes sense! I'm bummed out I didn't think of that. I've implemented a simple iterable to array util function, plus updated tests to use both built in `Set` and `List` from `immutable.js` to also have a test against #282 

This PR should be ready to be merged in now @developit.

P.S @robertknight  `typeof Symbol === 'function'` 🙃

developit:
Wow, awesome work guys.

src/h.js
```diff
@@ -1,6 +1,8 @@
 import { VNode } from './vnode';
 import options from './options';
+import { iterableToArray, isFunction } from './util';
 
+const hasSymbolSupport = isFunction(Symbol) && Symbol.iterator;
```
robertknight: `isFunction(Symbol)` triggers a `ReferenceError` if Symbol is undefined. Using `typeof Symbol === 'function'` is OK however.

developit:
@robertknight figured that'd throw, I should have said something. Might actually be able smaller to check `if (typeof ATTR_KEY!='string')` - [`ATTR_KEY` is a Symbol when Symbol is available](https://github.com/developit/preact/blob/master/src/constants.js#L10). Or maybe combine the checks into one `if()`.

FWIW I need to check a few things relating to size and perf before we merge this. Ideally I'd like to do that prior to this hitting master so it doesn't hold back `7.2`, which is probably able to be released at this point.

neeharv:
Changed the check to use `ATTR_KEY`. Slightly compacted the `iterableToArray` function. Code size diff with this PR

branch | minified + gzip  | minified  |
 ------ | ------------- | ------------- |
7.1.0      | `3921`  |`9300`   |
This PR| `3999`  | `9514`  |

developit:
Wow thanks for making those changes @neeharv! I think we're getting close now 👍

neeharv:
Any update on this? Fwiw this is one of the blockers for us moving over to preact for a major web app! If there are any concerns or changes @developit, happy to discuss and work on that.

developit:
@neeharv just size is an issue right now. I was holding off on this until the size could be played off against the fairly major reductions in 8.0, though it'd still be amazing if there was some way to further reduce things.  Any and all ideas super appreciated, your work here has been really helpful already.

src/h.js
```diff
@@ -28,6 +29,10 @@ export function h(nodeName, attributes) {
 		if ((child = stack.pop()) instanceof Array) {
 			for (i=child.length; i--; ) stack.push(child[i]);
 		}
+		else if (typeof ATTR_KEY === 'symbol' && child[Symbol.iterator]) {
```
robertknight: `typeof ATTR_KEY === 'symbol'` avoids a type error if `Symbol` is not natively provided but presumably it also means that this won't work if Symbol is polyfilled, since I presume `typeof` is not polyfilled. I think that's an OK limitation for an initial merge, but something to be aware of.

robertknight:
OK, I think I'm good with this now. In terms of the size impact, I think we probably want to come up with a project-wide means of making a decision about what to include and what to leave out. Perhaps something like a goal size to stay under (eg. 5KB min+gz) plus a checklist of questions to answer whether a change should go in Preact itself or be left to an external module?

robertknight:
Hmm, looks like there are failing tests that need looking into I'm afraid.

developit:
Total goal size is actually 3.5kb which we've been above for a long time.  I think that's why I've been a little wary of any changes that bump up the file size at all - Preact is supposed to be smaller than it currently is.  I'm going to publish some breaking changes to a branch for everyone to check out that should help cut size down in prep for 8.0.

JustinMGaiam:
Did this work ever make it into Preact?  We are trying to convert a project that uses Redux, React Redux,  React Router and ImmutableJS into Preact and have gotten everything to work except many of our component's children are ImmutableJS Lists. We end up with a string printed to the screen with the ImmutableJS structure instead of it iterating over the List.  We went with ImmutableJS since it was natively support in React so it has deeply embed itself in this project.

--- The Comment below was originally posted as a comment in #282 so I thought I would add it here --- 
To fix this issue if you check to see if the method `.toJS()` exists, it is available on all ImmutableJS elements and will return the plain Javascript equivalent of that object.  There is also a `.map()` and `.reduce()` method available if that is better implementation.

`List(['a', 'b', 'c']).toJS() == ['a', 'b', 'c']`
`List(['a', 'b', 'c']).map((v, k) => return v) == List(['a', 'b', 'c'])`
`List(['a', 'b', 'c']).reduce((r, v, k) => r = r + v) == 'abc'`

developit:
Not yet! It's on the roadmap for 8.

developit:
@JustinMGaiam also, I think this issue covers all Iterables, not just ImmutableJS.

src/h.js
neeharv: @robertknight what would you then suggest as a check for `Symbol` to fix this? This is also why the tests fail when Symbol is polyfilled.

robertknight: `typeof Symbol !== 'undefined'` should work with both native and polyfilled Symbol implementations.

JustinMGaiam:
@developit thanks! I look forward to version 8 so we can start looking at Preact as replacement for React.

robertknight:
The tests are failing after merging in changes from master. I don't suppose you'd be able to take a look @neeharv .

@developit what is your plan for this PR and Preact 8? Did you manage to make sufficient size savings elsewhere?

developit:
I've made some savings yes, but since we are aiming for 2kb it's still tight. I might have a way to cut down further though. If you're available on slack, ping me

src/h.js
Havunen: uh huh, I dont know how I have missed this comment. Its now failing in IE, lets fix it for next release.

neeharv:
Will take a look over this weekend and get the tests passing. Would love to
have this in v8!

developit:
@neeharv Agreed!

neeharv:
Did something in the latest master change wrt karma/phantomJS? I think I fixed the failing issue, but didn't touch any test code. Now phantomJS hangs completely when running the test suite, on local as well as on Travis. My knowledge of PhantomJS and Karma is limited, so any help would be appreciated.

developit:
I've been seeing this in other repos, I think it's a patch release of a dependency that just broke everything (annoying).  FWIW I usually use the Chrome runner:

```js
npm i -D karma-chrome-launcher

npm run test:karma:watch -- --browsers Chrome
```

neeharv:
I've been trying to merge the latest master into this and finally get this done, but I've run into an issue.

In `h.js`, we check if the child is an array, using

```
if ((child = stack.pop()) && child.pop!==undefined) 
```

whereas earlier there was an `instanceof Array` check. The problem with this is that third party collection libs may implement a `.pop()` method, (for eg- Immutable.List) and then this breaks. 

Suggestions on the best way to fix this? I have the code working fine with native collection objects like Set / Map using the ES6 iterator, but hard to find a fix for libs like Immutable without going back to the `instanceof` check.

developit:
Hmm. Might have to switch to `Array.isArray()` even though the performance is much worse :(

neeharv:
I can't seem to figure the browser tests out - everything else works fine in terms of testing, using phantom or chrome with karma just hangs the test and errors, locally or in CI. I can't see the exact errors, so don't know how to fix it. @developit do you think you could take a look? Everything else looks good.

I'd really love to get this in, will help us switch insider.in over to preact!

Emilios1995:
I also need this feature in order to switch to preact. Any new progress?

neeharv:
Ping. Any help @developit on this? Unable to figure how to get the tests to pass with Karma launcher, it just hangs for me.

developit:
The karma issue should be resolved thanks to downstream updates.

I know this is an unusual request, but if anyone is interested in Iterables as children, I'd ask you to please [join the Preact slack group](https://preact-slack.now.sh).  We have some stuff to talk about!

mq2thez:
Note that this unfortunately a blocker for Etsy migrating to Preact. While we would very much love to drop Immutable as a dependency, doing so is a long-term project due to the size of our codebase.

marvinhagemeister:
@mq2thez we're looking into it 👍

Thompson1985 (approved):

Issue #440, opened 2016-12-11
Dev Tools weird behaviour


ForsakenHarmony:
[Entry](https://github.com/ForsakenHarmony/feathers-test/blob/master/src/client/index.js) 

[Initial state of dev tools](https://files.gitter.im/developit/preact/usQe/blob)
[after changing route](https://files.gitter.im/developit/preact/r0dG/blob)

bundled with rollup

-----

Steps to reproduce:

- Create a Router with Routes
- Go into the Dev Tools

-----

[JSFiddle](http://jsfiddle.net/uz89k3f9/3/)

developit:
Did we determine that this was related to rollup?

ForsakenHarmony:
I've switched back to webpack and still have the same issue

robertknight:
Sorry I haven't gotten around to this yet. I'll try to take a look in the next couple of days.

robertknight:
OK, after some initial investigation it appears that the duplicate nodes happen without preact-router. This minimal test case also produces incorrect behavior:

```js
/** @jsx h */

const { h, render } = require('preact');
require('preact/devtools');

const Index = () => (
  <main>
   DevTools Test 
  </main>
);

const App = () =>
  <Index path='/'/>

render(<App />, document.body);
```

Result:

![devtools-dupe](https://cloud.githubusercontent.com/assets/2458/21674095/2bf2c96a-d324-11e6-9a8c-8d1809b2ae2d.png)

developit:
Maybe something to do with compositional components then, since both Router and your repro use it. Good find.  I'll see about pushing my devtools changes branch - I wasn't able to get it working (changed too much at once) but it might be a source of ideas.

developit:
Nice - good find. Maybe combine the two checks?

```js

return !component._parentComponent && !(component.base.parentNode && component.base.parentNode[ATTR_KEY]);
```

robertknight:
After some further investigation, it looks like the problem may be due to the order in which changes are reported to the devtools. With the default ordering, a change of route results in the update to the `Router` component being reported before the mounting of the child component for the active route.

Adding logic to sort updates so that mounts of new components in a given batch of updates are reported before updates to existing components, including the parent `Router` in this case, resolves the problem. I need to do a bit more digging to verify that this is actually the problem though.

robertknight:
So the problem is definitely due to the difference in ordering of lifecycle callbacks for children and parents when an update occurs. Given a structure like this:

```
<Parent>
[...]

PR #488, opened 2017-01-06
Fix detection of whether a component is a root or not

robertknight:
This fixes one of the issues reported in #440 where the devtools shows multiple roots instead of only one (`<App>`). The problem was incorrect reporting to the devtools about whether a component was a root or not.

When a root component whose only child was a composite was mounted, the
previous method of detecting a root component would incorrectly report
both the actual root and the child as root components.


robertknight:
The other problem reported in #440 where the devtools do not reflect the updates to the tree properly when changing routes still exists.

PR #491, opened 2017-01-08
Fix eslint warnings

mkxml:
I think we can remove those eslint warnings so we can spot real code mistakes easier.

I'm sending three commits just to make the changes easier to understand, could squash them later.

Build and tests OK in my local env.

Please review this and see if it's suitable for merge.

Thanks!

### DIFF
--- a/test/browser/performance.js
+++ b/test/browser/performance.js
@@ -21,7 +21,7 @@ function loop(iter, time) {
 
 
 function benchmark(iter, callback) {
-	let a = 0;
+	let a = 0; // eslint-disable-line no-unused-vars
 	function noop() {
 		try { a++; } finally { a += Math.random(); }
 	}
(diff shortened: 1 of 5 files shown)


test/browser/performance.js
```diff
@@ -21,6 +21,7 @@ function loop(iter, time) {
 
 
 function benchmark(iter, callback) {
+	/* eslint-disable no-unused-vars */
```
robertknight: `eslint-disable` is not block-scoped, so this applies to the entire file from this line on rather than just this function. You can scope the lint ignore to just the `a` variable using:

```
let a = 0; // eslint-disable-line no-unused-vars
```

mkxml: Oh, that's correct. Fixing it.

mkxml: Thanks! I pushed a new commit with the change.

Issue #493, opened 2017-01-10
render() can fail to unmount top-level components 


developit:
As found by @insin 🙇

PR #494, opened 2017-01-10
Fix top-level component unmount issue.

developit:
Fix issue where top-level components could be left hanging instead of being unmounted. Found by @insin, issue is #493.

### DIFF
--- a/src/vdom/component.js
+++ b/src/vdom/component.js
@@ -203,6 +203,7 @@ export function renderComponent(component, opts, mountAll, isChild) {
  */
 export function buildComponentFromVNode(dom, vnode, context, mountAll) {
 	let c = dom && dom._component,
+		originalComponent = c,
 		oldDom = dom,
 		isDirectOwner = c && dom._componentConstructor===vnode.nodeName,
 		isOwner = isDirectOwner,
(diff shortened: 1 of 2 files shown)


developit:
Demo of old broken behavior: http://jsfiddle.net/developit/p0zpf7th/

NekR:
Oh, so previously `c` was overriden by `c=c._parentComponent` even though it doesn't exist so `c`  was `underfined`/`null` at `if (c && !isDirectOwner) {`, yeah?

developit:
@NekR exactly. I think the `while` was added after the latter usage of `c`, and I forgot to store the old value.  I can't find anything that this breaks (checked the preactjs.com site and a few others) and it passes all the tests. Thinking it's pretty safe, particularly if the logic seemed sound.

NekR:
Yeah, seems like just a bug fix to me (important bug fix!). I don't see any breaking changes, so LGTM.

NekR:
Though I didn't review the test..

developit:
Test was just a port of @insin's findings. Mount a new component in place of an old one at the top level (from `render()`) and make sure `componentWillUnmount()` got called. It failed against master, passes with the change.

NekR:
@developit okay, seems fine assuming test fails without the change.

developit:
perfect. lets do this.

lukeed:
Looks good to me! Injected into my current projects (even tho no root unmount) & nothing breaks.

developit:
awesome, thanks a ton!

Issue #492, opened 2017-01-09
simple property assignment fails to set img width


enapupe:
I'm facing some incompatibility when replacing react with preact.
It seems the way preact sets attributes (https://github.com/developit/preact/blob/master/src/dom/index.js#L68 - simply by setting a property) is not safe.

My use case is a simple `img` component that receives `width` and breaks with the value of `100%` because of the the following:

```
const img = document.createElement('img');
img.width = '100%';
console.log(img.width); // 0
```

And what I think should be done (this is what react does)
```
const img = document.createElement('img');
img.setAttribute('width', '100%');
console.log(img.width); // expected value (automagically defined by the browser)
```

Unfortunately, changing `setProperty` to 
```
function setProperty(node, name, value) {
  try {
    node.setAttribute(name, value);
  } catch (e) { }
}
``` 
isn't enough.. I think we might have to do some checks before knowing which kind of property setting we should do (simple assignment or `setAttribute`)..  Any thoughts?

developit:
Interesting issue for sure. Preact's size basically means whitelists are impossible, so it seems like we'll need a way to determine this programmatically, or a full move (back) to attributes for setters. Maybe `String` values could trigger attribute setting?

robertknight:
For reference, the relevant part of the React code is https://github.com/facebook/react/blob/master/src/renderers/dom/shared/DOMPropertyOperations.js#L133 which references the attribute -> config bitflags defined in https://github.com/facebook/react/blob/master/src/renderers/dom/shared/HTMLDOMPropertyConfig.js

Looking at that second file, I'll note that the number of attributes which are marked as being required to be set as a property rather than an attribute is pretty small - only `checked`, `muted`, `multiple` and `selected`.

leeoniya:
@developit your best bet is probably sticking with attribute setters as the default and have a short whitelist of very common ones that can be set by prop. "class" and "id" bascially cover most of the perf-critical stuff.

also as @robertknight points out, the stuff that *should* be set by property assignment is also pretty small.

robertknight:
> Maybe String values could trigger attribute setting? (just pondering)

This idea seems worth exploring. It seems conceptually nicer to use property setters where possible and if Preact were to use `setAttribute` for boolean properties, you'd need some additional logic to translate the JS bool value to a string.

developit:
[...]

PR #495, opened 2017-01-10
Add test case that ensures `width` prop is properly set using setAttribute

enapupe:
related to #492

### DIFF
--- a/test/browser/render.js
+++ b/test/browser/render.js
@@ -406,6 +406,20 @@ describe('render()', () => {
 		expect(scratch.innerHTML, 're-set').to.equal('<div>'+html+'</div>');
 	});
 
+	it('should apply width property via setAttribute', () => {
+		class Thing extends Component {
+			render({width}) {
+				return <img width={width} />;
+			}
+		}
+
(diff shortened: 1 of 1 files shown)


enapupe:
Hey!
Today I crossed with this video https://www.youtube.com/watch?v=LY6y3HbDVmg that reminded me I should be using preact in the project that lead me to open this PR. 
But months passed and major version later and we still haven't fixed this image with issue? 
What the big deal? I mean, I really consider this to be a huge issue since it must affect many people trying to replace react with preact.. 
What's your PoV here? I'm confused about your goals.

developit:
Just an update on this: one of the goals of Preact is to avoid adding additional abstraction over the DOM. It's important that any means of fixing these inconsistencies with React can be accomplished without custom mappings/behavior.

marvinhagemeister:
Just checked the [spec](https://html.spec.whatwg.org/multipage/embedded-content-other.html#attr-dim-width) again and only pixel values are valid for `width`. Using anything else like percentages is invalid HTML and may lead to rendering differences in different browsers.

PR #502, opened 2017-01-12
update setState def. to support func.

vutran:
Small update to the TypeScript definitions for `setState()` to support for function signature.

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -45,7 +45,9 @@ declare namespace preact {
 
 		linkState:(name:string) => (event: Event) => void;
 
-		setState(state:StateType | ((prevState:StateType, props?:PropsType) => StateType), opts?:any):void;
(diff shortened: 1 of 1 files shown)


developit:
Good call! Thanks @vutran

src/preact.d.ts
```diff
@@ -45,7 +45,7 @@ declare namespace preact {
 
 		linkState:(name:string) => (event: Event) => void;
 
-		setState(state:StateType, opts?:any):void;
+		setState(state:StateType | ((prevState:StateType, props?:PropsType) => StateType), opts?:any):void;
```
SleeplessByte: In a callback, don't make something optional. You are always allowed to give a callback with less parameters.

SleeplessByte: It's clearer to do a function overload instead of a union type.

PR #506, opened 2017-01-13
Call componentDidMount for children before componentDidUpdate for parent

### LINKED ISSUE #440: Dev Tools weird behaviour

robertknight:
Previously the `componentDidUpdate` hook was called immediately after a
component was updated but `componentDidMount` was deferred until the end
of the render cycle in order to ensure that the corresponding DOM
element was actually present in the document when the callback was
invoked.

The result was that after an update to a component which mounted new
child component instances, the parent's `componentDidUpdate` would be
called before the child's `componentDidMount`. This broke the assumptions
of the React Dev Tools and possibly other projects.

Fix the problem by flushing the queue of mount callbacks before
executing any `componentDidUpdate` callbacks. See the commit message for a
more detailed explainer.

Fixes #440


developit:
Need as many eyes on this PR as we can. It passes tests, and @robertknight and I couldn't think of any reasons why it would be bad to flush early here. However, it's a major change to the deferral mechanism. Any comments, logs, inspections, etc are appreciated!

NekR:
I'm not sure I 100% get, but AFAIK `componentDidUpdate` shouldn't be called when `componentDidMount` is called, i.e. on initial render. Going to read commit message more carefully now though.

robertknight:
> I'm not sure I 100% get, but AFAIK componentDidUpdate shouldn't be called
> when componentDidMount is called, i.e. on initial render.

That is correct. Whenever a component is mounted, preact puts the component on a queue to trigger the `componentDidMount` callback. The reason for this deferral is that the callback cannot be called until the new element is actually added to the document. If a set of nested components are mounted, they can't be called until the top-most component's DOM element is inserted into the DOM.

Previously this queue was always flushed at the very end of the render cycle. However here I'm relying on the fact that given the current rendering order, it should also be safe to flush the queue whenever we get to updating an existing component.

In the original bug report, the structure was initially this:

```
<Router>
  <Index/>
</Router>
```

After a route change, the new structure looks like this:

```
<Router>
  <Login/>
</Router>
```

Previously, the `componentDidMount` for the `<Login/>` component would be fired _after_ `componentDidUpdate` for the `Router`. It will now be fired just before. At the point when we update `<Router/>`, we know that any new children in this update will have been mounted and inserted into the DOM.

developit:
Just reminding myself that this is still open. Need to verify some things and then merge.

developit:
@robertknight I think this might fix an compat issue with React Bootstrap.

robertknight:
Is the React Bootstrap issue filed anywhere?

developit:
https://github.com/developit/preact-compat/issues/308

developit:
@robertknight Let's get this into the next 8 beta. Then we can just watch for reports of any lifecycle oddities.

ustccjw:
just one thing: parent's componentDidUpdate, children' componentDidMount should be called after dom updated.

PR #510, opened 2017-01-15
Add README demo: preact-starter

lukeed:
Differs from `preact-boilerplate` because:

1. mine is built with Webpack2, 
2. has built-in service worker handlers, thus enabling offline support, and 
3. has a 100/100 lighthouse score ootb. (Hence the :100: emoji 😜)


developit:
@lukeed you should add it to preact-www too.

lukeed:
@developit Thanks, didn't even think about it.

PR #511, opened 2017-01-17
check for typeof attribute value, if is string use setAttribute

enapupe:
I wasn't able to run react tests against this change but I tried to change the logic as minimum as possible. 

ref #492

### DIFF
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -64,7 +64,7 @@ export function setAccessor(node, name, old, value, isSvg) {
 		}
 		l[name] = value;
 	}
-	else if (name!=='list' && name!=='type' && !isSvg && name in node) {
+	else if (typeof value !== 'string' && !isSvg && name in node) {
 		setProperty(node, name, value==null ? '' : value);
 		if (value==null || value===false) node.removeAttribute(name);
 	}


src/dom/index.js
```diff
@@ -64,7 +64,10 @@ export function setAccessor(node, name, old, value, isSvg) {
 		}
 		l[name] = value;
 	}
-	else if (name!=='list' && name!=='type' && !isSvg && name in node) {
+	else if (typeof value === 'string' && !isSvg && name in node) {
```
developit: Think it'd be possible to invert this conditional and have String fall through to the complete attribute add/update/remove flow on line 75?

enapupe: Done. Tests ✅

developit:
Looking at the travis build, it actually seems like this might be better for performance! (not sure how but I'll take it)

enapupe:
Don't forget to consider merging https://github.com/developit/preact/pull/495 after fixing this one.

developit:
@enapupe yup yup, I'll pair up the two.  These are likely to go into `8.0` as the behavior probably constitutes a breaking change, or at least it has a chance to be "breaking".

robdodson:
I was chatting with @developit about this PR today and he suggested adding a check to see if the value being set is an object and if so, falling back to using a property setter. I tried to outline the expected behavior in [this comment](https://github.com/developit/preact/issues/678#issuecomment-315218826) (sorry it's a bit long!)

andrewiggins:
Since this is opened against v8, and the original issue (#492) has been closed, I'm going to close this PR. If there is a another situation where this change would be beneficial in Preact X, open a new issue with that case, and we'll take a look.

PR #512, opened 2017-01-17
Allowing other types for key (Typescript definitions)

seveves:
When the id of my object is a number then I don't want to use id.toString() so I've added number and any as valid types for the key prop.


developit:
If anyone else wants to review, feel free. I am only barely learning TS 😆

vutran:
@seveves Is there a particular reason why `any` is included?

developit:
@vutran `key` can be any object, it just needs to be referentially equal across two renders in order for the diff to take it into account.  As an example, lots of people iterate over an Array of Objects and render components for each, passing the Objects themselves as the `key` prop.

vutran:
> passing the Objects themselves as the key prop.

Didn't know this was a thing. Makes sense now.

PR #515, opened 2017-01-18
Enforce a functional style to the hashToClassName

finico:
and saved some bytes 😄

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -45,14 +45,7 @@ export function isString(obj) {
  *	@private
  */
 export function hashToClassName(c) {
-	let str = '';
-	for (let prop in c) {
-		if (c[prop]) {
-			if (str) str += ' ';
-			str += prop;
-		}
-	}
-	return str;
+	return Object.keys(c).reduce((acc, x) => c[x] ? [...acc, x] : acc, []).join(' ');
 }
 
 


finico:
Coveralls informed that the coverage decreases.
But it looks weird.. the function works identically to the previous version

developit:
Hi @finico - I appreciate the PR!  I agree your implementation is much nicer to read. That being said, it is actually quite a bit slower due to the use of functions and increased memory allocation (the arrays and concatenation).  Also, I know the ES6 source looks a lot smaller, but the transpiled output that actually gets used is larger (shown [here in the Babel REPL](https://goo.gl/n9gcd2)).

I know this is a little counterintuitive and not as nice to read.  For performance-sensitive libraries like Preact, it's often the case that functional programming is out of the question as it can't be optimized to the same extent by the JS engine.

In order to make sure we're working from facts here, I've put together a fairly real-world benchmark showing the current implementation, and three variants of your suggested approach (reducing an array, a string, hoisting of the reducer).  I agree that they are much nicer looking code, but the benchmark shows the existing "C-like" implementation is worth keeping:

<a href="https://esbench.com/bench/587f9dee99634800a034763f"></a>

Benchmark: https://esbench.com/bench/587f9dee99634800a034763f

Please let me know your thoughts.

finico:
@developit thank you for your deep analysis!
I agree. I now tried `String::concat` but it's also slower..
The speed in this case is more important
I'm closing the PR

developit:
@finico It's the least I could do! I'm happy to have your eyes on the code, it's the best way to move things forward.  I noticed you're not already in the Preact Slack chat, you might find it an interesting place to hang out! https://preact-slack.now.sh

PR #516, opened 2017-01-19
add more typescript typings - component lifecycle methods

vutran:
More typings for component lifecycle methods. I've renamed the `props` to `nextProps` to make it more obvious when implementing.


robertknight:
LGTM

PR #518, opened 2017-01-20
add preact-codemod to libs

vutran:
- transforms React to Preact
- more exposure => more attention

😃


developit:
How was this not already in there? haha

PR #539, opened 2017-02-08
Switches empty component root from Text nodes to Comment nodes

### LINKED ISSUE #537: Text nodes can't accept properties in IE8
Hi guys!

First of all, thanks for you work, this library is awesome!

I know that IE8 is not officialy supported, but i have to and in your 'Browser support' page you say:

> It should work fine in IE7 and IE8

I've fixed it locally with this patch:
```javascript
let supportsAddTextNodeProps;

try {
  document.createTextNode('').test = true;
  supportsAddTextNodeProps = true;
} catch (e) {
  supportsAddTextNodeProps = false;
}

function getEmptyVNode() {
  if (!supportsAddTextNodeProps) {
    return document.createElement('p-empty');
  }

  return '';
}
```
```javascript
// empty values (null & undefined) render as empty Text nodes
if (vnode==null) vnode = getEmptyVNode();
```

So, can we fix it?
Or at least, can you remove `"It should work fine in IE7 and IE8"`?

gdorsi:
Fixes #537

### DIFF
--- a/src/vdom/diff.js
+++ b/src/vdom/diff.js
@@ -47,7 +47,7 @@ export function diff(dom, vnode, context, mountAll, parent, componentRoot) {
 		hydrating = dom && !(ATTR_KEY in dom);
 	}
 
-	let ret = idiff(dom, vnode, context, mountAll);
+	let ret = idiff(dom, vnode, context, mountAll, componentRoot);
 
 	// append the element if its a new parent
 	if (parent && ret.parentNode!==parent) parent.appendChild(ret);
@@ -63,7 +63,7 @@ export function diff(dom, vnode, context, mountAll, parent, componentRoot) {
 }
 
 
-function idiff(dom, vnode, context, mountAll) {
+function idiff(dom, vnode, context, mountAll, componentRoot) {
 	let ref = vnode && vnode.attributes && vnode.attributes.ref;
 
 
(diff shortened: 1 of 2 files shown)


src/vdom/diff.js
```diff
@@ -73,8 +73,14 @@ function idiff(dom, vnode, context, mountAll) {
 	}
 
 
-	// empty values (null & undefined) render as empty Text nodes
-	if (vnode==null) vnode = '';
+	if (vnode==null) {
+		if (componentRoot) {
+			if (dom) recollectNodeTree(dom);
+			return document.createComment('');
```
developit: I think we might have to check if `dom` is already a Comment node here and retain it, otherwise the Component will get recycled.

src/vdom/diff.js
```diff
@@ -73,8 +73,17 @@ function idiff(dom, vnode, context, mountAll) {
...
 
-	// empty values (null & undefined) render as empty Text nodes
-	if (vnode==null) vnode = '';
+	if (vnode==null) {
+		if (componentRoot) {
+			if (dom) {
+				if (dom.nodeType==8) return dom;
+				recollectNodeTree(dom);
+			}
+			return document.createComment('');
```
gdorsi: Should i create a function to avoid too many returns?

developit:
@gdorsi I might have fixed this through some other refactoring of the renderer, but I will merge the test regardless.

gdorsi:
Ah ok, tomorrow I'll make a try!

developit:
@gdorsi ping me on [slack](https://preact-slack.now.sh), want to run my thoughts by you.  I'm going to try to push the 8.0 alpha branch tonight, maybe you can take a look?

marvinhagemeister:
For Preact X we switched to a `Fragment` as a root node. This allows to mount multiple elements right at the root :+1: Closing this PR as we don't have a single HTML root anymore.

PR #542, opened 2017-02-09
Typescript: Add `class` & `className` to `SVGAttributes`

shakyShane:
There are many SVG attributes missing, but I think `class` should be considered essential - especially as its absence breaks the TS compilation.

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -88,6 +88,8 @@ declare namespace JSX {
 	}
 
 	interface SVGAttributes {
+		class?:string | { [key:string]: boolean };
+		className?:string | { [key:string]: boolean };
 		clipPath?:string;
 		cx?:number | string;
 		cy?:number | string;


src/preact.d.ts
```diff
@@ -88,6 +88,8 @@ declare namespace JSX {
 	}
 
 	interface SVGAttributes {
+		class?:string | { [key:string]: boolean };
+		className?:string | { [key:string]: boolean };
```
vutran: Not sure if I'm missing something, but is there a reason why `className` can accept an object?

developit: It currently supports accepting an object, and compiles a String from the keys with truthy values.  However, this is almost certainly being removed in Preact 8.

robertknight:
Thank-you for the contribution and I'm sorry it took such a long time to review. In the meantime another PR (#609) was submitted which added extensive SVG attributes to the type definitions, including `class` and `className` by extending `HTMLAttributes`.

shakyShane:
Perfect!

PR #564, opened 2017-02-26
closure compiler experiment 🗿

samccone:
Over the past week I have been investigating what kind of optimization wins we could get with preact, if instead of using uglify we switched to using closure compiler.

After getting the typings in place and cleaning up some invalid typings in the code, I ended up with a bundle that would compile and pass some of my naive tests.

### Size Details:

Current master gzip size w/o closure compiler
**GZIP SIZE:** 3923

Current master gzip size w closure compiler
**GZIP SIZE:** 3755

Compiled bundle diff (after pretty printing diff):
https://gist.github.com/samccone/7eb10742f632e4de08af594de35e4aca

--------

### Opportunities for additional shrinking

Looking at the bundle we can see places where we can get some major wins:

If anyone can think of a smaller way to represent or create either of these structures please do share:

```js
        sa = {
            boxFlex: 1,
            boxFlexGroup: 1,
            columnCount: 1,
            fillOpacity: 1,
            flex: 1,
            flexGrow: 1,
            flexPositive: 1,
            flexShrink: 1,
            flexNegative: 1,
            fontWeight: 1,
            lineClamp: 1,
            lineHeight: 1,
            opacity: 1,
            order: 1,
            orphans: 1,
            strokeOpacity: 1,
            widows: 1,
            zIndex: 1,
            zoom: 1
        },
```

```js
        fa = {
            blur: 1,
            error: 1,
            focus: 1,
            load: 1,
            resize: 1,
            scroll: 1
        },
```

--------

Now the question is, is a 4% savings really worth the overhead and maintenance cost of closure compiler.. maybe maybe not..

~~So on that note I will close this out and leave it as an engineering artifact to potentially be referenced in the future. 🕊~~

<details>
```
**3/4/17**

ok now we are at 1 warning which I am not sure why closure is not matching up the types.. but... I choose to ignore it for now

dist/preact.dev.js:83: WARNING - cannot instantiate non-constructor
        var p = new VNode(nodeName, attributes || undefined, children || EMPTY_CHILDREN);
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
</details>

### DIFF
--- a/externs.js
+++ b/externs.js
@@ -12,7 +12,7 @@ function VNode(nodeName, opt_attributes, opt_children) {}
 function NodeName() { }
 
 /** @type {Object} */
-NodeName.prototype.defaultProps
+NodeName.prototype.defaultProps;
 
 /** @type {(string|!NodeName)} */
 VNode.prototype.nodeName;
@@ -21,7 +21,7 @@ VNode.prototype.nodeName;
 VNode.prototype.attributes;
 
 /** @type {Array<VNode>|undefined} */
-VNode.prototypechildren;
+VNode.prototype.children;
 
 /** @type{boolean|string} */
 VNode.prototype.key;
@@ -45,21 +45,6 @@ Component.prototype.prevContext;
 Component.prototype.base;
 /** @type {?} */
 Component.prototype.nextBase;
-/** @type {?} */
-Component.prototype._parentComponent;
-/** @type {?} */
-Component.prototype._component;
-/** @type {?} */
-Component.prototype.__ref;
-/** @type {?} */
-Component.prototype.__key;
-/** @type {?} */
-Component.prototype._linkedStates;
-/** @type {?} */
-Component.prototype._renderCallbacks;
-
-/** @type {boolean} */
-Component.prototype._dirty;
 
 /** @type {Object} */
 Component.prototype.context;
--- a/externs.js
+++ b/externs.js
@@ -1,13 +1,21 @@
 /**
  * @constructor
- * @param {string|Function} nodeName
+ * @param {(string|!NodeName)} nodeName
  * @param {Object|undefined} attributes
  * @param {Array<VNode>|undefined} children
  * @return {undefined}
  */
 function VNode(nodeName, attributes, children) {}
 
-/** @type {string|Function} */
+/** @record
+ * @return {?}
+ */
+function NodeName() { }
+
+/** @type {Object} */
+NodeName.prototype.defaultProps
+
+/** @type {(string|!NodeName)} */
 VNode.prototype.nodeName;
 
 /** @type {Object|undefined} */
--- a/externs.js
+++ b/externs.js
@@ -1,0 +1,103 @@
+/**
+ * @constructor
+ * @param {string|Function} nodeName
+ * @param {Object|undefined} attributes
+ * @param {Array<VNode>|undefined} children
+ * @return {undefined}
+ */
+function VNode(nodeName, attributes, children) {}
+
+/** @type {string|Function} */
+VNode.prototype.nodeName;
+
+/** @type {Object|undefined} */
+VNode.prototype.attributes;
+
+/** @type {Array<VNode>|undefined} */
+VNode.prototypechildren;
+
+/** @type{boolean|string} */
+VNode.prototype.key;
+
+
+/** Base Component class, for the ES6 Class method of creating Components
+ * @constructor
+ * @param {Object} props
+ * @param {Object} context
+ * @return {undefined}
+ */
+function Component(props, context){}
+
+/** @type {?} */
+Component.prototype.prevState;
+/** @type {?} */
+Component.prototype.prevProps;
+/** @type {?} */
+Component.prototype.prevContext;
+/** @type {?} */
+Component.prototype.base;
+/** @type {?} */
+Component.prototype.nextBase;
+/** @type {?} */
+Component.prototype._parentComponent;
+/** @type {?} */
+Component.prototype._component;
+/** @type {?} */
+Component.prototype.__ref;
+/** @type {?} */
+Component.prototype.__key;
+/** @type {?} */
+Component.prototype._linkedStates;
+/** @type {?} */
+Component.prototype._renderCallbacks;
+
+/** @type {boolean} */
+Component.prototype._dirty;
+
+/** @type {Object} */
+Component.prototype.context;
+
+/** @type {Object} */
+Component.prototype.props;
+
+/** @type {Object|undefined} */
+Component.prototype.state;
+
+/**
+ *	@param {Object} nextProps
+ *	@param {Object|undefined} nextState
+ *	@param {Object} nextContext
+ *	@return {boolean} should the component re-render
+ */
+Component.prototype.shouldComponentUpdate = function(nextProps, nextState, nextContext){};
+
+/**
+ *	@param {string} key		The path to set - can be a dot-notated deep key
+ *	@param {string} [eventPath]	If set, attempts to find the new state value at a given dot-notated path within the object passed to the linkedState setter.
+ *	@return {Function} linkStateSetter(e)
+ */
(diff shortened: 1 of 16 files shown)


developit:
So amazing 😍😍

xtuc:
I'm curious about the difference in generated code. Awesome job @samccone.

externs.js
```diff
@@ -0,0 +1,111 @@
+/**
+ * @constructor
+ * @param {(string|!NodeName)} nodeName
+ * @param {Object|undefined} attributes
```
samthor: `@param {Object=} opt_attributes`

(= is "this or undefined". Also currently this is `null`, `Object` or `undefined`. You can say `{!Object=}` to be `Object` or `undefined`.

externs.js
```diff
@@ -0,0 +1,111 @@
...
+ * @param {Array<VNode>|undefined} children
+ * @return {undefined}
```
samthor: you don't need this if the function doesn't return.

externs.js
```diff
@@ -0,0 +1,111 @@
...
+ *	@param {Object|undefined} nextState
+ *	@param {Object} nextContext
+ *	@return {boolean} should the component re-render
+ */
+Component.prototype.shouldComponentUpdate = function(nextProps, nextState, nextContext){};
+
+/**
+ *	@param {string} key		The path to set - can be a dot-notated deep key
+ *	@param {string} [eventPath]	If set, attempts to find the new state value at a given dot-notated path within the object passed to the linkedState setter.
+ *	@return {Function} linkStateSetter(e)
```
samthor: sounds like this should be `{function(!Event)}`? (lower-case function implies never-null for functions, and then you can include args).

developit: nice, didn't know about that notation.
FWIW linkState() is being removed.

RReverser:
It appears that `_`-prefixed properties were not marked as private and thus Closure Compiler had to preserve them in generated code?

developit:
@rreverser interesting, I wonder how much affect that would have on gzipped output size.

RReverser:
@developit Me too, at least would be worth trying. cc @samccone

samccone:
hmm, I tried locally to make them private but it seems like they are not getting renamed, I will follow up with someone to see if I can get any more insight here.

RReverser:
Maybe it's because you have these properties in `externs.js`? (so CC thinks they're required to be public)

samccone:
That might make sense @RReverser, I do typically annotate inline in my code or use a tool like https://github.com/angular/tsickle to do it automatically for me, however babel + rollup likes to trip up annotation syntax like

`/** @type {Foo} */(foo).bar()`

to just
`/** @type {Foo} */foo.bar()`

which breaks :goat:, and makes this kind of experiment non-trivial :flags:

RReverser:
Closure Compiler is sensitive to parentheses? O_o

developit:
I think babel's is `(0, foo).bar()`

RReverser:
@developit Oh, but that's a totally different thing :)

externs.js
```diff
@@ -0,0 +1,119 @@
...
+NodeName.prototype.defaultProps
+
+/** @type {(string|!NodeName)} */
+VNode.prototype.nodeName;
+
+/** @type {Object|undefined} */
+VNode.prototype.attributes;
+
+/** @type {Array<VNode>|undefined} */
+VNode.prototypechildren;
```
RReverser: This also looks suspicious btw.

developit: lol

samccone:
@RReverser for inline type annotations yes, it is how it knows what the subject of the type annotation is.

externs.js
samccone: lol nice, I think this did not end up mattering really, since we are not using closure compiler for compilation safety and were only using it to describe the "external" shape for compilation

RReverser:
Tried to fix on top of @samccone's branch with privatised private props:

```bash
$ ls -al *.gz
-rw-r--r-- 1 Ingvar Ingvar 3743 Mar  8 09:33 preact.cc.js.gz
-rw-r--r-- 1 Ingvar Ingvar 4005 Mar  8 09:36 preact.min.js.gz
```

Barely smaller :)

src/vnode.js
```diff
@@ -1,14 +1,9 @@
...
-	/** @type {object<string>|undefined} */
-	this.attributes = attributes;
-
-	/** @type {array<VNode>|undefined} */
-	this.children = children;
-
-	/** Reference to the given key. */
+	this.attributes = opt_attributes;
+	this.children = opt_children;
 	this.key = attributes && attributes.key;
```
RReverser: Actually one more suspicious change: you renamed the parameter name, but not reference, so it can now references global instead: https://gist.github.com/samccone/e617a1d9021bb433942dbd75045d6d5c#file-compiler-diff-L62

RReverser:
Fixed VNode regression and squeezed just a little bit more :)

```bash
-rw-r--r-- 1 Ingvar Ingvar 3733 Mar  8 10:07 dist/preact.cc.js.gz
-rw-r--r-- 1 Ingvar Ingvar 3990 Mar  8 10:07 dist/preact.min.js.gz
```

RReverser:
Also looks like due to lack of annotations, CC mangled some DOM property names in `NON_DIMENSION_PROPS`: 

```javascript
qa={A:1,B:1,C:1,G:1,flex:1,flexGrow:1,I:1,flexShrink:1,H:1,fontWeight:1,J:1,lineHeight:1,opacity:1,order:1,orphans:1,K:1,widows:1,zIndex:1,zoom:1}
```

https://gist.github.com/samccone/e617a1d9021bb433942dbd75045d6d5c#file-compiler-diff-L654

developit:
5% is still pretty good. Any idea where the savings ends up coming from?  I'm already doing some refactoring that is geared towards size optimization (eg: the above map switched to a regex saves 100b), curious if there's overlap.

RReverser:
@developit My changes were quite small, mostly 1) changed VNode constructor to initialize attributes/children with empty object/array if not provided so that checks in other places can be avoided and 2) changed https://github.com/developit/preact/blob/b3845c17a00997f73c8e4530074dbb4efdf522e8/src/dom/index.js#L72-L80 to always use `*NS` methods (with `''` as URI for regular HTML) so that paths can be merged.

developit:
oooh - that second one is super smart, wish I'd thought of it! any idea if that might fail in old browsers?

RReverser:
@developit Not sure about it. I mean, in very old browsers with only DOM level 1 these methods simply don't exist, but then they would fail in SVG branch too (and I guess many other things would too). But worth checking.

samccone:
@RReverser mind opening a CL into my fork, so we can layer on the fixes here? thanks :balloon:

RReverser:
@samccone done https://github.com/samccone/preact/pull/1

FezVrasta:
Have you tried babili? I tested closure comp and babili on my ES6 library and babili performed much better.

RReverser:
@FezVrasta It achieves better compression because it generates ES6, which is not an option when you still need transpiling.

FezVrasta:
I got better results even with es5

robertknight:
Based on the conclusion at the end of the description about maintenance effort vs the achieved savings, I'm tempted to count this as an interesting experiment but close the issue for the moment. Does anyone object?

si14:
One thing to consider: an entire discussion on the size savings revolved around preact itself. However, savings can be much, much greater when you bring an entire app in picture, but for that preact should be compatible with Closure.

samccone:
@robertknight +1 to closing, however I think we might want to generate a correct externs file for preact so that as @si14 said.. we can closure compile preact code.

robertknight:
Thanks @samccone - that makes sense. I'm going to close this for now. A future PR that focused on just providing/generating an externs file so that Preact can be used with Closure compiler would be welcome as and when someone has time.

PR #566, opened 2017-02-26
Remove out-of-date comment

asolove:
Was reading through the source and noticed that this comment referred to old code that was removed in https://github.com/developit/preact/commit/6bfa2af9f5d2fb48c3e3b075b3f5e6dc85c4749c


developit:
heh - good catch, had this removed locally but I'll merge!

PR #573, opened 2017-03-01
Add basic typings for supported class-level (static) Component properties

jmrog:
This PR addresses a few relatively minor obstacles to using Preact "out of the box" -- i.e., without having to create your own types to handle what are really ecosystem-standard properties -- in a TypeScript project. The obstacles are related to the absence of typings for static (class-level) Component variables, namely `displayName` and `defaultProps`.

**Obstacle 1:**

Suppose that you are writing a higher-order component (HOC) that wraps a Preact Component, and you want to create a reasonable displayName for the HOC on the basis of the wrapped component's displayName. Trying to read the `displayName` property of the Preact Component results in complaints from the TypeScript compiler that that property does not exist on the type. For example:

```typescript
function getName(component: typeof preact.Component): string {
    return component.displayName || component.name || 'Component';
}
```

The compiler complains that `displayName` does not exist on the type.

**Obstacle 2:**

If you're writing an ordinary (not higher-order) Preact Component and you want type-checking for ecosystem-standard static properties (`displayName` and `defaultProps`), you won't have any. For example:

```typescript
class SomeComponent extends Component<SomeProps, SomeState> {
    static displayName = { lol: 'wut' };
    // . . . 
}
```

In that case, the compiler will not complain, which is sub-optimal. Likewise:

```typescript
class SomeComponent extends Component<SomeProps, SomeState> {
    // . . . 
}
SomeComponent.displayName = 'SomeComponent';
SomeComponent.defaultProps = { someProp: 'lol' };
```

Here, the compiler complains that the latter two properties do not exist on the type.

The PR fixes these issues. It validates `displayName` as a string, if present, and `defaultProps` as "any", if present (at the moment, anyway; I might prefer changing the type for `defaultProps` to "object", which is really what's desired, but I'm not sure whether this project wants its type definitions to be supported only by the latest version of TypeScript (2.2); if so, please let me know and I will update the PR!).

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -39,6 +39,9 @@ declare namespace preact {
 	abstract class Component<PropsType, StateType> implements ComponentLifecycle<PropsType, StateType> {
 		constructor(props?:PropsType);
 
+		static displayName?:string;
+		static defaultProps?:any;
+
 		state:StateType;
 		props:PropsType & ComponentProps;
 		base:HTMLElement;


src/preact.d.ts
```diff
@@ -39,6 +39,9 @@ declare namespace preact {
 	abstract class Component<PropsType, StateType> implements ComponentLifecycle<PropsType, StateType> {
 		constructor(props?:PropsType);
 
+		static displayName?:string;
+		static defaultProps?:any;
```
UnwrittenFun: `Partial<PropsType>` would be great, if typescript 2.1 is okay

jmrog: I had the same thought and even tried it out before this PR. Unfortunately, static members can't reference a class type's parameters. There are ways around this, but I was apprehensive about making the declaration file overly complex.

UnwrittenFun: I was worried that might be the case, shame ☹️

developit:
Thanks for such a detailed and well-explained overview of the change, really helps @jmrog 👍

robertknight:
LGTM. As an idle thought, I wonder whether it would be possible or make sense to derive Preact's typings from React's typings in future, as opposed to duplicating them - often with subtle differences.

developit:
@robertknight certainly seems nicer for maintenance. I wonder how different they are at this point?

jmrog:
IIRC, they are quite different. I had the same thought as @robertknight and did some (albeit brief/cursory) looking at React's typings before this PR (and the one in #576). Making Preact's typings similar to React's would require a decent amount of restructuring of the type declarations (again, IIRC; I haven't looked while writing this comment). It does seem like a sensible goal, though.

PR #576, opened 2017-03-02
Add TypeScript typings for functional components and inclusive components

jmrog:
This PR contains a few more proposed improvements to Preact's TypeScript declarations, related tangentially to some things mentioned in PR #573. In #573, I mentioned that the compiler complains at times when you do things with higher-order components, like so:

```typescript
function getName(component: typeof preact.Component): string {
    return component.displayName || component.name || 'Component';
}
```

The issue in #573 was with the `displayName` property. But _this_ PR addresses the fact that there also appears to be no out-of-the-box way to do similar things with _functional_ Preact components as opposed to "full-fledged" components (i.e., components with the full lifecycle implementation). That is, there seems to be no _out-of-the-box_ way (i.e., without creating my own custom types) to do something like this:

```typescript
// I avoided the name "GenericComponent" because the term "generic" is overloaded in the context of Typescript, so "InclusiveComponent" it is.
function getName(component: preact.InclusiveComponent<any, any>): string {
    return component.displayName || component.name || 'Component';
}
```

or even like this:

```typescript
function getName(component: typeof preact.Component | preact.FunctionalComponent<any>): string {
    return component.displayName || component.name || 'Component';
}
```

Ideally, I would be able to specify that a higher-order function accepts Preact components generally, whether those components are mere functional components or are "full-fledged" components, as in the first of the two preceding examples (without having to create my own types, since functional components are standard to the ecosystem and are widely used interchangeably with full-fledged components). This PR lets you do that, with type-checking/without the TypeScript compiler complaining.

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -1,5 +1,5 @@
 declare namespace preact {
-	interface ComponentProps<C extends Component<any, any>> {
+	interface ComponentProps<C extends (Component<any, any> | StatelessComponent<any>)> {
 		children?:JSX.Element[];
 		key?:string | number | any;
 		ref?:(el: C) => void;
@@ -37,15 +37,14 @@ declare namespace preact {
 		new (props?:PropsType):Component<PropsType, StateType>;
 	}
 
-	interface FunctionalComponent<PropsType> {
-		(props:PropsType & ComponentProps, context?:any):JSX.Element;
+	interface StatelessComponent<PropsType> {
+		(props:PropsType & ComponentProps<this>, context?:any):JSX.Element;
 		displayName?:string;
 		defaultProps?:any;
 	}
 
-	// An "inclusive" component is a component that is either a functional
-	// component or a more full-fledged component.
-	type InclusiveComponent<PropsType, StateType> = FunctionalComponent<PropsType> | typeof Component;
+    // Type alias for a component considered generally, whether stateless or stateful.
+	type AnyComponent<PropsType, StateType> = StatelessComponent<PropsType> | typeof Component;
 
 	abstract class Component<PropsType, StateType> implements ComponentLifecycle<PropsType, StateType> {
 		constructor(props?:PropsType);
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -36,6 +36,16 @@ declare namespace preact {
 		new (props?:PropsType):Component<PropsType, StateType>;
 	}
 
+	interface FunctionalComponent<PropsType> {
+		(props:PropsType & ComponentProps, context?:any):JSX.Element;
+		displayName?:string;
+		defaultProps?:any;
+	}
+
+	// An "inclusive" component is a component that is either a functional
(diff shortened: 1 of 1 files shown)


src/preact.d.ts
```diff
@@ -37,6 +37,16 @@ declare namespace preact {
 		new (props?:PropsType):Component<PropsType, StateType>;
 	}
 
+	interface FunctionalComponent<PropsType> {
```
robertknight: In the React typings, the equivalent interface is called `StatelessComponent`. Using the same name would probably be helpful to people migrating from one to the other. Thoughts?

src/preact.d.ts
```diff
@@ -37,6 +37,16 @@ declare namespace preact {
...
+		(props:PropsType & ComponentProps, context?:any):JSX.Element;
+		displayName?:string;
+		defaultProps?:any;
+	}
+
+	// An "inclusive" component is a component that is either a functional
+	// component or a more full-fledged component.
+	type InclusiveComponent<PropsType, StateType> = FunctionalComponent<PropsType> | typeof Component;
```
robertknight: Hmm ... this seems like a useful alias to provide but the name sounds a bit weird. I'm struggling to think of something much better. `AnyComponent`?

src/preact.d.ts
developit: This makes sense to me

jmrog:
Agreed on all counts. Updated PR.

robertknight:
@jmrog - Thanks for the update. Separately I merged https://github.com/developit/preact/pull/605 and forgot that this PR implemented very similar changes. Could you have a look at the current code in master and let me know whether there is anything from this PR that still needs to be added?

jmrog:
@robertknight Thanks for letting me know. Sure, I'll take a look and update as necessary, once time allows (probably very soon).

jmrog:
@robertknight: I updated this PR again to reflect the new material on master, though I do have one lingering question now. Earlier in this thread, you suggested renaming `FunctionalComponent` to `StatelessComponent` in order to follow the React naming conventions more closely. I did that; however, the other PR (which was merged) uses the name `FunctionalComponent`. So, for the moment, I've switched back to using `FunctionalComponent` for this PR. If `StatelessComponent` is still actually preferred, I can flip it back to that again. Is there a preference for that anymore?

The other remaining changes in this PR are small: 1) `displayName` and `defaultProps` are added as recognized properties for the `FunctionalComponent` type; 2) the `AnyComponent` type (a component that is either a functional component or a "class-y" component) is provided for convenience, since it will often be helpful for higher-order components; 3) code formatting is made more consistent in a few places.

Thanks!

robertknight:
> If StatelessComponent is still actually preferred, I can flip it back to that again. Is there a preference for that anymore?

No. I noticed that the [React docs](https://facebook.github.io/react/docs/components-and-props.html) use the term "functional component", so it is perhaps better to use that term even if it differs from the React typings.

Thanks for the contribution 👍

PR #591, opened 2017-03-16
Devtools: Protect against instantiating Map in browsers that don't support it

billneff79:
For pages that have react/devtools loaded, if you open them in older browsers like IE9, they fail because of the attempt to instantiate Map at the top level of the devtools file.  Just add some protection to only instantiate them if you can.  Since you can't add devtools to browsers that don't support Map, no harm is done with this.

### DIFF
--- a/devtools/devtools.js
+++ b/devtools/devtools.js
@@ -132,7 +132,7 @@ function createReactCompositeComponent(component) {
  * The same React*Component instance must be used when notifying devtools
  * about the initial mount of a component and subsequent updates.
  */
-let instanceMap = new Map();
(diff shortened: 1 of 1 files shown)


robertknight:
Originally we decided not to bother with supporting even loading of devtools in older browsers. If we change this then I think it would be a good idea to signpost it (eg. with a comment at the top of the file) so that others are aware.

developit:
For sure - I think this bit us because we sometimes use our `development` bundle for testing purposes, which includes the devtools extension.  With this check in place, it no longer errors on import, and then just never gets initialized (because the global hook isn't present).

Issue #434, opened 2016-12-07
forceUpdate should take a callback


Stuk:
As per the [React API docs](https://facebook.github.io/react/docs/react-component.html#forceupdate)

developit:
Good catch! This is a strange one since `forceUpdate()` is always synchronous, so there is no real purpose to the callback. However, it should be cheap/free to add (just duplicating [this line](https://github.com/developit/preact/blob/master/src/component.js#L80)) so we're probably fine to add it.

Stuk:
I think it's because React queues all component updates (e.g. in setState) and so in React it may not happen synchronously.

developit:
I don't see anything in their docs about it, but maybe it's something preact is doing unnecessarily.  Perhaps `forceUpdate()` doesn't need to be synchronous and just has to bypass `shouldComponentUpdate()`...

developit:
Released in 8.

PR #592, opened 2017-03-16
Add optional callback to forceUpdate method

wyze:
This PR aims to solve #434. I've also added a general `forceUpdate` is called and rerenders the component test since there didn't seem to be one. Let me know if I placed that in the right test file as I was unsure.

### DIFF
--- a/test/browser/spec.js
+++ b/test/browser/spec.js
@@ -124,4 +124,49 @@ describe('Component spec', () => {
 			expect(WithDefaultProps.prototype.getDefaultProps).to.be.calledOnce;
 		});
 	});
+
+	describe('forceUpdate', () => {
+		it('should force a rerender', () => {
+			let forceUpdate;
+			class ForceUpdateComponent extends Component {
+				componentWillUpdate() {}
+				componentDidMount() {
+					forceUpdate = () => this.forceUpdate();
+				}
+				render() {
+					return <div />;
+				}
+			}
+			sinon.spy(ForceUpdateComponent.prototype, 'componentWillUpdate');
+			sinon.spy(ForceUpdateComponent.prototype, 'forceUpdate');
+			render(<ForceUpdateComponent />, scratch);
+			expect(ForceUpdateComponent.prototype.componentWillUpdate).not.to.have.been.called;
+
+			forceUpdate();
+
+			expect(ForceUpdateComponent.prototype.componentWillUpdate).to.have.been.called;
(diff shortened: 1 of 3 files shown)


developit:
Awesome, and it's only 7 bytes!

NekR:
Callback is good, but I got to point out that if you are going to make forceUpdate() async someday, then that's major change (e.g. I rely on it to be sync).

developit:
Agreed - I was going to suggest that might be worth bringing up, but I'm definitely relying on it being synchronous as well.  I don't think we can reasonably do that in the near term.

I'd be curious what the filesize would be with a more naive callback solution since it's always synchronous:

```js
forceUpdate(callback) {
  renderComponent(this, FORCE_RENDER);
  if (callback) callback();
}
```

(not suggesting that's the right way to go here, just crossed my mind)

NekR:
@developit yeah, I thought about it too. There is no right way here, both are okay :-) (`callback && callback()` is smaller than `if` 😝)

test/browser/spec.js
```diff
@@ -124,4 +124,49 @@ describe('Component spec', () => {
...
+			}
+			sinon.spy(ForceUpdateComponent.prototype, 'componentWillUpdate');
+			sinon.spy(ForceUpdateComponent.prototype, 'forceUpdate');
+			render(<ForceUpdateComponent />, scratch);
+			expect(ForceUpdateComponent.prototype.componentWillUpdate).not.to.have.been.called;
+
+			forceUpdate();
+
+			expect(ForceUpdateComponent.prototype.componentWillUpdate).to.have.been.called;
+			expect(ForceUpdateComponent.prototype.forceUpdate).to.have.been.called;
```
robertknight: Thanks for adding these tests, although it doesn't actually check that the component got re-rendered. Ideally the test would do something mirroring a simplified real-world use of `forceUpdate`. Something like:

1. Create a component whose output depends on something other than props/state
2. Render
3. Change the external state
4. Force update
5. Check that the rendered output reflects the updated external state.

test/browser/spec.js
```diff
@@ -124,4 +124,49 @@ describe('Component spec', () => {
...
+		});
+
+		it('should add callback to renderCallbacks', () => {
```
robertknight: `renderCallbacks` is an implementation detail which we should avoid in the description. What the test actually _does_, checking that `forceUpdate` invokes its callback, is fine though.

Issue #556, opened 2017-02-22
Callback of `setState` is called at not expected time.


windyGex:
See the following code: 

```js
class A extends React.Component {
  componentWillMount() {
    this.setState({
        a: 'a'
    }, () => {
      console.log('setState');
    })
  }
  render() {
    return <B></B>
  }
}

class B extends React.Component{
  render() {
    return <div>B</div>;
  }
  componentDidMount() {
    console.log('didMount');
  }
}


ReactDOM.render(<A/>, mountNode);
```

**Preact** will log:

```
setState
didMount
```

**React** will log:

```
didMount
setState
```

I'm not sure if this is a bug？It will be error when access domNode of child component in callback.

developit:
Preact folds pre-mount `setState()` calls into the initial render by design. I hadn't thought of the callback issue though, that's cumbersome.

windyGex:
What's your opinion？ @robertknight

robertknight:
I can confirm this happens in Preact 10: https://codesandbox.io/s/lp5xyqq4vz

edimitchel:
@developit And now, callback is no more called !
https://codesandbox.io/s/556-nop2h

JoviDeCroock:
Hey @edimitchel, I can confirm that it works in `componentDidMount` and not in `componentWillMount`. I'll try to look into this later, for the time being you could use didMount instead .

edimitchel:
> Hey @edimitchel, I can confirm that it works in `componentDidMount` and not in `componentWillMount`. I'll try to look into this later, for the time being you could use didMount instead.

I used on `componentDidMount` on my personal project and this not working. Actually, you're right, on the demo, it works like expected. But for now, on my project, it doesn't.

JoviDeCroock:
@edimitchel it seems to work here: https://codesandbox.io/s/556-wlfp4

edimitchel:
@JoviDeCroock Yes, I changed the `Will` by `Did`.

JoviDeCroock:
> I execute another setState inside the callback (yes, I know, legacy code..), and the callback of this isn't called

I was talking about this part

edimitchel:
Sorry, I thought it was the same codesandbox url.
Can you see the issue here : https://codesandbox.io/s/556-glj62
The state doesn't change on the second callback.

JoviDeCroock:
Oh, the state doesn't seem to be updated in the callback but it is in the render, will try to look into it soon.

[...]

PR #593, opened 2017-03-17
Move when the callback of setState/forceUpdate is called

wyze:
This aims to solve #556. Not sure how I feel about the loop inside a loop. This does make the test pass though, so it should be a decent place to start from and iterate on if need be.


developit:
Nice, this is clever.  I'm not too worried about the nested loop since this only happens at the very end of each render pass, and often the number of pending component mounts there is small.

I'm curious though - does this break setState() callbacks after the first mount?  Seems like that loop is only for just-mounted components, whereas the setState() callback needs to be invoked for any update.

NekR:
From linked issue:
> I'm not sure if this is a bug？It will be error when access domNode of child component in callback.

Why callback to `setState` supposed to guarantee that everything will be mounted by that point? Is it the purpose of that callback?

From React docs:

> Instead, use componentDidUpdate or a setState callback (setState(updater, callback)), either of which are guaranteed to fire after the update has been applied. If you need to set the state based on the previous state, read about the updater argument below.
...
The second parameter to setState() is an optional callback function that will be executed once setState is completed and the component is re-rendered. Generally we recommend using componentDidUpdate() for such logic instead.

For me, this only says that `callback` is guaranteed to be called when **`this.state`** is updated and **component has been re-rendered**.

If this fix hurts something else or adds additional complexity then I don't it's worth it. If fix is going to be totally safe, then I'm okay with it.

Also, it's a breaking change, so for Preact 9.

NekR:
I mean, if it was merged prior Preact 8 went to `latest` then it could in v8, but now it's too late.

developit:
Agreed - if this ends up changing that timing we'd want a major bump. That said, I don't think major bumps are going to have the same scheduling and buildup going forward as they have in the past, they're just backwards-incompatible changes.

Issue #597, opened 2017-03-19
The type of render function


katyo:
As I see the prototype of `render` function is next:
```typescript
function render(node:JSX.Element, parent:Element, mergeWith?:Element):Element;
```

In `react` I render the root component (i.e. starting from `html` node) using the `document` object as container (i.e. parent) node and it works fine for this case. I tried to do same using `preact` and it also works as expected.

Actually I think the prototype of `render` function must be like that:
```typescript
function render(node:JSX.Element, parent:Element|Document, mergeWith?:Element):Element;
```

developit:
@katyo makes sense! Happy to accept a PR for this if you have the time.

marvinhagemeister:
Closing this, as this has been merged a while ago.

PR #599, opened 2017-03-20
Fixed type of render function

katyo:
Fix for issue #597


NekR:
Well, it looks okay, I just don't know why one would render directly into document. Doing that, you won't be able to escape Preact for anything else.

developit:
can still append non-preact nodes, but they might get garbage collected prior to 8.0 yup.

NekR:
@developit it's [`ParentNode`](https://developer.mozilla.org/en/docs/Web/API/ParentNode) interface. It doesn't have `innerHTML` and friends, so `document` doesn't have them too.

NekR:
Anyway. I think this change won't hurt :-)

NekR:
@developit well, they share `Node` interface of course too. It just was split into parts like `Node`, `ParentNode`, `ChildNode`, etc. `Element` implements most of them :-)

robertknight:
I'm not entirely convinced this change is a good idea. At the moment it works but as you noted, `Document` does not implement the `Element` interface so it is possible that Preact might end up trying to do things with the container that are not supported on a non-`Element` object in future. Still it got approval in the linked issue so I'm going to merge it.

PR #601, opened 2017-03-20
Conditional import of devtools based on NODE_ENV

t47io:
An attempt to address https://github.com/developit/preact/issues/431 to make `devtools` conditional and NOT in production.

### DIFF
--- a/devtools/index.js
+++ b/devtools/index.js
@@ -1,4 +1,4 @@
-import { initDevTools } from './devtools';
-
-initDevTools();
-
+if (NODE_ENV !== 'production') {
+	const initDevTools = require('./devtools').initDevTools;
+	initDevTools();
+}


devtools/index.js
```diff
@@ -1,4 +1,4 @@
-import { initDevTools } from './devtools';
-
-initDevTools();
-
+if (NODE_ENV !== 'production') {
```
NekR: `NODE_ENV` could be not provided at all, i.e. when used without bundlers.

t47io: Right, build failed. Should I use `process.env.NODE_ENV` instead?

t47io: I was thinking `process.env.NODE_ENV && process.env.NODE_ENV !== 'production'`

NekR: Also it's `process.env.NODE_ENV` I believe.

NekR: @t47io yes, it should be `process.env.NODE_ENV`. Not sure about using outside of a bundler.  @developit are builds for CDN somehow separate we could feed `process.env.NODE_ENV` into rollup there?

> I was thinking process.env.NODE_ENV && process.env.NODE_ENV !== 'production'

Well, in ideal case it should be `typeof process !== 'undefined' && process.env && process.NODE_ENV !== 'production'` but I not sure how `webpack` would handle that.

t47io:
What else should I look into?

devtools/index.js
```diff
@@ -1,4 +1,4 @@
-import { initDevTools } from './devtools';
-
-initDevTools();
-
+if (typeof process !== 'undefined' && process.env && process.env.NODE_ENV !== 'production') {
```
developit: Actually I think we might want to have this *on* by default when `process` is undefined - that'd be a browser context, which means someone is directly adding this to a page via a script tag.

```js
if (typeof process==='undefined' || process.env && process.env.NODE_ENV!=='production') {
  // ...
}
```

NekR:
Well, it seems good but I'm not confident enough to prove it until someone tests that it really works with webpack.

developit:
I just realized - this file is bundled via `rollup`, so the `require()` won't work.  I think instead we need to wrap the entirety of [devtools/devtools.js](https://github.com/developit/preact/blob/master/devtools/devtools.js) in the process check and leave this file as a simple import + invoke.

keemor:
Hi,
Is it probable that fix for disabling preact/devtools in production mode will be merged soon?

Thanks!

NekR:
Use preact/debug instead. It has this fix implemented. @_developit closing this one?

NekR:
Use preact/debug instead. It has this fix implemented. @developit closing this one?

PR #605, opened 2017-03-22
d.ts improvements

valotas:
- Introduced a `FunctionalComponent` type to describe functional components
- Added context to the Component's interfaces


src/preact.d.ts
```diff
@@ -33,18 +33,23 @@ declare namespace preact {
 		componentDidUpdate?(previousProps:PropsType,previousState:StateType,previousContext:any):void;
 	}
 
+	interface FunctionalComponent<PropsType, StateType> {
```
developit: I don't know enough about TS to say for sure, but I think this should context? The signature for Pure Functional Components is `fn(props, context) => jsx`

valotas: You are right. I'll fix that

src/preact.d.ts
```diff
@@ -57,7 +62,7 @@ declare namespace preact {
 		abstract render(props:PropsType & ComponentProps<this>, state:any):JSX.Element;
```
valotas: I believe here I should also add the context as a third argument, right?

developit: yup!

src/preact.d.ts
developit: sounds good to me :)

valotas:
I'll try to add some typescript compilation tests too

developit:
@valotas that would be *awesome*!

valotas:
@developit I think that this can be merged. Do you need anything else from me?

developit:
Nope, this all looks great to me. Would love a quick once-over by someone with more TS knowledge, I am mostly clueless about TS.

NekR:
I haven't used TS for a while tbh. Seems fine, but TS declaration format evolved from last time I used TS.

test/ts/prect-test.tsx
```diff
@@ -0,0 +1,32 @@
+import { h, render, Component } from 'preact';
```
robertknight: Typo in the name of this file "prect-test"

test/ts/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+	"compilerOptions": {
+		"module": "commonjs",
+		"lib": [
+			"es6",
+			"dom"
+		],
+		"noImplicitAny": true,
+		"noImplicitThis": true,
+		"strictNullChecks": false,
```
robertknight: Out of interest, would the tests pass if `strictNullChecks` were enabled?

PR #606, opened 2017-03-23
Update README.md

jpsc:
Change setState to keep consistency with the rest of examples.


developit:
hmm - it's actually a lot more common to just mutate state there, or assign to it, since it happens prior to initial render there's no need to use setState() to get the re-render triggering.

Maybe this would be better?

```js
this.state = {
  time: Date.now()
};
```

jpsc:
That can also be. I just didn't know about that difference :)

PR #611, opened 2017-03-27
Speed up children pairing by avoiding array holes & branching

asolove:
In my (admittedly, limited) testing, this looks like about a 2% win on the existing perf tests.

In certain circumstances, it does change the order in which key-less child vnodes get paired with child DOM nodes. Here's an example:

Previous contents were: `<a>1</a> <a>2</a> <div>3</div>`
New vnode children are: `<div>0</div> <a>1</a> <a>2</a>`

When searching for a child dom node to match the new first `div`, we end up finding `<div>3</div>` and reusing it. We then fill in its hole with `<a>1</a>`, meaning that the new nodes will be paired with the flipped DOM nodes.

So whether this is a win or loss depends in large part on the structure of changes inside each component.

### DIFF
--- a/src/vdom/diff.js
+++ b/src/vdom/diff.js
@@ -227,11 +227,11 @@ function innerDiffNode(dom, vchildren, context, mountAll, absorb) {
 			else if (!child && min<childrenLen) {
 				for (j=min; j<childrenLen; j++) {
 					c = children[j];
-					if (c && isSameNodeType(c, vchild)) {
+					if (isSameNodeType(c, vchild)) {
 						child = c;
-						children[j] = undefined;
-						if (j===childrenLen-1) childrenLen--;
-						if (j===min) min++;
+						children[j] = children[min];
+						children[min] = undefined;
+						min++
 						break;
 					}
 				}
--- a/src/vdom/diff.js
+++ b/src/vdom/diff.js
@@ -231,7 +231,7 @@ function innerDiffNode(dom, vchildren, context, mountAll, absorb) {
 						child = c;
 						children[j] = children[min];
 						children[min] = undefined;
-						min++
+						min++;
 						break;
 					}
 				}


developit:
Lol sorry

robertknight:
> So whether this is a win or loss depends in large part on the structure of changes inside each component.

Indeed, and the impact of this is going to far outweigh the engine-level optimizations here. We don't have a good set of benchmarks that represent real-world UIs, so I'm going to conjecture.

I would guess that a reasonable assumption is that in unkeyed lists it is most common for the list of nodes to be the same between two renders except for nodes added or removed. A common example would be a UI element that is added or removed in some state while the rest of the UI stays the same. In that case, a decent strategy is to match vchildren against existing child nodes in the same order. That strategy is also easy to understand and explain.

The moving around of array elements that happens here means this doesn't happen, so my hypothesis is that this could lead to a lot more work.

src/vdom/diff.js
```diff
@@ -227,11 +227,11 @@ function innerDiffNode(dom, vchildren, context, mountAll, absorb) {
...
 					c = children[j];
-					if (c && isSameNodeType(c, vchild)) {
+					if (isSameNodeType(c, vchild)) {
 						child = c;
-						children[j] = undefined;
-						if (j===childrenLen-1) childrenLen--;
-						if (j===min) min++;
+						children[j] = children[min];
+						children[min] = undefined;
+						min++;
```
robertknight: To explain this and check my understanding:

1. `children` is the list of child DOM nodes which were created based on a vdom Element in the previous render but did not have a `key` prop assigned.
2. This loop iterates over the unkeyed vdom elements and unkeyed DOM elements and matches them up based on type. Once a DOM element is chosen for re-use, it can't be used for another vdom element, so the code previously set the array item to `undefined` so that it would be skipped over when processing other vdom nodes.
3. There is a cost to checking the falsyness of `undefined` entries in the array and skipping over them, which is what this PR set out to change.
4. The way it now works is by _moving_ the first unused element of the `children` array to the position of the 'used' DOM element and then clearing the first unused position. This means that all the `undefined` entries in the `children` array are at the start in positions less than `min`. Since the loop starts iterating with `j=min`, `c = children[j]` will never be undefined and the check can be removed.

A side effect is that elements in the `children` array can get reordered as the list of vdom children is processed.

robertknight:
A simple and safe optimisation might be to change `if (c && ...)` to `if (c !== undefined && ...)` though?

asolove:
So this PR emerged from several hours of staring at the child pairing loop to understand how it worked. At that time, it seemed like there was room for optimization and, having only stared at that code for quite a whole, that this was the most important thing in the world.

In the cold light of another day, whatever small, occasional performance benefit and code golf dork points this PR generates, it is more than overruled by making an already-complex process even more complex, and making it even harder to reason about the performance impacts of small changes to client component code.

robertknight:
> So this PR emerged from several hours of staring at the child pairing loop to understand how it worked. 

Heh - I had to read/step through it a few times to grok it this morning. What it does isn't that complex but I think the implementation could benefit from at least some additional or revised comments to clarify things.

developit:
I think at the very least there's a takeaway here - it would be good to find a way to deal with these short-lived objects (the key mapping and children array) in some lighter way.

PR #621, opened 2017-04-02
Ts definitions overhaul

Hotell:
- refactors typings to 2.x version style
- adds/fixes missing defs for full Preact api usage ( HOC, Providers, Context )
- adds proper test suite for TS
- adds formatting scripts for TS
- add TS strict mode ( strictNullChecks )

Partially closes: https://github.com/developit/preact/issues/639

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,11 @@
     "test:karma": "karma start test/karma.conf.js --single-run",
     "test:mocha:watch": "npm run test:mocha -- --watch",
     "test:karma:watch": "npm run test:karma -- no-single-run",
-    "lint": "eslint devtools src test",
+    "lint": "npm run lint:js && npm run lint:ts",
+    "lint:js": "eslint devtools src test",
+    "lint:ts": "npm run format:ts -- --verify",
+    "lint:ts:fix": "npm run format:ts -- -r",
+    "format:ts": "tsfmt --useTsconfig test/ts/tsconfig.json",
     "prepublish": "npm run build",
     "smart-release": "npm run build && npm test && git commit -am $npm_package_version && git tag $npm_package_version && git push && git push --tags && npm publish",
     "release": "cross-env npm run smart-release"
@@ -105,6 +109,7 @@
     "sinon": "^1.17.4",
     "sinon-chai": "^2.8.0",
     "typescript": "^2.2.2",
+    "typescript-formatter": "5.1.2",
     "uglify-js": "^2.7.5",
     "webpack": "^1.13.1"
   },
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "test:karma": "karma start test/karma.conf.js --single-run",
     "test:mocha:watch": "npm run test:mocha -- --watch",
     "test:karma:watch": "npm run test:karma -- no-single-run",
-    "lint": "npm run lint:js && npm run lint:ts",
+    "lint": "npm-run-all lint:js lint:ts",
     "lint:js": "eslint devtools src test",
     "lint:ts": "npm run format:ts -- --verify",
     "lint:ts:fix": "npm run format:ts -- -r",
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -1,5 +1,9 @@
 // Type definitions for Preact v7.0
 // TypeScript Version: 2.2
+
+export = preact;
+export as namespace preact;
+
 declare namespace preact {
 
 	type PreactType = string | ComponentClass<any, any> | SFC<any>;
@@ -31,8 +35,7 @@ declare namespace preact {
 	 * ```
 	 */
 	interface ComponentProps {
-		// interface ComponentProps {
-		children?: JSX.Element[];
+		children?: VNode<any>[];
 		key?: Key;
 		ref?: Ref<any>;
 	}
@@ -67,7 +70,7 @@ declare namespace preact {
 
 	type SFC<Props> = StatelessComponent<Props>;
 	interface StatelessComponent<Props> {
-		(props: Readonly<Props> & Readonly<ComponentProps>, context?: any): JSX.Element;
+		(props: Readonly<Props> & Readonly<ComponentProps>, context?: any): VNode<any>;
 		defaultProps?: Partial<Props>;
 		displayName?: string;
 	}
@@ -110,7 +113,7 @@ declare namespace preact {
 		setState<K extends keyof State>(state: Pick<State, K>, callback?: () => void): void;
 		setState<K extends keyof State>(fn: (prevState: State, props: Props) => Pick<State, K>, callback?: () => void): void;
 
-		forceUpdate(callBack?: () => any): void;
+		forceUpdate(): void;
 
 		abstract render(props?: Readonly<Props> & Readonly<ComponentProps>, state?: Readonly<State>, context?: any): JSX.Element | null;
 	}
@@ -120,7 +123,7 @@ declare namespace preact {
 	}
 
 	function h<PropsType>(node: ComponentClass<PropsType, ComponentState> | SFC<PropsType>, params: PropsType, ...children: (JSX.Element | JSX.Element[] | string)[]): JSX.Element;
-	function h(node: string, params: JSX.HTMLAttributes & JSX.SVGAttributes & { [propName: string]: any }, ...children: (JSX.Element | JSX.Element[] | string)[]): JSX.Element;
+	function h(node: string, params: HTMLAttributes & SVGAttributes & { [propName: string]: any }, ...children: (JSX.Element | JSX.Element[] | string)[]): JSX.Element;
 	function render(node: JSX.Element, parent: Element, mergeWith?: Element): Element;
 	function rerender(): void;
 	function cloneElement(element: JSX.Element, props: any): JSX.Element;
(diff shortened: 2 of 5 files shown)


robertknight:
Hello. Thank-you for the contribution and time you've spent on this @Hotell but this PR is way to large for anyone to meaningfully review. It also inevitably conflicts with other smaller changes that other contributors have been making.

Could you please have a think about how you can split up this work into smaller/simpler chunks which are easier for the rest of us to grok.

developit:
Definitely a ton of good stuff in here though, I wonder what the best way to diff this would be? I think @Hotell was aiming for React parity in the interfaces, which is something we've bounced around before.

Hotell:
Hi guys sorry for my super delayed response... I was busy with conferences, oss etc.

@robertknight 
> Could you please have a think about how you can split up this work into smaller/simpler chunks which are easier for the rest of us to grok.

I'm using conventional changelog conventions for commits ( which I highly recommend btw ) -> I also chunked changes in separate atomic commits so it's you can read what it does.

@developit 
> @Hotell was aiming for React parity in the interfaces, which is something we've bounced around before.

Partialy yes but that's just one step.

I should've created and issue for this before submittin PR, my mistake, I was kind of a hurry.  

will create it now.

But as I see it, Preact core team doesn't have experienced TS guy ( pls correct me if I'm wrong ), so maybe we should outsource definitions to definitely typed.

Hotell:
Pls see issue https://github.com/developit/preact/issues/639

screendriver:
> But as I see it, Preact core team doesn't have experienced TS guy ( pls correct me if I'm wrong ), so maybe we should outsource definitions to definitely typed.

I would leave it here in preact. Microsoft [recommend](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) to add typings directly to every library itself and only extract it to DefinitelyTyped when this is not possible.

marvinhagemeister:
+1 on leaving the typings in preact 💯 

The changes here look great! Awesome work @Hotell !

src/preact.d.ts
```diff
@@ -1,349 +1,414 @@
...
+		// @TODO - preact-compat addition
+		// this is needed when used with react libs via preact-compat ( because React has refs on Component instance )
+		// refs?: any;
 
-		forceUpdate(): void;
+		constructor(props?: Props, context?: any);
 
-		abstract render(props?:PropsType & ComponentProps<this>, state?:StateType, context?: any):JSX.Element;
-	}
+		linkState: (name: string) => (event: Event) => void;
```
jmrog: Might want to mark this as deprecated, since (in Preact 8+) `linkState` is now in a separate module and not part of `Component` itself.

src/preact.d.ts
```diff
@@ -1,349 +1,414 @@
...
-	interface ElementClass extends preact.Component<any, any> {
+	interface ChildContextProvider<CC> {
+		getChildContext(): CC,
 	}
 
-	interface ElementAttributesProperty {
-		props:any;
-	}
+	//
+	// Peact Nodes
```
jmrog: Typo.

src/preact.d.ts
```diff
@@ -361,390 +426,413 @@ declare namespace JSX {
...
+		accept?: string;
+		acceptCharset?: string;
+		accessKey?: string;
+		action?: string;
+		allowFullScreen?: boolean;
+		allowTransparency?: boolean;
+		alt?: string;
+		async?: boolean;
+		autocomplete?: string;
+		autofocus?: boolean;
```
jmrog: I've got a separate PR to make this change, so maybe we can just merge that in after this... _But_, if you want, I believe `autoComplete` and `autoFocus` should be camel-cased (as I just wrote them).

src/preact.d.ts
```diff
@@ -361,390 +426,413 @@ declare namespace JSX {
...
+		autoPlay?: boolean;
+		capture?: boolean;
+		cellPadding?: number | string;
+		cellSpacing?: number | string;
+		charSet?: string;
+		challenge?: string;
+		checked?: boolean;
+		class?: string | { [key: string]: boolean };
+		className?: string | { [key: string]: boolean };
```
jmrog: Use of the object form for `class`/`className` is removed in Preact 8. Could we also mark these as deprecated in some way?

src/preact.d.ts
```diff
@@ -361,390 +426,413 @@ declare namespace JSX {
...
-		stop:SVGAttributes;
-		symbol:SVGAttributes;
-		text:SVGAttributes;
-		tspan:SVGAttributes;
-		use:SVGAttributes;
+}
+
+declare global {
+
+	// augment global Function
```
jmrog: I suppose this is mostly harmless, but it still feels a little dirty to me. Is it really necessary now that these properties are already on `SFC`?

test/ts/prect-test.tsx
```diff
@@ -1,32 +1,425 @@
-import { h, render, Component } from 'preact';
+import { h, render, Component, SFC, ChildContextProvider, ComponentProps, ComponentClass, VNode } from 'preact';
```
jmrog: Can you rename this file so that it is `preact-test.tsx`?

test/ts/prect-test.tsx
```diff
@@ -1,32 +1,425 @@
...
 		}
 	}
 
-	render({ initialInput }: DummyProps, { input }: DummyState) {
-		return <DummerComponent initialInput={initialInput} input={input} />
+	render(props: DummyProps, state: DummyState) {
+		const { initialInput } = this.props;
+		const { input } = this.state;
+		return (
+			<DummerComponent
```
jmrog: Now I'm _really_ nitpicking, but it'd be a little less jarring if the typo ("Dummer") were corrected here, too (to "Dumber"). Not a big deal, though.

test/ts/prect-test.tsx
```diff
@@ -1,32 +1,425 @@
...
+const DummerComponent2: SFC<DummerComponentProps> = ({ initialInput, input }) => (
+	<div>Input: {input}, initial: {initialInput}</div>
+);
+DummerComponent2.defaultProps = { input: 'yay' };
+DummerComponent2.displayName = 'Hello2';
+
+// Children Projection
+
+type LayoutProps = {
+	children?: [LayoutProjecitionFn]
```
jmrog: The typo here might also cause some dev headaches. Should be `LayoutProjectionFn`.

NaridaL:
May I suggest the following improvement:
Don't have JSX.HTMLAttributes extend PreactHTMLAttributes, which allows for more accurately typed `h()`
```ts
	interface PreactHTMLAttributes<T extends Element = Element> {
		dangerouslySetInnerHTML?:DangerouslySetInnerHTML;
		key?:string;
		ref?:(el?: T) => void;
	}
// first overload as is
function h<T extends keyof HTMLElementTagNameMap>(node:T,
          params:JSX.HTMLAttributes & PreactHTMLAttributes<HTMLElementTagNameMap[T]>&{[propName: string]: any},
         ...children:(JSX.Element|JSX.Element[]|string)[]):JSX.Element;
// third definition for SVG
```
`HTMLElementTagNameMap` is part of lib.es6.d.ts

matthewmueller:
hey all, is there some way to use this file to overwrite the existing distributed typescript file?

matthewmueller:
Been playing around with this version a bit. Definitely an improvement over the previous definitions, but it doesn't currently seem well suited for passing `props` through `render(props)`:

```js
interface Props {
  href: string
}

export default class Link extends Component<Props, {}> {
  render(props: Props) {
    props.children // error: no children
  }
}
```

To fix this:

```js
interface Props extends ComponentProps {
  href: string
}

export default class Link extends Component<Props, {}> {
  render(props: Props) {
    props.children // no children
  }
}
```

But ComponentProps is supposedly deprecated.

---

Additionally, I'm not sure there's something that can be done about this, but it's quite tedious to pass children into methods like `h(...)`, because `PreactChildren` is not assignable to `VNode<any>`, however it does seem to work naturally if you use the JSX syntax.

package.json
```diff
@@ -26,7 +26,11 @@
     "test:karma": "karma start test/karma.conf.js --single-run",
     "test:mocha:watch": "npm run test:mocha -- --watch",
     "test:karma:watch": "npm run test:karma -- no-single-run",
-    "lint": "eslint devtools src test",
+    "lint": "npm-run-all lint:js lint:ts",
```
valotas: I think here `--parallel` could be used

src/preact.d.ts
```diff
@@ -1,349 +1,414 @@
...
+	type ComponentState = {} | void;
+	interface Attributes {
+		key?: Key;
+	}
+	interface ClassAttributes<T> extends Attributes {
+		ref?: Ref<T>;
+	}
+
+	/**
+	 * @deprecated.
```
valotas: This looks like a copy & paste from `react.d.ts`. Is it really needed?

src/preact.d.ts
```diff
@@ -1,349 +1,414 @@
...
+		__html: string;
 	}
 
 	interface PreactHTMLAttributes {
-		dangerouslySetInnerHTML?:DangerouslySetInnerHTML;
-		key?:string;
-		ref?:(el?: Element) => void;
+		dangerouslySetInnerHTML?: DangerouslySetInnerHTML;
+		key?: Key;
+		ref?: Ref<Element>;
```
valotas: This is not supported by preact if I am not mistaken

developit: correct, string refs are only patched into preact-compat, preact core is just function refs.

marvinhagemeister:
I just merged https://github.com/developit/preact/pull/737 because it seemed a bit more up to date. Nonetheless I'd love to bring in additional improvements from this PR 👍 

@Hotell Could you rebase this PR against master? 🐱

PR #625, opened 2017-04-06
Switch back to using .parentNode.removeChild() for IE compat

### LINKED ISSUE #624: Element.remove() is unsupported in Internet Explorer
While testing **8.0-beta** in IE11 I've run into a problem in `unmountComponent()` calling `.remove()` on `base`:

https://github.com/developit/preact/blob/8/src/vdom/component.js#L266

Changing that line to:

```js
base.parentNode.removeChild(base);
```

...fixes the problem. I'm not sure if 8.0 is supposed to support IE, but worth keeping in mind.

KrofDrakula:
Fixes #624

### DIFF
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -102,7 +102,7 @@ function eventProxy(e) {
 /** Remove the child node from parent if attached
  * @param {Element} el
  */
-export default function removeNode(el) {
+export function removeNode(el) {
 	if (el.parentNode !== null)
 		el.parentNode.removeChild(el);
 }
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -98,3 +98,11 @@ function setProperty(node, name, value) {
 function eventProxy(e) {
 	return this._listeners[e.type](options.event && options.event(e) || e);
 }
+
+/** Remove the child node from parent if attached
+ * @param {Element} el
+ */
+export default function removeNode(el) {
+	if (el.parentNode !== null)
+		el.parentNode.removeChild(el);
+}
--- a/src/vdom/component.js
+++ b/src/vdom/component.js
@@ -5,8 +5,7 @@ import { enqueueRender } from '../render-queue';
 import { getNodeProps } from './index';
 import { diff, mounts, diffLevel, flushMounts, recollectNodeTree, removeChildren } from './diff';
 import { createComponent, collectComponent } from './component-recycler';
-
-
+import remove from './remove';
 
 /** Set a component's `props` (generally derived from JSX attributes).
  *	@param {Object} props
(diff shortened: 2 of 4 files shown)


KrofDrakula:
@robertknight This should do the trick. I've externalized the function to DRY, but could probably be better if we just inlined it to `.parentNode.removeChild()`, but that requires checking that `parentNode` isn't `null`.

src/vdom/remove.js
```diff
@@ -0,0 +1,3 @@
+export default function remove(el) {
+	el.parentNode && el.parentNode.removeChild(el);
```
NekR: JS abstraction leaks so this should be `if (el.parentNode !== null) { el.parentNode.removeChild(el) }` @developit 😄

KrofDrakula: Sure, I'll change that.

NekR: I'm not really sure, was just making fun of that situation with such conditions. @developit knows more which is faster and which one gzips better. Sorry for confusion!

src/vdom/component.js
```diff
@@ -5,8 +5,7 @@ import { enqueueRender } from '../render-queue';
 import { getNodeProps } from './index';
 import { diff, mounts, diffLevel, flushMounts, recollectNodeTree, removeChildren } from './diff';
 import { createComponent, collectComponent } from './component-recycler';
-
-
+import remove from './remove';
```
developit: Thinking this should be `removeNode()` exported from `dom/index.js`. Thoughts?

KrofDrakula: Sure, makes more sense. I just wanted to get the ball rolling. 🎢

src/vdom/remove.js
developit: There are a few of these cases now in the codebase so the gzipped size should be fine with the explicit null check.  I generally try to benchmark to determine where it's worthwhile, but I'm fairly certain in this case it would be since it's immediately accessing `.removeChild` on the object value.

developit:
Ok filesize: right now we're sitting at 10b for this (which is actually quite good).

src/dom/index.js
```diff
@@ -98,3 +98,11 @@ function setProperty(node, name, value) {
 function eventProxy(e) {
 	return this._listeners[e.type](options.event && options.event(e) || e);
 }
+
+/** Remove the child node from parent if attached
+ * @param {Element} el
+ */
+export default function removeNode(el) {
```
developit: whoop ~~default~~ 😁

KrofDrakula: ARGH... sorry, one sec.

PR #626, opened 2017-04-06
Remove unnecessary object check in setAccessor

NekR:
Fixes https://github.com/developit/preact/issues/401

@developit do we need tests for this?


NekR:
_(created PR from `OpenJSX` because I already have fork in my profile which I cannot delete just yet)_

developit:
Ah - we'll need this done against the `8` branch, since this line is now inline `typeof`'s.

NekR:
I'll re-PR, no problem.

PR #628, opened 2017-04-07
Remove unnecessary object check in setAccessor

NekR:
Fixes https://github.com/developit/preact/issues/401


NekR:
We'll need tests for that, I'll do them a bit later, so do not merge just yet (unless someone wants to PR tests instead).

PR #631, opened 2017-04-07
Added two minor changes to jsdocs

tomaszlakomygp:
Hey,

After seeing this tweet: https://twitter.com/_developit/status/849840997282570240 I got really curious how preact actually works and how is it so small.

While reading the source code I noticed two places which didn't have updated JSdocs, this change fixes that :)

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -48,7 +48,7 @@ extend(Component.prototype, {
 
 	/** Update component state by copying properties from `state` to `this.state`.
 	 *	@param {object} state		A hash of state properties to update with new values
-	 *  	@param {function} callback	A function to be called once component state is updated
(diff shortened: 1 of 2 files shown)


developit:
Awesome, thanks!

src/component.js
```diff
@@ -48,7 +48,7 @@ extend(Component.prototype, {
 
 	/** Update component state by copying properties from `state` to `this.state`.
 	 *	@param {object} state		A hash of state properties to update with new values
-	 *  	@param {function} callback	A function to be called once component state is updated
+	 *	@param {function} callback	A function to be called once component state is updated
```
tomaszlakomygp: This probably was my editor behaving a bit funky, I didn't reconfigure it to work with preact codebase, sorry :)

PR #632, opened 2017-04-07
Changed ATTR_KEY value to be a bit shorter and save 12 bytes of bandwidth

tomaszlakomygp:
This probably won't get merged but I just had this crazy idea that preact might save some bandwith by reducing the `ATTR_KEY` internal property value to be a bit shorter

Before:
```js
MacBook-Pro-2:preact TomaszLakomy$ npm run build

> preact@8.0.1 build /Users/macos/Stuff/preact
> npm-run-all --silent clean transpile copy-flow-definition copy-typescript-definition strip optimize minify size

⚠️   'default' is imported from external module 'rollup-plugin-node-resolve' but never used

Inlining constant: NO_RENDER=0
Inlining constant: SYNC_RENDER=1
Inlining constant: FORCE_RENDER=2
Inlining constant: ASYNC_RENDER=3
Inlining constant: ATTR_KEY='__preactattr_'
gzip size: 3372
```

After:
```js
MacBook-Pro-2:preact TomaszLakomy$ npm run build

> preact@8.0.1 build /Users/macos/Stuff/preact
> npm-run-all --silent clean transpile copy-flow-definition copy-typescript-definition strip optimize minify size

⚠️   'default' is imported from external module 'rollup-plugin-node-resolve' but never used

Inlining constant: NO_RENDER=0
Inlining constant: SYNC_RENDER=1
Inlining constant: FORCE_RENDER=2
Inlining constant: ASYNC_RENDER=3
Inlining constant: ATTR_KEY='__preattr_'
gzip size: 3360
```

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -6,7 +6,7 @@ export const FORCE_RENDER = 2;
 export const ASYNC_RENDER = 3;
 
 
-export const ATTR_KEY = '__preactattr_';
+export const ATTR_KEY = '__preattr_';
 
 // DOM properties that should NOT have "px" added when numeric
 export const IS_NON_DIMENSIONAL = /acit|ex(?:s|g|n|p|$)|rph|ows|mnc|ntw|ine[ch]|zoo|^ord/i;


developit:
I'd thought about doing this as a breaking change in 8.0, but it would cause a lot of headaches for people.  Reason it's stayed the same is because other libraries like `preact-router` use that property to check if a given DOM element was rendered by Preact, so its name is actually part of Preact's API.

I'm surprised it was 12 bytes though. Wish we could change it but it's probably too late haha. FWIW I feel like `__props` might have been nice, but then it's not really specific enough to preact.

tomaszlakomygp:
There's nothing I hate more in web development than breaking changes in libraries/frameworks I use so I guess we can close it. I suspected this might be a breaking change, but I thought that I'd clarify it with you, @developit 😄 

On the other hand it might be moved to some config that might be injected into `preact-router` and other libraries in the future.

BTW: Speaking of constants.js, my brain hurts when trying to process `IS_NON_DIMENSIONAL` regexp, what the hell 😄

tomaszlakomygp:
Closing this PR, perhaps some another time we can revisit this change.

developit:
lol I should have left an apology comment above that regex, I know.  I generated a list of CSS properties with and without units by querying for 0-values on a div, then regex-golfed that using a visual tool to get the smallest pattern that matched most/all values and didn't provide false-positives.

I'd considered having preact-router actually pull in the constants file. Might be worthwhile, but currently that would cause some bundlers to import the whole thing, which is prohibitive.  Some day haha.

PR #633, opened 2017-04-08
preact/debug

NekR:
`preact/debug` is here 🎉

## Usage

To use `preact/debug`, you have to set `process.env.NODE_ENV` to `'development'` for your bundler, e.g. for webpack:
```js
new webpack.DefinePlugin({
  'process.env': {
    'NODE_ENV': JSON.stringify('development')
  }
})
```

Also, `preact/debug` imports `preact/devtools` automatically 👌

### Initialization

**Basic**
```js
require('preact/debug');

// or

import 'preact/debug';
```

**Configurable**
```js
require('preact/debug')({
  stringRefs: false
});

// or

import preactDebug from 'preact/debug';

preactDebug({
  stringRefs: false
});
```

## Config options
_(all turned on by default)_

* `undefinedComponents`: `boolean`, throw or not on undefined components
* `stringRefs`: `boolean`, throw or not on string refs (this throws anyway, but better track trace with the option)
* `arrayChildren`: `boolean`, warn or not about missing `key` properties on elements in array children

Warning example:

image

## Side notes

* Source directory is `debug_src/` and is transpiled to `debug/`. Ideally source directory should be `debug/` and transpiled to `debug.js`, but I don't know anyway to do so with requiring only `'./debug.js'` module in it an ignoring everything else (can be done with browserify but WAT 😨).
* Another way might be to publish separate package called `preact-debug`, but there is a problem that webpack makes ES6+ imports getter only so you won't be able to override `preact.h` property. If this could be worked around, then probably is the way to go.
* `preact.createElement` isn't hooked just yet


NekR:
If this landed, https://github.com/developit/preact/pull/601 might not be necessary.

NekR:
image

About folder structure. Also if we are going `preact-debug` package way, we can make it export preact itself (this way we solve webpack imports problem) and it could be aliased in dev mode. Though it's easier to include debug in the code rather than in config :-)

lukeed:
The `debug_src` can be dropped & everything moved to a single file (`debug.js`) if everything were wrapped within the `NODE_ENV !== 'development'` check.

robertknight:
What is the rationale for making each of these warnings configurable? I would be inclined to make the debug package zero or minimally configurable until we have a clear use case, since that reduces the risk of having to make a breaking change in future.

debug_src/debug.js
```diff
@@ -0,0 +1,78 @@
...
+	h: preact.h
+};
+
+const config = {
+	undefinedComponents: true,
+	stringRefs: true,
+	arrayChildren: true
+};
+
+preact.h = function(nodeName, attributes) {
```
robertknight: Could this be built on Preact's existing `options.vnode` hook rather than patching `h` directly? Then it should work for `createElement` as well. This would also be consistent with how the devtools module works - hooking into Preact's global hooks.

NekR: It could, but that way you won't get good stack trace, i.e. the call before throwing the error will point directly to the place of `h()` invocation. If this won't provide good stack traces I'm not really interested in having this package myself.

NekR:
@robertknight I for example need only `undefinedComponents` option, all other are annoying to me. Maybe `stringRefs` are fine, but `arrayChildren` for sure would annoy me.

NekR:
@robertknight 
> I would be inclined to drop the configurability in the initial release of this unless you have a clear use case for it.

@developit suggests to not use `forEach((a, i) => <div key={i} />)` because it already use indexes internally and specifying them that way may more hurt than help. In that case, I would expect people to disable that warning.

This is in addition to that this option is just annoying for myself.

debug_src/debug.js
developit: I want to check if we can prune the stack trace - we'd know it's 2 levels up, and the stack is just a String so we could remove the first two lines.  My reasoning is that `preact` is actually a frozen object when bundled via Webpack 2 - this is already causing issues with React Ho Loader since it does the patching technique used here, and I'm worried it's going to become an issue.  Thoughts?

developit:
I'm inclined to agree about the keys option, only because I've got codebases that would probably crash the console from the sheer number of warnings that would generate 😇

@NekR one option would be to hold off on releasing the `key` check since that's a pretty huge assertion on our part and a common cause of FUD around how keys work.

So maybe if we turned off key warnings for now we could release with no options?  Still keep it a constructor but have the defaults be to warn both for undefined components and string refs.

BTW - I wonder if we need to detect `preact-compat` and disable the string ref warning if it's in-place, since that adds support for them?

NekR:
> till keep it a constructor but have the defaults be to warn both for undefined components and string refs.

As I said in PR comment all debug checks are ON by default :-) Config is there just to disable things.

> BTW - I wonder if we need to detect preact-compat and disable the string ref warning if it's in-place, since that adds support for them?

Makes sense, not sure how  to so it though.

debug_src/debug.js
NekR: Stacks aren't that simple. Firefox have weird stack format and you cannot just modify it. There is no docs about the format and people in Mozilla IRC doesn't answer really to that question. There might be some lib by Fx extensions team, but I don't know about it.

robertknight: > It could, but that way you won't get good stack trace

Not sure I follow. Surely throwing in the vnode hook would just result in a trace with the point of the `h(...)` invocation being one more entry up in the stack trace?

developit:
I just know the keys warning is something I've complained about in React a lot, I'm kinda tempted to avoid promoting that people go crazy with keys out-of-the-box haha.

NekR:
@lukeed, @robertknight, @developit fine, fine, I'll do as you all want 😇

NekR:
Okay, I did requested changed but didn't test them :-)

debug.js
```diff
@@ -0,0 +1,90 @@
+'use strict';
```
developit: we should `.gitignore` the built files here

debug/index.js
```diff
@@ -0,0 +1,85 @@
...
+				}
+
+				keys[key] = true;
+			});
+		}
+
+		return oldVnodeOption.call(this, vnode);
+	};
+
+	const inspectChildren = (children, inspect) => {
```
developit: `children` should always be pre-flattened (though I'm fine leaving this here)

package.json
```diff
@@ -39,10 +40,13 @@
   },
   "files": [
     "devtools",
+    "debug",
```
developit: Thinking this might cause issues since there is both `debug.js` and `debug/index.js`? Not sure.

NekR:
@developit I did with `debug` the same what you do with `devtools`. Also, there is `require('preact')` in `debug`, how would that work in browser?

developit:
Devtools seems to be [UMD](https://github.com/developit/preact/blob/master/config/rollup.config.devtools.js#L7) - https://unpkg.com/preact@8.1.0/devtools.js

The preact import gets transformed to a detection, should work nicely.

NekR:
I mean with folder/generation. You have `devtools/index.js` folder and `devtools.js` file.

developit:
True, and that's working (guess that's my answer haha).

developit:
I think we're ready to merge this :)

NekR:
@developit yeah, I think so. But if you wish it to be UMD then I can try that.

debug/index.js
NekR: I'll remove that eventually, too lazy right now.

NekR:
I merged it ¯\_(ツ)_/¯

trueadm:
In React, we make warnings only fire once so the console isn't spammed. Might be worth doing it here too :)

NekR:
For the same element/component or in general write to console once for a each warning?

trueadm:
For each warning usually. Spamming a console is known to have an adverse affect on DX. We had really great feedback when we changed how this worked.

developit:
I like that idea. Keeping it global keeps things simple.

NekR:
I have nothing against it, but I won't have time for it for some time.
Someone may want to create a new issue for it and assign it to me :-)

rupav:
@NekR I see a different and strange behavior of React-dev-tools using Preact/debug. React-dev-tools are active with this config only (as suggested in this PR description), in our project, we are using Preact with React as an alias.
I also would like to know if Redux-store dev tools can be configured for Preact? 😅

marvinhagemeister:
@rupav the redux devtools should work right out of the box as they are not framework dependent.

rupav:
Forgive me @marvinhagemeister  for asking without checking again. I had not configured setup redux-dev-tools on configureStore. Yeah its working great. But what is your opinion  React-dev-tools with Preact/debug? How it can be best utilized.

marvinhagemeister:
@rupav easiest way to do it is by including `preact/debug`. If something isn't working correctly you should file a bug :)

PR #636, opened 2017-04-09
Reduced minimied & gzipped preact size by 25 bytes

tomaszlakomygp:
Before:
```js
MacBook-Pro-2:preact TomaszLakomy$ npm run build

> preact@8.0.1 build /Users/macos/Stuff/preact
> npm-run-all --silent clean transpile copy-flow-definition copy-typescript-definition strip optimize minify size

⚠️   'default' is imported from external module 'rollup-plugin-node-resolve' but never used

Inlining constant: NO_RENDER=0
Inlining constant: SYNC_RENDER=1
Inlining constant: FORCE_RENDER=2
Inlining constant: ASYNC_RENDER=3
Inlining constant: ATTR_KEY='__preactattr_'
gzip size: 3372
```

After:
```js
MacBook-Pro-2:preact TomaszLakomy$ npm run build

> preact@8.0.1 build /Users/macos/Stuff/preact
> npm-run-all --silent clean transpile copy-flow-definition copy-typescript-definition strip optimize minify size

⚠️   'default' is imported from external module 'rollup-plugin-node-resolve' but never used

Inlining constant: NO_RENDER=0
Inlining constant: SYNC_RENDER=1
Inlining constant: FORCE_RENDER=2
Inlining constant: ASYNC_RENDER=3
Inlining constant: ATTR_KEY='__preactattr_'
gzip size: 3347
```

All unit tests are passing 👍

### DIFF
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -32,6 +32,7 @@ export function removeNode(node) {
  *	@private
  */
 export function setAccessor(node, name, old, value, isSvg) {
+	let i;
 	if (name==='className') name = 'class';
 
 
@@ -51,9 +52,9 @@ export function setAccessor(node, name, old, value, isSvg) {
 		}
 		if (value && typeof value==='object') {
 			if (typeof old!=='string') {
-				for (let i in old) if (!(i in value)) node.style[i] = '';
+				for (i in old) if (!(i in value)) node.style[i] = '';
 			}
-			for (let i in value) {
+			for (i in value) {
 				node.style[i] = typeof value[i]==='number' && IS_NON_DIMENSIONAL.test(i)===false ? (value[i]+'px') : value[i];
 			}
 		}
@@ -61,7 +62,7 @@ export function setAccessor(node, name, old, value, isSvg) {
 	else if (name==='dangerouslySetInnerHTML') {
 		if (value) node.innerHTML = value.__html || '';
 	}
-	else if (name[0]=='o' && name[1]=='n') {
+	else if (name.slice(0,2)==='on') {
 		let useCapture = name !== (name=name.replace(/Capture$/, ''));
 		name = name.toLowerCase().substring(2);
 		if (value) {
@@ -73,7 +74,12 @@ export function setAccessor(node, name, old, value, isSvg) {
 		(node._listeners || (node._listeners = {}))[name] = value;
 	}
 	else if (name!=='list' && name!=='type' && !isSvg && name in node) {
-		setProperty(node, name, value==null ? '' : value);
+        /** Attempt to set a DOM property to the given value.
+         *	IE & FF throw for certain property-value combinations.
+         */
+		try {
+			node[name] = value==null ? '' : value;
+		} catch (e) { }
 		if (value==null || value===false) node.removeAttribute(name);
 	}
 	else {
@@ -88,17 +94,6 @@ export function setAccessor(node, name, old, value, isSvg) {
 		}
 	}
 }
-
-
-/** Attempt to set a DOM property to the given value.
- *	IE & FF throw for certain property-value combinations.
- */
-function setProperty(node, name, value) {
-	try {
-		node[name] = value;
-	} catch (e) { }
-}
-
 
 /** Proxy an event to hooked event handlers
  *	@private
--- a/src/h.js
+++ b/src/h.js
@@ -4,15 +4,13 @@ import options from './options';
 
 const stack = [];
 
-const EMPTY_CHILDREN = [];
-
 /** JSX/hyperscript reviver
 *	Benchmarks: https://esbench.com/bench/57ee8f8e330ab09900a1a1a0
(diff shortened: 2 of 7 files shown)


src/h.js
```diff
@@ -36,7 +34,7 @@ export function h(nodeName, attributes) {
 			if (simple && lastSimple) {
 				children[children.length-1] += child;
 			}
-			else if (children===EMPTY_CHILDREN) {
+			else if (children===[]) {
```
UnwrittenFun: This is not the same
```js
> [] === []
false
```

tomaszlakomygp: You're right, I'll revert that change

NekR: > All unit tests are passing 👍

I think there should be a test to catch this thing.

src/dom/index.js
```diff
@@ -51,17 +52,17 @@ export function setAccessor(node, name, old, value, isSvg) {
...
+			for (i in value) {
 				node.style[i] = typeof value[i]==='number' && IS_NON_DIMENSIONAL.test(i)===false ? (value[i]+'px') : value[i];
 			}
 		}
 	}
 	else if (name==='dangerouslySetInnerHTML') {
 		if (value) node.innerHTML = value.__html || '';
 	}
-	else if (name[0]=='o' && name[1]=='n') {
+	else if (name.slice(0,2)==='on') {
```
developit: This ends up being way slower, we benchmarked it a while back.

src/dom/index.js
```diff
@@ -73,7 +74,12 @@ export function setAccessor(node, name, old, value, isSvg) {
 		(node._listeners || (node._listeners = {}))[name] = value;
 	}
 	else if (name!=='list' && name!=='type' && !isSvg && name in node) {
-		setProperty(node, name, value==null ? '' : value);
+        /** Attempt to set a DOM property to the given value.
+         *	IE & FF throw for certain property-value combinations.
+         */
+		try {
```
developit: Moving this inline deopts setAccessor(), which is called a lot. It was in a separate function so that it would only deopt when called.

src/h.js
developit: Agreed, need a test there.

src/vdom/component.js
```diff
@@ -78,12 +78,8 @@ export function renderComponent(component, opts, mountAll, isChild) {
...
 		component.state = previousState;
 		component.context = previousContext;
-		if (opts!==FORCE_RENDER
-			&& component.shouldComponentUpdate
-			&& component.shouldComponentUpdate(props, state, context) === false) {
-			skip = true;
-		}
-		else if (component.componentWillUpdate) {
+		skip = opts!==FORCE_RENDER && component.shouldComponentUpdate && component.shouldComponentUpdate(props, state, context) === false;
+		if (component.componentWillUpdate) {
```
developit: This will call `componentWillUpdate()` for mounts because the else clause is removed. It should only be called for updates.

src/vdom/diff.js
```diff
@@ -254,7 +254,7 @@ export function recollectNodeTree(node, unmountOnly) {
 	else {
 		// If the node's VNode had a ref function, invoke it with null here.
 		// (this is part of the React spec, and smart for unsetting references)
-		if (node[ATTR_KEY]!=null && node[ATTR_KEY].ref) node[ATTR_KEY].ref(null);
```
developit: The null check here improves performance because it avoids casting the object to a boolean.

src/vdom/component-recycler.js
```diff
@@ -9,8 +9,7 @@ const components = {};
 
 /** Reclaim a component for later re-use by the recycler. */
 export function collectComponent(component) {
-	let name = component.constructor.name;
-	(components[name] || (components[name] = [])).push(component);
+	(components[component.constructor.name] || (components[component.constructor.name] = [])).push(component);
```
NekR: Is this really smaller? Maybe gzipped better

developit: Might be after gzip yeah, but I worry about the duplicated property access for performance reasons.

tomaszlakomygp:
Thanks for the review, I've learned a lot 👍 I'll amend this PR and write some unit tests when I'll have some time :)

developit:
I should put a guide together for running the performance tests too, that's something that is basically undocumented right now.

src/h.js
probablyup: This also brings in 2 new allocations, which seems unnecessary.

developit: This one actually breaks `shouldComponentUpdate` for leaf nodes, since `props.children` will be different every time, even if it's empty.

tomaszlakomygp:
Looks like I'll wait for the performance test guide before trying to improve performance/bundle size :) Thanks for all the reviews though!

developit:
Sorry for the wait on that guide! It's on my list 😇
Appreciate the PR, it was good to call these things out and discuss them - sometimes I do perf optimizations and they just slip in unnoticed!

PR #641, opened 2017-04-11
Add `onSearch` event definition

jmfirth:
This adds the `onSearch` event definition.


idchlife:
This is no standard... is it?

jmfirth:
Not at this time, but none-the-less used.  https://developer.mozilla.org/en-US/docs/Web/Events/search. I don't particularly have interest in seeing this definition added as much as I have interest in seeing `preact-router` have working definitions and sensible test cases for them.  Perhaps the easier solution is changing the demo example `preact-router` uses.

developit:
I'd be fine adding `onSearch`, it's fairly widely used.

PR #647, opened 2017-04-15
Use slice instead of substring.

braddunbar:
Besides being shorter, the semantics of slice are usually preferred in my experience. 🎉


developit:
Works for me! Need to double check gzipped size (sometimes these things are crazy).

developit:
Funky - I'm seeing `slice()` show up as 1b larger when gzipped. Maybe because substring is used elsewhere?

PR #652, opened 2017-04-19
Add test for rendering a component with text nodes

braddunbar:
The implementation of `h` [suggests](https://git.io/vSApQ) that textual children of components should not be merged. However, there is no test coverage of this behavior. So I added some!

Is that actually the case? And if so, why are text nodes of components not merged?


developit:
This changed in 8. I'm actually wondering if we could bring back at least some of the behavior (assuming perf & size don't bloat as a result), since your other issue (`foo {0} -> ["foo ", 0]`) is kinda awkward to work around.  We need to gather some input, and maybe check what React does in this case.  I looked into this a while back and I believe they just pass `children` through without any normalization, which is why I made that switch in 8.

developit:
👍👍

PR #657, opened 2017-04-20
yarn support

suhasdeshpande:
@developit I started building a side project with preact and I didn't see yarn and thought I will raise my first PR.


lukeed:
From what I know, @developit doesn't use Yarn so I think this is intentional.

PR #664, opened 2017-04-23
Remove component recycling

NekR:
I went ahead 😱


NekR:
First hand info: this hurts performance.

NekR:
What about `shouldRecycle() { return false }` life-cycle method?

developit:
Can get some perf back by removing [this line](https://github.com/developit/preact/pull/664/files#diff-86702919356dad82cf58b1e4150288f5R219) - right now anything that was triggering a component recycle is discarding that component's DOM.  With recycling removed, it can instead keep it and diff it.  It's funny, that is actually how Preact worked prior to 4.x 😅

NekR:
@developit I'm still not sure about the right way here. I didn't test with your perf-test branch yet, but I'll eventually.

developit:
All good - this is something we have to be super careful with, it'll take a while and some new tests to verify we're not regressing performance (or if we are, that it's controlled).

developit:
Leaving a note here for myself and others. Here's the work I did to adjust the rest of the diff so that it actually leverages the recycler removal for some performance gains:
https://github.com/NekR/nekr-preact/compare/v8-no-recycling...developit:scratch/perf

PepsRyuu:
Is there any update on this? Had an issue with Preact showing old images until the new src loaded, and had to resort to embedding Preact into the project with "collectComponent()" commented out to fix it. Would be great to have a temporary option to disable component recycling until the proper removal and performance tweaks are in place.

kurtextrem:
This issue has been mentioned in the blog post about Uber Mobile: https://eng.uber.com/m-uber/. Just fyi.

marvinhagemeister:
Good news: We completely removed the recycler in Preact X. An alpha went out just a few hours ago 🎉

PR #667, opened 2017-04-24
replace for...in to Object.keys

ustccjw:
for...in will check all `__proto__` objects，use Object.keys will have better performance.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -3,8 +3,6 @@
  *	@private
  */
 export function extend(obj, props) {
-	for (let i in props) obj[i] = props[i];
+	Object.keys(props).forEach(key => obj[key] = props[key]);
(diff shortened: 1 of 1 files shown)


developit:
Sadly `Object.keys()` creates an intermediary Array, which makes it significantly slower than `for..in`.

Monomorphic benchmark:
https://esbench.com/bench/5870af58d5686f990078bc45

Megamorphic benchmark:
https://esbench.com/bench/58fea76899634800a0347e1c

In both cases, `for..in` is much faster.

ustccjw:
@developit thank you for reply.

developit:
No problem, thanks for the PR! Happy to have more people investigating performance 😊

PR #669, opened 2017-04-24
fix: allow render function to return null

KnisterPeter:
In react components are allowed to return null from the render function

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -1,5 +1,5 @@
 declare namespace preact {
-	interface ComponentProps<C extends Component<PropsType, any> | FunctionalComponent<PropsType>, PropsType> {
+	interface ComponentProps<C extends Component<any, any> | FunctionalComponent<any>> {
 		children?:JSX.Element[];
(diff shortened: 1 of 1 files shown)


developit:
Any TS folks like @mseddon want to approve?

mseddon:
LGTM!

src/preact.d.ts
```diff
@@ -1,5 +1,5 @@
 declare namespace preact {
-	interface ComponentProps<C extends Component<any, any> | FunctionalComponent<any>> {
+	interface ComponentProps<C extends Component<PropsType, any> | FunctionalComponent<PropsType>, PropsType> {
```
UnwrittenFun: What's the reason for adding PropsType here?

KnisterPeter: Isn't it just more specific/correct than 'any'?

UnwrittenFun: As PropsType isn't used in this interface and isn't constrained in any way, it provides no more safety than simply any, and only adds complexity.

Issue #671, opened 2017-04-25
Component attempting to render unexpected value renders <undefined></undefined> 


dharFr:
Components attempting to render something that is neither a component, or a `null`/`undefined` value, results in an `<undefined></undefined>` node rendered to the DOM.

e.g. : 
 - `const Zero = () => (0)` => renders `<undefined></undefined>`
 - `const Number = () => (42)` => renders `<undefined></undefined>`
 - `const False = () => (false)` => renders `<undefined></undefined>`

Here is a [test case](https://jsfiddle.net/dhar/2kx4f3oq/8/) to reproduce.

Not sure how it is supposed to behave, but rendering `<undefined></undefined>` does feel appropriate to me and might lead to unexpected behaviors.

dharFr:
Just tested out of curiosity, and React seems to [throw an error](https://facebook.github.io/react/docs/error-decoder.html?invariant=105&args[]=Zero) in such situations.

developit:
Ah yep, this makes sense.  In 7 or 8 (don't remember), we moved type handling into `h()`, but the root of a component is not passed through `h()` if it's a literal value.  Good call.

Seems like the solution is just to extend this condition a bit:
https://github.com/developit/preact/blob/master/src/vdom/diff.js#L69

Something like:

```js
if (vnode==null || vnode===false || vnode===true) vnode = '';

if (typeof vnode==='string' || typeof vnode==='number') {
  // ...
```

dharFr:
So the idea would be to avoid throwing an error by silently ignoring booleans + implicitly converting numbers, right?

Doesn't seem that difficult. I'll try to send a PR.

PR #673, opened 2017-04-25
Fix Render unexpected vnode types

dharFr:
A fix proposal for #671, including unit-test to reproduce the behavior.


dharFr:
You're welcome 👍

PR #674, opened 2017-04-26
Use build code from preact-compat

btd:
I used code from preact-compat (you said you are ok with it). So i used buble instead of babel etc.
Also i used umd as export target and gzip size increased to +40 bytes. Is this ok? (i know you care about size).


developit:
Awesome, thanks for putting this together.  I'm going to check it out locally and try to figure out why it's adding 40 bytes.  That's not too huge of a size increase, but it's worth doing a diff to check. Will post back here when I've been able to find anything.

btd:
@developit 40 bytes added because at the end build become umd instead of modules+browser globals (so it is amd support added it).

developit:
Ahh, okay. I usually keep `umd` for the `main` entry but use a simplified wrapper for the minified one.

btd:
@developit ping. If you think it should be updated i can do this. Just point me to wrapper for umd you are using.

xMartin:
Works for me. Would love to see AMD support!

developit:
It might have to be custom. I can do it post-merge, not sure I have anything that would work out-of-the-box.

xMartin:
Looks like UMD support was removed again by https://github.com/developit/preact/commit/5b04dbf0f04042638b1f3310274442707b9dd6ad.

PR #675, opened 2017-04-26
ES Module build

### LINKED ISSUE #670: Web-compatible ES module build
Hey @developit! I was wondering if you would be open to a new build that is ES modules compatible for the web?  I have a branch of this working locally. It uses a small custom babel plugin that turns stuff like `'./util'` into `'./util.js'`;

It works in Safari:

screen shot 2017-04-25 at 10 34 22 am

In my branch I run the babel plugin and put everything into `dist/es/`.   This took me like 5 minutes, super easy. **If** you are open to adding this, there are a couple of options:

## `dist/es` or some such folder

This would be the way I did it for my proof of concept. This means that browser users would use Preact with something like:

```js
import { Component } from './node_modules/preact/dist/es/preact.js';
```

This would require no changes to your `src/` codebase. If you wanted to keep src/ the way it is, this is the way to go.

## `preact.js` in root folder

A bit nicer solution for web users would be if you added `preact.js` to the root folder and all it did was:

```js
export * from './src/preact.js';
```

Or it could point to `'./es/dist/preact.js'`. If you are willing to change your src/ modules to add the `.js` at the end and point to exact files in that way, then no Babel transformations would be needed.

Users would only have to do:

```js
import { Component } from './node_modules/preact/preact.js';
```

Which is just a tiny bit easier.

Let me know what you think!  cc @treshugart

matthewp:
This adds an ES module build into `dist/preact.esm.js` as [described here](https://github.com/developit/preact/issues/670#issuecomment-297201594). Closes #670


developit:
😍😍

developit:
Hoping to be able to merge this and get a release out in the next couple of days. Sorry for the delay, life stuff taking priority for a little bit!  Anyone else able to review is much appreciated.

matthewp:
No worries, thanks!

jescalan:
I was just about to PR this myself -- looking forward to it!

developit:
Me too! I should have time to get this merged after Google I/O.

aduth:
Would this need also a `module` entry in `package.json` to allow it to be detected by bundlers?

https://github.com/rollup/rollup/wiki/pkg.module

developit:
indeed!

matthewp:
What should "module" point to? I'll update the PR with that if it will help move this along. Anything else needed?

aduth:
>What should "module" point to?

`dist/preact.esm.js` I believe

matthewp:
Added the field.

FredKSchott:
+1, would love to have this. I have some tooling in my build pipeline that needs the "module" field. 

While it's no longer required in Node's proposed ESM support plan, it's still the current standard for denoting an ESM entrypoint.

developit:
I really am hoping to get this into 8.2, sorry for the horrible delay everyone :(

developit:
so... finally merged!

matthewp:
Yay!

PR #676, opened 2017-04-26
Upgrade to webpack 2

gdub22:
Karma is using webpack to bundle the source for testing.


developit:
I'm surprised this works with Babel 5! haha

gdub22:
Ha yea, preact is still depending on `babel-loader` 5.x which uses babel 5.x.  Webpack is essentially just passing the options through.

PR #677, opened 2017-04-27
Fix attributes diff on DOM node

dmail:
### Expected behaviour 

Diff should remove attributes when they are no longer present in the next html state (see [unit test](https://github.com/developit/preact/compare/master...dmail-fork:master?expand=1#diff-a1195645838de4f080fcb358b5132a4cR499))

### Actual behaviour

It behaves as expected when the merged node is created using `h`.
It preserves removed attributes when the merged node comes from the DOM.

### Possible solutions

- I suggest to approve this pull request :)

### DIFF
--- a/src/vdom/diff.js
+++ b/src/vdom/diff.js
@@ -121,8 +121,16 @@ function idiff(dom, vnode, context, mountAll, componentRoot) {
 
 
 	let fc = out.firstChild,
-		props = out[ATTR_KEY] || (out[ATTR_KEY] = {}),
+		props,
 		vchildren = vnode.children;
+
+	if (out[ATTR_KEY]) {
+		props = out[ATTR_KEY];
+	}
+	else {
+		props = out[ATTR_KEY] = {};
+		for (let a=out.attributes, i=a.length; i--; ) props[a[i].name] = a[i].value;
+	}
 
 	// Optimization: fast-path for elements containing a single TextNode:
 	if (!hydrating && vchildren && vchildren.length===1 && typeof vchildren[0]==='string' && fc!=null && fc.splitText!==undefined && fc.nextSibling==null) {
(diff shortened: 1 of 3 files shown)


src/vdom/diff.js
```diff
@@ -124,6 +124,8 @@ function idiff(dom, vnode, context, mountAll, componentRoot) {
 		props = out[ATTR_KEY] || (out[ATTR_KEY] = {}),
 		vchildren = vnode.children;
 
+	for (let a=out.attributes, i=a.length; i--; ) props[a[i].name] = a[i].value;
```
developit: we'll need to wrap this in a check for props. If the propcache already exists on an element, skipping attribute checks is a fairly major performance gain.

developit: I'm fine doing this post-merge if you want.

dmail:
@developit code updated to loop over attributes only when not cached : https://github.com/developit/preact/pull/677/files#diff-a0896fa5d65b1d80ca4d496badbae50fR127

src/vdom/diff.js
```diff
@@ -123,9 +123,17 @@ function idiff(dom, vnode, context, mountAll, componentRoot) {
...
 	let fc = out.firstChild,
-		props = out[ATTR_KEY] || (out[ATTR_KEY] = {}),
+		props,
 		vchildren = vnode.children;
 
+	if (out[ATTR_KEY]) {
+		props = out[ATTR_KEY];
+	}
+	else {
+		props = out[ATTR_KEY] = {};
```
developit: I'll probably consolidate this post-merge, but it's good!

dmail:
Hi @developit, is there a planned release with this fix ?

developit:
Yup, 8.2 out soon.

timmywil:
@developit would love that release!

developit:
Me too, heh.

PR #683, opened 2017-05-02
Babel 6 upgrade

gdub22:
Building & testing working 100%.
~~Diff with old babel 5 output: https://www.diffchecker.com/VmO4dPJb~~
Output is identical to babel 5 (via 1 added codemod)

TODO:
- [x] Investigate why code coverage dropped
- [x] Restore gzip size, which went up slightly.  I *think* this is because of the renamed let variables, which has no effect on the minified size, but being minified to different characters effects the gzip size?


gdub22:
Update: Same output as babel 5! https://www.diffchecker.com/VmO4dPJb
(They just underscore a few local variables, which doesn't matter when minified)

xtuc:
Awesome :+1: 

Note that Babel 7 will be released soon and the upgrade should be easy.

gdub22:
Not a big code coverage guy, so anyone that can shed light on why that dropped let me know.  

As far as I can see it's complaining about no coverage for `cloneElement`, which there are indeed no tests for.  Therefore, I assume the code coverage was incorrect before and this uncovered the problem.

gdub22:
@developit this is good to go for review now

config/codemod-let-name.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Restores var names transformed by babel's let block scoping
+ */
+export default (file, api) => {
+	let j = api.jscodeshift;
+	let code = j(file.source);
+
+	code.findVariableDeclarators().filter(d => /^_i/.test(d.value.id.name)).renameTo('i');
```
developit: Slight chance this might collide with other variables, right?

gdub22: Yes it would be possible if someone added a var with that prefix. No scope analysis. 

I personally wouldn't add this mod at all. I just wanted to make the output identical.

developit:
So excited to have this at size parity! 💃

treshugart:
I'm in the process of adding custom element support (<CustomElement />) and need the fixed class transpile that Babel 6 gives us. Anything I can do to help move this along, please let me know!

xtuc:
Congrats 💃

PR #694, opened 2017-05-10
TypeScript definitions update: allow tabIndex value to be string

winding-lines:
A lot of devs set tabIndex="1", this diff allows that use case in addition to
tabIndex = {1}.

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -574,7 +574,7 @@ declare namespace JSX {
 		step?:number | string;
 		style?:any;
 		summary?:string;
-		tabIndex?:number;
+		tabIndex?:number | string;
 		target?:string;
 		title?:string;
 		type?:string;


marvinhagemeister:
Thanks for your PR. As a preact ts user myself I'd very much prefer it to be of type `number`. The spec says that only [valid integer values](https://www.w3.org/TR/html5/editing.html#sequential-focus-navigation-and-the-tabindex-attribute) are supported.

Allowing strings would prevent TS from throwing an error for the following code which is invalid jsx:

```tsx
function MyComponent() {
  // No compile error is raised here, although "abcde" is obviously invalid :/
  return <input tabIndex="abcde" type="text"/>
}
```

developit:
One thought: technically the value could be `null`, `undefined` or `""` and those would be valid.

marvinhagemeister:
@developit hm, the spec is quite explicit on only allowing integers. My guess is that `null` or `undefined` are ignored by the browser and effectively the same as not specifying `tabindex` at all?

robertknight:
Thank-you for the contribution @winding-lines, but I'm in agreement with @marvinhagemeister. The spec is [quite clear](http://w3c.github.io/html/editing.html#the-tabindex-attribute) that this value is supposed to be an integer and it is in [the React typings](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f4aad91a85398172a488724e1ef776a3ca75de67/types/react/index.d.ts#L2197) too.

PR #695, opened 2017-05-11
Typings: camel-cases two attributes, alphabetizes, and adds htmlFor

jmrog:
This PR corrects the case of two properties in the `JSX.HTMLAttributes` interface in the type declarations. It also moves the `for` property to the appropriate spot (alphabetically) so that it's easier to locate. Finally, it adds the `htmlFor` property. I'm not sure whether that last change is desired -- since `for` is in the type declarations, maybe it's intended that we just use that in Preact rather than `htmlFor`? -- so I'm open to removing it. I'll just note that, IMO, including it provides more consistency with React's typings, as well as with Preact's decision to allow both `class` and `className`.


marvinhagemeister:
@jmrog Great find! Thanks a lot for your PR, much appreciated 🎉

PR #696, opened 2017-05-16
Use typeof to test booleaness.

bmeurer:
Using the typeof check is shorter, and in case of Chrome also faster at
this point as shown in https://esbench.com/bench/591b46f299634800a03481f8
(upstream bug to address this at http://crbug.com/v8/6403). I personally
also find the typeof version more readable, as it makes it clear what
the intent was.


developit:
Any impact on filesize? That was the original reasoning behind the strict equality checks, but maybe the impact is small enough it's worthwhile.

PR #699, opened 2017-05-17
bugfix: children doesn't support ['one', , ] array

luolonghao:
children doesn't support `['one', , ]` array, the array become to `['one', undefined, undefined]`

the bug breaks antd-mobile picker component .

* https://github.com/ant-design/ant-design-mobile/blob/master/components/picker/index.tsx#L57
* https://github.com/react-component/m-cascader/blob/master/src/Cascader.tsx#L66


src/h.js
```diff
@@ -22,7 +22,9 @@ export function h(nodeName, attributes) {
 	}
 	while (stack.length) {
 		if ((child = stack.pop()) && child.pop!==undefined) {
-			for (i=child.length; i--; ) stack.push(child[i]);
+			[].concat(child).reverse().forEach((item) => {
```
developit: An `if` statement would be better here. Functions, concat and reverse are all very expensive.

developit:
FWIW we'll need to check what React does here. AFAIK children is supposed to be passthrough for Components.

src/h.js
luolonghao: now use `if` statement, reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach?v=example

developit:
`in` check seems reasonable. Need to check for filesize.

luolonghao:
@developit are you still need to check something? I think the changed code is compatible and necessary.

andrewiggins:
I believe this is fixed in Preact X - I think we just pass `props.children` as is to components so gonna close this PR since it is opened against Preact 8. If you still have issues with using the libraries mentioned in the description, let us know!

PR #706, opened 2017-05-24
Accurate code coverage + more tests

### LINKED ISSUE #685: Code coverage is not accurate
The karma runner is compacting the src before running code coverage, therefore dead code eliminating anything that isn't used:

https://github.com/developit/preact/blob/9001436a453d736267359538e2d822ccf23e3115/test/karma.conf.js#L124

For example, `cloneElement` is never tested/imported.  Disable that line and you'll see coverage drop.  Uncovered during https://github.com/developit/preact/pull/683

gdub22:
Fixes #685 
The `compact` babel option in karma code coverage put all of the code onto only 15 lines so they weren't actually all covered.

This removes `compact` and adds missing tests.


gdub22:
Coveralls coverage is expected to drop/fail

developit:
Just checked the docs and AFAIK `compact` just removes whitespace within inlined babel helpers - it doesn't do any kind of dead code elimination or anything fancy like that.  It seems like the coverage dropped because there are now more lines of code added by Babel that can't be tested?

FWIW the `cloneElement()` coverage is missing because that file is never imported in the tests.  If we added an empty test with just an import, it would drop accordingly.  Maybe that's the more direct option here?  I think `cloneElement()` is the only case where that is true.

gdub22:
Yup you're right, compact just doesn't add whitespace.  I incorrectly assumed it meant 'minify'.  Unsure why that would cause coverage to drop. I'll dig some more and add the cloneElement test.

gdub22:
Ok, so code coverage is just counting 'lines' of code.  So when compact is on, *all* the code is only on 15 lines.  A line is covered as long as anything on it is accessed, not necessarily all the functions on that line.  Therefore, removing compact is still important.

Importing/calling cloneElement in the tests gets the 'functions' coverage up to 100%.  Not sure which other lines still need to be hit.  Anyone?

With `compact: true`
```
Functions    : 97.44% ( 38/39 )
Lines        : 100% ( 15/15 )
```

Without `compact`:
```
Functions    : 97.44% ( 38/39 )
Lines        : 98.36% ( 481/489 )
```

Without `compact` & cloneElement tested:
```
Functions    : 100% ( 39/39 )
Lines        : 98.57% ( 482/489 )
```

developit:
This is good, and happy to have that last mile testing for sure.  Think we're good to merge now?

gdub22:
Good to go!

PR #717, opened 2017-06-01
Only execute appendChild when vnode length greater than origin node.

### LINKED ISSUE #705: Input losing focus
Input losing focus after insert elements before it. Example:
```javascript
export class TodoList extends Component {
    state = { todos: [], text: '' };
    setText = e => {
        this.setState({ text: e.target.value });
    };
    addTodo = (e) => {
	if (e.keyCode === 13) {
	    let { todos, text } = this.state;
            todos = todos.concat({ text });
            this.setState({ todos, text: '' });
	}
    };
    render({ }, { todos, text }) {
        return (
            <div onKeyDown={ this.addTodo }>
                { todos.map( todo => (<div>{todo.text}</div> )) }
		<input value={text} onInput={this.setText} />
            </div>
        );
    }
}
```

windyGex:
Fix #705 .

### DIFF
--- a/src/vdom/diff.js
+++ b/src/vdom/diff.js
@@ -222,7 +222,7 @@ function innerDiffNode(dom, vchildren, context, mountAll, isHydrating) {
 			child = idiff(child, vchild, context, mountAll);
 
 			if (child && child!==dom) {
-				if (i>=len) {
+				if (i>len) {
 					dom.appendChild(child);
 				}
 				else if (child!==originalChildren[i]) {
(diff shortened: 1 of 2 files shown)


developit:
Awesome - and awesome test!  I'd love to duplicate this Todos example and run some more things against it too.

e1ectronic:
The fix doesn't work in the following example (where each child consists of three elements): https://jsfiddle.net/o402afwp/

developit:
@e1ectronic good catch! - that's because `len` is the length before any children get added, so it doesn't take middle insertions into account.  I have a fix that avoids using lengths entirely, and your demo was very useful in figuring it out!  It's now working wondefully:

preview

PR #719, opened 2017-06-03
Add Play.cash to demos list

feross:
My new music site is built using Preact! https://play.cash


developit:
Epic! I had meant to add this when I first saw it ;)

PR #726, opened 2017-06-08
Add preact-render-to-json

nathancahill:
Useful for Jest snapshot testing Preact/JSX components.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ Preact supports modern browsers and IE9+:
 
 - :raised_hands: [**preact-compat**](https://git.io/preact-compat): use any React library with Preact *([full example](http://git.io/preact-compat-example))*
 - :page_facing_up: [**preact-render-to-string**](https://git.io/preact-render-to-string): Universal rendering.
+- :loop: [**preact-render-to-json**](https://git.io/preact-render-to-json): Render for Jest Snapshot testing.
 - :earth_americas: [**preact-router**](https://git.io/preact-router): URL routing for your components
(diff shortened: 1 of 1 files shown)


developit:
Oh wow, thanks for putting that together!  I was wondering - how much did you have to change from preact-render-to-string to get it to output JSON?  Just wondering about repos and that kind of stuff.  Happy to have it though!

nathancahill:
I kept all of the logic and stripped out the JSX and string stuff. Basically returns:

```
{
    type: nodeName,
    props: attributes,
    children: pieces
}
```

developit:
one thing I noticed - we can rip out the encodeEntities entirely, JSON doesn't require HTML entity escapement

nathancahill:
You're right of course. I'll update that.

oren:
Is that the preferred way to test components in Preact? If so this should be part of the official docs. It will be very helpful for beginners to see a good way of testing in Preact.

nathancahill:
Well, I literally created the project/repo a couple hours ago, so "preferred" is relative 🙂

But I did explore the docs looking for a section on testing first, before forking preact-render-to-string to create this project. So it might be good to add a docs section on testing.

developit:
@oren It's for Jest only, but a great solution.  I'd like to include it in preact-boilerplate and potentially preact-cli if it works as well as it looks like it should.

oren:
I understand that it is best only. I would think that having it as part of the official docs is valuable since we currently have nothing about testing.

Issue #620, opened 2017-04-02
Remove ComponentLifecycle interface in favour of individual interfaces


UnwrittenFun:
I'd like to propose removing the [ComponentLifecycle](https://github.com/developit/preact/blob/master/src/preact.d.ts#L25-L34) interface and replace it with individual interfaces, e.g.
```ts
interface WillMount {
    componentWillMount(): void;
}

interface WillReceiveProps<PropsType, ContextType> {
    componentWillReceiveProps(nextProps: PropsType, nextContext: ContextType): void;
}
```

This way instead of Component implementing the ComponentLifecycle interface where everything is marked optional, the consumer would be able to explicitly implement WillMount, for example.

This allows for greater type safety as TypeScript will be able to know if the interface is implemented incorrectly, something it can't do with the current solution.
![](https://i.uwf.io/nwaIQg.png)

This has the added benefit of allowing vscode (and presumably other editors) to auto implement the interfaces. (And I wouldn't have to flick to the docs because I forgot the signature literally every time I need to implement a lifecycle event 😳)

I realise this would deviate from the React typings, but I am hoping that is not a major concern.

robertknight:
Unless I'm missing something, splitting up the interface shouldn't be necessary to get type safety. Since JavaScript does not support overloaded methods (ie. there can only be one implementation with a given name), then TypeScript knows that if a particular optional method is implemented then it must conform to the signature of any method with the same name in implemented interfaces.

```ts
import { Component } from 'preact';

class Thing extends Component<any,any> implements preact.ComponentLifecycle<any,any> {
       // error TS2420: Class 'Thing' incorrectly implements interface 'ComponentLifecycle<any, any>'
	componentWillMount(a: string): number {
		return 42;
	}

	render() {
		return null;
	}
}
```

One slightly surprising thing I noticed here is that `implements` is not transitive, so although `Component` is declared to implement `ComponentLifecycle`, subclasses of `Component` are not automatically required to implement it.

UnwrittenFun:
Yeah if you implement the ComponentLifecycle interface on your class that extends Component it will bring back that type safety, however as it stands Component already implements ComponentLifecycle and it would be odd to implement it twice.

The ComponentLifecycle interface as it is, isn't very useful as implementing it gives no guarantees to the methods on the class and when implementing the methods, gives no type info for the parameters (though separating them out also does not provide this, it does allow editor codegen)
![](https://i.uwf.io/rxxUvO.png)

developit:
[...]

PR #730, opened 2017-06-13
Preparation for typescript 2.4

kwoon:
[Typescript Changes](https://blogs.msdn.microsoft.com/typescript/2017/06/12/announcing-typescript-2-4-rc/)

Please read about Week Types.

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -45,7 +45,7 @@ declare namespace preact {
     // Type alias for a component considered generally, whether stateless or stateful.
 	type AnyComponent<PropsType, StateType> = FunctionalComponent<PropsType> | typeof Component;
 
-	abstract class Component<PropsType, StateType> extends ComponenLifecycle<PropsType, StateType> {
+	abstract class Component<PropsType, StateType> extends ComponentLifecycle<PropsType, StateType> {
 		constructor(props?:PropsType, context?:any);
 
 		static displayName?:string;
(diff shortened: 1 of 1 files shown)


UnwrittenFun:
I feel like this further supports #620

marvinhagemeister:
This PR fixes an issue we just stumbled upon at work:

```bash
48  abstract class Component<PropsType, StateType> implements ComponentLifecycle<PropsType, StateType> {
                   ~~~~~~~~~

node_modules/preact/dist/preact.d.ts(48,17): error TS2559: Type 'Component<PropsType, StateType>' has no properties in common with type 'ComponentLifecycle<PropsType, StateType>
```

+1 on merging this PR

marvinhagemeister:
@developit Any chance this might get merged? This makes preact work again with the latest TypeScript version.

disintegrator:
Great stuff! I also wanted a fix TS 2.4 compile error. While this does overcome the weak type overlap compile error, it does result in an unnecessary abstract class. `ComponentLifecycle<P,S>` isn't really an appropriate abstract class to expose to consumers as part of the public API. Furthermore, it is not referred to or extended by any type other than `Component<P,S>`. Perhaps, the more appropriate solution is to place all the lifecycle methods under `Component<P,S>` and then deleting `ComponentLifecycle<P,S>`. What do y'all think?

marvinhagemeister:
That sounds great as well! The `ComponentLifeCycle` interface only has one implementation anyway.

valotas:
Not directly related, but shouldn't the props be of type `PropsType & ComponentProps<this>` for the ComponentLifecycle methods?

Issue #732, opened 2017-06-19
ComponentWillUnmount throwing error, when refering to a DOM node


Ninerian:
# Description
## 1. 
When a component is replaced with another one, the DOM is updated, before the `componentWillUnmount` event is called. This prevents the execution of scripts, depending on the DOM Elements of the component. 

I created an [example](http://jsfiddle.net/ninerian/v4qx9zvz/2/). When you change from the Home tab to another one,  you should get an alert, with the result of the shown query. Inside the console, I log the trace of the call.

## 2.
Additionally the `componentWillUnmount` is iterated in a top-down approach, so if a child of a component has a `componentWillUnmount` set, it's fired, when the containing component is already removed, resulting in the same error.

This [example](http://jsfiddle.net/ninerian/v4qx9zvz/4/) already contains the fix for No. 1 and shows the problem of the nested component in the same way, as ex 1.

# Expected behavior
`componentWillUnmount` should be called, before the DOM Nodes are removed.

# Shown behavior
`componentWillUnmount` is called, after the DOM Nodes are removed.

# Source of problem
## 1.
In [vdom/component.js](https://github.com/developit/preact/blob/master/src/vdom/component.js#L146) the child is replaced and after that replacement, 'componentWillUnmount' is called.

## 2. 
In [vdom/component.js](https://github.com/developit/preact/blob/master/src/vdom/component.js#L267) the base node is removed and the the children are collected and removed. 

# Solution
## 1.
Change [vdom/component.js](https://github.com/developit/preact/blob/master/src/vdom/component.js#L146) to 
```
if (baseParent && base!==baseParent) {

  baseParent.replaceChild(base, initialBase);
  
  if (!toUnmount) {
    initialBase._component = null;
    recollectNodeTree(initialBase, false);
  } else {
    baseParent.insertBefore(initialBase,base);
    unmountComponent(toUnmount);
    toUnmount = null;
   }
}
```

## 2.
The traversal of the DOM tree should be bottom top, while calling `componentWillUnmount`. After reaching the top node, which has to be removed, it could be removed without removing each child separately.

PR #733, opened 2017-06-20
Fix: Invoke componentWillUnmount before the DOM element is removed

Ninerian:
This PR changes fixes the Issue #732.

- it removes the Node in `unmountComponent` and `recollectNodeTree` after the iteration steps
- it replaces the children in `renderComponent` only when the component isn't unmounting
- it extends a test to cover multiple nested components

The change I commented in component.js was discovered by accident. As the tests are passing and a quick test with a small app too, I don't know, if it's breaking something else. So someone with more preact knowledge could estimate the consequences better than me.

### DIFF
--- a/src/vdom/component.js
+++ b/src/vdom/component.js
@@ -144,12 +144,18 @@ export function renderComponent(component, opts, mountAll, isChild) {
 		if (initialBase && base!==initialBase && inst!==initialChildComponent) {
 			let baseParent = initialBase.parentNode;
 			if (baseParent && base!==baseParent) {
-				baseParent.replaceChild(base, initialBase);
 
 				if (!toUnmount) {
+					baseParent.replaceChild(base, initialBase);
 					initialBase._component = null;
 					recollectNodeTree(initialBase, false);
 				}
+				// @TODO Formerly the initialBase was always replaced by the base
+				//		 but somehow all tests passes also if its only called at not unmounting.
+				//       These else condition was the original fix.
+				// else {
+				// 	baseParent.insertBefore(base, initialBase);
+				// }
 			}
 		}
 
@@ -264,10 +270,11 @@ export function unmountComponent(component) {
 
 		component.nextBase = base;
 
-		removeNode(base);
 		collectComponent(component);
-
 		removeChildren(base);
+
+		// Remove Node as last, so all childs exists when componentWillUnmount is called
+		removeNode(base);
 	}
 
 	if (component.__ref) component.__ref(null);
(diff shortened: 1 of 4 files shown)


src/vdom/diff.js
```diff
@@ -262,11 +262,12 @@ export function recollectNodeTree(node, unmountOnly) {
 		// (this is part of the React spec, and smart for unsetting references)
 		if (node[ATTR_KEY]!=null && node[ATTR_KEY].ref) node[ATTR_KEY].ref(null);
 
+		removeChildren(node);
```
developit: moving this here will be very very expensive since it makes tree deconstruction depth-first.

src/vdom/component.js
```diff
@@ -144,12 +144,18 @@ export function renderComponent(component, opts, mountAll, isChild) {
...
 			let baseParent = initialBase.parentNode;
 			if (baseParent && base!==baseParent) {
-				baseParent.replaceChild(base, initialBase);
 
 				if (!toUnmount) {
+					baseParent.replaceChild(base, initialBase);
 					initialBase._component = null;
 					recollectNodeTree(initialBase, false);
 				}
+				// @TODO Formerly the initialBase was always replaced by the base
```
developit: This is probably more nuanced than I can feasibly check in a PR review, haha.
Was the issue here that `replaceChild()` removes `initialBase` from the DOM before `recollectNodeTree()` invokes `componentWillUnmount()` on the owning component?

Ninerian: Yeah, this was the issue. My solution was to prepend the new element before the existing one, as this gets removed in the ‘unmountComponent’ function. But it seems, that the replacements is additionally done later in the stack. This would explain the passing tests.

src/vdom/diff.js
Ninerian: Then I would suggest to either call only the unmountComponent depth first, or wait until the most upper dirty node, to remove it.

Juul:
I can confirm that this is still an issue and that it is fixed by this pull request. It's really a pretty major problem that one of the major lifecycle functions is defective in preact.

Zjcompt:
Any word of getting this into Preact officially? componentWillUnmount is currently broken and is a basic component lifecycle method.

The docs are currently incorrect or at least needs to be updated to reflect the bug discussed in this PR: https://preactjs.com/guide/api-reference & https://preactjs.com/guide/getting-started
It states under Lifecycle Methods & The Component Lifecycle that componentWillUnmount is called prior to removal from the DOM. 

https://preactjs.com/guide/external-dom-mutations
States clearly in code that:
`  componentWillUnmount() {
    // component is about to be removed from the DOM, perform any cleanup.
  }
`

marvinhagemeister:
The code in `master` is very different now that we merged Preact X. We removed the reconciler because it was the source of many bugs :+1: 

@Zjcompt Updating the docs takes the same amount of time as just fixing this, so it's a more worthwile investment to just fix the issue.

PR #737, opened 2017-06-24
Ts improvements 2

valotas:
Tried to improved a little bit the types. Mostly copying React's definitions where needed. There is a breaking change though. I removed `ComponentProps` as in my opinion is not needed. The recommended way with react and ts2 is to use the ClassAttributes interface.


valotas:
On top of that, I believe that we need some more tests. Any ideas? In my opinion the best way to have properer testing would be to write tests in ts, compile them and then run them to also test some behavior. For example I use `JSX.Element[]` as the type of the children but I am not sure if this is what I get back.

src/preact.d.ts
```diff
@@ -1,39 +1,54 @@
 declare namespace preact {
-	interface ComponentProps<C extends Component<any, any> | FunctionalComponent<any>> {
-		children?:JSX.Element[];
+	type Key = string | number;
+	type Ref = (el?: Element) => void;
```
UnwrittenFun: This is not correct when used on a component.

src/preact.d.ts
```diff
@@ -51,19 +66,24 @@ declare namespace preact {
...
+		props:Readonly<PropsType> & Readonly<{children?: JSX.Element[]}>;
 		context:any;
 		base:HTMLElement;
 
+		/**
+		 * @deprecated
+		 *
+		 * Not part of the core preact anymore
+		 */
 		linkState:(name:string) => (event: Event) => void;
```
UnwrittenFun: Considering the typings are shipped with Preact itself, this can just be removed

src/preact.d.ts
valotas: Sorry about that. What is the right type in the case of a component ref?

src/preact.d.ts
valotas: Gone! It would make sense though to add it to the new module.

src/preact.d.ts
UnwrittenFun: An instance of the component itself

valotas:
@developit I am in need of a review and maybe ideas one complex typescript usage. In theory this branch is ready for merging.

matthewmueller:
@valotas this seems to work really well for me! Quick question though, I didn't see any examples of how to use `children` with a regular component.

Is something like this possible?

```js
render(props: Props) {
  return (
    <div class="page">
      <div class="player">{props.children}</div>
    </div>
  )
}
```

valotas:
Sure, I'll add a test for it once I'll get back online (back from vacation)

valotas:
@matthewmueller I had to introduce the `RendableProps` for what you need. You can see an example usage at `Component-test.jsx`

JKillian:
👍 big fan of getting something like this merged and released! The current types have enough issues that it makes it very difficult to use Preact with TS. (For example, `props` being optional where it shouldn't be, `string`s not being allowed to be mixed with `Element`s properly, etc.)

matthewmueller:
@valotas thanks 😄! I'll be testing this out again next week.

@JKillian right now I'm just copying these into my project as `types/preact.d.ts` and adding definitions in my tsconfig:

```json
"paths": {
  "*": [
    "types/*"
  ]
}
```

a bit manual, but works fine.

valotas:
I forgot that. I'll try to fix the merge conflict this weekend.

valotas:
Merge conflicts fixed. As I saw previously the `base` attribute of the Component was not optional. Is this the case? Is there a guarantee that with a `Component` instance `base` will always be accessible?

matthewmueller:
@valotas nice just had a chance to try the updates. everything seems to be working well, but I had a quick question:

what's the reason to go with `RendableProps` over just attaching it to attributes? something like this:

```ts
interface Attributes {
    key?: string | number | any
    children?: JSX.Element[]
  }
```

Then using it:

```ts
export interface Props extends Attributes {
  filepath: string
  head: any
  style: any
  page: any
}

export default class Document extends Component<Props, void> {

}
```

developit:
^ attributes bit makes sense to me.

Regarding `base` - it's always added by Preact, though it is `undefined` in the constructor and `componentWillMount()` (as part of mount lifecycle timing).

developit:
If you guys greenlight this I think we can test them out in a release. Any type definition breakages tend to get noticed and reported quickly haha.

matthewmueller:
Oh I get it now, attributes are for the class itself (in the example above: `Document`). Nonetheless, I think it'd be a bit easier to remember if we could use extend.

Rather than:

```ts
interface Props {
  filepath: string
  head: any
  style: any
  page: any
}

export default class Document extends Component<Props, void> {
  render({ children }: RendableProps<Props>) {
    // ...
  }
}
```

Unfortunately, this is really due to a limitation around Typescript definitions (while inherited classes are type-checked, the types don't flow through).

Just for others, in the case above, you can do the types yourself pretty easily, which is what I'd probably end up doing:

```ts
export interface Props {
  filepath: string
  head: any
  style: any
  page: any
  children?: JSX.Element[]
}

export default class Document extends Component<Props, void> {
  render({ children }: RendableProps<Props>) {
    // ...
  }
}
```

But, i still think it'd be better if it was something like this:

```ts
import { h, Component, Props } from 'preact'

interface DocumentProps extends Props {
  filepath: string
  head: any
  style: any
  page: any
}

export default class Document extends Component<DocumentProps, void> {
  render({ children }: DocumentProps) {
    // ...
  }
}
```

valotas:
Sorry for my late answer.

@matthewmueller I believe you answered the question regarding `Attributes` yourself :) I will rename Rendable to `Renderable`. Yet, if someone has a better idea regarding this, let me know. I do not like the name either. Regarding it's use, I tend to use `this.props` instead which inherits the types :).

Apart from that I do not understand the problem with `preact/devtools`. Never saw that error before.

matthewmueller:
> @matthewmueller I believe you answered the question regarding Attributes yourself :) I will rename Rendable to Renderable. Yet, if someone has a better idea regarding this, let me know. I do not like the name either. Regarding it's use, I tend to use this.props instead which inherits the types :).

Ah I see. My vote would be it being an interface as opposed to a type with a generic. So we could do:

```jsx
interface Props extends Renderable {
  name: string
}
```

Could also be an additional type.

> Apart from that I do not understand the problem with preact/devtools. Never saw that error before.

I think happens if you copy + paste the type into your project. Shouldn't be a problem once this gets merged!

**Two more things:**

1. Noticed a bug with the VNode. `children` can also accept as string (for textnodes). Should be:

```ts
interface VNode<P> {
    nodeName: string | ComponentClass<P, any> | SFC<P>
    attributes: P
    children: Array<VNode<any> | string>
    key: Key | null | undefined
  }
```

2. What do you think about adding an additional property `jsx?: boolean` to `HTMLAttributes`? 

Basically to support styled-jsx a bit better. I wish they didn't use something custom, but at this point, prettier also has styled-jsx auto formatting via that tag, so seems like it's getting standardized sorta.

---

Happy to help push this forward if you don't have time. It's just a bit weird on Github branching off a branch.

developit:
@matthewmueller for Strings, `children` boxes them into an Array:

```js
h('div', null, 'a string').children // ['a string']
```

matthewmueller:
@developit This would be related to creating raw VNodes. e.g. 

```ts
const style: VNode = {
  nodeName: "style"
  attributes: {
    id: "some-id"
  },
  children: ["hi there"]
}
```

Right now the type definition only allows something like this (which I don't think would work):
```ts
const style: VNode = {
  nodeName: "style"
  attributes: {
    id: "some-id"
  },
  children: [{
     nodeName: "hi there"
  }]
}
```

developit:
^ correct, that would result in preact rendering a `<hi there>` element (which would throw an exception).  I'd definitely recommend always passing things through `h()` - it does children normalization that the rest of the diff kinda expects to be already done by the time it runs.

--edit--:  I totally misread your type definition - you're right, it needs to be `Array<VNode | string>`!  I had read it as "an array of vnodes, or a string".

valotas:
Ok, fixed the `VNode` types. Can someone explain me the jsx property thingy? I think I didn't understand that or never stumbled upon this pattern.

<hr/>

Apart from that. I also do not like the `RendableProps` approach and for sure it's name. But I can not think of something better. In react's types there is no type/interface exposure. The only reason that I define a type here is to allow proper typed destructuring on the render function. I am also not fun of the base props class as children is something that you should not care about when you define the properties of your component. You get it for free from the framework and that is how they should be defined. Have in mind that react used to have the a base class and not anymore.

Never the less I am happy to remove the `RendableProps`, inline the types at the render function and provide a `Props` interface for easy access in the render function. I have to state again though that I do not like the idea. 

My workaround is to destructure `this.props` in the render function instead of the the function's arguments.

<hr />

Finally, please let me know of any name changes. Changing this afterwords would make preact backwards incompatible. I'll check again against the react's type names. I believe it makes sense to be as close as possible there.

matthewmueller:
@valotas awesome! I'm much more interested in seeing this get in there than the specific types can be done in PRs later.

After closer inspection of the TS test, it's not terrible to just add, when needed:

```
render({ initialName, children }: RendableProps<SimpleComponentProps>) {
```

I'd still prefer it to be called RenderableProps, but not a big deal either way.

valotas:
@matthewmueller sorry, I forgot to rename it. `RenderableProps` it is now!

valotas:
One more thing that I mentioned above: what should I do with the `jsx: boolean`? I haven't seen its use anywhere at the moment. Can someone give me an example or try to explain it to me?

matthewmueller:
Awesome :-) I'd say @developit's call there. Doesn't really hurt, but it's also nonstandard.

https://github.com/zeit/next.js uses it via https://github.com/zeit/styled-jsx, so it'd be handy for next.js folks. 

The main issue for me is that prettier doesn't support formatting without the jsx field, so I was sorta falling into a pit of despair. That's likely to not be an issue soon though: https://github.com/prettier/prettier/pull/2774

valotas:
Added `jsx` attribute too. Shall I squash the commits?

matthewmueller:
+1, thanks for all your hard work @valotas!

valotas:
@developit Is this going to be merged anytime soon? Is there anything else needed? It is open for 3 months now.

hkrutzer:
Could this be rebased and merged? 🎉

valotas:
Rebased. Merging is not up to me

valotas:
@UnwrittenFun can you review that again?

valotas:
@developit I wouldn't like to bother you with ts stuff, but if you are not willing to merge this, I would propose to just close the pr. It doesn't make sense to clean conflicting files every week.

kristoferbaxter:
Spoke with @developit and I’ll take a look at this over the next day or so.

Thank you for your continued persistence and patience.

hkrutzer:
Maybe the definitions can be moved to DefinitelyTyped.

marvinhagemeister:
My vote would go for `DefinitelyTyped` as well, if this really won't be merged. There is really great stuff in this PR and it would be a shame to leave it to collect dust

arusanov:
+1 for moving typings to `DefinitelyTyped`. Theres a ton of PR open with TS improvements none of them got merged for months and current state of them is far from React typings. Please let people more familiar with TS manage them as they found more suitable.

marvinhagemeister:
@valotas I've just become part of the core preact team and I'm currently going through all TypeScript related PRs. Would really love to merge this. Any chance you could rebase the changes one last time?

valotas:
I'll have a look at it

marvinhagemeister:
@valotas Awesome! Thank you so much for your patience and sorry that it took so long🎉

PR #740, opened 2017-06-27
Correct the definition of ComponentLifecycle

RyanCavanaugh:
There is no value named `ComponentLifecycle`, so this should be represented as an interface rather than a class.

The correct way to copy down the optional members of an interface into a class declaration is to use declaration merging (see diff).

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -22,7 +22,7 @@ declare namespace preact {
 		key:string;
 	}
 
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
Awesome to see a core contributer of the TypeScript team here! I've always struggled with the different concept of values and declarations in the past, and your explanation nails it 👍

gnarf:
Can we push a release of this?

PR #741, opened 2017-06-29
Add library size check in CI

siddharthkp:
This will help in **keeping** preact tiny.

It's basically,

this in the terminal: 

```
 PASS  ./dist/preact.min.js: 3.37kB < threshold 4kB gzip
```

and this in the pull request: 

1. Setting the file and threshold:

    The config sits inside `package.json`, the threshold is set to 4kB right now.
You can add multiple files here. ([docs](https://github.com/siddharthkp/bundlesize/blob/master/README.md#configuration))

2. Authorize bundlesize:

    `bundlesize` needs access to github API with scope `repo:status`, You can follow the [steps here](https://github.com/siddharthkp/bundlesize/blob/master/README.md#build-status)

Happy to hear your feedback 😄

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "babel-eslint": "^6.1.0",
     "babel-loader": "^5.3.2",
     "babel-runtime": "^5.8.24",
-    "bundlesize": "^0.5.5",
+    "bundlesize": "^0.5.6",
     "chai": "^3.4.1",
     "copyfiles": "^1.0.0",
     "core-js": "^2.4.1",
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "babel-eslint": "^6.1.0",
     "babel-loader": "^5.3.2",
     "babel-runtime": "^5.8.24",
-    "bundlesize": "^0.5.6",
+    "bundlesize": "^0.5.7",
     "chai": "^3.4.1",
     "copyfiles": "^1.0.0",
     "core-js": "^2.4.1",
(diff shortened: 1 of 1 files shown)


robertknight:
This is awesome 👍 

I think it would be helpful to make it clear in the message that the file sizes are not the literal sizes of the bundle on disk but the size after gzipping.

siddharthkp:
@robertknight That's a great suggestion, done!

siddharthkp:
@robertknight Do you need some help with merging this?

lukeed:
I was just about to open this same PR. @developit you will like this.

developit:
Wow, I read about this earlier this morning and was hoping to have time to PR it today. You people are amazing.

Anything to set up for travis or permissions on my end?

package.json
```diff
@@ -120,5 +122,11 @@
       "babel-runtime",
       "jscodeshift"
     ]
-  }
+  },
+  "bundlesize": [
+    {
+      "path": "./dist/preact.min.js",
+      "threshold": "4Kb"
```
developit: I think "threshold" is the amount of bytes the build is allowed to change in a given PR/commit, so for Preact we can set this super low since the size isn't allowed to increase:  maybe something like 100b.

lukeed: I'm pretty sure it's the maximum size allowed for the lib. I believe it will just diff against the last-known size of `master` branch & do diff on `current - master`

siddharthkp:
> Wow, I read about this earlier this morning and was hoping to have time to PR it today.

Ha! I made this PR before posting it on twitter 😋 

> I think "threshold" is the amount of bytes the build is allowed to change

Nope, @lukeed is right, it's the absolute threshold after which the build will fail. But, I understand how that might be confusing. Happy to change it to something more obvious, suggestions?

siddharthkp:
> I've auth'd and added the `GITHUB_TOKEN` travis env var. Anything else I need to to?

Nope, just merge it 😄

You'll see the comparison with the next PR that comes up.

lukeed:
@siddharthkp 😅 Feel free to completely disregard, but I would have expected the current `threshold` behavior to be named `max`, and then use `threshold` as the permissible diff-gain before throwing an error.

Totally usable & workable as is though 🎉

siddharthkp:
@lukeed yeah, thinking back, threshold is a really bad name. I'll fix that soon.

developit:
I think I agree with @lukeed - happy to merge this and change if that gets updated.  Maybe to avoid having to rename you could add a new option called `maxDelta`, which sortof clarifies `threshold` is the absolute upper bound.

siddharthkp:
@developit This is not configured properly. I was using `github_token`, not `GITHUB_TOKEN`.

Now accepting both as valid in 0.6.1

I raised a PR for that: https://github.com/developit/preact/pull/744

developit:
and here I thought env vars were case-insensitive :P

siddharthkp:
So did I!

PR #743, opened 2017-06-29
style attribute update logic optimized

briefy:
I changed the style attribute update logic,
it is easier to understand and less code

### DIFF
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -47,13 +47,11 @@ export function setAccessor(node, name, old, value, isSvg) {
 		node.className = value || '';
 	}
 	else if (name==='style') {
-		if (!value || typeof value==='string' || typeof old==='string') {
+		if (!value || typeof value === 'string') {
 			node.style.cssText = value || '';
 		}
-		if (value && typeof value==='object') {
-			if (typeof old!=='string') {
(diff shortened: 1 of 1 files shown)


src/dom/index.js
```diff
@@ -47,13 +47,11 @@ export function setAccessor(node, name, old, value, isSvg) {
...
-		if (!value || typeof value==='string' || typeof old==='string') {
+		if (!value || typeof value === 'string') {
 			node.style.cssText = value || '';
 		}
-		if (value && typeof value==='object') {
-			if (typeof old!=='string') {
-				for (let i in old) if (!(i in value)) node.style[i] = '';
-			}
+		if (value && typeof value === 'object') {
+			node.style.cssText = '';
```
developit: This wipes CSS on every render, which would be very expensive.

developit:
I pulled the performance numbers from the two Travis builds and it looks like the impact is quite severe:

| Before | After
|---|----|
| screen shot 2017-07-01 at 3 03 49 pm | screen shot 2017-07-01 at 3 03 37 pm

briefy:
@developit 
I wonder why the empty diff drop so much, I see the performance test,
and on my computer,  I don't see such a drop.

marvinhagemeister:
Looks like this PR is abandoned.

PR #754, opened 2017-07-11
Fix preact/debug accessing attributes that don't have toString

plusCubed:
In some rare cases (e.g. `Object.create(null)`), a prop value doesn't have the `toString()` function, and thus cannot be cast to a string, throwing `TypeError: Cannot convert object to primitive value`. This should fix that.


developit:
I've looped @NekR in since he wrote this. Looks good to me though.

PR #772, opened 2017-07-26
Update createNode to use prevSvgMode for foreignObject

grahm:
Updating idiff createNode to use prevSvgMode instead of isSvgMode when the element is foreignObject because foreignObject is still an SVG node

https://github.com/developit/preact/issues/771


grahm:
This PR doesn't cover all uses of isSVG. attributes don't work.

PR #773, opened 2017-07-26
Updated diff to set isSVGMode to false for children of foreignObject …

grahm:
…not its self.

Because foreignObject is it's self an SVG element only its children should exit SVG Mode.

Git issue: https://github.com/developit/preact/issues/771

### DIFF
--- a/src/vdom/diff.js
+++ b/src/vdom/diff.js
@@ -102,7 +102,7 @@ function idiff(dom, vnode, context, mountAll, componentRoot) {
 
 
 	// Tracks entering and exiting SVG namespace when descending through the tree.
-	isSvgMode = vnodeName==='svg' ? true : vnodeName==='foreignObject' ? false : isSvgMode;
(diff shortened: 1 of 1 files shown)


developit:
Ah interesting. Might need to play with this a bit to golf the size down.

developit:
What about this?

```js
	if (vnodeName==='svg') isSvgMode = true;

	vnodeName = String(vnodeName);
	if (!dom || !isNamedNode(dom, vnodeName)) {
		out = createNode(vnodeName, isSvgMode);
		if (dom) {
			while (dom.firstChild) out.appendChild(dom.firstChild);
			if (dom.parentNode) dom.parentNode.replaceChild(out, dom);
			recollectNodeTree(dom, true);
		}
	}

	if (vnodeName==='foreignObject') isSvgMode = false;
```

marvinhagemeister:
Send me this was included in Preact X👍

PR #774, opened 2017-07-26
Don't add undefined children to props

webdeb:
When you use `preact-compat` or remove children from the vnode manually, children are still provided as an undefined key inside the props.

This produces a problem, when you use a library like `paypal-checkout` which is checking the props-keys against their limited props definitions.

bildschirmfoto 2017-07-27 um 18 02 12

> Uncaught Error: [paypal-button] Invalid prop: children
>     at Component.error (index.js:343)
>     at Object.validateProps [as c] (validate.js:88)
>     at ParentComponent.setProps (index.js:358)
>     at new ParentComponent (index.js:49)
>     at Component.init (index.js:355)
>     at _class2.componentDidMount (react.js:28)
>     at flushMounts (app.js:398)
>     at diff (app.js:427)
>     at render (app.js:1059)
>     at app.js:1368
> 

So, this PR will not add children to props, if children are `undefined` inside the `vnode`. And actually nothing will change, `props.children === undefined` is still `true`.

### DIFF
--- a/src/vdom/index.js
+++ b/src/vdom/index.js
@@ -35,7 +35,9 @@ export function isNamedNode(node, nodeName) {
  */
 export function getNodeProps(vnode) {
 	let props = extend({}, vnode.attributes);
-	props.children = vnode.children;
+	if (vnode.children!==undefined) {
+		props.children = vnode.children;
+	}
 
 	let defaultProps = vnode.nodeName.defaultProps;
 	if (defaultProps!==undefined) {


src/vdom/index.js
```diff
@@ -36,6 +36,7 @@ export function isNamedNode(node, nodeName) {
 export function getNodeProps(vnode) {
 	let props = extend({}, vnode.attributes);
 	props.children = vnode.children;
+	if (props.children === undefined) delete props.children;
```
marvinhagemeister: I think a better way would be to put the condition around the line above:

```js
if (vnode.children !== undefined) {
  props.children = vnode.children;
}
```

The `delete` keyword is known to cause a deopt in V8.

webdeb:
added a better PR description to demonstrate the problem

marvinhagemeister:
LGTM, can confirm that this fixes the issue. [paypal-checkout](https://github.com/paypal/paypal-checkout) indeed loop over `props` and throw on unknown keys. (as in the screenshot).

@developit This is can be merged.

package.json
```diff
@@ -120,13 +120,21 @@
   },
   "babel": {
     "presets": [
-      ["env", {
-        "loose": true,
-        "exclude": ["transform-es2015-typeof-symbol"],
-        "targets": {
-          "browsers": ["last 2 versions", "IE >= 9"]
+      [
```
marvinhagemeister: These whitespace changes should not be part of the PR.

webdeb: Oh it was an accident :), I've removed package.json from PR at all.

webdeb:
Thanks @marvinhagemeister!

developit:
What's the size difference to master here?

developit:
Looks like it adds 9 bytes. Not the worst but I'm looking into ways to make it more free.

webdeb:
@developit Thank you for your feedback. 

I've opened another optional PR, which extends the extend util and adds `-8 bytes` (fixes also that undefined children problem) https://github.com/developit/preact/pull/822

NekR:
> Looks like it adds 9 bytes. Not the worst but I'm looking into ways to make it more free.

@developit This is also a breaking change, no matter how light this looks. Anyone who is using `children.map(...)` this days without conditions would get a broken code. And, to be honest, I don't really like that this won't be possible anymore.

developit:
@NekR it does scare me a little to make children sometimes empty

webdeb:
> Anyone who is using children.map(...) this days without conditions would get a broken code..

This is wrong, If children is `undefined` then `children.map` would't work anyway, right? If you are not using `preact-compat` then you will not face the issue with `undefined` children at all, so nothing to worry about.

webdeb:
Anyway, I think this one is just a quick fix.. This PR #822 is actually better (performance & code size) ..so I am closing this PR, thanks for your attention guys

NekR:
I'm okay with the change as far as it doesn't affect normal `preact` itself.

marvinhagemeister:
Closing, because in our next version the `children` property will only be set when the component actually received any children. This was a positive side-effect when we completely rewrote `props` handling :tada:

PR #775, opened 2017-07-27
Fix flow type definition

iWader:
Currently the flow type definition is importing the `render` method from `react`, but it should be `react-dom`

```
Error: node_modules/preact/src/preact.js.flow:3
  3: import { createElement as h, cloneElement, Component, render } from 'react';
                                                           ^^^^^^ Named import from module `react`. This module has no named export called `render`.
```


developit:
Hi there - I had commented on #782, but I'll ask here too - do you think we could define render() ma happy here instead?  It's different from React's render since it allows a third argument.

developit:
I think the arguments have to have names (though I really don't know for sure):

```js
import { createElement as h, cloneElement, Component } from 'react';

declare function VNode(): any
declare var HTMLElement: any;
declare function render(vnode: VNode, parent: HTMLElement, replace: HTMLElement): HTMLElement;

export { h, cloneElement, Component, render };
export default { h, cloneElement, Component, render };

declare export function rerender(): void;
declare export var options: Object;
```

marvinhagemeister:
We removed the flow types from Preact with 10.x as nobody was interested in them anymore. Users we spoke to switched to TypeScript instead.

PR #782, opened 2017-08-01
Fix flowtype definitions

DimitryDushkin:


ooflorent:
@DimitryDushkin Could you also export `createElement` since `preact` exports it?

DimitryDushkin:
done

PR #789, opened 2017-08-03
Add option to disable component recycling

gpoitch:
Not sure if/when component recycling removal will land so I'm throwing this idea out there so users can choose to opt out.

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -11,6 +11,13 @@ export default {
 	 */
 	//syncComponentUpdates: true,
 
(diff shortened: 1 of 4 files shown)


developit:
Likely soon - the issue here would be that turning off recycling without adjusting the diff to reuse old DOM when swapping components will produce a less optimal diff.

marvinhagemeister:
Closing: We completely removed the recycler as it was the cause of a lot of hard to fix bugs 👍

PR #804, opened 2017-08-10
Move donation message to scripts

### LINKED ISSUE #803: Can't install when npm is not available
This fd1df721a374e1ed97694cc5e837a08f4dcf5855 actually breaks things. In environments where only `yarn` is present Preact can't be installed.

andrejsm:
This fixes the issue where installation would fail in environments where `npm` is not available.

Fixes #803


andrejsm:
Please do check if it works under Windows

developit:
Seems like it doesn't work under Win.

andrejsm:
Thanks for checking. I'll then rewrite it to `node` script. Will get to this within a couple of hours.

andrejsm:
Moved from Shell script to Nodejs script. Works on Windows as well.

developit:
Awesome - tested that out locally too and it works well.  Only thing I'd like to see (and I don't mind doing it myself post-merge) is instead of using `./scripts`, put the script in the existing `config/` dir.

andrejsm:
Moved to `config/`

developit:
Works great!

PR #806, opened 2017-08-11
nextState use new object

ustccjw:
just to make https://github.com/ustccjw/immutable-component/blob/master/src/index.js works with preact.(react do like this)


developit:
Looks good so far, need to check filesize.

ustccjw:
@developit hi  how about this PR ?

developit:
hey @ustccjw - sorry for the super slow lead time on this. I need to do some fine-grained perf testing on this PR (and a few others) before we can merge, since it change how mutation is used.  For this one I actually think there might be a slight improvement in perf!

andrewiggins:
Thanks for the PR! I believe this functionality is present in Preact X - setState produces a new state object every time. If you can, try out Preact X and let us know if it works. If not open an issue and we'll investigate :)

PR #819, opened 2017-08-20
Add support for componentDidCatch Component method

rpetrich:
Allow components to catch errors that occur during child components' constructors or lifecycle methods. For simplicity's sake does not pass a second "info" argument like React does. If no component catches and handles a thrown exception, DOM will be left in an inconsistent state as it is in react 8.2.1


developit:
Hiya! Thanks for all the work here - just reading up on your solution and forming some thoughts.

I had a question - was the existing `_parentComponent` property not working for you?  It's set here:
https://github.com/developit/preact/blob/master/src/vdom/component.js#L122

test/browser/lifecycle.js
```diff
@@ -455,6 +455,154 @@ describe('Lifecycle methods', () => {
 		});
 	});
 
+	describe('#componentDidCatch', () => {
+
+		it('should be called when child fails in constructor', () => {
+			class ErrorReceiverComponent extends Component {
+				componentDidCatch(error) {
+					this.setState({ "error": error });
```
paranoidjk: eslint error cause ci failed https://travis-ci.org/developit/preact/builds/266889109?utm_source=github_status&utm_medium=notification

rpetrich:
@developit `_parentComponent` wasn't set in the path where components are created inside `buildComponentFromVNode`. Perhaps I've misinterpreted the meaning of `_parentComponent` and should actually introduce a new property?

test/browser/lifecycle.js
rpetrich: @paranoidjk Thanks for the tip. Resolved.

developit:
Not sure - `_parentComponent` only indicates a compositional parent (where a component has rendered the current component at its root), not grandparents (where there are DOM elements between the two).

rpetrich:
Ah, that makes sense. Error handling should pass through non-component ancestors to be compatible with React.

Would you prefer this PR keep the adjusted behaviour of `_parentComponent`, or be amended to restore the previous behaviour of `_parentComponent` and implement a different approach of finding the deep ancestor components?

developit:
@rpetrich definitely need to use a second property - using the current one will break higher order components (the existence of that property is used to determine if a component is a HoC).

rpetrich:
@developit This is ready for an updated review. I took the liberty of naming the new property `_ancestorComponent` (minifying to `__a`)

developit:
@rpetrich perfect, that's exactly the name I would have chosen! I'll re-review this.

src/vdom/component.js
```diff
@@ -158,15 +175,8 @@ export function renderComponent(component, opts, mountAll, isChild) {
...
-		if (base && !isChild) {
-			let componentRef = component,
-				t = component;
-			while ((t=t._parentComponent)) {
-				(componentRef = t).base = base;
-			}
-			base._component = componentRef;
-			base._componentConstructor = componentRef.constructor;
-		}
+		base._component = component;
```
developit: Removing the conditional & crawl here seems like it would break component composition. When re-rendering a component that is a child of another component, it will update the `_component` property in the DOM do point to itself, which is then incorrect.

developit:
Would it be possible to split up some of the changes here? Right now this adds `componentDidCatch()`, but also changes some pretty fundamental things about the diff that seem like they would break real-world apps (not sure why the tests still pass, seems like we're either missing tests or I'm missing something haha).  I'm super curious about those other changes and want to figure out if we can merge them too, but they would represent a major version bump and a lot of scrutiny, whereas the `componentDidCatch` stuff could be merged right away.

src/vdom/component.js
adamsilverstein: I removed these changes in https://github.com/developit/preact/pull/886 and verified the tests still pass. Wasn't sure how/if I could amend this PR directly.

gaearon:
Here is the React error boundary test suite. It took us a long time to iron out the edge cases so it would be great to aim for at least partial parity.

https://github.com/facebook/react/blob/master/src/renderers/__tests__/ReactErrorBoundaries-test.js

thysultan:
@gaearon Would a recursive React error boundary continue to bubble up the tree or bail?

gaearon:
It stops at the closest boundary. If that boundary throws attempting to re-render, it bubbles up again, and the process repeats until we reach the root. If we have reached the root, and the error is still unhandled, we unmount the whole thing.

thysultan:
What happens to errors thrown in `componentWillUnmount `, do they also try to bubble/recover or are they passive?

gaearon:
Yes, errors in all lifecycle hooks are caught, including `componentWillUnmount`, and trigger the closest error boundary. However, there is some nuance there that is easy to miss.

Both `componentDidMount` and `componentDidUpdate` methods often set up subscriptions. Whenever we‘re unmounting due to an error, we need to at least *try* to execute `componentWillUnmount` for every component that has had (now or previously) its `componentDidMount` called. That is our best effort at preventing a memory leak (and dead subscriptions causing future errors).

However, since `componentWillUnmount` itself might throw, we need to isolate them from each other, and still execute them despite potential further errors.

Here is [an example](https://github.com/facebook/react/blob/master/src/renderers/__tests__/ReactErrorBoundaries-test.js#L1715-L1734) of the interplay between a broken `componentDidMount` in one component and a broken `componentWillUnmount` in another one. Hope this helps!

hassanbazzi:
@gaearon Hey Dan! I'm attempting to get this going again. Is there an updated link to that test suite?

gaearon:
Our latest thinking on this is here: https://github.com/acdlite/react/blob/b71fca18671e199158ad8e59281a6d871e742459/packages/react-dom/src/__tests__/ReactErrorBoundaries-test.js

developit:
Awesome - looks like we can pretty much use that as-is. Nice work @gaearon + team

gaearon:
Note the call order doesn't match 1:1 what's in the latest npm release, but that's the intended one that'll go into master soon. The one on npm has some minor differences caused by some funkiness we've been able to refactor away recently.

hassanbazzi:
@gaearon Thank you so so much! That's one awesome test suite haha. As Jason said, we'll be able to just drop that in. :)

taylor-cedar:
@hassanbazzi are you still working on this PR?

hellsan631:
+1 for this. It has been one of the biggest pains in switching to preact from react recently, is the lack of the ```componentDidCatch``` error handler.

peterjwest:
Does this have to be perfect to be merged? It sounds like there are some edge cases left to deal with, but this is already much better than the current behaviour - which silently fails and breaks the whole app.

src/vdom/component.js
```diff
@@ -179,7 +196,11 @@ export function renderComponent(component, opts, mountAll, isChild) {
 		// flushMounts();
 
 		if (component.componentDidUpdate) {
-			component.componentDidUpdate(previousProps, previousState, previousContext);
+			try {
```
adamsilverstein: try/catch is expensive, can we avoid doing this unless the component or an ancestor implements componentDidCatch?

hellsan631: I believe the check would be more expensive because you have to check every ancestor of the component, which means the complexity goes up as your tree gets deeper. Its really a question of O(n) vs O(1), and with modern browsers, try/catch isn't even that expensive.

https://jsperf.com/try-catch-jsperf

Created a JSPerf to check, and it seems like the checking for  the existence of the method on just the component is more expensive then a standard try-catch

adamsilverstein: 👍 thanks for checking

hassanbazzi:
Can you possibly add the test suite @gaearon mentioned to this and see if it passes? It may show us things we didn't account for. If we're adding this expecting people to switch seamlessly from React to Preact, then it should definitely pass completely.

rpetrich:
I've added support for additional cases uncovered by the upstream React test suite linked by @gaearon. Lifecycle event ordering doesn't match what's dictated, but this will require more changes than reasonably belong in this pull request. The tests are imported and failing in a separate [componentDidCatch-with-ReactErrorBoundary-tests](https://github.com/rpetrich/preact/tree/componentDidCatch-with-ReactErrorBoundary-tests) branch.

@developit, will need guidance as to how to proceed

_(and yes, I do realize how long I've let this pull request gather dust)_

adamsilverstein:
Note: some size and performance metrics on https://github.com/developit/preact/pull/886 which I closed in favor of this PR.

src/vdom/component.js
```diff
@@ -49,6 +49,18 @@ export function setComponentProps(component, props, opts, context, mountAll) {
 	if (component.__ref) component.__ref(component);
 }
 
+export function catchErrorInComponent(error, component) {
+	for (; component; component = component._ancestorComponent) {
+		if (component.componentDidCatch) {
+			try {
```
lukeed: Do we _need_ another try/catch here? I would think that if `componentDidCatch` exists, the error is passed to it & it ends there. If the method does not exist, it's uncaught so the component (tree) is unmounted.

src/render-queue.js
```diff
@@ -16,6 +16,12 @@ export function rerender() {
 	let p, list = items;
 	items = [];
 	while ( (p = list.pop()) ) {
-		if (p._dirty) renderComponent(p);
+		if (p._dirty) {
+			try {
```
lukeed: I'm worried about the try/catch for each cycle of the while loop. 

Looking at `renderComponent`, it [_already_ has its own try/catch](https://github.com/rpetrich/preact/blob/3a3d7cb1213bdd643e4a35323cc4478065e4e061/src/vdom/component.js#L157) block that we can & should reuse.

Instead, `catchErrorInComponent` should be passed to `renderComponent` somehow (/cc @developit) -- or, better, yet, we import the function into `vdom/component.js` & call it directly from the `catch` phase.

src/vdom/component.js
```diff
@@ -38,7 +38,7 @@ export function setComponentProps(component, props, opts, context, mountAll) {
 	component._disable = false;
 
 	if (opts!==NO_RENDER) {
-		if (opts===SYNC_RENDER || options.syncComponentUpdates!==false || !component.base) {
+		if ((opts&SYNC_RENDER) || options.syncComponentUpdates!==false || !component.base) {
```
lukeed: Is the change to `&` necessary? Not super familiar with all the incoming `opts`, but I do know that this is a deopt. Runtime engine has to infer that `1` or `0` are booleans, rather than just receive booleans directly. 

I think this might matter in performance given that this is the Component class & rendering is the default/assumed behavior.

src/vdom/component.js
```diff
@@ -105,40 +118,49 @@ export function renderComponent(component, opts, mountAll, isChild) {
 		let childComponent = rendered && rendered.nodeName,
 			toUnmount, base;
 
-		if (typeof childComponent==='function') {
-			// set up high order component link
+		try {
```
lukeed: This is the try/catch I was referring to in my other review comment.

We can/should call `catchErrorInComponent` from within here directly to avoid the extra layer(s) of try/catch blocks.

src/vdom/component.js
```diff
@@ -105,40 +118,49 @@ export function renderComponent(component, opts, mountAll, isChild) {
...
+				toUnmount = initialChildComponent;
+				if (toUnmount) {
+					cbase = component._component = null;
+				}
 
-			// destroy high order component link
-			toUnmount = initialChildComponent;
-			if (toUnmount) {
-				cbase = component._component = null;
+				if (initialBase || (opts&SYNC_RENDER)) {
```
lukeed: Same comment before about `&` implicit coercion.

src/vdom/component.js
```diff
@@ -105,40 +118,49 @@ export function renderComponent(component, opts, mountAll, isChild) {
...
-			if (initialBase || opts===SYNC_RENDER) {
-				if (cbase) cbase._component = null;
-				base = diff(cbase, rendered, context, mountAll || !isUpdate, initialBase && initialBase.parentNode, true);
+		} catch (e) {
+			if (opts == REATTEMPT_RENDER) {
+				throw e;
 			}
+			base = initialBase || document.createTextNode("");
+			exception = e;
+			hasException = true;
```
lukeed: The `catchErrorInComponent` would live here (directly). If that function doesn't / can't unmount the tree directly (not sure), then `hasException` can be renamed to `toDismount` & trigger the flush + re-render below.

src/render-queue.js
ryan-petrich-mpf: Errors in render/child components should finish mounting/updating the component before dispatching `componentDidCatch` and potentially re-rendering. This is the try/catch around render/child components inside `renderComponent` and won't catch errors in the components mounting/update methods. Maybe a single try/catch inside `renderComponent` and booleans representing how much progress into the method has completed is a better approach? (since the error needs to be handled differently depending on where it comes from)

rpetrich: Errors in render/child components should finish mounting/updating the component before dispatching `componentDidCatch` and potentially re-rendering. This is the try/catch around render/child components inside `renderComponent` and won't catch errors in the components mounting/update methods. Maybe a single try/catch inside `renderComponent` and booleans representing how much progress into the method has completed is a better approach? (since the error needs to be handled differently depending on where it comes from)

src/vdom/component.js
rpetrich: I've added a new `REATTEMPT_RENDER` opt. It behaves like `SYNC_RENDER`, but rethrows any errors that occur in child components. A reattempt path is required to cause any errors that occur in child components of an error handling path to bubble up to ancestor components instead of triggering an infinite rerender loop.

To clarify, imagine this scenario: Component renders a tree of child components. This child tree throws an error. Component catches the error and triggers a rerenders with a separate tree of child components describing the error. This tree also throws an error (yikes!). If we redispatched to the component's `componentDidCatch` it would likely trigger a rerender that generates the same erroring render tree. Instead preact should now treat that second error as having come from the component itself and pass it up to an ancestor so it can be handled at a higher level.

We can add a comparison with 0 or compare against both constants explicitly to remove the deopt path (at the expense of some additional bytes in either approach)

src/vdom/component.js
rpetrich: Errors that occur in `componentDidCatch` should bubble to ancestors, regardless of what lifecycle method generated the error or from what path that lifecycle method occurred. Error paths _should_ be rare, so it's fine for it to be slow.

src/render-queue.js
lukeed: Got it 👍 What would the boolean approach look like? Would it be a single key assignment on every component, every render?

rpetrich: Boolean approach evolved into a nested try/catch with some additional state tracking on the component itself (see: the new `_caught` property) to properly handle cases where a component fails repeatedly and should bubble its errors to an ancestor.

I also merged the latest master and added tests for `getDerivedStateFromProps` and `getSnapshotBeforeUpdate` to ensure they're properly integrated with `componentDidCatch`

rpetrich:
Here is an updated run of the performance tests @andrewiggins posted in #886: https://gist.github.com/rpetrich/19e259bfd7fc66c10ce5b339cf895283

src/vdom/component.js
```diff
@@ -266,11 +304,18 @@ export function buildComponentFromVNode(dom, vnode, context, mountAll) {
 export function unmountComponent(component) {
 	if (options.beforeUnmount) options.beforeUnmount(component);
 
-	let base = component.base;
-
+	if (component._disable) return;
```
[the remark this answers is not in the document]
rpetrich: Definitely an oversight that the check against `_disable` wasn't the first operation in the function.

src/vdom/diff.js
```diff
@@ -39,7 +45,7 @@ export function flushMounts() {
  * @param {object} context The current context
  * @param {boolean} mountAll Whether or not to immediately mount all components
  * @param {Element} parent ?
- * @param {boolean} componentRoot ?
+ * @param {import('../component').Component} componentRoot ?
```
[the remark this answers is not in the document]
rpetrich: This parameter previously was a boolean that represented whether or not the node being diffed was rooted inside a component. I've repurposed it to instead pass the nearest ancestor component instance instead, if any and the truthiness is thus preserved.

I'd like to follow up by renaming this in a later PR and avoid having yet more changes packed into this one.

test/browser/lifecycle.js
```diff
@@ -1,6 +1,16 @@
-import { h, render, rerender, Component } from '../../src/preact';
+import { h, render, options, Component } from '../../src/preact';
 /** @jsx h */
 
+// Wrapper rerender function that fully drains the queue
+const renderQueue = [];
+options.debounceRendering = (callback) => renderQueue.push(callback);
```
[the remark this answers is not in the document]
rpetrich: For the current render queue implementation in preact any render operations enqueued while the render queue is being drained will be dispatched in a separate microtask.

Before verifying the final state of the DOM the tests need all of the render operations (regardless of how/when they're performed) to be processed. It seemed cleaner to build a simple render queue than to attempt to guess at how many of the standard `rerender` calls would have been necessary.

It may be more correct for preact to flush all render events in a single microtask. This would be a simple adjustment, but I think such a change deserves separate discussion.

src/vdom/component.js
```diff
@@ -79,129 +96,150 @@ export function renderComponent(component, renderMode, mountAll, isChild) {
...
-		}
-		component.props = props;
-		component.state = state;
-		component.context = context;
-	}
-
-	component.prevProps = component.prevState = component.prevContext = component.nextBase = null;
-	component._dirty = false;
+		rendered, inst, cbase,
+		exception, clearCaught = component._caught;
```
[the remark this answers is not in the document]
rpetrich: The `_caught` property and `clearCaught` variable are to coordinate when a component rerenders in response to an error from one of its children, but the new children it returns in response to the first error also throw an error. Without handling this case specifically, the component would receive componentDidCatch again and continue to ask for the same failing child node tree to be rendered. Instead, the error should bubble further up the chain to be handled by another component.

The "should bubble on repeated errors" and "should bubble on ignored errors" tests cover these cases. Repeated errors is the normal case to handle, as it isn't possible for an error boundary component to anticipate every error that could occur in fallback child tree. Components that implement `componentDidCatch` but fail to even ask to be re-rendered upon catching an error are completely derelict in their duty, but it's useful to have a test for them anyway.

test/browser/lifecycle.js
```diff
@@ -866,6 +876,405 @@ describe('Lifecycle methods', () => {
...
+				componentDidCatch(error) {
+					this.setState({ error });
+				}
+				render() {
+					return <div>{this.state.error ? String(this.state.error) : this.props.children}</div>;
+				}
+			}
+			class ErrorAdapterComponent extends Component {
+				componentDidCatch() {
+					// Ignore the error
```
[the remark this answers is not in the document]
rpetrich: Generally call `setState` and render a new node tree that is different from the previous one that yielded an error. Commonly this will include stashing the error in the state so that a description of the tree can be rendered.

test/browser/lifecycle.js
```diff
@@ -866,6 +876,405 @@ describe('Lifecycle methods', () => {
...
+			sinon.spy(ErrorAdapterComponent.prototype, 'componentDidCatch');
+			sinon.spy(ErrorReceiverComponent.prototype, 'componentDidCatch');
+			render(<ErrorReceiverComponent><ErrorAdapterComponent><ErrorGeneratorComponent/></ErrorAdapterComponent></ErrorReceiverComponent>, scratch);
+			rerender();
+			expect(ErrorAdapterComponent.prototype.componentDidCatch).to.have.been.called;
+			expect(ErrorReceiverComponent.prototype.componentDidCatch).to.have.been.called;
+			expect(scratch).to.have.property('textContent', 'Error: Error!');
+		});
+
+		it('should bubble on ignored errors', () => {
```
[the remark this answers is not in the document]
rpetrich: Added.

rpetrich:
@andrewiggins I could use some tips on what to do with the ReactErrorBoundary tests. They're integrated on a separate branch, but failing. Parts that are failing:
1. React will unmount the entire tree if an error is not handled by any component. This PR preserves the existing preact behaviour of leaving a partially rendered tree in the DOM when an error is unhandled.
2. Render order differs between preact and react for certain operations and it would be a substantial effort to reform this.

hassanbazzi:
I wonder if you can maybe just run the test suite against preact-compat. I think compat has the order changes you’re talking about within it.

hassanbazzi:
Oh and did I say? Amazing work on this PR!

test/browser/lifecycle.js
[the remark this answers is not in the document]
andrewiggins: Thanks! That was very helpful

src/vdom/component.js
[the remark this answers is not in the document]
andrewiggins: Thanks for taking the time to explain this - it helps!

andrewiggins:
@rpetrich Just took a look at the branch and I see what you mean. Let's figure out what to do about those test in a separate PR/issue.

sampotts:
Not to be a pain with the usual "+1" but what's the status on this? It'd be awesome to be able to use this for error logging 👍

rpetrich:
@sampotts Currently it's stable and usable by referencing the branch as an npm dependency, but the PR is somewhat stalled (which is understandable considering that it's a substantial adjustment, increases the API surface area, and has performance implications)

andrewiggins:
@lukeed Is there a way to unset your review to unblock this PR? Or if you have time, could you take another look and leave more comments or approve?

studentIvan:
What is the state of this issue? is it in the 8.3.0 release?

studentIvan:
@rpetrich just tested your current branch code, seems it throws

```
preact.js?10a9:324 Uncaught (in promise) TypeError: Failed to execute 'replaceChild' on 'Node': parameter 1 is not of type 'Node'.
    at renderComponent (preact.js?10a9:324)
    at rerender (preact.js?10a9:37)
```

at this line `baseParent.replaceChild(base, initialBase);`

on error, but componentDidCatch works

rpetrich:
@studentIvan Do you have a test case I could examine?

studentIvan:
@rpetrich sorry it's on my project, so only if we can skype/etc and I will share my screen to you.
what I have done is simple added componentDidCatch method to root App component with latest preact code base and throw something while render.
By the way - if (void 0 === base) return; saves the situation, can you tell me what will broke with this code? If nothing can we just add this simple line into it?)

studentIvan:
@rpetrich https://github.com/studentIvan/preact/commit/d15a82b48cc615b57683789dae1f598121f12148

rpetrich:
@studentIvan `base` should not be undefined at that point and I'm wary to add a check there if it's only hiding problems elsewhere. Is it possible to extract a smaller test case from your code? Alternatively, feel free to reach out to me on the Preact Slack team—I have the same handle.

photz:
Until this is merged, is there some workaround? I need to be able to catch exceptions in my root component and respond to them.

studentIvan:
> Until this is merged, is there some workaround? I need to be able to catch exceptions in my root component and respond to them.

I use preact from studentIvan/preact#componentDidCatch-support before merge and base errors resolution

roberttod:
I am also having to use studentIvan/preact#componentDidCatch-support. What needs to be done to merge this?

johnhaitas:
What about support for `static getDerivedStateFromError(error)`? https://reactjs.org/docs/react-component.html#static-getderivedstatefromerror

sampotts:
Will this get merged in 2019 do we think?

marvinhagemeister:
@sampotts `componentDidCatch` will ship with Preact X. We need the PR a while back there. Our current these estimate is early March and we'll likely merge the 2 repos before that 🎉

sampotts:
Awesome. Can't wait 👍

huhuanming:
Awesome!

pierpo:
Fantastic! 🎉
Any news about the "early March" estimate? 😊

marvinhagemeister:
@pierpo An alpha release is scheduled for March 4th✌️ See: https://twitter.com/marvinhagemeist/status/1097973028426788864

pierpo:
@marvinhagemeister Amazing! 😍 
Thank you for the very quick answer!

marvinhagemeister:
The alpha release is out :tada:

PR #821, opened 2017-08-20
Add rugby board as an example

crispgm:
Rugby News Board is my side project, which is fully powered by preact & preact-router. I am not sure whether is it a good example. Thanks.


crispgm:
😸

PR #822, opened 2017-08-20
No undef childs with extended extend, POC

webdeb:
@developit This PR is just another approach to deal with the undefined props (children) problem.
It extends the extend util to reuse it when assigning default props, because the main logic of the loop is the same, except we do not allow to override defined `obj` props.

### DIFF
--- a/src/vdom/index.js
+++ b/src/vdom/index.js
@@ -34,17 +34,10 @@ export function isNamedNode(node, nodeName) {
  * @returns {Object} props
  */
 export function getNodeProps(vnode) {
-	let props = extend({}, vnode.attributes);
-	props.children = vnode.children;
+	let props = vnode.children!==undefined ? { children: vnode.children } : {};
+	props = extend(props, vnode.attributes);
 
-	let defaultProps = vnode.nodeName.defaultProps;
-	if (defaultProps!==undefined) {
-		for (let i in defaultProps) {
-			if (props[i]===undefined) {
-				props[i] = defaultProps[i];
-			}
-		}
-	}
-
-	return props;
+	return vnode.nodeName.defaultProps!==undefined
+		? extend(props, vnode.nodeName.defaultProps, true)
+		: props;
 }
--- a/src/vdom/index.js
+++ b/src/vdom/index.js
@@ -34,10 +34,12 @@ export function isNamedNode(node, nodeName) {
  * @returns {Object} props
  */
 export function getNodeProps(vnode) {
-	let props = vnode.children!==undefined ? { children: vnode.children } : {};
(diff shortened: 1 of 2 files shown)


src/vdom/index.js
```diff
@@ -34,17 +34,10 @@ export function isNamedNode(node, nodeName) {
...
-		for (let i in defaultProps) {
-			if (props[i]===undefined) {
-				props[i] = defaultProps[i];
-			}
-		}
-	}
-
-	return props;
+	return vnode.nodeName.defaultProps!==undefined
+		? extend(props, vnode.nodeName.defaultProps, true)
```
webdeb: the last param => `true`, means do not override defined props. So we can remove the above loop, and reuse `extend`

src/vdom/index.js
```diff
@@ -34,17 +34,10 @@ export function isNamedNode(node, nodeName) {
  * @returns {Object} props
  */
 export function getNodeProps(vnode) {
-	let props = extend({}, vnode.attributes);
-	props.children = vnode.children;
+	let props = vnode.children!==undefined ? { children: vnode.children } : {};
```
webdeb: instead of checking afterwards we can perform the check for the children before, and create the initial props either `{}` or `{children: vnode.children}`

The next line just takes previously created props as the first argument

webdeb:
@developit this one adds only 7bytes if you compare the gz versions

leeoniya:
`void 0` is shorter than `undefined`. may add up to a bit if used across the codebase. may even help to alias it. `const UNDEF = void 0;`

webdeb:
my last commit made it 8 bytes less than the master ;)

webdeb:
> void 0 is shorter than undefined. may add up to a bit if used across the codebase. may even help to alias it: const UNDEF = void 0;

maybe, could you try it on master and check agains the master its gzipped *3458 bytes* at the moment

developit:
The babel setup in this repo converts all `undefined` to `void 0` (see [output here](https://unpkg.com/preact@8.2.1/dist/preact.min.js)).

webdeb:
@developit Thanks for quick feedback. You mean skipping `undefined` props? Hmm `extend` is also used in `setState`, and you may want to set some state explicitly to `undefined`, so this wouldn't work, if I understood correctly.

Regarding performance, what are you using to measure?

webdeb:
In the meanwhile I made some tests with the built-in benchmark tool. The results are actually quite  same for preact & preact(-forked with this PR).. Sometimes the master is faster, sometimes the fork.. really, no remarkable difference...

marvinhagemeister:
This is fixed in Preact X :tada:

Issue #820, opened 2017-08-20
Error running npm install in preact dependency.


jamieowen:
Hi,

I'm used to using React but decided to give preact a go, along with the preact material ui components. It seems awesome so far...

When I installed both preact along with the jsx babel transform, things went ok.
When I installed preact-material-components as well, and imported a single component directly: 

```
import Button from 'preact-material-components/Button';
```
preact complained about the lack of the transform-object-rest-spread transform.

I could install this directly in my folder but i'd assume that preact should handle those deps. And declare them in the package.json browserify transform options.

I had to do an npm install within the node_modules/preact folder - but this then failed as at the end of the install the preact build command fails ( as there was a lack of config folder ) as its not included in the default npm install.

So i had to clone the preact repo directly in my node_modules and run install to get things working.
Not ideal - but just thought i would ask if this was all expected behaviour! :)

Cheers,

Jamie

developit:
Hi there - that sounds pretty tedious! I figured out what caused the issue - by default, Babel actually recursively inspects `package.json` files for a `babel` configuration option (similar to its handling of `.babelrc).  Preact uses its package.json to store babel configuration, which means your setup is actually trying to apply those plugins to preact's `dist`.  I think we can simply add a prerelease script that strips [unnecessary keys](https://github.com/developit/preact/blob/master/package.json#L121-L151) from the package.json.

rubenmoya:
I've been doing some research about stripping keys from the package.json before releasing, but I had no luck at all, haven't seen any examples or anyone doing it...

How would you approach this?

developit:
I think something like this would work:
https://gist.github.com/developit/11bf91c3e4240ea3b615ac80684351d0

just as a prepublish script:

```js
  "prepublish": "node config/prepublish.js",
```

rubenmoya:
The only thing I don't like is the `package.json` is still modified after the code has been published, maybe doing something like `"prepublish": "node config/prepublish.js && git checkout package.json",`? what do you think?

developit:
Hmm - that is funky yeah. The git checkout freaks me out a bit - maybe `cp` it to a dotfile and then back?

```js
  "prepublishOnly": "cp package.json .package.json; node prepublish.js",
  "postpublish": "mv -f .package.json package.json"
```

PR #825, opened 2017-08-22
Add prepublishOnly script

rubenmoya:
Related to: #820

This PR adds a `prepublishOnly` script which removes unnecessary keys from the `package.json`.

Feedback is welcome :)

### DIFF
--- a/config/prepublish.js
+++ b/config/prepublish.js
@@ -1,0 +1,20 @@
+const path = require('path').join(__dirname, '..', 'package.json');
+const pkg = require(path);
+
+const UNNECESSARY_KEYS = [
+	'babel',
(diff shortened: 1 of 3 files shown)


config/prepublish.js
```diff
@@ -0,0 +1,20 @@
...
+
+const UNNECESSARY_KEYS = [
+	'babel',
+	'greenkeeper',
+	'bundlesize',
+	'devDependencies',
+	'eslintConfig'
+];
+
+UNNECESSARY_KEYS.forEach(key => delete pkg[key]);
```
developit: probably best to keep this as a normal function to support Node 6

PR #827, opened 2017-08-22
#666 support css variables

### LINKED ISSUE #666: Add support for CSS Custom Properties
Cross-references with React:
- PR adding support for CSS Custom Props: https://github.com/facebook/react/pull/9302
- Issue closed by previous PR: https://github.com/facebook/react/issues/6411

React will soon be able to support CSS Custom Properties (CSS Variables) by using `.style.setProperty(<prop>, <value>)` instead of directly modifying `.style.<prop> = <value>` for custom properties (detected by seeing if the property starts with two dashes, e.g., `--foo-bar`).

It seems that, for objects passed into the `style` attribute of Preact components, this is currently not the case, as it is still directly setting the style properties: https://github.com/developit/preact/blob/9e1dd185bb1d6505c35477d8101c886c25dc51c8/src/dom/index.js#L57

Furthermore, the use of `.style.setProperty(prop, value)` seems to be _much_ faster in most environments (except for IE9-11, where it is slightly slower): https://jsperf.com/dom-style-vs-setproperty - shouldn't matter anyway if you choose to use `.setProperty` for only custom properties.

The end-goal is to have this work:

```jsx
<div style={{ '--foo': 'green', color: 'var(--foo, red)' }}> 
  ...
</div>
```

What do you think? 😄 It's a fairly minor change and I can make a PR if needed.

tkh44:
This is most likely unacceptable as far as size goes, but I wanted to get the process started.

Note that PhantomJS does not support cssText with custom properties in it. No idea what to do about this other than to drop css vars from that test.

fixes #666


src/dom/index.js
```diff
@@ -52,10 +52,21 @@ export function setAccessor(node, name, old, value, isSvg) {
 		}
 		if (value && typeof value==='object') {
 			if (typeof old!=='string') {
-				for (let i in old) if (!(i in value)) node.style[i] = '';
+				for (let i in old) if (!(i in value)) {
+					if (i.indexOf('--') === 0) {
```
developit: I think we can always use `setProperty()` / `removeProperty()` here, which saves a bunch of bytes.  AFAIK `style.setProperty('color', 'red')` is equivalent to `style.color='red'`.

developit: just re-read the original thread, i was wrong since setProperty() uses CSS case rather than camelCase. poop!

developit:
Awesome. 

Regarding the testing, it's probably a good time to switch the default launcher to be Headless Chrome, I don't really use Phantom for Preact anymore since the performance numbers are unreliable.

developit:
the bundle size thing didn't kick in, but it looks like this goes from `3444` -> `3482` (+38b).  Honestly that's not too bad, and I really want CSS variables support.

tkh44:
> do you think style.setProperty('--foo', '') would work like style.removeProperty('--foo')?

It works fine. According to the spec this should work `style.setProperty('--foo')`, but it errors in chrome. `style.setProperty('--foo', '')` works fine though.

src/dom/index.js
```diff
@@ -58,10 +58,21 @@ export function setAccessor(node, name, old, value, isSvg) {
...
+					}
+				}
 			}
 			for (let i in value) {
-				node.style[i] = typeof value[i]==='number' && IS_NON_DIMENSIONAL.test(i)===false ? (value[i]+'px') : value[i];
+				if (~i.indexOf('-')) {
+					node.style.setProperty(i, value[i]);
+				}
+				else {
+					node.style[i] = typeof value[i]==='number' && IS_NON_DIMENSIONAL.test(i)===false ? (value[i]+'px') : value[i];
```
smlmrkhlms: Could you turn `IS_NON_DIMENSIONAL.test(i)===false` into `!IS_NON_DIMENSIONAL.test(i)`

JeremyVSR:
What about using the ternary operator ?
```
	 ~i.indexOf('-')
		 ? node.style.setProperty(i, value[i]) 
		 : node.style[i] = typeof value[i]==='number' && IS_NON_DIMENSIONAL.test(i)===false 
			? (value[i]+'px') 
			: value[i];
```

src/dom/index.js
```diff
@@ -58,10 +58,21 @@ export function setAccessor(node, name, old, value, isSvg) {
 		}
 		if (value && typeof value==='object') {
 			if (typeof old!=='string') {
-				for (let i in old) if (!(i in value)) node.style[i] = '';
+				for (let i in old) if (!(i in value)) {
```
lukeed: Challenge: https://twitter.com/preactjs/status/998981752130588673

```js
  let i; // placeholder
  if (typeof old!=='string') {
    for (i in old) {
      if (!(i in value)) node.style.setProperty(i, '');
    }
  }
  for (i in value) {
    node.style.setProperty(i, typeof value[i]==='number' && IS_NON_DIMENSIONAL.test(i)===false ? (value[i]+'px') : value[i]);
  }
}
```

src/dom/index.js
lukeed: Damn GitHub collapsing comments. Missed this

src/dom/index.js
PepsRyuu: This would be backwards incompatible. Doesn't work for properties that are defined using camel-casing.

```javascript
node.style.backgroundColor = value; // works
node.style['background-color'] = value; // works
node.style.setProperty('background-color', value); // works
node.style.setProperty('backgroundColor', value); // doesn't work
```

lukeed: @PepsRyuu aye, saw this after 🙈 /delete

andrewiggins:
@tkh44 Do we need the `if` checks at all? I removed the if statements (shown below) and the PR size was reduced by 17 bytes, so its size is now only +9 bytes on top of master. All tests pass. I quickly checked in a handful in browsers and I think this should work. It appears setProperty doesn't throw an error on invalid property names (which will happen if a component defines something like `backgroundColor` like you've pointed out).

```js
	if (typeof old!=='string') {
		for (let i in old) if (!(i in value)) {
			node.style.setProperty(i, '');
			node.style[i] = '';
		}
	}
	for (let i in value) {
		node.style.setProperty(i, value[i]);
		node.style[i] = typeof value[i]==='number' && IS_NON_DIMENSIONAL.test(i)===false ? (value[i]+'px') : value[i];
	}
}
```

tkh44:
@andrewiggins that should work. There was a bug in the older code that I just realized.

```
typeof value[i] ===
				'number' &&
				IS_NON_DIMENSIONAL.test(i) === false ? (value[i] + 'px') : value[i])
```
needs to be in both the setProperty and the assignment. 

abstracting it would look like

```
function a(b) { typeof b ===
				'number' &&
				IS_NON_DIMENSIONAL.test(i) === false ? (b + 'px') : b)
```

Not sure where this puts us in terms of bytes.

src/dom/index.js
```diff
@@ -83,10 +83,21 @@ export function setAccessor(node, name, old, value, isSvg) {
...
+					if (~i.indexOf('-')) {
```
developit: We should check what the size is when doing `if (i[0]=='-')` - we only care about using setProperty for custom properties, so it'll always be a leading `--`.

marvinhagemeister:
Closing, we finally have support for css variables in Preact X 🎉💯

PR #828, opened 2017-08-23
Move babel settings to a .babelrc

### LINKED ISSUE #820: Error running npm install in preact dependency.

rmacklin:
Fixes #820 

Since the .babelrc doesn't match the package's `files` list, it won't be included in the published package. This means that users who don't exclude node_modules from transpilation will be able to use preact without it requiring them to have preact's internal babel presets and plugins installed (which isn't necessary if they're depending on the pre-transpiled version in `dist`)

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,23 @@
+{
+  "presets": [
+    [
+      "env",
+      {
+        "loose": true,
+        "exclude": [
+          "transform-es2015-typeof-symbol"
+        ],
+        "targets": {
+          "browsers": [
+            "last 2 versions",
+            "IE >= 9"
+          ]
+        }
+      }
+    ]
+  ],
+  "plugins": [
+    "transform-object-rest-spread",
+    "transform-react-jsx"
+  ]
+}
--- a/package.json
+++ b/package.json
@@ -118,29 +118,6 @@
     "uglify-js": "^2.7.5",
     "webpack": "^2.4.1"
   },
-  "babel": {
-    "presets": [
-      [
-        "env",
-        {
-          "loose": true,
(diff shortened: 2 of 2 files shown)


rmacklin:
This is an alternative solution to #820 from the existing PR, #825. The benefit of this approach is that it's simpler in that we don't need to maintain a custom `prepublishOnly` script. A drawback is that it adds another file to the package root, which @developit doesn't like. I'm opening the PR so @developit can make the call for which approach to use. Feel free to close this if you prefer the `prepublishOnly` approach - as long as #820 is fixed I'll be happy.

developit:
@rmacklin I'm super torn. Thinking maybe we can try one release with #825 since it also removes a few other unnecessary properties, but if that causes issues I think we should revert and go with the .babelrc.  Sound reasonable?

rmacklin:
Sure, sounds reasonable. Thanks!

PR #835, opened 2017-08-24
preact pass through promised components to allow async rendering

3axap4eHko:
related to the https://github.com/developit/preact-render-to-string/pull/41


developit:
Hmm - this doesn't work for non-static children though.

3axap4eHko:
@developit it handle case when child is a promise so in the usual situation there are no promise components and it does not affect any existing cases

developit:
Preact doesn't support Promises or async components at all. If it did, we'd need to modify the entire render pipeline to do so.

3axap4eHko:
And this is good point to be better than react. With my 2 PR my projects work like a charm with async components, I just use async action with my component. For eaxample:
``` javascript
...
import { myAction } from '../redux/actions';

@withService({
    service: props => myAction(props.match.slug),
    mapToProps: post => ({ post }),
})
export default class MyComponent extends Component {
    render() {
        return <article>{this.props.post.content}</article>;
   }
}
```
this approach works the same way with client side rendering and SSR without any voodoo magic

3axap4eHko:
@developit could you take a look on this example https://github.com/3axap4eHko/preact-async-example please

PR #842, opened 2017-08-28
Add support for create-react-class Components

aloukianova:
This is to handle create-react-class components such as rc-tooltip, which don't have constructors or component.base defined, causing the page to break when loading dev tools.

### DIFF
--- a/devtools/devtools.js
+++ b/devtools/devtools.js
@@ -83,7 +83,7 @@ function typeName(element) {
  */
 function createReactCompositeComponent(component) {
 	const _currentElement = createReactElement(component);
-	const node = component.base;
+	const node = component.base || component.rootNode;
 
 	let instance = {
 		// --- ReactDOMComponent properties


developit:
Hmm - `create-react-class` doesn't work out-of-the-box with Preact (and can't). It should be used with `preact-compat`, which adds `.base` and the other requisite properties.

loukitzanna:
We are using preact-compat with the webpack alias specifically for create-react-class

developit:
If you're getting components without a `.base` property, you're ending up with `React` classes. That means something's off in your config.  It sounds like you're using the original `create-react-class` module - you'll want to instead use the aliased preact-compat version:

```js
alias: {
  'react': 'preact-compat',
  'react-dom': 'preact-compat',
  'create-react-class': 'preact-compat/lib/create-react-class'
}
```

loukitzanna:
We're already doing this, but it doesn't seem to work with react-component for some reason. I can't think of any other configs that would make it completely ignore preact though... However, this is probably not a problem with devtools and I'll look into fixing it on the tooltips instead

paranoidjk:
@aloukianova Hi, i am the owner of rc-tooltip, can you give a reproducible demo?  I'd like to help to figure out what's going wrong.

developit:
Maybe create-react-class is being inlined into rc-tooltip?

developit:
I can confirm - `rc-tooltip` is shipping an embedded copy of `create-react-class`, which it shouldn't be doing.  You can see it here - there's no import for it, and the source is embedded in the bundle:
https://unpkg.com/rc-tooltip@3.4.9/dist/rc-tooltip.js

paranoidjk:
@developit Actually, It's because that rc-tooltip relyed on rc-trigger, which used create-react-class: https://github.com/react-component/trigger/blob/master/package.json#L60

But i think if users do not use rc-tooltip umd bundle, but require or import it from npm, with webpack alias setting in preact-compat doc, it should works fine? https://github.com/react-component/trigger/blob/master/package.json#L26

PR #843, opened 2017-08-28
extend() uses Object.assign() 

stephenjfox:
Instead of manually doing the loop, use the widely support Object.assign
Ref: http://kangax.github.io/compat-table/es6/#test-Object_static_methods_Object.assign


developit:
@stephenjfox We're using a custom implementation because it performs better.

PR #861, opened 2017-09-08
Try: Fix preact diff algorithm causing the constantly unmouting/remounting bug

### LINKED ISSUE #857: Preact is constantly remounting subcomponents
While trying to migrate from React to preact, we noticed some mounting/rerendering issues;
Not sure if you consider this as a bug or not, but compared to react there's a difference in the way preact decides which component got dropped and which one got reordered when two siblings have the same type but not the same children.

Code examples are clearer:

 - react: https://codesandbox.io/s/422o58o3j4
 - preact: https://codesandbox.io/s/j9r725q2w
 - preact with workaround: https://codesandbox.io/s/6z431l219n

Notice that the subcomponent in the preact version is constantly remounting, if we hide/show the previous `<div>` while it's mounted only once in React. I found a workaround by setting a `key` prop to identify the `<div>`s.

youknowriad:
fixes #857

This PR includes the following changes to preact:

 - When diffing children, assign keys to all the children automatically if no key is provided (based on the index of the element in the children array)
 - To make it work properly, I needed to drop the "merge string siblings" from the `h` implementation, because this was causing nodes to have different keys depending on whether the previous node is a string or not (Not sure, why this merge was being done, I need context here)
 - with these changes, old elements (non preact elements) are being dropped when rerendering (Same here, I lack context on why this was implemented in the first place).

I'm new to preact, please excuse my lack of context :)

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -7,6 +7,7 @@ export const ASYNC_RENDER = 3;
 
 
 export const ATTR_KEY = '__preactattr_';
+export const ELT_KEY_PREFIX = 'preact-';
 
 // DOM properties that should NOT have "px" added when numeric
 export const IS_NON_DIMENSIONAL = /acit|ex(?:s|g|n|p|$)|rph|ows|mnc|ntw|ine[ch]|zoo|^ord/i;
--- a/src/vdom/diff.js
+++ b/src/vdom/diff.js
@@ -1,5 +1,5 @@
-import { ATTR_KEY } from '../constants';
-import { isSameNodeType, isNamedNode } from './index';
+import { ATTR_KEY, ELT_KEY_PREFIX } from '../constants';
+import { isNamedNode } from './index';
 import { buildComponentFromVNode } from './component';
 import { createNode, setAccessor } from '../dom/index';
 import { unmountComponent } from './component';
@@ -140,7 +140,7 @@ function idiff(dom, vnode, context, mountAll, componentRoot) {
 	}
 	// otherwise, if there are existing or new children, diff them:
 	else if (vchildren && vchildren.length || fc!=null) {
-		innerDiffNode(out, vchildren, context, mountAll, hydrating || props.dangerouslySetInnerHTML!=null);
+		innerDiffNode(out, vchildren, context, mountAll);
 	}
 
 
@@ -160,18 +160,14 @@ function idiff(dom, vnode, context, mountAll, componentRoot) {
  *	@param {Array} vchildren		Array of VNodes to compare to `dom.childNodes`
  *	@param {Object} context			Implicitly descendant context object (from most recent `getChildContext()`)
  *	@param {Boolean} mountAll
- *	@param {Boolean} isHydrating	If `true`, consumes externally created elements similar to hydration
  */
-function innerDiffNode(dom, vchildren, context, mountAll, isHydrating) {
+function innerDiffNode(dom, vchildren, context, mountAll) {
 	let originalChildren = dom.childNodes,
-		children = [],
 		keyed = {},
 		keyedLen = 0,
-		min = 0,
 		len = originalChildren.length,
-		childrenLen = 0,
 		vlen = vchildren ? vchildren.length : 0,
-		j, c, f, vchild, child;
+		f, vchild, child;
 
 	// Build up a map of keyed children and an Array of unkeyed children:
 	if (len!==0) {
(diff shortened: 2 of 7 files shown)


developit:
Hiya - it looks like this breaks type-based diffing, which is a really important part of how Preact works.

For the adjacent string merging - the reason that is done is because, for both SSR and some client contexts, adjacent Text nodes are merged together in the DOM.  By merging ahead-of-time, Preact matches what the DOM does.

youknowriad:
>  it looks like this breaks type-based diffing, which is a really important part of how Preact works.

Can I ask why it's important? This was the main issue causing the unmouting/mounting IMO

> For the adjacent string merging - the reason that is done is because, for both SSR and some client contexts, adjacent Text nodes are merged together in the DOM.

this is annoying, because we can't rely on the index of the dom node to match previous/new child, unless we save the old vnode somewhere

developit:
The unmounting was caused by Preact not knowing it should be appending or replacing, but we can probably use children.length to have a better idea of that.

I know it seems odd, but it's actually impossible to know the index of the DOM node that needs to be replaced, because VNodes are immutable and each render pass is given a complete new set of them as children.  Consider this example:

```js
// in render 1:
let children = [
  { nodeName: 'div', children: ['some text'] },
  { nodeName: 'div', children: [ { nodeName: HugeComponent } ] }
];

// in render 2:
let children = [
  { nodeName: 'div', children: [ { nodeName: HugeComponent } ] }
];
```

When diffing based on index alone, this would remount the component.

youknowriad:
@developit 

but how will you write the `render` function returning these elements, maybe something like this:

```js
return (
  { someCondition && <div>some text</div> }
  <div><HugeComponent />
);
```

this will produce 

```js
let children = [
  { nodeName: 'div', children: ['some text'] },
  { nodeName: 'div', children: [ { nodeName: HugeComponent } ] }
];
```
or
```js
let children = [
  '',
  { nodeName: 'div', children: [ { nodeName: HugeComponent } ] }
];
```

It gets trickier if it's not a condition but rather a loop (`filter`) or something, and I guess that's why we need explicit `key`s in this case.

developit:
^ your last point is the kicker there, and also the place where people tend to use keys incorrectly (simply passing the loop index, which does nothing since that's already the default).  Ideally, the diff should realize that an insertion is better because the next item is of like-type (or because the child count is increasing with this render).

youknowriad:
Question regarding this existing unit test https://github.com/developit/preact/blob/master/test/browser/components.js#L247

Changing the state value in this "Bad Container" and rerendering it causes the diff algorithm to reuse the previously mounted `Comp`. In my understanding, this is an error, the test should expect `F` instead of reusing `D`. Am I wrong?

developit:
No, if there's a matching component in the existing set of children it will always be reused. That case does seem slightly odd though - the first two `null` holes are collapsed into a single empty Text node, meaning the third component is counted as being at position 2 instead of 3.

src/constants.js
```diff
@@ -7,6 +7,7 @@ export const ASYNC_RENDER = 3;
 
 
 export const ATTR_KEY = '__preactattr_';
+export const ELT_KEY_PREFIX = 'preact-';
```
aduth: Do you need the prefix?

test/browser/render.js
```diff
@@ -472,8 +472,8 @@ describe('render()', () => {
 	it('should reorder child pairs', () => {
 		let root = render((
 			<div>
-				<a>a</a>
-				<b>b</b>
+				<a key="a">a</a>
```
aduth: Maybe there ought to be a test without the keys to verify that reordering still works, just not with strict equality of the last two assertions (since without the keys, the resulting nodes will be correct, just not reused).

src/vdom/diff.js
```diff
@@ -160,32 +160,24 @@ function idiff(dom, vnode, context, mountAll, componentRoot) {
...
 				props = child[ATTR_KEY],
 				key = vlen && props ? child._component ? child._component.__key : props.key : null;
-			if (key!=null) {
-				keyedLen++;
-				keyed[key] = child;
-			}
-			else if (props || (child.splitText!==undefined ? (isHydrating ? child.nodeValue.trim() : true) : isHydrating)) {
-				children[childrenLen++] = child;
-			}
+			key = key || ( ELT_KEY_PREFIX + i );
```
aduth: If `key` was explicitly assigned as `0` (or empty string), this would reassign to the prefixed index. Is that intended? Guessing this is what would have been previously respected by `key!=null` defaulting only in the case of `null` and `undefined`.

src/constants.js
youknowriad: It was a way to avoid ambiguity if the user explicitly set numeric keys for some of the elements (but not all). I don't know if we should really optimize for this use-case though

src/vdom/diff.js
youknowriad: Good catch

youknowriad:
Closing this as this is not likely the direction preact wants to move into.

Issue #850, opened 2017-09-03
Range Input: Order of Attributes Bug


mikewesthad:
I stumbled upon a bug while working with Preact today. If you create an input range, the initial render of the slider position may be wrong, depending on the order of the attributes. The `value` attribute needs to be set after `min`/`max`/`step`. If `value` is applied before `min`/`max`/`step` are properly set, the value will be forced to match the default min=0/max=100/step=1 on a slider.

E.g. if you create two sliders:

```js
// This one renders with the slider at 0
<input type="range" value={0.5} min="0" max="1" step="0.05" />

// This one renders with the slider at 0.5
<input type="range" min="0" max="1" step="0.05" value={0.5} />
```

![image](https://user-images.githubusercontent.com/1131802/30006088-12453c1e-90b6-11e7-8efd-06d275848719.png)

See minimal code to reproduce [here](https://codepen.io/mikewesthad/pen/gxEmmx)

When creating an dom element from a vnode, it looks like attributes are applied one at a time via a `for...in` loop. Since the ordering of keys in an object is arbitrary, it seems like [`diffAttributes`](https://github.com/developit/preact/blob/master/src/vdom/diff.js#L294) needs to apply the attributes in a sorted way. I can't think of other potential ordering pitfalls with attributes (maybe the react source code would have insights), but at the very least, applying `value` last should fix this issue.

developit:
Interesting! We might also be able to work around this by setting `value` as an attribute.

mikewesthad:
@developit - isn't `value` already being passed as an attribute to `diffAttributes`? Or did you mean pulling `value` out into its own property during `h`?

developit:
I mean special-casing `value` like we do `type` and `list`:
https://github.com/developit/preact/blob/master/src/dom/index.js#L76

Basically, forcing a `value` prop from VDOM/JSX to be applied to the DOM using setAttribute() instead of as a property.

mikewesthad:
Hm, but the problem is the order in which `value` is applied relative to other attributes, not whether it's applied via prop or attribute ([vanilla js example](https://codepen.io/mikewesthad/pen/qXGyrx?editors=0010)). I might be missing something, but by the time we're in `setAccessor`, it's too late to adjust the order without doing something async.

developit:
Interesting, I would have thought the step validation would only apply to the value property, not the attribute itself. You're right, this likely requires a bit of insane workaround magic in diffAttributes.

mikewesthad:
[...]

PR #865, opened 2017-09-11
Fix for #850: Range Input order of attributes bug

mikewesthad:
Summary of issue:
When creating a range `input` by setting one attribute at a time, the order in which the attributes are set matters. The `value` attribute should be applied last, otherwise it may be clipped/rounded to match the browser default min/max/step values. See bug example in #850.

PR:
This PR fixes that by adjusting `diffAttributes` to delay setting value on `input` elements until all other attributes are handled. The associated test passes with this RP (and fails without it).

Background:
@gpoitch pointed out the relevant react code [here](https://github.com/facebook/react/blob/8ab56e5c8b8f797a6b2bdd4188b4082c55a5190b/src/renderers/dom/stack/client/wrappers/ReactDOMInput.js#L58-L71), which has similar logic for setting value last. (Thanks!)


paranoidjk:
@MicheleBertoli typo. should be example in https://github.com/developit/preact/issues/850

src/vdom/diff.js
```diff
@@ -293,6 +293,7 @@ export function removeChildren(node) {
  */
 function diffAttributes(dom, attrs, old) {
 	let name;
+	const isInput = dom.tagName.toLowerCase() === "input";
```
developit: Need to use `dom.normalizedNodeName || dom.nodeName` here

mikewesthad: Ah, thanks for taking a look. Realized `isNamedNode` is there for this exact purpose, so I dropped that in place of the line in question.

src/vdom/diff.js
```diff
@@ -303,8 +304,14 @@ function diffAttributes(dom, attrs, old) {
...
 	for (name in attrs) {
-		if (name!=='children' && name!=='innerHTML' && (!(name in old) || attrs[name]!==(name==='value' || name==='checked' ? dom[name] : old[name]))) {
+		if (name==='children' || name==='innerHTML' || (isInput && name==='value')) continue;
+		if (!(name in old) || attrs[name]!==(name==='value' || name==='checked' ? dom[name] : old[name])) {
 			setAccessor(dom, name, old[name], old[name] = attrs[name], isSvgMode);
 		}
 	}
+
+	// update value last so that browser does not reject it for not fitting the default min/max/step
+	if (isInput && ("value" in attrs) && (!("value" in old) || attrs.value !== dom.value)) {
```
developit: I bet we could just always special-case `value`, even for non-inputs.

developit:
I'd like to get this into the next non-patch release, just need to golf it a bit to get the size hit down.

mikewesthad:
Closing this now that this has been merged in elsewhere in #1435

Issue #770, opened 2017-07-25
Flow check cause error


jakubzaba:
I got this error when I check Flow
```
Error: node_modules/preact/dist/preact.js.flow:3
  3: import { createElement as h, cloneElement, Component, render } from 'react';
                                                           ^^^^^^ Named import from module `react`. This module has no named export called `render`.
```
I can ignore it with `<PROJECT_ROOT>/node_modules/preact/.*` in `.flowconfig` or do you have any other suggestions how to fix it?

developit:
Looks like we need to update the definitions.

marvinhagemeister:
We removed support for flow in `master` as the typings were unmaintained and nobody took up the torch to update them. For now contributing definitions to `flow-typed` is much more future proof :+1:

PR #868, opened 2017-09-12
Fix flow types

sgrowe:
Currently the flow types import `render` from `react` but they are actually exported by 'react-dom' (at least as far as flow is concerned: https://github.com/facebook/flow/blob/master/lib/react.js#L257). See issue #770.

This PR fixes that and also adds a test to check flow types are correct (`npm run test:flow`).

Here's an example of the test output with incorrect flow types:

```
SROWE262:preact sr$ npm run test:flow

> preact@8.2.5 test:flow /Users/sr/preact
> flow check

Error: src/preact.js.flow:3
  3: import { createElement as h, cloneElement, Component, render } from 'react';
                                                           ^^^^^^ Named import from module `react`. This module has no named export called `render`.


Found 1 error
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! preact@8.2.5 test:flow: `flow check`
npm ERR! Exit status 2
npm ERR!
npm ERR! Failed at the preact@8.2.5 test:flow script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/sr/.npm/_logs/2017-09-12T12_57_31_562Z-debug.log
```


developit:
Good find!  This actually brings up a good point though:  `render()` has a different signature in preact than in React's version:

```js
render(vnode: Object, parent: Element, toReplace?: Element)
```

Maybe we should take this opportunity to update it here?

sgrowe:
Yeah, good shout! I've added that in but with vnode as a [`React.Node`](https://flow.org/en/docs/react/types/#toc-react-node) rather than an object just to be a bit more specific.

KenanY:
Since this seems like the latest stab at this issue, just connecting the dots: https://github.com/developit/preact/pull/775, https://github.com/developit/preact/issues/781, https://github.com/developit/preact/pull/782

PR #869, opened 2017-09-12
Breaking: update TypeScript definition for Component.children

niedzielski:
- Update ComponentProps.children typing to allow string arrays and
  JSX.Element arrays of arrays. e.g., this allows strictly typed
  components to be passed a child text node such as "foo". Note that
  children are always [arrays] and this isn't a functionally breaking
  change.

- Add new ComponentChildren type which is a useful fundamental type
  union to expose.

[arrays]: https://preactjs.com/guide/differences-to-react#what-s-missing-

---

I'm not sure I've understood any of this right so don't hesitate to
close it if it looks wrong. I think this would be a potentially type
breaking change for consumers that had extended ComponentProps and
redefined children.

The use case I'm trying to solve is
`render(Foo({fooProperty: "foo", children: ["bar"]}))` where:

```tsx
import { ComponentProps, h } from "preact";

export interface Props extends ComponentProps<any> {
  fooProperty: string;
}

export function Foo({
  fooProperty,
  children
}: Props): JSX.Element {
  return (
    <div>
      <div>{fooProperty}</div>
      <div>{children}</div>
    </div>
  );
}
```

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -4,6 +4,8 @@ export as namespace preact;
 declare namespace preact {
 	type Key = string | number;
 	type Ref<T> = (instance: T) => void;
+	type ComponentChild = JSX.Element | string | number | null;
+	type ComponentChildren = ComponentChild[];
 
 	/**
 	 * @deprecated
@@ -29,7 +31,7 @@ declare namespace preact {
 	}
 
 	interface PreactDOMAttributes {
-		children?: JSX.Element[];
+		children?: ComponentChildren;
 		dangerouslySetInnerHTML?: {
 			__html: string;
 		};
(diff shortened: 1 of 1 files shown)


developit:
Hi @niedzielski - this looks like a good improvement! The only thing I think we might want to change is that Preact will never pass `children` as arrays of arrays - it's always flattened by the time it gets passed to a component.

niedzielski:
@developit, done and thanks! Please let me know if I've misunderstood you.

developit:
Looks good to me! I usually like to have someone more familiar with typescript review before merging.

calebeby:
I'd really like to see this merged!

src/preact.d.ts
```diff
@@ -1,6 +1,9 @@
 declare namespace preact {
+	type ComponentChild = JSX.Element|string;
```
marvinhagemeister: Type `number` and `null` is supported as well :)

niedzielski:
Rebased and squashed :+1:

marvinhagemeister:
@niedzielski Awesome! Thank you so much for your contribution 🎉

PR #870, opened 2017-09-13
Change "it's" to "its" in README.md

jdanford:


developit:
Oh dear, I'm the worst with [that specific slip-up](http://www.its-not-its.info/), there are probably many instances of it haha.

PR #876, opened 2017-09-16
property to check if component is class-based or stateless

alfrcr:
Shamelessly copying the reactjs way


alfrcr:
can do the check by using `Component.prototype.render` as discussed on Twitter

PR #880, opened 2017-09-17
wrap render in `try` when component defines `componentDidCatch`

adamsilverstein:
First pass at https://github.com/developit/preact/issues/799

### DIFF
--- a/src/vdom/component.js
+++ b/src/vdom/component.js
@@ -95,7 +95,20 @@ export function renderComponent(component, opts, mountAll, isChild) {
 	component._dirty = false;
 
 	if (!skip) {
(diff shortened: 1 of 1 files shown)


developit:
I think `componentDidCatch` needs to catch errors in child render methods and bubble them up if we're to match React's implementation.  @rpetrich implemented something that does this in #819, mind taking a look?

adamsilverstein:
Good point, I agree it should work that way. I'll take a look, thanks for the reference.

adamsilverstein:
Ah, #819 is much farther along, closing this favor of that PR and I'll see if I can contribute there.

Issue #648, opened 2017-04-18
Post-mount with componentDidMount


ustccjw:
#In 8.1.0, componentDidMount/componentDidUpdate can not promise to be called after dom mount/update.

related resource: https://developmentarc.gitbooks.io/react-indepth/content/life_cycle/birth/post_mount_with_component_did_mount.html

related PR:
https://github.com/developit/preact/pull/506#issuecomment-293879820

developit:
@ustccjw would you be able to put together a jsfiddle that shows the ordering issue?  I'm not sure I understand the problem.

ustccjw:
demo: https://jsfiddle.net/ustccjw/uzyh4r01/1/

when you click update button，child `componentDidUpdate` will execute before parent DOM update.

robertknight:
I'll describe the issue as I understand it. Given this component tree:

```
Parent
   |
   Child
```

In Preact 8.1.0, triggering an update on `Parent` which re-renders both `Parent` and `Child` will result in the child's `componentDidUpdate` callback being triggered _after_ the `Child`'s DOM has updated but _before_ the `Parent`'s DOM has been updated.

React on the other hand queues both `componentDidMount` and `componentDidUpdate` callbacks and executes them all once the entire tree has been updated ([in the Fiber renderer](https://github.com/facebook/react/blob/master/src/renderers/shared/fiber/ReactFiberScheduler.js#L509), [in the stack Renderer](https://github.com/facebook/react/blob/master/src/renderers/dom/stack/client/ReactReconcileTransaction.js#L67)).

The [componentDidUpdate](https://facebook.github.io/react/docs/react-component.html#componentdidupdate) docs don't specify this behaviour, but if the child runs code in `componentDidUpdate` which might be invalidated by the parent's update in any way (eg. accessing the parent directly, measuring the child's own size) then the result will be wrong unless the child's `componentDidUpdate` is deferred until the whole tree is updated.

I suggest that in Preact we queue `componentDidUpdate` callbacks in the same queue that handles mount callbacks and flush them all once the entire tree is updated. I think we deferred doing this last time because it increased the size slightly, but I don't see an easy way around this. Thoughts @developit ?

developit:
I agree. We can work some magic to keep the size down. Maybe this ties into the idea of a more general queue (though we can't really use it as a general callback queue due to size/perf).

ustccjw:
@developit @robertknight Are there any solutions ?

developit:
[...]

PR #884, opened 2017-09-19
Fix issue #648: add queue for componentDidUpdate

markselby9:
Fix this issue by adding a queue for componentDidUpdate lifecycles and flush it after rendering

before this PR: gzip size: 3444
after this PR: gzip size: 3503

Issue: https://github.com/developit/preact/issues/648


marvinhagemeister:
Dang, I wish we'd seen this PR sooner! Our code in `master` is similar to the changes in this PR. Since I started working on preact (about a year ago) we've always had a queue for `componentDidMount`. Not sure when it was added and why this PR wasn't merged back then :S

PR #885, opened 2017-09-19
Add error message about render method in debug

markselby9:
1. Added error message `parentElement` === `mergeElement ` in render method
2. Added error message when both parentElement and mergeElement be undefined

Issue: https://github.com/developit/preact/issues/815


NekR:
@developit you gotta CC me to this. You cannot `require('preact')` in webpack, because it wraps modules into ES6 Modules interop.

PR #890, opened 2017-09-21
check child is meaningless

MrErHu:
Check child is meaning. when code run here, child must be null!


developit:
I don't think this is true for children with mixed keyed/unkeyed.

marvinhagemeister:
@developit We reassign `child` to `null` just a few lines above this check https://github.com/developit/preact/blob/master/src/vdom/diff.js#L195 . Seems like this can be indeed removed.

Issue #844, opened 2017-08-28
dangerouslySetInnerHTML, children and setState quirks


molily:
Hello,

first of all, thank you for this great project. After having built several React apps, we’ve decided to give Preact a try. We’ve built an app using preact-cli and almost everything worked flawlessly. I’m really excited about this project!

We ran into one strange issue that took me hours to track down. This is the smallest test case I was able to produce:

```js
import { h, render, Component } from 'preact';
import PropTypes from 'prop-types';

class Container extends Component {

  constructor() {
    super();
    this.state = { state: 'initial' };
  }

  componentDidMount() {
    setTimeout(() => {
      this.setState({ state: 'updated' });
    }, 2000);
  }

  render() {
    const { children } = this.props;
    const { state } = this.state;
    return (
      <div>
        <div>Container state: {state}</div>
        {children}
      </div>
    );
  }

}

Container.propTypes = {
  children: PropTypes.element.isRequired
};

const App = () => (
  <Container>
    <div>App</div>
    <div>
      →
      <span dangerouslySetInnerHTML={{ __html: '<u>HTML</u>' }} />
      ←
    </div>
  </Container>
);

render(<App />, window.document.body);
```

Expected behavior: After the state change, the content set via dangerouslySetInnerHTML remains unchanged.

Actual behavior: After the state change, the content set via dangerouslySetInnerHTML is removed from the DOM completely.

I’m a newbie to the codebase so I can explain why this happens but I cannot come up with a fix.

Under normal circumstances, when the state of a component changes, the call stack is `rerender` → `idiff` → `innerDiffNode` (which removes the subtree added by `dangerouslySetInnerHTML`) → `diffAttributes` → `setAccessor` (which sets `innerHTML` again).

Under normal circumstances, `setAccessor` is called because this condition …

`attrs[name]!==(name==='value' || name==='checked' ? dom[name] : old[name]))`

yields true, in short

`attrs[name]!==old[name]`

yields true. The old value (an object `{ __html: '<u>HTML</u>' }`) is a equal but not identical to the new value (also an object `{ __html: '<u>HTML</u>' }`). This is because the object is created anew with each `App#render` call.

In contrast, in the test case above, `App#render` is not called again when `Container` re-renders, so it’s the same old attributes object. Therefore `diffAttributes` does not detect a difference so it does not call `setAccessor` to set `innerHTML` again.

[...]

PR #894, opened 2017-09-22
Fix dangerouslySetInnerHTML being reapplied repeatedly whenever a vnode renders

rpetrich:
Two adjustments had to be made to prevent the DOM subtree from unnecessary updating on a node with `dangerouslySetInnerHTML` set:
1. Only reapply dangerouslySetInnerHTML when the inner __html value changes
2. Don't diff children for vnodes that have a dangerouslySetInnerHTML attribute


src/dom/index.js
```diff
@@ -60,7 +60,7 @@ export function setAccessor(node, name, old, value, isSvg) {
 		}
 	}
 	else if (name==='dangerouslySetInnerHTML') {
-		if (value) node.innerHTML = value.__html || '';
+		if (typeof value !== "undefined") node.innerHTML = value;
```
developit: I think `if (value!==undefined)` is a bit cheaper (gzip/runtime)

developit:
@rpetrich - any chance the `value.__html !== old.__html` check fixes this issue on it's own? It seems like, while it might end up resulting in setAccessor being called erroneously like it is today, it would prevent re-applying `innerHTML` when the value has not changed.

rpetrich:
@developit It doesn't. `innerHTML` creates a DOM subtree that's unmanaged by preact, which will then be destroyed by `innerDiffNode` if a check isn't added.

_(at least that's my memory of this patchset—it's been awhile since I've reviewed it in depth)_

cristianbote:
This is addressing #844

marvinhagemeister:
This is fixed with Preact X :tada:

PR #900, opened 2017-10-03
Fix issue of unmounting component which only renders `null`

laysent:
Using current implementation, when component renders `null`, it will be actually rendered as an empty text node. However, if that component is the only child inside a parent component, when removed, `componentWillUnmount` will not be triggered. (See unit test for details)
The reason is that children with falsy value will be treated as empty string when VNode is created, and that is the same as how component returns null is treated.
Using comment node instead of empty text node will fix this.

All tests passed, but I am not sure if the implementation is good. Any suggestion is very welcome.

### DIFF
--- a/src/vdom/diff.js
+++ b/src/vdom/diff.js
@@ -65,8 +65,19 @@ function idiff(dom, vnode, context, mountAll, componentRoot) {
 	let out = dom,
 		prevSvgMode = isSvgMode;
 
-	// empty values (null, undefined, booleans) render as empty Text nodes
-	if (vnode==null || typeof vnode==='boolean') vnode = '';
+	// empty values (null, undefined, booleans) render as comment nodes
+	if (vnode==null || typeof vnode==='boolean') {
+		const comment = ' preact empty ';
+		if (!dom || dom.nodeName !== '#comment' || dom.textContent !== comment) {
+			out = document.createComment(comment);
+			if (dom) {
+				if (dom.parentNode) dom.parentNode.replaceChild(out, dom);
+				recollectNodeTree(dom, true);
+			}
+		}
+		out[ATTR_KEY] = true;
+		return out;
+	}
 
 
 	// Fast case: Strings & Numbers create/update Text nodes.
--- a/test/browser/components.js
+++ b/test/browser/components.js
@@ -251,6 +251,24 @@ describe('Components', () => {
 
 	});
 
+	it('should trigger componentWillUnmount when unmounting component that renders null', () => {
+		const componentWillUnmount = sinon.spy(() => { });
+		class Comp extends Component {
+			componentWillUnmount() {
+				componentWillUnmount();
+			}
+			render() {
+				return null;
+			}
+		}
+		const Parent = ({ show }) => <div>{show && <Comp />}</div>;
+		render(<Parent show />, scratch);
+		expect(scratch.lastChild.innerHTML).to.equal('<!-- preact empty -->');
+		render(<Parent show={false} />, scratch, scratch.lastChild);
+
+		expect(componentWillUnmount).to.have.been.calledOnce;
+		expect(scratch.lastChild.innerHTML).to.equal('');
+	});
 
 
 	describe('props.children', () => {
--- a/test/browser/render.js
+++ b/test/browser/render.js
@@ -91,22 +91,22 @@ describe('render()', () => {
 
 	it('should not render null', () => {
 		render(null, scratch);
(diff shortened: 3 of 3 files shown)


developit:
Hi @laysent - good catch!  I've been thinking, and it might be possible to solve this more simply and without involving comments, by adding a check on [L136 of diff.js](https://github.com/developit/preact/blob/master/src/vdom/diff.js#L136):

```diff
- if (!hydrating && vchildren && vchildren.length===1 && typeof vchildren[0]==='string' && fc!=null && fc.splitText!==undefined && fc.nextSibling==null) {
+ if (!hydrating && vchildren && vchildren.length===1 && typeof vchildren[0]==='string' && fc!=null && fc.splitText!==undefined && fc._component==null && fc.nextSibling==null) {
```

laysent:
Hi @developit thanks a lot for the review.
I would still recommend to use comment instead of EmptyTextNode, because I think there is
a very weird issue in IE (the actual reason why I made this PR). Please look at the following code:

```javascript

const Empty = () => null;

const container = document.createElement('div');
container.id = 'root';
document.body.appendChild(container);;

render(<Empty />, container);

console.log(
  'preact componet is in EmptyTextNode after render: ',
  document.getElementById('root').firstChild._component,
);

setTimeout(() => {
  console.log(
    'It might be undefined in IE after a while without doing anything: ',
    document.getElementById('root').firstChild._component,
  );
}, 5 * 1e3);
```

When I use IE 11 to run the above code, I sometimes (4 out of 5) notice that the second
`console.log` will output `undefined` instead of the actual `Empty` component. This is
not 100% reproducable, but it happens a lot.

If `node._component` is undefined, it will not be unmounted correctly, because in following
code, condition is not truthy:

```javascript
export function recollectNodeTree(node, unmountOnly) {
	let component = node._component;
	if (component) {
		// if node is owned by a Component, unmount that component (ends up recursing back here)
		unmountComponent(component);
	}
```

One thing to notice is that, when preact creates EmptyTextNode, if we store it in any global
variable, the above issue will disappear. It seems to me that IE does not always keep
EmptyTextNode, but creates one when you access it via code. (I can't find any evidence to proof
this, it's just my theory)

When you use comment, it seems there is no such issue.

I hope I have made myself clear. Let me know if you have any question or if I made any
stupid mistake. ;-P

developit:
I wonder if a self-reference on the text node would be sufficient to keep it from being garbage-collected?

laysent:
@developit Hi, thanks for the feedback.

I have tried it, seems IE still garbage-collects it. Here is how I test:

```javascript
// update if it's already a Text node:
if (dom && dom.splitText!==undefined && dom.parentNode && (!dom._component || componentRoot)) {
	if (dom.nodeValue!=vnode) {
		dom.nodeValue = vnode;
	}
}
else {
	// it wasn't a Text node: replace it with one and recycle the old Element
	out = document.createTextNode(vnode);
	out.__selfPointer= out; // This is the line I add
        /**
         * If I change it to following, seems it will not be garbage-collected
         * window.__pointer = out;
         */
	if (dom) {
		if (dom.parentNode) dom.parentNode.replaceChild(out, dom);
		recollectNodeTree(dom, true);
	}
}
```

The above code is around `src/vdom/diff.js` line 84

laysent:
@developit Hi, any update on this? If there is anything I can update/improve, feel free to let me know

marvinhagemeister:
I can't reproduce the described issue in `master`. The test from this PR is passing there. It was most likely fixed when we pushed the code for Preact X where we don't diff against the DOM anymore :tada:

jasongornall:
@laysent I just ran into this exact issue... Did you end up going with this fix?

marvinhagemeister:
@jasongornall Sounds like you're on an older version of Preact. Please upgrade to Preact `10.5.3`.

jasongornall:
We are in the process of upgrade a monolith codebase... But do you think this patch would help us out in the interim?

marvinhagemeister:
Not sure, I've never really worked on the 8.x code-base, except maybe for a little bug fix here and there.

I'm curious though what's preventing you from upgrading to the `10.x` release line. We had a lot of positive feedback from users working on large to small upgrades that they were able to upgrade easily.

laysent:
> @laysent I just ran into this exact issue... Did you end up going with this fix?

@jasongornall Hi, it's been quite a long time, not sure how exactly I dealt with it. But returning something like `<div />` will by pass it, hopefully this workaround helps.

PR #905, opened 2017-10-10
[ts] stateful Component type should be assignable to AnyComponent type

### LINKED ISSUE #903: Can't assign component class to type AnyComponent in typescript
This typescript snippet produces an error, where I'd expect it to work:

```typescript
interface Props { name: string; }
interface State { value: number; }
class MyComponent extends preact.Component<Props, State> {
  render() { return null; }
}

let MyComponentVar: preact.AnyComponent<Props, State> = MyComponent;
```

In preact.d.ts there is the line
https://github.com/developit/preact/blob/f7834ec9fb7848581c6fbc7833bcbe2f2db9fe61/src/preact.d.ts#L46
which seems to lose type information in the case of `typeof Component`. I think this should be
```
type AnyComponent<PropsType, StateType> = FunctionalComponent<PropsType> | ComponentConstructor<PropsType, StateType>;
```
instead. This looks related to Typescript FAQ item here:
https://github.com/Microsoft/TypeScript/wiki/FAQ#why-cant-i-write-typeof-t-new-t-or-instanceof-t-in-my-generic-function

preact version: 8.2.5

mike-north:
Fixes #903
Related: https://github.com/developit/preact-redux/issues/24

Small example of what happens in the current version of preact
```ts
// A (potentially) stateful component
class C extends Component<{foo: 'bar'}, {}> {
  render() {return (<div>Hi</div>); }
}
// This will throw an error on TypeScript compilation!
let x: AnyComponent< {foo: 'bar'}, {}> = C;
/*
Type 'typeof C' is not assignable to type 'AnyComponent<{ foo: "bar"; }, {}>'.
  Type 'typeof C' is not assignable to type 'FunctionalComponent<{ foo: "bar"; }>'.
    Type 'typeof C' provides no match for the signature '(props?: ({ foo: "bar"; } & ComponentProps<FunctionalComponent<{ foo: "bar"; }>>) | undefined, context?: any): Element'.
*/ 
```
Use of typeof here

https://github.com/developit/preact/blob/f7834ec9fb7848581c6fbc7833bcbe2f2db9fe61/src/preact.d.ts#L46

is inappropriate. The end result is that currently, only `FunctionalComponent<P, S>` can be assigned to `AnyComponent<P, S>`.


shantanuraj:
@developit @mike-north  Can we get this merged? The code looks good, and fixes class components / preact-redux integration.

GuskiS:
Merge please :(

PR #911, opened 2017-10-13
Optimize defer

jjant:
Optimize `defer` function by removing unnecessary duplicate call of `Promise.resolve`.

According to https://jsperf.com/preact-defer-perf, it's almost twice as fast.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -10,4 +10,5 @@ export function extend(obj, props) {
 /** Call a function asynchronously, as soon as possible.
  *	@param {Function} callback
  */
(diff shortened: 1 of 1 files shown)


developit:
Hiya @jjant! This looks good and the perf boost is nice, but it also adds 11 bytes.  There might be a way to reorganize things to avoid that.

bmeurer:
@developit How about this?

```js
export const defer = typeof Promise=='function' ? Promise.prototype.then.bind(Promise.resolve()) : setTimeout;
```

There's no need to create a `Promise` instance, just to get to the `then` method on the prototype.

marvinhagemeister:
We've merged the suggested snippet with #1370 :+1:

PR #912, opened 2017-10-14
refactor: remove condition which is always true

clarkdo:


pl12133:
Thanks @clarkdo! However this looks identical to #890.

PR #913, opened 2017-10-14
refactor isSameNodeType in vdom

clarkdo:


marvinhagemeister:
The only change of value is line 10, where two conditional checks have been modified to iterate over an array instead. Although I haven't Benchmarked this, I'd assume that the iteration is a tiny tiny bit slower.

Issue #898, opened 2017-09-30
a diff bug 


aaawhz:
`component.shouldComponentUpdate(props , state, context) === false`

‘===’  ==》 ‘==’

NaN false null undefined 0  can work

yaodingyd:
First of all, `NaN`, `null` and `undefined` is not `==` to `false`.
Second, `shouldComponentUpdate` should return either `true` or `false`.

PR #916, opened 2017-10-15
make the preact as same as react in shouldComponentUpdate return value

huangming1994:
Hey, l found there is some difference between `React` and `Preact` about `shouldComponentUpdate`.
if you return a value like `''`,`0`,`false`,`NaN`,`undefined` in `React` , they all skip render. but only `false` skip render in `Preact`.

### DIFF
--- a/src/vdom/component.js
+++ b/src/vdom/component.js
@@ -80,7 +80,7 @@ export function renderComponent(component, opts, mountAll, isChild) {
 		component.context = previousContext;
 		if (opts!==FORCE_RENDER
 			&& component.shouldComponentUpdate
-			&& component.shouldComponentUpdate(props, state, context) === false) {
+			&& !component.shouldComponentUpdate(props, state, context)) {
 			skip = true;
 		}
 		else if (component.componentWillUpdate) {


yaodingyd:
It's intended. See #898.

thysultan:
@yaodingyd Either i'm reading this wrong but i think the comment by @developit in #898 might come of as confusing. The intention seems to be to follow React in the heuristics of `shouldComponentUpdate`, which is appropriately matched by what @huangming1994 has presented.

huangming1994:
@yaodingyd ok, l didn't see the issue. :D

developit:
Do we know what values React supports for this?  If anyone has time to set up a jsfiddle test, that would be useful.  If we're going to change Preact's behavior here, it only makes sense to bring it more in-line with React's.

Kanaye:
As suggested [here](https://jsfiddle.net/s9szt0sf/1/) a jsfiddle.
Looks like react treats everything `truthy` as `true` and everything `falsy` as `false`.
Also react logs a warning `Warning: TestComponent.shouldComponentUpdate(): Returned undefined instead of a boolean value. Make sure to return true or false.` for `undefined` but not for other non boolean values 😕

developit:
What if we changed the check to this?  It'd get us `0, '', false` equalling false:

```js
&& component.shouldComponentUpdate(props, state, context) == false) {
```

marvinhagemeister:
This is fixed in Preact X :tada:

PR #923, opened 2017-10-24
add TypeScript support

k1r0s:
Hi @developit 

There are many developers that choose Preact with TypeScript. Also there are lot of preact-typescript starters here on gh. It would be nice to make its life easier.


developit:
useful!

PR #935, opened 2017-11-17
Fix #924

### LINKED ISSUE #924: `preact/debug` error only in v8.2.6
After upgraded my project from preact@v8.2.5 to v8.2.6, I had this error when loading the page:

```js
debug.js:49 Uncaught TypeError: Cannot set property render of #<Object> which has only a getter

// debug.js:L49
preact.render = function (vnode, parent, merge) {
```

If I stop requiring `preact/debug` it would be OK. I require it like this:
```ts
window['Promise'] = require('promise-polyfill')

if (process.env.ENV === 'production') {
  // Production
} else {
  // Development
  Error['stackTraceLimit'] = Infinity
  require('preact/debug')
}
```

When using v8.2.5 it was all good.

I'm using only preact with `webpack@2` without `preact-compat` or any react related stuff.

developit:


NekR:
So what happens when there is an error? It just doesn't apply the debug feature? Seems good then, until we find a better solution. I actually thought that you already merged this.

developit:
Yup.

PR #943, opened 2017-11-25
Add *Capture TypeScript type definitions

andybons:
Since properties support `on<Event>Capture` props, support them in TS.


developit:
Approved! we may want to check if this interacts poorly with the other TS PR.

andybons:
@developit do you mean this one? https://github.com/developit/preact/pull/933

developit:
Yup

andybons:
This and #933 alter different parts of the file that are unrelated.

Issue #915, opened 2017-10-15
componentWillUnmount() fires too late


purplecode:
Hi. I have noticed that `componentWillUnmount()` in not fired immediately after the component is removed from the DOM in case when it was rendered into the same element type as the one that is replacing it (comments: 1, 2). It is fired much later, when the element is replaced with a completely different content (comments: 3, 4, 5).

It is similar to [#392](https://github.com/developit/preact/issues/392) and [#373](https://github.com/developit/preact/issues/373), but reproducible with the latest version.

*[jsfiddle](https://jsfiddle.net/niespammnie/a22awx1y/1/)*

```html
<!DOCTYPE html>
<html>
<script src="https://cdnjs.cloudflare.com/ajax/libs/preact/8.2.5/preact.js"></script>
<body></body>
<script>
const { h, render, Component } = preact;

class App extends Component {
  componentDidMount() { console.log('componentDidMount') }
  componentWillUnmount() { console.log('componentWillUnmount') }
  render() { return h('div', {}, '[component]') }
}

render(h(App), document.body);

// 1. This is not firing componentWillUnmount() 
render(h('div', {}, '[div]'), document.body, document.body.lastChild);
// 2. Neither this
render(h('div', {}, h('div', {}, '[another diff]')), document.body, document.body.lastChild);

// 3. I need to render null
// render(null, document.body, document.body.lastChild);

// 4. Or a different element
// render(h('span', {}, '[span]'), document.body, document.body.lastChild);

// 5. Or a different component
// class AnotherApp extends App {}
// render(h(AnotherApp), document.body, document.body.lastChild);
</script>
</html>
```

yaodingyd:
Seems like `recollectNodeTree` is never called for a simple `vnode` with same node name as current `dom`'s node name in `diff` process.

developit:
Hi there - `componentWillUnmount()` is called before unmounting/removal, not after.
This is definitely a bug though!  Here's the offending line:
https://github.com/developit/preact/blob/0a88752e4890d391b5e23809d9432f625a833ae8/src/vdom/diff.js#L110

That line needs to be amended to check `dom._component`, similar to the check for Text nodes here:
https://github.com/developit/preact/blob/0a88752e4890d391b5e23809d9432f625a833ae8/src/vdom/diff.js#L76

NekR:
I believe it affects `refs` order too, I had a repro somewhere..

narendrashetty:
[...]

PR #952, opened 2017-12-06
 fixes the bug to trigger componentWillMount when same component is replaced

narendrashetty:
This fixes this issue #915 where the componentWillMount needs to be triggered when it is replaced with similar component.

I have also added a test case for the same.

Let me know if I need to do any other checks


developit:
Quick glance and this looks good! Thanks a million @narendrashetty.

narendrashetty:
@developit Thank you! It's my pleasure to contribute to Preact. Hoping this is just a start :)

developit:
Me too!!

andrewiggins:
Since the original issue was closed cuz it is fixed in Preact X, I'm gonna close this PR. Let us know if you still have issues with this in Preact X!

Issue #836, opened 2017-08-25
Child components become undefined in HOC after rerender component


vyshkov:
Hi. I have a strange problem, which is not reproducing if i am using React.
I have such code:

```
const SomeComponet = props => (
    <div>{props.children}</div>
);

function connect(WrappedComponent) {
    return class Connect extends Component {
        componentDidMount() {
            setTimeout(() => this.forceUpdate(), 3000)
        }
        render() {
            console.log("children:", this.props.children)
            return <WrappedComponent {...this.props}/>;
        }
    };
}

const HOC = connect(SomeComponet);

export default () => (
    <HOC>
        <span>!!!!!!!!!</span>
    </HOC>
);
```

After this.forceUpdate() all props become empty, so in console i see such output:

> children: [VNode]
> children: undefined

And that span `<span>!!!!!!!!!</span>` is not rendering any more.

Please help me understand whats going wrong.

pabla:
I have the same problem.

developit:
This is because Babel transpiles `<a {...obj}>` to `h('a', obj)` without cloning `obj`.  Preact then [mutates the props passed to h() in-place](https://github.com/developit/preact/blob/master/src/h.js#L21), which is probably shouldn't be doing!  For now, passing any other prop to the child component will mitigate the effect, because it will trigger Babel to clone the props when expanding them onto the JSX element.

Here's a JSFiddle showing the fix (adding an "a" prop):
https://jsfiddle.net/developit/7utpc2gp/

vyshkov:
Thank you @developit

paranoidjk:
I think this is a bug of the babel. Since spread operator should do a shallow copy.

marvinhagemeister:
Just tried the snippet in the original comment and I'm unable to reproduce the bug in  Preact X :tada:

PR #953, opened 2017-12-06
Cloning attributes object to avoid mutation in h()

narendrashetty:
This fixes the issue #836 which is caused when using `<a {...obj}>` where Babel transpiles to `h('a', obj)` without cloning the `obj`.

So I now clone the attributes object before any operation is done.

This PR also breaks the test `should render components with props` where it compares the passed props with the component props. Since now it doesn't mutate, the `children` property is still part.

What do you think I should do about it @developit.


developit:
This is probably ly too much of a performance impact to do in every `h()`. Thinking...

narendrashetty:
@developit That's right. Is it then better to fix this issue from Babel side? Can't think of any better solution for this.

React seems to be cloning the object, that's where I got the idea from.

developit:
I would love to fix it in Babel yeah. It's interesting though, I tend to prefer `<div {...props} />` because I know it avoids a clone, now we'd be changing that semantic.

Perhaps if performance testing reveals this clone is cheap enough in preact, we could merge. The function is already there at least.

narendrashetty:
Having the props mutable may give some weird bugs. Because I would never expect the passed props when changed internally effect the outer value.

I come from the idea that if anything in the props needs to be changed, let the outer component know. May be I am wrong. Let me know what you think.

developit:
That's definitely the correct philosophy, so I can see how props being sometimes mutable is problematic. Just want to make sure we solve it in the right place.

hellsan631:
This is one of the nastiest bugs i've ever encountered. Completely giving up on preact because of this. None of my UI libraries work in production any more.

hellsan631:
This kills a lot of compatibility with newer UI react libraries. Where you would expect them to work, they don't. It also is problematic for using a lot of HOC that really help with productivity.

To note: this seems to happen when using preact-cli, and not when creating an alias on a react project.

Case in point, i use storybook to build UI's with preact, and didn't have any issue inside storybook. But in the preact-cli environment, I am unable to use the components that I built. This caused a lot of frustration.

Mitranim:
Just ran into a bug caused by the props mutation. Found this PR. Would love to help it move along, if possible. Mutating user-defined inputs leads to very surprising behavior and should be avoided.

Perhaps we should clone props only when mutating:

```js
	if (attributes && attributes.children != null) {
		if (!stack.length) stack.push(attributes.children);
                 attributes = extend({}, attributes);
		delete attributes.children;
	}
```

In other words, treat them as immutable, but avoid _needless_ copies.

marvinhagemeister:
The issue that this PR resolves is fixed in Preact X. The test from this PR passes in `master` without any additional changes :+1:

PR #963, opened 2017-12-27
[WIP] New context API

jtmthf:
This PR is an implementation of React's proposed new context API for Preact. This implementation is built on top of the existing context API. React's corresponding PR is https://github.com/facebook/react/pull/11818 and the RFC is https://github.com/reactjs/rfcs/pull/2.

These are the following tasks and considerations I've identified that will need to be accomplished before merging.

- [ ] Add tests. I'm planning for now on recreating the existing context based tests, but using the new API.
- [ ] TypeScript definitions.
- [ ] Flow definitions.
- [ ] Devtools support warning on rendering a `Provider` without any `Consumer`'s.
- [ ] Code documentation improvements.
- [x] Wait for https://github.com/reactjs/rfcs/pull/2 to be approved.
- [ ] Examine size reduction opportunities. Currently this adds 200 bytes to the (min + gzip) size.
- [ ] Consider refactoring to remove the old context API and only support this new API. This would be inline with Preact traditionally only supporting the most modern parts of React's API. Currently it appears that Facebook may deprecate and remove old context at some future point.

### DIFF
--- a/src/create-context.js
+++ b/src/create-context.js
@@ -14,7 +14,7 @@ export function createContext(key) {
 	function Provider(props, context) {
 		Component.call(this, props, context);
 
-		c = [];
+		this.c = [];
 	}
 
 	extendComponent(Provider.prototype, {
@@ -29,7 +29,7 @@ export function createContext(key) {
 
 		componentWillReceiveProps(nextProps) {
 			if (!Object.is(this.props.value, nextProps.value)) {
-				c.forEach(subscriber => {
+				this.c.forEach(subscriber => {
 					subscriber(nextProps.value);
 				});
 			}
--- a/src/create-context.js
+++ b/src/create-context.js
@@ -1,0 +1,64 @@
+import { Component } from "./component";
+import { extendComponent } from "./util";
+
+/**
+ *
+ * @param {string} key
+ */
+export function createContext(key) {
+	/** Provider Component class.
+	 *	prop `value` is provided to every corresponding `Consumer` component.
+	 *	@public
+	 *
+	 */
+	function Provider(props, context) {
+		Component.call(this, props, context);
+
+		c = [];
+	}
+
+	extendComponent(Provider.prototype, {
+		s(subscriber) {
+			this.c.push(subscriber);
+			subscriber(this.props.value);
+		},
+
+		getChildContext() {
+			return { [key]: this.s };
+		},
+
+		componentWillReceiveProps(nextProps) {
+			if (!Object.is(this.props.value, nextProps.value)) {
+				c.forEach(subscriber => {
+					subscriber(nextProps.value);
+				});
+			}
+		},
+
+		render() {
+			return this.props.children;
+		}
+	});
+
+  /**  Consumer Component class.
+	 *	`children` must be a function that accepts the value from the `Consumer`'s
+	 		corresponding `Provider` component.
+	 *	@public
+	 *
+	 */
+	function Consumer(props, context) {
+		Component.call(this, props, context);
+
+		context[key](value => {
+			this.setState({ value });
+		});
+	}
+
+	extendComponent(Consumer.prototype, {
+		render() {
+			return this.props.children(this.state.value);
+		}
+	});
+
+	return { Provider, Consumer };
+}
--- a/src/preact.js
+++ b/src/preact.js
@@ -1,6 +1,7 @@
 import { h, h as createElement } from './h';
 import { cloneElement } from './clone-element';
 import { Component } from './component';
+import { createContext } from './create-context';
 import { render } from './render';
 import { rerender } from './render-queue';
 import options from './options';
@@ -10,6 +11,7 @@ export default {
 	createElement,
 	cloneElement,
 	Component,
+	createContext,
 	render,
 	rerender,
 	options
@@ -20,6 +22,7 @@ export {
 	createElement,
 	cloneElement,
 	Component,
+	createContext,
 	render,
 	rerender,
 	options
--- a/src/util.js
+++ b/src/util.js
@@ -1,3 +1,5 @@
+import {Component} from './component';
+
 /**
  *  Copy all properties from `props` onto `obj`.
  *  @param {Object} obj		Object onto which properties should be copied.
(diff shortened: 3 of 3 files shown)


zeromake:
my context api realization

[link](https://github.com/zeromake/preact/tree/new-context)
[demo](https://jsfiddle.net/vk5rnoaq/)

create-context.js
``` javascript
import { Component } from "./component";
import { extendComponent } from "./util";

export function createContext(value) {
    var context = {
        default: value
    }
    function Provider(p, c) {
        Component.call(this, p, c);
        this.c = [];
        this.p = this.p.bind(this);
        this.u = this.u.bind(this);
    }
    extendComponent(Provider, {
        p(subscriber) {
            this.c.push(subscriber);
            return this.props.value;
        },
        u(subscriber) {
            this.c = this.c.filter(function (i) { return i !== subscriber; });
        },
        getChildContext() {
            var provider = {
                push: this.p,
                pop: this.u,
                context: context,
            };
            let providers = this.context.providers;
            if (providers) {
                providers.push(provider);
            } else {
                providers = [provider];
            }
            return { providers };
        },
        componentWillReceiveProps(nextProps) {
            if (this.props.value !== nextProps.value) {
                this.c.forEach(function (subscriber) {
                    subscriber(nextProps.value);
                });
            }
        },
        render() {
            return this.props.children && this.props.children[0];
        }
    });
    function Consumer(p, c) {
        Component.call(this, p, c);
        this.updateContext = this.updateContext.bind(this);
        if (c.providers) {
            for (var i = c.providers.length - 1; i >= 0; i--) {
                var provider = c.providers[i];
                if (provider.context === context) {
                    var value = provider.push(this.updateContext);
                    this.state = {
                        value: value,
                    };
                    this.popSubscriber = provider.pop;
                    break;
                }
            }
        }
    }
    extendComponent(Consumer, {
        updateContext(val) {
            this.setState({ value: val });
        },
        componentWillUnmount() {
            if (this.popSubscriber) {
                this.popSubscriber(this.updateContext);
            }
        },
        render() {
            return this.props.children && this.props.children[0](this.state.value);
        }
    });
    context.Provider = Provider
    context.Consumer = Consumer
    return context;
}
```
util.js

``` javascript
export function extendComponent(obj, prop) {
	function __() { this.constructor = obj; }
	__.prototype = Component.prototype
	obj.prototype = new __();
	extend(obj.prototype, prop)
}
```

test/test.html
``` html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <script src="../dist/preact.js"></script>
    <!-- <script src="../node_modules/preact/dist/preact.dev.js"></script> -->
    <title>Document</title>
</head>
<body>
    <div id="app">
    </div>
    <script>
        'use strict';
        var h = preact.h;
        var render = preact.render;
        var Component = preact.Component;
        const ThemeContext = preact.createContext({
            background: 'red',
            color: 'white'
        });
        const ContextMap = {
            1: { background: 'red', color: 'white' },
            2: { background: 'green', color: 'white' }
        }
        class App extends Component {
            constructor() {
                super()
                this.state = {
                    map: 1
                }
            }
            select() {
                this.setState({
                    map: this.state.map === 1 ? 2: 1
                })
            }
            render () {
                return (
                    h("div", null,
                        h(ThemeContext.Provider, {value: ContextMap[this.state.map === 1 ? 2: 1]},
                            h(Header, {map: this.state.map})
                        ),
                        h(ThemeContext.Provider, {value: ContextMap[this.state.map]},
                            h(Header, {map: this.state.map === 1 ? 2: 1})
                        ),
                        h("button", {onClick: () => this.select()}, "测试")
                    )
                    
                );
            }
        }
        class Header extends Component {
            // shouldComponentUpdate() {
            //     return false
            // }
            render () {
                return h(Title, this.props, "Hello React Context API")
            }
        }
        class Title extends Component {
            render () {
                return h(ThemeContext.Consumer, null,
                    context => h(this.props.map === 1 ? "h1": "h2", {style: {background: context.background, color: context.color}},
                        this.props.children
                    )
                )
            }
        }
        render(
            h(App), 
            document.getElementById('app')
        );

    </script>
    <script src="../devtools.js"></script>
</body>
</html>
```

developit:
Hi @jtmthf & @zeromake!

What would you two think of combining these and publishing it as a dedicated module?  That way we can link to it from Preact core.  Seem reasonable?

zeromake:
@developit I think we can, will non-invasive react new api publishing it as a dedicated module.

valotas:
What is the status on this? Since I would love to make use of it, I do have an implementation [here](https://github.com/valotas/preact-context). It is in typescript and lacks some tests.

@developit  Since I mostly used the code here and wouldn't like to hijack anything, I am ok on moving my repository anywhere that would make sense and also giving ownership to either @jtmthf or @zeromake.

All I want is to make use of this feature :)

valotas:
Having a closer look at it. If we do not always need to use `Consumer` "inside" a `Provider`, the [implementation](https://github.com/valotas/preact-context/blob/contextless/src/context.ts) can be simplified and doesn't depend on preact's context at all. From the [rfcs](https://github.com/acdlite/rfcs/blob/new-version-of-context/text/0000-new-version-of-context.md):

```
If a consumer is rendered without a matching provider as its ancestor, it receives the default value passed to createContext, ensuring type safety.
```

it looks like that this should be the case.

@developit maybe it makes more sense to ditch context in favor of the new api.

kontrollanten:
Thanks for your work! Are there any news on this?

gerhardsletten:
This can be used by overriding preact-compat in webpack-alias:

```js
resolve: {
  alias: {
    'react': path.resolve(__dirname, '../src/helpers/preact-compat'),
    'react-dom': 'preact-compat'
  }
}
```

And in your `preact-compat.js` override:

```
import {
  version,
  DOM,
  PropTypes,
  Children,
  render,
  createClass,
  createFactory,
  createElement,
  cloneElement,
  isValidElement,
  findDOMNode,
  unmountComponentAtNode,
  Component,
  PureComponent,
  unstable_renderSubtreeIntoContainer,
  __spread
} from 'preact-compat'
import { createContext } from 'preact-context'

export {
  version,
  DOM,
  PropTypes,
  Children,
  render,
  createClass,
  createFactory,
  createElement,
  cloneElement,
  isValidElement,
  findDOMNode,
  unmountComponentAtNode,
  Component,
  PureComponent,
  unstable_renderSubtreeIntoContainer,
  __spread,
  createContext // <-- Add it here
}
export default {
  version,
  DOM,
  PropTypes,
  Children,
  render,
  createClass,
  createFactory,
  createElement,
  cloneElement,
  isValidElement,
  findDOMNode,
  unmountComponentAtNode,
  Component,
  PureComponent,
  unstable_renderSubtreeIntoContainer,
  __spread,
  createContext // <-- And here
}
```

marvinhagemeister:
Good news: `createContext` is part of Preact X :tada:

jackbravo:
So, is the new context API already part of preact? https://github.com/valotas/preact-context is no longer needed?

marvinhagemeister:
yup

PR #971, opened 2018-01-04
Changing the README to match the website

darol100:
##  Problem
In the website said, "Fast 3kB alternative to React, with the same ES6 API."

However, in the README said, "Fast 3kB alternative to React, with the same ES2015 API."

## Solution
Changing the README text, so it matches the website.


developit:
We're going to adopt the new tagline across all sites:
"Fast 3kB alternative to React with the same modern API."

PR #974, opened 2018-01-08
fix: preact-compat unstable_renderSubtreeIntoContainer throw error 

paranoidjk:
## Reproduce
```bash
git clone git@github.com:ant-design/ant-design-mobile.git
cd ant-design-mobile
npm install
npm start
open localhost:8002/components/input-item
// or open http://localhost:8002/components/picker
```

preact-compat unstable_renderSubtreeIntoContainer throw error when use preact/devtools

close https://github.com/ant-design/ant-design-mobile/issues/2139

### DIFF
--- a/devtools/devtools.js
+++ b/devtools/devtools.js
@@ -248,7 +248,7 @@ function createDevToolsBridge() {
 		const prevRenderedChildren = [];
 		let instance = instanceMap.get(component);
 		if (instance) {
-			visitNonCompositeChildren(instanceMap.get(component), childInst => {
+			visitNonCompositeChildren(instance, childInst => {
 				prevRenderedChildren.push(childInst);
 			});
 		}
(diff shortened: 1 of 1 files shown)


developit:
I think the failing test is correct. Perhaps this would work?

```js
	/** Notify devtools that a component has been updated with new props/state. */
	const componentUpdated = component => {
		const prevRenderedChildren = [];
		let instance = instanceMap.get(component)
		if (instance) {
			visitNonCompositeChildren(instance, childInst => {
				prevRenderedChildren.push(childInst);
			});
		}

		// Notify devtools about updates to this component and any non-composite
		// children
		const instance = updateReactComponent(component);

		Reconciler.receiveComponent(instance);
		visitNonCompositeChildren(instance, childInst => {
```

paranoidjk:
@developit  You are right 👍 CI is pass.

developit:
This will go into 9.0 :)

PR #978, opened 2018-01-13
Add option to support NativeScript event handling

Hizoul:
In the exploration of using [preact with nativescript](https://github.com/Hizoul/preact-to-nativescript) I had to [adjust eventHandling](https://github.com/staydecent/nativescript-preact/issues/4#issuecomment-355989276) to get events working.

In this pull request I added a boolean called `nativeScript` in the `options`. I also added a test to cover the correct behaviour if said option is truthy.

I am unsure wether:
1. platform specific code like this should even be in preact
2.  this way to solve this is mergeworthy but it seemed like the most uninitrusive solution for exisiting behaviour to me

### DIFF
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -63,8 +63,11 @@ export function setAccessor(node, name, old, value, isSvg) {
 		if (value) node.innerHTML = value.__html || '';
 	}
 	else if (name[0]=='o' && name[1]=='n') {
-		let useCapture = name !== (name=name.replace(/Capture$/, ''));
-		name = name.toLowerCase().substring(2);
+		// In NativeScript the third addEventListener Parameter is a self reference https://docs.nativescript.org/api-reference/classes/_data_observable_.observable#addeventlistener
+		let useCapture = options.nativeScript ? node : name !== (name=name.replace(/Capture$/, ''));
+		// in NativeScript Events are case sensitive hence we only lowercase the first character of the eventName
+		// (e.g. onTextChanged would become textchanged but needs to be textChanged)
+		name = options.nativeScript ? name.substring(2, 3).toLowerCase() + name.substring(3) : name.toLowerCase().substring(2);
 		if (value) {
 			if (!old) node.addEventListener(name, eventProxy, useCapture);
 		}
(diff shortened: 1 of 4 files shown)


developit:
Hmm - I don't think it's a good idea to merge something specific to NativeScript into Preact itself. Preact is specifically targeted at rendering DOM.

I'd suggest it might be better to instead provide a library that patches Preact to work in NativeScript.  It'd look something like this:

```js

let old = options.vnode;
options.vnode = vnode => {
  // only mutate element nodes (not components)
  if (typeof vnode.nodeName==='string' && vnode.attributes) {
    for (let i in vnode.attributes) {
      if (i[0]==='o' && i[1]==='n') {
        vnode.attributes[i].toLowerCase = unLowercase;
      }
    }
  }
  if (old) old(vnode);
}

let oldE = options.event;
options.event = e => {
  e.type = e.type || e.eventName;
  return oldE && oldE(e) || e;
}

function unLowercase() {
  return this[0].toLowerCase() + this.substring(1);
}
```

Hizoul:
Thank you very much for the quick reply and suggestion. I believe not touching preact at all would be cleaner and should be the way to go as expressed in my first question.

Unfortunately, after giving your suggestion a go, I fear that this wont be fully possible.

While the suggested code does eliminate the adjustment to the [name = options.nativeScript line](https://github.com/developit/preact/pull/978/files#diff-4935bb00e75b7854383877cbfd9d770fR70) there still would be code changes necessary within preact to make it work.
1. eventProxy will still throw an NPE. options.event is only called after the listener has been looked up at e.type of the original event. Hence it would have to be overriden first like this (which i tested and does work if 2. is fullfilled as well)
```javascript
    function eventProxy(e) {
        e = options.event && options.event(e) || e
        return this.__l[e.type](e);
    }
```
2. If the third argument on the addEventListener-Call is not adjusted then the this-Context in the eventProxy-Method is not correct and leads to a null pointer exception trying to access this.__l. That is why in the pull-request I [overwrote the useCapture argument](https://github.com/developit/preact/pull/978/files#diff-4935bb00e75b7854383877cbfd9d770fR67) with node if options.nativeScript = true.

marvinhagemeister:
Thank you for your PR 👍 As @developit said, we're only focusing on the DOM for now.

PR #982, opened 2018-01-20
fix cloneElement test

decadef20:
The `clone` and `instance` are both VNode instances ,why to test prototype property here?
Maybe we should test whether clone.nodeName.prototype equals to instance.nodeName.prototype ,or I missed something?


developit:
Hmm - I think the `expect()` there is actually just verifying that both are `VNode` instances. A constructor check would seem much more appropriate.

decadef20:
Great. I also think so.

Issue #945, opened 2017-11-27
Boolean attributes are treated inconsistently


call-a3:
When setting attributes to boolean values, preact interprets falsey values as "they don't need to be set". This makes sense and is good behaviour.

However, I would similarly expect that preact would also interpret `true` as the opposite "it simply needs to be set". It does do this, but it renders the attribute with value `"true"`. I would expect that it would be rendered with value `""` (empty string), which fits better with HTML's default syntax for boolean attributes.

For an example of what I mean: https://codesandbox.io/s/llq38jyrzz

Related to #83

marvinhagemeister:
I checked against Preact X in master and we're correctly setting `boolean` values 🎉

RobinMalfait:
👌

PR #983, opened 2018-01-21
[945] Update boolean attributes to be empty value for consistency

timurista:
## Change boolean attributes to have empty value

In response to #945 
I added an explicit check in case the value === "true" which simply sets attribute to "" which is inline with web standard in moz docs: https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute

> Boolean attributes are considered to be true if they're present on the element at all, regardless of their actual value; as a rule, you should specify the empty string ("") in value (some people use the attribute's name; this works but is non-standard). See the example below for a practical demonstration.

I also updated the test to make sure it renders empt value instead of ="true"

### DIFF
--- a/devtools/devtools.js
+++ b/devtools/devtools.js
@@ -248,7 +248,7 @@ function createDevToolsBridge() {
 		const prevRenderedChildren = [];
 		let instance = instanceMap.get(component);
 		if (instance) {
-			visitNonCompositeChildren(instanceMap.get(component), childInst => {
+			visitNonCompositeChildren(instance, childInst => {
 				prevRenderedChildren.push(childInst);
 			});
 		}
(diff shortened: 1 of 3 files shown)


timurista:
@developit did you want me to still merge this? Which part still needs to be compressed? Are we talking the elseif statement or something else?

developit:
Not sure yet - let's leave it open for a bit, we won't be cutting a release until next week.

timurista:
sure, np 👍  sounds good to me

timurista:
@developit any news on the perf-review?

developit:
I'm concerned that this may cause issues with some DOM Boolean Attributes, but haven't been able to determine if that's likely or not.

timurista:
@developit you want me to close this then?

developit:
no - I want to write some tests around it and was hoping to have time to research the effects. I actually think we should merge this, I just need to be sure of the implications.

timurista:
@developit this is very out of date, you want me to close it?

PR #988, opened 2018-01-24
Should fix developit/preact-router#198

developit:
This change should fix the preact-router issue here:
https://github.com/developit/preact-router/issues/198


larrybotha:
Any chance this fix can be published?

Issue #967, opened 2017-12-30
[bug?] Why props are optional in FunctionalComponent?


alexeypetrushin:
The TypeScript typing for functional component have optional argument `props?`, why? 

```
interface FunctionalComponent<PropsType> {
  (props?:PropsType & ComponentProps<this>, context?:any):JSX.Element;
  displayName?:string;
  defaultProps?:any;
}
```

It should be `props:PropsType` not `props?:PropsType`.

P.S.

Why it's a bug? Because if that would be true this example from preact docs would be incorrect and could blow up if `props` undefined

```
const Link = ({ children, ...props }) => (
    <a {...props}>{ children }</a>
);
```

According to the typings it should be written as

```
const Link = (props) => (
    if (props) {
        const { children, ...props } = props
        return <a {...props}>{ children }</a>
    } else {
        return <span>props undefined...</span>
    }
);
```

developit:
True! This won't complain if the function doesn't use the props argument, I think it was just a misinterpretation.

PR #990, opened 2018-01-27
make props required in functional component #967

timurista:
this is a fix for #967 based on feedback that props should be required for a functional component


andrewiggins:
Wooo! @timurista thanks for doing this. Was just about to submit a PR myself :)

PR #1005, opened 2018-02-14
Specify types directive in package.json for TypeScript

jrf0110:
This allows the TS language service to automatically pull in types for Preact. This benefits non-TypeScript users as well. If you're using VSCode and just plain vanilla JS, it will pickup Preact's types automatically after installing and provide autocompletion


developit:
Beauty! I thought that was already in there.

PR #1015, opened 2018-02-24
Fix(TS): Declare preact/debug module

ovr:
Hey!

Example of usage:

image

Actual error (before fix):

```
src/components/app.tsx(10,12): error TS7016: Could not find a declaration file for module 'preact/debug'. '/Users/ovr/projects/startup/notify/frontend/node_modules/preact/debug.js' implicitly has an 'any' type.
  Try `npm install @types/preact` if it exists or add a new declaration (.d.ts) file containing `declare module 'preact';`
```

Expected:

works :)

Thanks


marvinhagemeister:
@ovr That's awesome! Thanks a lot for your contribution 🎉

bz2:
@marvinhagemeister This seems to have regressed in current preact releases, or something else needs settting up? There's an `/preact/debug/src/internal.d.ts` but not a (maybe empty) public type export?

PR #1016, opened 2018-02-25
Set component instance constructor to make it work nicely with react-hot-loader

arqex:
Fixes https://github.com/developit/preact/issues/1007


developit:
Need to check if there are any performance issues with this technique

BlakeTurner:
@developit any update? I'm making the switch to preact (saves me 44kb gzipped!) and this is one of the final hurdles for me.

developit:
I think we are good to give this a try. Tests are passing and I doubt there'll be a massive performance impact.

reznord:
How about releasing a RC version before we directly go to main release?

marvinhagemeister:
@developit Agree, doesn't seem like this would affect performance.

@reznord Perhaps a simple patch release is enough? Just went through the commit history and it only contains bugfixes and updates to our `README`.

developit:
Ok - did a quick perf test. There is impact - not sure how we proceed:

**Native ES Classes:**


**Babel transpiled classes:**


Bench: https://esbench.com/bench/5abb9bd7f2949800a0f619c1

marvinhagemeister:
Oh no. I merged too quickly I guess :/

PR #1017, opened 2018-02-26
remove unused third jsdoc param

k1r0s:


developit:
I wonder if there's a commonly accepted way of indicating rest params in JSDoc (3)?

k1r0s:
since this is a optional argument I guess you can use ts notation

```
/**
 * Clones the given VNode, optionally adding attributes/props and replacing its children.
 * @param {VNode} vnode	The virtual DOM element to clone
 * @param {Object} props	Attributes/props to add when cloning
 * @param {VNode[]} ...rest:	Any additional arguments will be used as replacement children.
 */
export function cloneElement(vnode, props) {
```

or 

use a commonly accepted way to represent aditional parameters

```
/**
 * Clones the given VNode, optionally adding attributes/props and replacing its children.
 * @param {VNode} vnode	The virtual DOM element to clone
 * @param {Object} props	Attributes/props to add when cloning
 * @param {VNode[]} [, rest]:	Any additional arguments will be used as replacement children.
 */
export function cloneElement(vnode, props) {
```
I guess everybody will understand both.

k1r0s:
sry, just useless observation xd

PR #886, opened 2017-09-20
Add support for componentDidCatch Component method

adamsilverstein:
Amends https://github.com/developit/preact/pull/819 removing these [changes](https://github.com/developit/preact/pull/819/files/b616cc88c012e57a89c5aa4d5afcf567183c0593#diff-86702919356dad82cf58b1e4150288f5L161)  - tests still pass.

PR #1018, opened 2018-02-27
fix failing render side-effects

PhilippMi:
should fix developit/preact#958


src/vdom/diff.js
```diff
@@ -43,15 +43,19 @@ export function diff(dom, vnode, context, mountAll, parent, componentRoot) {
...
-
-	// append the element if its a new parent
-	if (parent && ret.parentNode!==parent) parent.appendChild(ret);
+	let ret;
+	try {
+		ret = idiff(dom, vnode, context, mountAll, componentRoot);
+
+		// append the element if its a new parent
+		if (parent && ret.parentNode !== parent) parent.appendChild(ret);
+	} finally {
```
developit: Are the additional changes here needed, or just the try/finally?

PhilippMi: What do you mean by additional changes? I need to adjust `diffLevel` and `hydrating` in the finally block so that it will be executed even when an error is thrown. Aside from that, there are no additional changes.

k1r0s: Guess `ret` should be initialized though, to prevent other bugs

PhilippMi: For clarification:
For this change, there are 2 important scenarios how the code will execute:
1. The `render` works as expected and the code is executed as before. `diffLevel` and `hydrating` are calculated in a different spot but the result is the same.
2. The `render` throws an error, which leads to `idiff` throwing an error which leads to this function aborting execution and not executing anything below line 48 - except for the finally block. This is necessary to get preact into a sane state again before leaving the function.

So I don't think it is necessary to initialise `ret`. It will be initialised in the next line and if `idiff` throws an error then  `ret` won't be used anymore.

k1r0s: noice

marvinhagemeister:
@developit Since this changes how preact handles errors during render, I let you decide if this should be merged or not.

PhilippMi:
> Since this changes how preact handles errors during render, I let you decide if this should be merged or not.

The change does not change the error handling. It still lets errors through.

If you are worried about the performance implications of the `try`/`finally`, there are other solutions to this problem. For instance, you could always set `diffLevel` and `hydrating` to their initial value before starting the recursive diffing.

marvinhagemeister:
@PhilippMi No, try/finally is perfect and much clearer. That was just a remark on past browser behaviours.

Just noticed that I read the `try/finally` as `try/catch`. So my comment about error handling makes no sense. Sorry for the confusion 😄

developit:
It seems like this should get combined with #886 (componentDidCatch implementation).  Anyone else care to confirm?  To me, if we're going to start catching render errors, it seems worth providing a way to hook them at the same time.

PhilippMi:
Fixed with 10.0.0

PR #1027, opened 2018-03-05
Make render node arg nullable

### LINKED ISSUE #996: TypeScript: `render`'s first arg needs to be nullable
In order to remove an element, I need to be able to do `render(null, parentEl, elToRemove)`. In the [typings file](https://github.com/developit/preact/blob/master/src/preact.d.ts#L79), `render`’s first argument is typed `Element`, which is non-nullable, precluding passing in `null`. This is preventing me from removing elements from a TypeScript project. The [source](https://github.com/developit/preact/blob/master/src/vdom/diff.js#L69) suggests that this argument can be null, undefined, a boolean (?), a string or a number.

~~(I haven't gotten this far yet, but I assume that if `render` returns the rendered element, and I'm removing things, it an also return `null`, which means the function’s return type also needs to be nullable.)~~ (Sounds like no problem on the return type.)

wuweiweiwu:
Fixes #996 

- [x] test that passing in `null` will render an empty text node

Hopefully my first contribution! :)

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -98,7 +98,7 @@ declare namespace preact {
 
 	function h<PropsType>(node: ComponentFactory<PropsType>, params: PropsType, ...children: (ComponentChild | ComponentChildren)[]): JSX.Element;
 	function h(node: string, params: JSX.HTMLAttributes & JSX.SVGAttributes & { [propName: string]: any }, ...children: (ComponentChild | ComponentChildren)[]): JSX.Element;
-	function render(node: JSX.Element, parent: Element | Document, mergeWith?: Element): Element;
+	function render(node: ComponentChild, parent: Element | Document, mergeWith?: Element): Element;
 	function rerender(): void;
 	function cloneElement(element: JSX.Element, props: any): JSX.Element;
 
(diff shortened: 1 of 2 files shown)


src/preact.d.ts
```diff
@@ -98,7 +98,7 @@ declare namespace preact {
 
 	function h<PropsType>(node: ComponentFactory<PropsType>, params: PropsType, ...children: (ComponentChild | ComponentChildren)[]): JSX.Element;
 	function h(node: string, params: JSX.HTMLAttributes & JSX.SVGAttributes & { [propName: string]: any }, ...children: (ComponentChild | ComponentChildren)[]): JSX.Element;
-	function render(node: JSX.Element, parent: Element | Document, mergeWith?: Element): Element;
+	function render(node: JSX.Element | null, parent: Element | Document, mergeWith?: Element): Element;
```
marvinhagemeister: We just recently merged a PR that introduces the `ComponentChild` type:

wuweiweiwu: @marvinhagemeister Awesome! I didn't realize render also took in numbers. 

Just updated the PR!

marvinhagemeister:
@wuweiweiwu Merged 🎉 Thanks a lot for your contribution 👍

wuweiweiwu:
No problem :) I'll pick up some other issues as well!

PR #1031, opened 2018-03-11
Fix wrong source mapping in tests for src/

marvinhagemeister:
Noticed that although breakpoints are correctly hit, the line mapping was completely off.

This is caused by additional code created by coverage reporting. Tried replacing the now deprecated `isparta-loader` with `istanbul-instrumenter-loader` to see if this fixes things, but I couldn't get the latter to work.

### Before
screen shot 2018-03-11 at 11 51 52

### After

screen shot 2018-03-11 at 11 53 13

Launch settings for vscode used for the above screenshot (run `test:karma:watch` first):

```json
{
  "version": "0.2.0",
  "configurations": [
    {
      "type": "chrome",
      "request": "attach",
      "name": "Attach Karma Chrome",
      "address": "localhost",
      "port": 9333,
      "pathMapping": {
        "/": "${workspaceRoot}",
        "/base/": "${workspaceRoot}/"
      }
    }
  ]
}
```

I do think that the debugger working by default is more useful for new contributors than coverage enabled. But this may be just my personal preference. Coverage is of course still collected on our CI. This PR only affects the local dev environment. 

### Changes:

- [x] invert `COVERAGE` feature flag to only be enabled when explicitly set to `true`. Before it was always enabled and only disabled when set to `false`
- [x] ignore common editor specific configuration (`.vscode/` + `.idea` for now)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,7 @@ package-lock.json
 /devtools.js.map
 /debug.js
 /debug.js.map
+
+# Editor and IDE configuration
+.vscode/
+.idea/
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,6 @@ install:
 
 script:
   - npm run build
-  - npm run test
+  - COVERAGE=true npm run test
   - COVERAGE=false FLAKEY=false PERFORMANCE=false npm run test:karma
   - ./node_modules/coveralls/bin/coveralls.js < ./coverage/lcov.info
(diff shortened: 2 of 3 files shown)


k1r0s:
If I understand properly this change only affects developers will not see coverage report locally before pushing to remote, right?

marvinhagemeister:
@k1r0s yep 👍

PR #1019, opened 2018-02-28
update dependencies and tests to latest

hassanbazzi:
Just part of the changes needed. WIP.

PR #1032, opened 2018-03-11
Upgrade dependencies

marvinhagemeister:
This PR replaces #1019, which seems to be abandoned.

Changes:

- [x] upgrade all dependencies to the newest versions
- [x] fix linting errors
- [x] sinon [renamed](https://github.com/sinonjs/sinon/issues/1390) `reset()` to `resetHistory()`
- [x] Fix invalid `expect(...).to.have.deep.property(...)` checks. Seems like chai is broken here. imo there is no need for chai here anyways, just access the property directly
- [x] Fix broken test in `devtools.js`.

The last one is quite curious as the test assertion implies that it checks for `callcount === 2` by using `.twice`. But because `.twice` is actually `undefined` nothing is asserted. A simply breakpoint just before that prints a call count of `5`, even in master.

Seems like this is a bug which we need to fix (out of scope for this PR though, because it's already wrong in master)


test/browser/devtools.js
```diff
@@ -193,7 +193,7 @@ describe_('React Developer Tools integration', () => {
 		node._component.setState({count: 4});
 		node._component.forceUpdate();
 
-		expect(renderer.Reconciler.mountComponent).to.have.been.called.twice;
+		expect(renderer.Reconciler.mountComponent).to.have.been.called.callCount(5);
```
marvinhagemeister: As mentioned: `.twice` was `undefined` in master. That's why nothing was checked here. The actual call count is 5. This likely is a bug.

marvinhagemeister:
ping @developit @k1r0s Rebased and upgraded against latest dependencies. There were some additional api changes for `uglify` and `webpack`

config/properties.json
```diff
@@ -1,27 +1 @@
...
-      "$__preactattr_": "__preactattr_"
-    }
-  },
-  "vars": {
-    "cname": -1,
-    "props": {}
-  }
-}
\ No newline at end of file
+{"props":{"props":{"$_dirty":"__d","$_disable":"__x","$_listeners":"__l","$_renderCallbacks":"__h","$__key":"__k","$__ref":"__r","$normalizedNodeName":"__n","$nextBase":"__b","$prevContext":"__c","$prevProps":"__p","$prevState":"__s","$_parentComponent":"__u","$_componentConstructor":"_componentConstructor","$__html":"__html","$_component":"_component","$__preactattr_":"__preactattr_","$vnode":"i","$debounceRendering":"t","$defaultProps":"l","$afterMount":"o","$componentDidMount":"u","$base":"s","$dangerouslySetInnerHTML":"v","$ref":"h","$render":"p","$componentWillMount":"g","$componentWillReceiveProps":"m","$props":"k","$syncComponentUpdates":"S","$shouldComponentUpdate":"P","$componentWillUpdate":"T","$getChildContext":"_","$componentDidUpdate":"j","$afterUpdate":"C","$beforeUnmount":"H","$componentWillUnmount":"L","$setState":"M","$forceUpdate":"$","$h":"I","$cloneElement":"N","$Component":"O","$rerender":"U","$exports":"q","$preact":"A"}},"vars":{"props":{}}}
```
marvinhagemeister: I was confused by the duplicate `props` key, but this file is autogenerated that way

k1r0s: do you've any idea about why this doens't keep previous format?

src/dom/index.js
```diff
@@ -78,7 +78,7 @@ export function setAccessor(node, name, old, value, isSvg) {
 		if (value==null || value===false) node.removeAttribute(name);
 	}
 	else {
-		let ns = isSvg && (name !== (name = name.replace(/^xlink\:?/, '')));
+		let ns = isSvg && (name !== (name = name.replace(/^xlink:?/, '')));
```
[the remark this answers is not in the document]
marvinhagemeister: Changed this because eslint detected this as an unnecessary escape character:

test/browser/devtools.js
k1r0s: @marvinhagemeister I'd prefer to skip this test but keeping desired output rather than hardcoding the actual result..

src/dom/index.js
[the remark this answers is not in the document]
k1r0s: its okay, next time would be nice to have lint rules on a separate commit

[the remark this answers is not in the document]
marvinhagemeister: That's a great suggestion. Just rebased and extracted this change into its own commit.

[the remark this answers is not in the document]
k1r0s: N O I C E :+1:

config/properties.json
marvinhagemeister: Unfortunately not. The build output seems to be unaffected though, as far as I can tell.

test/browser/devtools.js
marvinhagemeister: @k1r0s You're right, that's the better way to go. Pushed a new commit that skips this test and adds a todo comment.

developit:
Awesome work. We'll need to look at that devtools test, I think there might even be an open bug relating to it.

Issue #1030, opened 2018-03-07
Type-hinting functional components?


mindplay-dk:
I'm trying to write a simple function-wrapper around the `<button>` element.

According to the [test-suite](https://github.com/developit/preact/blob/master/test/ts/preact.tsx#L35):

```ts
const ComponentWithChildren: FunctionalComponent<DummerComponentProps> = ({ input, initialInput, children }) => {
	return <div>
		<span>{initialInput}</span>
		<span>{input}</span>
		<span>{children}</span>
	</div>
}
```

Am I missing something, or why does this approach not work for me at all?

```ts
import { h, FunctionalComponent } from "preact";

export interface ButtonProps {
    title: string;
}

export const Button: FunctionalComponent<ButtonProps> = ({ title, children, ...props }) => {
    return <button {...props}>{title}</button>;
};
```

I get a bunch of errors:

```
Type '(ButtonProps & ComponentProps<FunctionalComponent<ButtonProps>>) | undefined' has no property 'title' and no string index signature.

Type '(ButtonProps & ComponentProps<FunctionalComponent<ButtonProps>>) | undefined' has no property 'children' and no string index signature.

Type '{ key?: any; ref?: ((el: FunctionalComponent<ButtonProps>) => void) | undefined; }' is not assignable to type 'HTMLAttributes'.
  Types of property 'ref' are incompatible.
    Type '((el: FunctionalComponent<ButtonProps>) => void) | undefined' is not assignable to type '((el?: Element | undefined) => void) | undefined'.
      Type '(el: FunctionalComponent<ButtonProps>) => void' is not assignable to type '((el?: Element | undefined) => void) | undefined'.
        Type '(el: FunctionalComponent<ButtonProps>) => void' is not assignable to type '(el?: Element | undefined) => void'.
          Types of parameters 'el' and 'el' are incompatible.
            Type 'Element | undefined' is not assignable to type 'FunctionalComponent<ButtonProps>'.
              Type 'undefined' is not assignable to type 'FunctionalComponent<ButtonProps>'.
```

I tried copy/pasting the entire contents of the test, and it gives similar errors.

I looked at the [last successful build](https://travis-ci.org/developit/preact/builds/348865971) from 3 days ago, and the [tests](https://github.com/developit/preact/blob/d0a2735ac1a04064a3a3f0ee6fb8ca0b1ed943ca/test/ts/preact.tsx) didn't look any different.

[...]

PR #1033, opened 2018-03-11
Add support for strict mode in TypeScript

marvinhagemeister:
As mentioned in #1030 we lacked support for TypeScripts `strict` mode which is as the name implies less forgiving on mismatched types.

Changes:

- [x] Enable `strict` mode in `tsconfig.json`
- [x] Make props and state generic optional in `Component` interface
- [x] Fix issues related to `strict` mode

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -54,50 +54,61 @@ declare namespace preact {
 
 	type RenderableProps<P> = Readonly<P> & Readonly<{ children?: ComponentChildren }>;
 
-	interface FunctionalComponent<PropsType> {
-		(props: RenderableProps<PropsType>, context?: any): VNode<any>;
+	interface FunctionalComponent<P = {}> {
+		(props: RenderableProps<P>, context?: any): VNode<any> | null;
 		displayName?: string;
-		defaultProps?: any;
+		defaultProps?: Partial<P>;
 	}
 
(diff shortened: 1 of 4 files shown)


mindplay-dk:
Thanks for jumping on this! 😄 

Why did you change the type parameter to single letters?

IMO that works when you have a singe type `T` and it's use/meaning is obvious from the generic class' name - hiding the meaning of the type-arguments in this case doesn't really help with readability or comprehension.

If I may, I'd suggest using the `T` prefix convention, e.g. `TProps`, `TState`, which is come conventional when you have more than one type argument.

Or just leave it as-is. (Which may actually make more sense in the context of this PR, as it's largely unrelated to the actual changes you're proposing.)

marvinhagemeister:
That was more to align with react typings. They use single characters for generics throughout the whole codebase. In my opinion this makes it easier to read as well, although opinions on that obviously differ.

marvinhagemeister:
TGFM = Total Gay Frat Move (according to the [urban dictionary](https://www.urbandictionary.com/define.php?term=TGFM)) 🤔 😂

k1r0s:
LOOOO

totally messed xd

Looks good for me

image

marvinhagemeister:
@k1r0s No worries, had a good laugh. That's always a good thing to start the week with :)

PR #1035, opened 2018-03-15
Refactors

Artikash:
Made a few small refactors

### DIFF
--- a/src/clone-element.js
+++ b/src/clone-element.js
@@ -1,4 +1,3 @@
-import { extend } from './util';
 import { h } from './h';
 
 /**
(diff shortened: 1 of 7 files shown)


yaodingyd:
Thanks for contributing. IMO this doesn't benefit preact much though, as preact tries to use ES6 syntax as less as possible to minimize transpiled file size.

Artikash:
Ah, fair enough.

marvinhagemeister:
@yaodingyd Agree, although I'd love to see more experiments with ES2015. Assuming that there are more and more projects out there which do not have to support old browser, we should explore what size benefits a ES2015-only version of preact would give us.

My guess is that it would be even smaller than what we have today. This would be great for more modern projects which only need to support ES2015+ browsers.

PR #1037, opened 2018-03-16
super important PR

k1r0s:
We should allow any serializable value to be asigned to `value` html prop 

fix:
```
whatever is not assignable to type 'HTMLAttributes'. Types of property 'value' are incompatible. Type 'number' is not assignable to type 'string | string[]'.	
```


marvinhagemeister:
Oh well, just noticed that the commit message could be a bit more descriptive. We should aim for more clear commit logs in the future

PR #1051, opened 2018-03-30
Add back missing attributes types to RenderableProps

andrewiggins:
The existing types don't allow a component to add Preact common props that any component can specify (e.g. `key`). I've updated the types to add back these common props to `RenderableProps`.

I also changed some types from `type` to `interface` to allow other libraries to extend them with their own common props (e.g. preact-router adds `path`).

See the tests I added for an example of the code that previously didn't work.

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -10,16 +10,18 @@ declare namespace preact {
 	/**
 	 * @deprecated
 	 *
-	 * Use Attributes instead
+	 * Use Attributes instead. This type must remain an interface to not break
+	 * other libraries' types (e.g. preact-router)
 	 */
-	type ComponentProps = Attributes;
+	interface ComponentProps extends Attributes {}
 
 	/**
 	 * @deprecated
 	 *
-	 * Use ClassAttributes instead
+	 * Use ClassAttributes instead. This type must remain an interface to not break
+	 * other libraries' types (e.g. preact-router)
 	 */
-	type PreactHTMLAttributes = ClassAttributes<any>;
+	interface PreactHTMLAttributes extends ClassAttributes<any> {}
 
 	interface Attributes {
 		key?: string | number | any;
@@ -52,7 +54,8 @@ declare namespace preact {
 		key?: Key | null;
 	}
 
-	type RenderableProps<P> = Readonly<P> & Readonly<{ children?: ComponentChildren }>;
+	type RenderableProps<P> = Readonly<P> & Readonly<Attributes> & Readonly<{ children?: ComponentChildren }>;
+	type ClassRenderableProps<P> = RenderableProps<P> & Readonly<ClassAttributes<any>>;
 
 	interface FunctionalComponent<P = {}> {
 		(props: RenderableProps<P>, context?: any): VNode<any> | null;
@@ -87,7 +90,7 @@ declare namespace preact {
 		static defaultProps?: any;
 
 		state: Readonly<S>;
-		props: RenderableProps<P>;
+		props: ClassRenderableProps<P>;
 		context: any;
 		base?: HTMLElement;
 
@@ -96,7 +99,7 @@ declare namespace preact {
 
 		forceUpdate(callback?: () => void): void;
 
-		abstract render(props?: RenderableProps<P>, state?: Readonly<S>, context?: any): JSX.Element | null;
+		abstract render(props?: ClassRenderableProps<P>, state?: Readonly<S>, context?: any): JSX.Element | null;
 	}
 
 	function h<P>(
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -10,18 +10,16 @@ declare namespace preact {
 	/**
 	 * @deprecated
 	 *
-	 * Use Attributes instead. This type must remain an interface to not break
-	 * other libraries' types (e.g. preact-router)
+	 * Use Attributes instead.
 	 */
-	interface ComponentProps extends Attributes {}
+	type ComponentProps = Attributes;
 
 	/**
 	 * @deprecated
 	 *
-	 * Use ClassAttributes instead. This type must remain an interface to not break
-	 * other libraries' types (e.g. preact-router)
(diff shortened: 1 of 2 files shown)


src/preact.d.ts
```diff
@@ -52,7 +54,8 @@ declare namespace preact {
 		key?: Key | null;
 	}
 
-	type RenderableProps<P> = Readonly<P> & Readonly<{ children?: ComponentChildren }>;
+	type RenderableProps<P> = Readonly<P> & Readonly<Attributes> & Readonly<{ children?: ComponentChildren }>;
```
marvinhagemeister: ```Readonly<P & Attributes & { children?: ComponentChildren }>;```

src/preact.d.ts
```diff
@@ -96,7 +99,7 @@ declare namespace preact {
 
 		forceUpdate(callback?: () => void): void;
 
-		abstract render(props?: RenderableProps<P>, state?: Readonly<S>, context?: any): JSX.Element | null;
+		abstract render(props?: ClassRenderableProps<P>, state?: Readonly<S>, context?: any): JSX.Element | null;
```
marvinhagemeister: I'm not sure if this is correct. My understanding is that `html` tags can have `refs`, but not a component per se.

andrewiggins:
@k1r0s of course! :)

I scanned through issues that turned up in searches for "typescript", "key", and "attribute" and didn't find any (did find some other issues I might take a stab at). If you'd like a more complete demonstration of the problem, let me know. Happy to oblige.

test/ts/preact.tsx
```diff
@@ -29,7 +35,7 @@ function DummerComponent({ input, initialInput }: DummerComponentProps) {
 	return <div>Input: {input}, initial: {initialInput}</div>;
 }
 
-render(h(DummerComponent, { initialInput: "The input", input: "New input" }), document);
+render(h(DummerComponent, { initialInput: "The input", input: "New input", key: "1", jsx: false }), document);
```
marvinhagemeister: A quick search doesn't turn up any mention of a special `jsx` attribute. Seems like this is an error in our typings? /cc @developit

andrewiggins: Hmm agreed. It seems it was added in #737. @valotas, do you know what the `jsx` prop defined in the `Attributes` interface is meant to do? Doesn't appear Preact does anything with it. Does TypeScript use it for something?

src/preact.d.ts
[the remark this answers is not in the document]
andrewiggins: Agreed. Updated to the suggested syntax.

src/preact.d.ts
andrewiggins: I think Preact supports refs on class components ([example](https://codesandbox.io/s/52p18zyj4n)). Check the console output of that window. If this isn't supposed to be used however, I'll remove it. Just let me know.

andrewiggins: FYI, the linked codesandbox is just a copy of the relevant components from the preact.tsx test file to demonstrate usage of refs.

src/preact.d.ts
```diff
@@ -10,16 +10,18 @@ declare namespace preact {
...
 
 	/**
 	 * @deprecated
 	 *
-	 * Use ClassAttributes instead
+	 * Use ClassAttributes instead. This type must remain an interface to not break
+	 * other libraries' types (e.g. preact-router)
 	 */
-	type PreactHTMLAttributes = ClassAttributes<any>;
+	interface PreactHTMLAttributes extends ClassAttributes<any> {}
```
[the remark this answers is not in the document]
andrewiggins: Yea, that's fair. Submitted developit/preact-router#274 to update those types to address these breaking changes. I reverted my change to make these an interface.

FYI, once these types are released, any library that tries to extend `PreactHTMLAttributes` or `ComponentProps` will be broken since you can't extend from a `type` declaration. Could be a good forcing function to make other libraries update to the new type names, if you'd like to do that.

andrewiggins:
@marvinhagemeister Thanks for taking the time to look at my PR! I've responded to all your comments. Let me know what you think.

test/ts/preact.tsx
valotas: @matthewmueller [suggested](https://github.com/developit/preact/pull/737#issuecomment-328071941) the `jsx` attribute. I personally never used it, but it looks like some frameworks do use this. It does nothing as fas as I understood but having it defined in the types helps making use of it (when needed). Of course you can always define it in your props.

Since #737 has not been published yet, we could remove it if @developit and/or @marvinhagemeister does not like it.

src/preact.d.ts
```diff
@@ -52,7 +52,8 @@ declare namespace preact {
 		key?: Key | null;
 	}
 
-	type RenderableProps<P> = Readonly<P> & Readonly<{ children?: ComponentChildren }>;
+	type RenderableProps<P> = Readonly<P & Attributes & { children?: ComponentChildren }>;
+	type ClassRenderableProps<P> = RenderableProps<P> & Readonly<ClassAttributes<any>>;
```
valotas: Maybe it makes more sense to have the ref defined right next to the `children` at the `RenderableProps`. That way we do not need the ClassRenderableProps and we have ref working for the SFC. What do you think about that?

src/preact.d.ts
marvinhagemeister: You're right. Thanks for the follow up :)

src/preact.d.ts
marvinhagemeister: @andrewiggins what are your thoughts? Apart from this the pr is ready to be merged 🎉👍

andrewiggins: @marvinhagemeister whoops! Thought I had pushed an update for this, but I forgot to. I've done what @valotas suggested and added a line in the test file to validate that refs work for SFCs. Thanks for following up.

P.S. Congrats on the 8.2.8 release! Looking forward to integrating it into my project :)

andrewiggins:
No worries! Thanks for circling back :)

PR #1063, opened 2018-04-13
Uglifyjs fix

ocboogie:
Note: I'm not quite sure if Preact is the problem.

I came across a problem using [react-slick](https://github.com/akiran/react-slick) and preact. I got a `Uncaught TypeError: Cannot read property 'className' of undefined`, but the odd thing was I only got this error in production mode. So I looked into it and it turns out preact-compat wasn't mutating the `VNode`. And the reason for that was Uglifyjs was not exposing `VNode` anymore. Since `VNode` is an empty class and only used once, in the `h` function, Uglifyjs makes it inline, like `let p = new function() {}();` so preact-compat can't access it.

You can see this issue in action [here](https://github.com/ocboogie/preact-uglyjs-bug).

So my fix(hack) for this issue was setting a property on `VNode` after it was declared, this way Uglifyjs doesn't make it inline. They're very well might be a more elegant way to fix this.

### DIFF
--- a/src/vnode.js
+++ b/src/vnode.js
@@ -1,4 +1,5 @@
 /** Virtual DOM Node */
 export function VNode() {}
 
-VNode.undefined = undefined
+// For more info look at issue #1063
+VNode.undefined = undefined;
--- a/src/vnode.js
+++ b/src/vnode.js
@@ -1,2 +1,4 @@
 /** Virtual DOM Node */
 export function VNode() {}
+
+VNode.undefined = undefined


yaodingyd:
Did you try using preact-compat? I think this line https://github.com/developit/preact-compat/blob/master/src/index.js#L41 covers your issue.

ocboogie:
@yaodingyd Unfortunately that's not the case. Since the `VNode` class is inline in the `h` function, it creates a new class each time `h` is ran. So when preact-compat is mutating the constructor, it only mutates that Instance.

yaodingyd:
@ocboogie Just realize I didn't fully read your first comment. My bad! Will investigate more.

ocboogie:
Just so you know the compression option that's causing problems is the `unused` option. You can see it [here](https://github.com/mishoo/UglifyJS2#compress-options).

yaodingyd:
@ocboogie your example's console log stays the same for development and production build. Can you explain how your code shows the uglifyjs issue?

kaisermann:
What version of uglifyjs are you using? If it's >= 3, turn off the `compress.reduce_funcs` options and see if it works.

yaodingyd:
It looks like setting either `unused` or `reduce_funcs` to false both works.

developit:
I would prefer to solve this by documenting the issue - adding weight and cost to Preact to fix an Uglify bug isn't something we can do.  Anyone want to take a swing at adding it to the docs?

developit:
Anyone know if source position matters to Uglify here?  Also is there an upstream bug we can track to know when Uglify fixes this?

developit:
Hi all! This was fixed in Preact 8.3.0 thanks to @wardpeet's clever trick (#1153).

PR #1071, opened 2018-04-21
Remove object-notation for class/className from TS typings

levrik:
That feature was removed in 8.0.0


k1r0s:
THanks @levrik for your help!

PR #1073, opened 2018-04-24
allow optional arguments even if not used

k1r0s:
when providing a callback many people can just, rather than use an arrow function () => this.whatever(), they can provide the whole function as a callback and that may trigger TS errors since that callback can host arguments.

With this change we allow any kind of callback which IMO is more correct

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -91,10 +91,10 @@ declare namespace preact {
 		context: any;
 		base?: HTMLElement;
 
-		setState<K extends keyof S>(state: Pick<S, K>, callback?: () => void): void;
-		setState<K extends keyof S>(fn: (prevState: S, props: P) => Pick<S, K>, callback?: () => void): void;
+		setState<K extends keyof S>(state: Pick<S, K>, callback?: (...args) => void): void;
+		setState<K extends keyof S>(fn: (prevState: S, props: P) => Pick<S, K>, callback?: (...args) => void): void;
 
-		forceUpdate(callback?: () => void): void;
+		forceUpdate(callback?: (...args) => void): void;
 
 		abstract render(props?: RenderableProps<P>, state?: Readonly<S>, context?: any): JSX.Element | null;
 	}


marvinhagemeister:
This should already work right out of the box via casting:

```ts
function foo (callback: () => void) {}
function bar(arg1: string, arg2: number) {}

// Simply cast the function reference as `any`
foo(bar as any);
```

Example on [TypeScript Playground](https://www.typescriptlang.org/play/index.html#src=function%20foo%20(callback%3A%20()%20%3D%3E%20void)%20%7B%0D%0A%20%20%20%20console.log(%22hey%22)%0D%0A%7D%0D%0A%0D%0Afunction%20listener(arg1%3A%20string%2C%20arg2%3A%20number)%20%7B%0D%0A%0D%0A%7D%0D%0A%0D%0Afoo(listener%20as%20any))

Changing the type of the callback of both `forceUpdate` and `setState` is incorrect in my opinion, because we don't call it anywhere with arguments.

developit:
I'm ambivalent here. I guess if I had a concern, it'd be that there are probably a lot of these same cases in the definitions we'd have to update.

k1r0s:
I agree that this can be avoided but IMO ts definitions shouldn't be restrictive as you're force to explicitly specify your edge cases with `as any` and since `any` it self is kinda wildcard. 

ig types should only specify that arguments may actually exist but preact doesn't give any info about.

What do you think?

developit:
Would it be easier to define a type `UserCallback` that we use everywhere a generic callback function is allowed, and then we could allow arbitrary arguments to UserCallback?

marvinhagemeister:
-1 I'm against including such a change, because it would open a can of worms and will hide errors. Example:

```ts
// `arg1` is non-optional, and the function won't work when
// it isn't supplied. TS makes sure that this is the case.
function foo(arg1: string) {
   // do something with arg1, eg throws when arg1 is not present
}

// Although foo will definitely throw, this error could not be caught anymore by TS.
forceUpdate(foo);
```

k1r0s:
@marvinhagemeister I don't understand your message :\ 

`(...args)` makes arguments optional because isn't up to preact interact with'em

k1r0s:
> Changing the type of the callback of both forceUpdate and setState is incorrect in my opinion, because we don't call it anywhere with arguments.

ofc we never call callbacks with arguments, but user could provide anyother callback that has arguments defined for whatever reason. Imo we should provide restrictive definitions if needed.

k1r0s:
in your comments here: https://github.com/developit/preact/pull/1073#issuecomment-385001656

in such a case user should be aware of his own code and that parameter must be optional. Basically he wants some function to be called when render_callbacks came to an end and he doesn't need to wrap its finisher with another function which doesn't receive parameters bcoz we (preact) told TS to complain if any user do such a thing, which IMO is quite common (code reusing) :\

marvinhagemeister:
@k1r0s If the user knowingly supplies a callback that expects arguments, than that would be a bug. It is definitely up to us as to how the function signature should look like, because we are the one who will call it after all.

```ts
// changed signature as suggested in this PR
function forceUpdate(cb: (...args: any[]) => void) {
  cb();
}

function foo(arg1: string) {
  if (arg1 === undefined) {
    throw new Error("Should never happen");
  }
}

// Will throw, although types would say that this is valid
forceUdpate(foo);
```

k1r0s:
@marvinhagemeister, thanks to be here. I think u're completely right. And I was wrong. If some developers are used to provide own defined callbacks, then they should use optional parameters to explicitly specify that behavior. 

[correct code on typescript playground](https://www.typescriptlang.org/play/index.html#src=function%20foo%20(callback%3A%20()%20%3D%3E%20void)%20%7B%0D%0A%20%20%20%20console.log(%22hey%22)%0D%0A%7D%0D%0A%0D%0Afunction%20listener(arg1%3F%3A%20string)%20%7B%0D%0A%0D%0A%7D%0D%0A%0D%0Afoo(listener))

PR #1077, opened 2018-04-26
Rollback uglify3 -> uglify2

### LINKED ISSUE #1075: Latest release is completely broken
It looks like there was an update to the minification script which caused all variables to be renamed, which probably was to save some extra bytes in the package. However, it renamed all the exports, and it also renamed "exports" as well, so now the package exports an empty object, which causes anything using preact to fail completely. Here's how the main export looks, built from the latest release directly:

For now, we are pinning to `8.2.7`, which fixes the problem. Probably something should be done about this quickly, as it's a fairly critical issue.

marvinhagemeister:
Seems like uglify3's behaviour has changed with regards to mangling properties. This reverts uglify back to v2.

Fixes #1075


marvinhagemeister:
Here is a comparison between 8.2.7 and 8.2.8. Only difference in the build artefact is an unnecessary regex escape that has been removed. Apart from that everything is identical:

screen shot 2018-04-27 at 01 48 14

marvinhagemeister:
@developit We have one that checks if the non-transpiled files have the correct exports, but not one that checks the transpiled output

marvinhagemeister:
I've played around with `uglifyjs@3` a bit more, but still haven't had any success. There is a flag to list properties that should not be mangled (`--mangle-props reserved=['foo','bar']`) but I have trouble getting the `options` object and checks to reliably work.

I'll merge this, so that we can make a `8.2.9` hotfix release. After that we can look how to upgrade to `uglifyjs@3` or if we should switch to GCC.

PR #1088, opened 2018-05-03
Fix AnyComponent Type problem

zD98:
AnyComponent type should be FunctionalComponent or ComponentConstructor.

```
class Test extends preach.Component<any , any> {
    render(){}
}
// typeof Test is ComponentConstructor
```


marvinhagemeister:
We could merge both `AnyComponent` and `ComponentFactory` as they are the same with the changes this PR.

zD98:
preact-redux `connect` use preact.AnyComponent  to check `Component Type`. So should the `preact-redux` change `AnyComponent` to `ComponentConstructor` ?

scurker:
@zD98 Do you think we could add a test for this? `test/ts/hoc-test.tsx` currently only tests the definition using `ComponentFactory/Component` constructor. It should be fairly straightforward to add a test for type checking `AnyComponent`:

```ts
function HOCWrapper<T>(Wrappable: AnyComponent<T>): AnyComponent(T) {
  ...
}
```

marvinhagemeister:
@zD98 I'd love to merge this as this is the right fix. Would you be willing to add a tiny test case for this? Something like @scurker suggested or:

```ts
class Foo extends Component {
	render() {
		return <div>Hello</div>
	}
}

function Bar() {
	return <div>World</div>
}

const a: AnyComponent = Foo;
const b: AnyComponent = Bar;
```

marvinhagemeister:
Closing this PR as @scurker made a new one with the tests added. Thanks to everyone for working on a fix and getting to the bottom of it :)

PR #1102, opened 2018-05-16
Adding pointer event definitions

jakearchibald:
There's a bit of a complication here. Two of the pointer event events end in `Capture`, so this PR also defines `onGotPointerCaptureCapture` for the capturing variant of this event.

Does this even work?


developit:
This is just a React API oversight. Not only is there no way to listen to an event name ending in `Capture` without it also being capturing, `onFooCaptureCapture` is required.

Issue #411, opened 2016-11-21
Preact and Relay


yasserkaddour:
Hi,
I just tried to switch to preact on my relay project using preact-compat, it's compile but I have this error on the console.

> GraphQLStoreChangeEmitter.js:82Uncaught TypeError: this._batchUpdate is not a function
>     at e._processBroadcasts (http://localhost:9000/assets/app.js:31:3416)
>     at http://localhost:9000/assets/app.js:31:3104
>     at i (http://localhost:9000/assets/app.js:10:19989)
>     at http://localhost:9000/assets/app.js:10:20732
>     at MutationObserver.r (http://localhost:9000/assets/app.js:13:20214)
> e._processBroadcasts @ GraphQLStoreChangeEmitter.js:82
> (anonymous function) @ GraphQLStoreChangeEmitter.js:63
> i @ core.js:37
> (anonymous function) @ core.js:123
> r @ browser-raw.js:52

I think preact is awesome, I reduced my gzip bundle from 160k to 124k with it, it's just to bad it's doesn't work. Is there something you can do about it? 

Thanks

developit:
Hi Yasser - sorry I didn't reply to this earlier. Do you have an example that you started from or anything?  I've not used Relay (though I'm extremely interested in it) so it's a little hard to reproduce.

YasserKaddour:
Hi Jason, Thank you for considering working on Relay ;) 
I added preact to the relay-starter-kit repo here https://github.com/YasserKaddour/relay-starter-kit-preact
I got the same error message as in my relay project, I hope this will help, thank you

developit:
Awesome, I'll take a look when I get a little bit of spare time!

twhid:
I poked around at this a bit and it looks like Relay ([here](https://github.com/facebook/relay/blob/master/src/tools/relayUnstableBatchedUpdates.js)) is relying on `react-dom` to have a property called `unstable_batchedUpdates`. [This function](https://github.com/facebook/react/blob/master/src/renderers/shared/stack/reconciler/ReactUpdates.js#L94) is what is what is [assigned to that property](https://github.com/facebook/react/blob/master/src/renderers/native/ReactNativeStack.js#L47).

It would seem that preact-compat would need to implement this in order to support using preact with Relay. I'm new to both Relay and the innards of React and have no clue how to fix this, but thought the leg work in tracking it down might be helpful.

I'm also interested as it reduces a vendor bundle gzipped by almost 40% (148 -> 93k).

[...]

PR #1104, opened 2018-05-17
(Relay compatibility) Additional render for PFCs that don't return a VNode

kevlened:
This should help with Relay compatibility (#411)

[`relay`](https://github.com/facebook/relay) adds the concept of Containers to React. A Container defines the data needed to render a React Component with a graphql query. In Relay Modern, Components are wrapped in a Container in 3 ways: [`createFragmentContainer`](https://facebook.github.io/relay/docs/en/fragment-container.html#) (most common), [`createRefetchContainer`](https://facebook.github.io/relay/docs/en/refetch-container.html), and [`createPaginationContainer`](https://facebook.github.io/relay/docs/en/pagination-container.html).

They're all built on [`buildReactRelayContainer`](https://github.com/facebook/relay/blob/bd0b53ff60a41eb415fbd606b8ba4810afe736af/packages/react-relay/modern/buildReactRelayContainer.js#L35-L102) which returns a `Container` ~~constructor, but doesn't instantiate one~~ HOC. This means [when `preact` tries to normalize the component](https://github.com/developit/preact/blob/master/src/vdom/component-recycler.js#L17-L30), it treats the `buildReactRelayContainer` function as a PFC, adding `doRender` as a mock render function. Unfortunately, when `doRender` is called, it returns a `Container` instance instead of a `VNode`, so nothing is rendered.

This solution (courtesy of [an investigation by @adamdickinson](https://github.com/developit/preact/issues/411#issuecomment-373216817)) calls the PFC, then calls `render` if it exists. He notes [there may still be an issue with recycling](https://github.com/developit/preact/issues/411#issuecomment-379134904), but I'm assuming that'll be resolved if/when recycling is removed in #664 or optional in #789.

edit: The skinny is that Preact expects PFCs to return VNodes instead of Components, but React allows PFCs to return Components.

### DIFF
--- a/src/vdom/component-recycler.js
+++ b/src/vdom/component-recycler.js
@@ -45,5 +45,6 @@ export function createComponent(Ctor, props, context) {
 
 /** The `.render()` method for a PFC backing instance. */
 function doRender(props, state, context) {
-	return this.constructor(props, context);
+	let c = this.constructor(props, context);
+	return c && c.render ? c.render() : c;
 }
--- a/test/browser/components.js
+++ b/test/browser/components.js
@@ -80,6 +80,29 @@ describe('Components', () => {
 	});
 
 
+	it('should render functional components that return nested components', () => {
+		const PROPS = { foo:'bar', onBaz:()=>{} };
+
+		class C4 extends Component {
+			render() {
+				return <div>C4</div>;
+			}
+		}
+		sinon.spy(C4.prototype, 'render');
+
+		const C4hoc = sinon.spy( props => <C4 {...props} /> );
+
+		render(<C4hoc {...PROPS} />, scratch);
+
+		expect(C4.prototype.render)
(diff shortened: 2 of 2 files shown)


developit:
Alright, looping back on this - here's a demo showing the undocumented React behaviour:
https://jsfiddle.net/developit/7cnbug1t/

Personally, I don't think this is something we should add to Preact - it's something we should ask Relay to stop doing. AFAIK this is not an officially supported React feature, and it's basically identical to JSX composition:

```js
const Outer => props => new Inner()
// should be written as:
const Outer => props => <Inner {...props} />
```

andrewiggins:
The original issue this PR was opened to address has been closed and since this PR is open against Preact v8 I'm going to close this. If you still have issues with using Relay, do open a new issue with your repro. Thanks!

PR #1110, opened 2018-05-20
Fix spellcheck attr not set to false

### LINKED ISSUE #651: Default-true boolean attributes mistakenly removed from nodes?
Thanks for all your work on this library!

When I write
```jsx
<input spellCheck={false}/>
```
I expected to get a `<input>` DOM element with `spellcheck` set to false --- but instead it appears to be set to `null`, and so effectively spellchecking remains on for that input element. Is this expectation sensible?

If so, does something else need to happen in the vicinity of https://github.com/developit/preact/blob/master/src/dom/index.js#L75-L87 ? There are two things going on with `spellCheck` in particular; one is that the (p)react name for it is camel-cased, but the actual DOM attribute is lowercase, so we fall into https://github.com/developit/preact/blob/master/src/dom/index.js#L80-L87 --- because while `"spellcheck" in node` is true, `"spellCheck" in node` isn't. Two is that all falsy values seem to be rendered into the DOM as absent attributes, which doesn't seem to make sense for attributes that are default-true, (like `spellcheck`) and require an explicit false setting to clear them.

marvinhagemeister:
I'm not too keen on checking directly against the attribute name, but this seems to be the only case in the spec where the attribute should not be removed if the value is `false`.

```js
const el = document.createElement("input");

el.removeAttribute("spellCheck");
el.spellcheck // logs `true` :/

el.setAttribute("spellCheck", false);
el.spellcheck // logs `false`
```

Fixes #651.

### DIFF
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -81,10 +81,13 @@ export function setAccessor(node, name, old, value, isSvg) {
 	}
 	else if (name!=='list' && name!=='type' && !isSvg && name in node) {
 		setProperty(node, name, value==null ? '' : value);
-		if (value==null || value===false) node.removeAttribute(name);
+		if ((value==null || value===false) && name!='spellcheck') node.removeAttribute(name);
 	}
 	else {
 		let ns = isSvg && (name !== (name = name.replace(/^xlink:?/, '')));
+		// spellcheck is treated differently than all other boolean values and
+		// should not be removed when the value is `false`. See:
+		// https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-spellcheck
 		if (value==null || value===false) {
 			if (ns) node.removeAttributeNS('http://www.w3.org/1999/xlink', name.toLowerCase());
 			else node.removeAttribute(name);
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -690,7 +690,7 @@ declare global {
 			sizes?: string;
 			slot?: string;
 			span?: number;
-			spellCheck?: boolean;
+			spellcheck?: boolean;
 			src?: string;
 			srcset?: string;
 			srcDoc?: string;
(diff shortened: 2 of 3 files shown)


marvinhagemeister:
Locally, without code-coverage enabled the performance difference is much smaller than in our CI:

| test | before | after |
|---|---|---|
| empty diff | 87939/s (30 ticks) | 88119/s (31 ticks) |
| repeat diff | 10988/s (250 ticks) | 11709/s (228 ticks) |
| large VTree | 43701/s (63 ticks) | 43082/s (62 ticks) |
| style/prop mutation | 20460/s (134 ticks) | 20316/s (133 ticks) |
| SSR Hydrate | 1376/s (1967 ticks) | 1315/s (2020 ticks) |

src/dom/index.js
```diff
@@ -85,7 +85,10 @@ export function setAccessor(node, name, old, value, isSvg) {
 	}
 	else {
 		let ns = isSvg && (name !== (name = name.replace(/^xlink:?/, '')));
-		if (value==null || value===false) {
+		// spellcheck is treated differently than all other boolean values and
+		// should not be removed when the value is `false`. See:
+		// https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-spellcheck
+		if ((value==null || value===false) && name !== "spellCheck") {
```
developit: I think the name is `"spellcheck"`, right?  In that case it actually wouldn't fall through to here, but rather would pass the `in` check on line 82.  It seems like the special-casing belongs there:

```js
// replaces L84:
if ((value==null || value===false) && name!='spellcheck') node.removeAttribute(name);
```

marvinhagemeister: Went the `setAttribute` way, but forcing lowercase seems better. Will change that.

marvinhagemeister:
@developit Updated with your suggestion 🎉

PR #1111, opened 2018-05-20
WIP: Normalize jsdoc formatting

### LINKED ISSUE #1106: Update JSDoc
In trying to better understand how Preact works, I've noticed that the JSDocs of some of the internal methods [have become out of date](https://github.com/developit/preact/blob/4f57ba02bd4dee37d3278f2e75ce18073a36deac/src/vdom/component.js#L13). I'd like to take a stab at updating these in an effort to better understand Preact. But before I do, could I get other's opinions on what style they prefer? I see a couple different styles in the codebase today. Some options are outlined below.

**Comment start**
*Most common in Preact*
Comment starts on same line as `/**`. ([example in Preact](https://github.com/developit/preact/blob/4f57ba02bd4dee37d3278f2e75ce18073a36deac/src/vdom/component.js#L10)):
```js
/** Description
 *	@param {Object} props Description
 */
```

Comment starts on line after `/**` ([example in Preact](https://github.com/developit/preact/blob/4f57ba02bd4dee37d3278f2e75ce18073a36deac/src/util.js#L2)) (Note: this appears to be the style on [usejsdoc.org](http://usejsdoc.org/about-getting-started.html))
```js
/**
 * Description
 * @param {Object} props Description
 */
```

With both of these, the intention is that `@tag` aligns to the top `Description`.

**Parameter description alignment**
Tabs between `@tag {type} name` and `description` to align descriptions ([example in Preact](https://github.com/developit/preact/blob/4f57ba02bd4dee37d3278f2e75ce18073a36deac/src/vdom/component.js#L13)):
```js
/**
 * Description
 * @param {Object} props	Description
 * @param {Number} props2	Description2
 */
```

Single space between `@tag {type} name` (i.e. don't align descriptions) ([example in Preact](https://github.com/developit/preact/blob/4f57ba02bd4dee37d3278f2e75ce18073a36deac/src/vdom/index.js#L40))
```js
/**
 * Description
 * @param {Object} props Description
 * @param {Number} props2 Description2
 */
```

@developit @marvinhagemeister - let me know if you have an opinion or not - it doesn't matter to me. I just wanted to see if it matters to you.

marvinhagemeister:
(no description)

### DIFF
--- a/src/vdom/component.js
+++ b/src/vdom/component.js
@@ -7,11 +7,14 @@ import { diff, mounts, diffLevel, flushMounts, recollectNodeTree, removeChildren
 import { createComponent, collectComponent } from './component-recycler';
 import { removeNode } from '../dom/index';
 
-/** Set a component's `props` (generally derived from JSX attributes).
- *	@param {Object} props
- *	@param {Object} [opts]
- *	@param {boolean} [opts.renderSync=false]	If `true` and {@link options.syncComponentUpdates} is `true`, triggers synchronous rendering.
- *	@param {boolean} [opts.render=true]			If `false`, no render will be triggered.
+/**
+ * Set a component's `props` (generally derived from JSX attributes).
+ * @param {Object} props
+ * @param {Object|number} [opts] If opts is a `number` it is equal to the render
+ *  mode (see constants.js for available options). Otherwise it is an options
+ *  object.
+ * @param {boolean} [opts.syncComponentUpdates] triggers synchronous rendering.
+ * @param {boolean} [opts.render=true] If `false`, no render will be triggered.
  */
 export function setComponentProps(component, props, opts, context, mountAll) {
 	if (component._disable) return;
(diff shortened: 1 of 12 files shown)


src/vdom/component.js
```diff
@@ -7,11 +7,14 @@ import { diff, mounts, diffLevel, flushMounts, recollectNodeTree, removeChildren
...
 
-/** Set a component's `props` (generally derived from JSX attributes).
- *	@param {Object} props
- *	@param {Object} [opts]
- *	@param {boolean} [opts.renderSync=false]	If `true` and {@link options.syncComponentUpdates} is `true`, triggers synchronous rendering.
- *	@param {boolean} [opts.render=true]			If `false`, no render will be triggered.
+/**
+ * Set a component's `props` (generally derived from JSX attributes).
+ * @param {Object} props
+ * @param {Object|number} [opts] If opts is a `number` it is equal to the render
```
andrewiggins: I don't think the `opts` parameter of `setComponentProps` is used as an options object anymore. It appears to just be the render mode. At least I didn't see any usages of `opts` as an object.

marvinhagemeister: Good catch!

PR #1112, opened 2018-05-20
Add getSnapshotBeforeUpdate lifecycle hook

marvinhagemeister:
- [x] Add getSnapshotBeforeUpdate tests
- [x] Port unit tests from react
- [x] Add unit tests

### DIFF
--- a/src/vdom/component.js
+++ b/src/vdom/component.js
@@ -78,6 +78,9 @@ export function renderComponent(component, opts, mountAll, isChild) {
 		snapshot = previousContext,
 		rendered, inst, cbase;
 
+	// FIXME: Do we have an issue with previous state??
+	const prev = extend({}, previousState);
+
 	if (component.constructor.getDerivedStateFromProps) {
 		state = component.state = extend(state, component.constructor.getDerivedStateFromProps(props, state));
 	}
@@ -112,7 +115,7 @@ export function renderComponent(component, opts, mountAll, isChild) {
 		}
 
 		if (isUpdate && component.getSnapshotBeforeUpdate) {
-			snapshot = component.getSnapshotBeforeUpdate(previousProps, previousState);
+			snapshot = component.getSnapshotBeforeUpdate(previousProps, prev);
 		}
 
 		let childComponent = rendered && rendered.nodeName,
@@ -192,7 +195,8 @@ export function renderComponent(component, opts, mountAll, isChild) {
 		// flushMounts();
 
 		if (component.componentDidUpdate) {
-			component.componentDidUpdate(previousProps, previousState, snapshot);
+			// FIXME: Do we have an issue with previous state!?
+			component.componentDidUpdate(previousProps, prev, snapshot);
 		}
 		if (options.afterUpdate) options.afterUpdate(component);
 	}
--- a/src/vdom/component.js
+++ b/src/vdom/component.js
@@ -68,7 +68,7 @@ export function renderComponent(component, opts, mountAll, isChild) {
 		state = component.state,
 		context = component.context,
 		previousProps = component.prevProps || props,
-		previousState = component.prevState || state,
+		previousState = extend({}, component.prevState || state),
 		previousContext = component.prevContext || context,
 		isUpdate = component.base,
 		nextBase = component.nextBase,
@@ -77,9 +77,6 @@ export function renderComponent(component, opts, mountAll, isChild) {
 		skip = false,
 		snapshot = previousContext,
 		rendered, inst, cbase;
-
-	// FIXME: Do we have an issue with previous state??
-	const prev = extend({}, previousState);
 
 	if (component.constructor.getDerivedStateFromProps) {
 		state = component.state = extend(state, component.constructor.getDerivedStateFromProps(props, state));
@@ -115,7 +112,7 @@ export function renderComponent(component, opts, mountAll, isChild) {
 		}
 
 		if (isUpdate && component.getSnapshotBeforeUpdate) {
-			snapshot = component.getSnapshotBeforeUpdate(previousProps, prev);
+			snapshot = component.getSnapshotBeforeUpdate(previousProps, previousState);
 		}
 
 		let childComponent = rendered && rendered.nodeName,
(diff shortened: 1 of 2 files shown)


test/browser/lifecycle.js
```diff
@@ -218,6 +218,181 @@ describe('Lifecycle methods', () => {
 		// [should not override state with stale values if prevState is spread within getDerivedStateFromProps](https://github.com/facebook/react/blob/25dda90c1ecb0c662ab06e2c80c1ee31e0ae9d36/packages/react-dom/src/__tests__/ReactComponentLifeCycle-test.js#L1035)
 	});
 
+	describe.only("#getSnapshotBeforeUpdate", () => {
```
marvinhagemeister: TODO: remove only

src/vdom/component.js
```diff
@@ -75,8 +75,12 @@ export function renderComponent(component, opts, mountAll, isChild) {
 		initialBase = isUpdate || nextBase,
 		initialChildComponent = component._component,
 		skip = false,
+		snapshot = previousContext,
 		rendered, inst, cbase;
 
+	// FIXME: Do we have an issue with previous state??
+	const prev = extend({}, previousState);
```
marvinhagemeister: Dirty workaround: I have the suspicion that previousState is mutated somewhere. Caught this with the tests for getSnapshotBeforeUpdate by react.

TODO: Check what's happening here. Seems fishy.

src/vdom/component.js
```diff
@@ -68,13 +68,14 @@ export function renderComponent(component, opts, mountAll, isChild) {
 		state = component.state,
 		context = component.context,
 		previousProps = component.prevProps || props,
-		previousState = component.prevState || state,
+		previousState = extend({}, component.prevState || state),
```
marvinhagemeister: Although this is needed for `getSnapshotBeforeUpdate` this also fixes a bug with `componentDidUpdate`. During rendering, the `state` variable is mutated. This has the consequence, that `previousState` would never hold the previous state, but the most current one.

marvinhagemeister:
This PR is now ready to review 🎉

src/vdom/component.js
developit: where were we mutating previousState?

src/vdom/component.js
```diff
@@ -68,13 +68,14 @@ export function renderComponent(component, opts, mountAll, isChild) {
...
 		previousContext = component.prevContext || context,
 		isUpdate = component.base,
 		nextBase = component.nextBase,
 		initialBase = isUpdate || nextBase,
 		initialChildComponent = component._component,
 		skip = false,
+		snapshot = previousContext,
```
developit: Q: were we deviating from React here by passing `context` as a 3rd argument to `componentDidUpdate`?  If so, maybe we correct that here.

marvinhagemeister: Yep, we were/are deviating from React there. React only never had a third `context` parameter. Same for `shouldComponentUpdate` where we pass an additional context arg too. I'd prefer to these changes in a separate PR, because they are independent from `getSnapshotBeforeUpdate`.

developit: Agreed regarding the separate PR, but we should make sure they go out in the same release - we'll save a few bytes by dropping that reassignment.

test/browser/lifecycle.js
```diff
@@ -218,6 +218,181 @@ describe('Lifecycle methods', () => {
 		// [should not override state with stale values if prevState is spread within getDerivedStateFromProps](https://github.com/facebook/react/blob/25dda90c1ecb0c662ab06e2c80c1ee31e0ae9d36/packages/react-dom/src/__tests__/ReactComponentLifeCycle-test.js#L1035)
 	});
 
+	describe("#getSnapshotBeforeUpdate", () => {
+		it('should call nested new lifecycle methods in the right order', () => {
```
andrewiggins: Love this test! This might be out of scope for this PR, but could we move this test to be directly under the "Lifecycle methods" `describe` block or put it in its own `describe` block? This test specificies the behavior of more than just `getSnapshotBeforeUpdate`. It specifies the order of (potentially all) lifecycle methods. I think it would be valuable to expand this test to include the order of the lifecycle methods for all 4 main render "events": mounting, props update, state update, and unmounting.

I started [a test similar to this](https://github.com/andrewiggins/preact/blob/884457b5833799a9f98ad87acbd530430d31eb52/test/browser/lifecycle.js#L59) when implementing `getDerivedStateFromProps` but pulled it out as it was getting messy. This implementation is much more succinct. I'd be happy to expand this test in another PR if you think its useful.

src/vdom/component.js
```diff
@@ -75,10 +75,12 @@ export function renderComponent(component, opts, mountAll, isChild) {
...
 		rendered, inst, cbase;
 
 	if (component.constructor.getDerivedStateFromProps) {
-		state = component.state = extend(state, component.constructor.getDerivedStateFromProps(props, state));
+		previousState = extend({}, previousState);
```
andrewiggins: Good catch - thanks!

test/browser/lifecycle.js
```diff
@@ -218,6 +218,181 @@ describe('Lifecycle methods', () => {
...
+				'outer getSnapshotBeforeUpdate',
+				'inner getDerivedStateFromProps',
+				'inner shouldComponentUpdate',
+				'inner getSnapshotBeforeUpdate',
+				'inner componentDidUpdate',
+				'outer componentDidUpdate'
+			]);
+		});
+
+		it('should pass the return value from getSnapshotBeforeUpdate to componentDidUpdate', () => {
```
andrewiggins: I appreciate the mirroring of the React UT suite 😄

test/browser/lifecycle.js
marvinhagemeister: That's a great suggestion! Moved the test to the top `describe`-block. I'm a bit hesitant to add `componentWillUpdate`, `componentWillMount` and `componentWillReceiveProps` to the test, because we'll follow react and mark them as deprecated with our next release.

andrewiggins: Yea, that makes total sense. No need to add them

src/vdom/component.js
marvinhagemeister: mh I have a branch where I removed the non-standard context arguments, but I get the exact same gzip size as in master :S

PR #1113, opened 2018-05-20
eager init of _renderCallbacks to avoid checks on it

valotas:
size also improves a little bit


developit:
Nice! We should grab perf numbers for this locally, but I think this is the way to go!

marvinhagemeister:
Results for my Macbook Pro 15" 2016:

| Test | master | PR |
|---|---|---|
| empty diff | 87744/s (25 ticks) | 89288/s (30 ticks) |
| repeat diff | 11319/s (228 ticks) | 11034/s (249 ticks) |
| large VTree | 43565/s (59 ticks) | 43455/s (62 ticks) |
| style/prop mutation | 19518/s (132 ticks) | 20533/s (132 ticks) |
| SSR Hydrate | 1377/s (1880 ticks) | 1356/s (1984 ticks) |

Issue #1008, opened 2018-02-16
Typescript declarations: children gets type-checked as an attribute?


mindplay-dk:
I've run into a peculiar type-checking problem with functional components.

```
const Example = ({ title, children }: { title?: string, children: VNode[] }) => (
    <div class="example">
        <div class="example__source">
            <pre><code class="language-html">{renderHTML(children)}</code></pre>
        </div>
        <div class="example__content">{children}</div>
    </div>
);

const doc = (
    <div>
        <h1>Typography</h1>
        <Example>
            <h1>Heading 1</h1>
            <h2>Heading 2</h2>
            <h3>Heading 3</h3>
        </Example>
    </div>
);
```

Every call to `<Example>` complains that `children` is missing, which obviously isn't true.

I've always thought it was odd that `children`, being completely unrelated to `props`, is mixed into the `props` object, rather than being passed as a second argument - presumably that's what causes this problem? But I guess we inherited this weirdness from React 🙄 

I could of course type-hint as `children?` any time a functional component accepts `children`, making it optional - but in this example (and others) it *isn't* optional, and the the problem just moves to `renderHTML(children)` where I then have to write `renderHTML(children!)` which also isn't strictly correct, or `renderHTML(children || [])` which *would* be correct if `children` was really optional in the first place, so this is even more confusing.

Is there any way to improve the type-hinting of `children` somehow?

I suspect the only way this could ever improve is if the maintainers of React could be convinced to untangle `children` from `props` - which is hardly worth even propositioning to them, as that would break literally every component in every project.

I thought I'd ask anyway 😃

developit:
That's likely due to the definitions being based on React's. We pull `children` out of `props`, but that's actually the original location of it, and might be changing in a future version.

Preact's definition shouldn't have children listed as optional - it's always present and always an Array.

mindplay-dk:
> We pull `children` out of `props`

You do? I've been getting `children` via `props.children`, and that works. Removing it from `props` would be breaking vs React, right? (Which I suppose might be okay in a major release, assuming the React compatibility-package can restore it?)

> Preact's definition shouldn't have children listed as optional - it's always present and always an Array

Good to know.

On that note, I've had `attributes` come up as `null` - I've had to do e.g. `attributes || {}` in several places. It would be nice if this consistently came through as an Object as well, but I suppose that's a (however small) breaking change, too?

developit:
[...]

PR #1116, opened 2018-05-21
Add typescript child type checking

### LINKED ISSUE #1086: Typescript JSX child type checking
Typescript allows type checking of element children if you tell it the property name to use, would be very useful for type inference of expressions passed to a component, for example preact-router's [Match](https://github.com/developit/preact-router#active-matching--links)

https://www.typescriptlang.org/docs/handbook/jsx.html#children-type-checking

Alexendoo:
Closes #1086

It fixes the original issue from #1008, it will now type check correctly

It will also allow type inference, a basic example is provided in the tests that infers the type of `num` but it will be useful for anything like developit/preact-router#269

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -4,8 +4,8 @@ export as namespace preact;
 declare namespace preact {
 	type Key = string | number;
 	type Ref<T> = (instance: T) => void;
-	type ComponentChild = JSX.Element | string | number | null;
-	type ComponentChildren = ComponentChild[];
+	type ComponentChild = VNode<any> | string | number | null;
+	type ComponentChildren = ComponentChild[] | ComponentChild | object | string | number | null;
 
 	/**
 	 * @deprecated
@@ -104,12 +104,12 @@ declare namespace preact {
 	function h<P>(
 		node: ComponentFactory<P>,
 		params: Attributes & P | null,
-		...children: (ComponentChild | ComponentChildren)[]
+		...children: ComponentChildren[]
 	): VNode<any>;
 	function h(
 		node: string,
 		params: JSX.HTMLAttributes & JSX.SVGAttributes & Record<string, any> | null,
-		...children: (ComponentChild | ComponentChildren)[]
+		...children: ComponentChildren[]
 	): VNode<any>;
 
 	function render(node: ComponentChild, parent: Element | Document, mergeWith?: Element): Element;
@@ -134,6 +134,10 @@ declare global {
 
 		interface ElementAttributesProperty {
 			props: any;
+		}
+
+		interface ElementChildrenAttribute {
+			children: any;
 		}
 
 		interface SVGAttributes extends HTMLAttributes {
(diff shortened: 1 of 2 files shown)


src/preact.d.ts
```diff
@@ -4,8 +4,8 @@ export as namespace preact;
 declare namespace preact {
 	type Key = string | number;
 	type Ref<T> = (instance: T) => void;
-	type ComponentChild = JSX.Element | string | number | null;
-	type ComponentChildren = ComponentChild[];
+	type ComponentChild = VNode<any> | string | number | null;
+	type ComponentChildren = ComponentChild[] | ComponentChild | {} | string | number | null;
```
andrewiggins: Is this correct? I was under the impression that the children prop is always an array.

Alexendoo: Not sure now it's been pointed out... it didn't want to type check correctly without changing children in some form though.

I ended up doing it a similar way to the DT react definitions, if there's a type difference between them there may be a more accurate way of doing it for preact but I'm not sure what it would be

Alexendoo:
Sure thing, changed it to `object` @marvinhagemeister

Alexendoo:
Yeah this is correct @valotas it is always an array at runtime, this definition is necessary to appease the assumptions TSX makes, it's summed up nicely here https://github.com/developit/preact/pull/1116#discussion_r191113051

It may be possible in TypeScript 3 using LibraryManagedAttributes to do this correctly on both sides however, but I haven't explored this yet

valotas:
@Alexendoo thanks a lot for the explanation! I'll have a look at ts3

src/preact.d.ts
sant123: @marvinhagemeister why the use of `object` here? Taking a look on ReactNode it has the following types:

```ts
type ReactNode =
        | ReactElement
        | string
        | number
        | Iterable<ReactNode>
        | ReactPortal
        | boolean
        | null
        | undefined
        | DO_NOT_USE_OR_YOU_WILL_BE_FIRED_EXPERIMENTAL_REACT_NODES[
            keyof DO_NOT_USE_OR_YOU_WILL_BE_FIRED_EXPERIMENTAL_REACT_NODES
        ];
```

sant123: Perhaps we may refactor it like:

```ts
type ComponentChild = VNode<any> | string | number | Iterable<ComponentChild> | bigint | boolean | null | undefined
```

Alexendoo: At the time React had an object in its definition also but I do not remember why (this is a 6 year old PR 😅), type definitions could also not refer to themselves like that back then

sant123: Yeah pretty old 😅 now that is possible to refer to a type by itself; would you mind if I create a PR to update ComponentChild types? `object` is pretty wide and we can change it by `Iterable<ComponentChild>`.

sant123: Just created the PR, @Alexendoo thank you so much for answering.

Issue #1007, opened 2018-02-15
Preact flushing entire components using react-hot-loader


arqex:
Hi guys,

I had a tough day trying to make preact work with react-hot-loader v3 but at least I have discovered why.

The problem I had was that preact was flushing an entire form component when typing on an input, it was supposed to just re-render the input. Unmounting and mounting the component make the input lose the focus and the form is not usable anymore.

After much investigating the problem was that react-hot-loader uses proxies in order to create instances of the components, see:
https://github.com/gaearon/react-hot-loader/blob/master/src/patch.dev.js#L106

Using RHL, when a component is created using `inst = new Ctor(props, context)` by preact we are not getting an instance of Ctor, but an instance of a copy of Ctor created by the proxy.

That makes that `inst.constructor !== Ctor` so the next time that the component get rendered, preact tries to build it from the VNode and finds that `dom._componentConstructor !== vnode.nodeName` and it decides to unmount it.

I understand this is not preact's fault, but it's so easy to fix just modifying one line of code of preact, overriding `inst.constructor = Ctor` for all components:

In [component-recycler.js](https://github.com/developit/preact/blob/52bb91e17a56e6f3dba050c3ae001292b6ff2285/src/vdom/component-recycler.js):

```js
if (Ctor.prototype && Ctor.prototype.render) {
  inst = new Ctor(props, context);
  Component.call(inst, props, context);
} else {
  inst = new Component(props, context);
  inst.render = doRender;
}

// Override the constructor even if the instance was created by `new Ctor()`
// This will make preact work ok with react-hot-loader using proxies to create instances
inst.constructor = Ctor;
```

If I create a PR would you merge it? It's my missing piece to finally make preact and react-hot-module work together!

Thanks for preact, it's great!

arqex:
A temporary hacky fix I use to make RHL work properly with preact in the meanwhile, remember that if you use `componentDidMount` in your component you will need to add this piece of code there too:

```js
import {Component} from 'preact';

// Fix react-hot-loader with preact
Component.prototype.componentDidMount = function(){
	var vnode = React.createElement(this.constructor);
	this.base._componentConstructor = vnode.nodeName;
	this.constructor = vnode.nodeName;
}
```

developit:
Ahhh, interesting! We could probably justify adding that line as long as it doesn't have a huge performance impact. I'd be happy to look at a PR if you want to submit one.

marvinhagemeister:
[...]

PR #1120, opened 2018-05-25
Add a new option to control element comparison

theKashey:
Just adding an option (to pair `vnode`) to hook element comparison. Required by #1007 
Another option is to use @arqex's code - https://github.com/developit/preact/compare/master...arqex:master  with setComponentComparator function.  It is just a matter of taste.

Looking forward to make Preact quite 🔥[loadable]
(😕I did not found tests for vnode, and did not add my ones next to them  😭)


theKashey:
New version of React-Hot-Loader with partial Preact support was released.
Waiting for this stuff for full compliance.

src/vdom/index.js
```diff
@@ -1,5 +1,18 @@
...
 
+/**
+ * Checks that two components are equal
+ * @param {Component} component1 First component to compare
+ * @param {Component} component2 Second component to compare
+ * @return {boolean}
+ * @private
+ */
+export function areComponentsEqual(component1, component2) {
+	return options.areComponentsEqual
```
marvinhagemeister: One could remove the `if` check here by making it the default implementation for `options.areComponentsEqual`

theKashey: Pro - speed matters
Cons - you will be unable to revert configuration. Not a big deal, but could be nasty.

theKashey: Probably one `if` here is way faster than one more function call you are asking for.

theKashey: @marvinhagemeister - you were right.
My variant is 4x times slower than yours, and yours are equal to the base line.

https://jsperf.com/call-or-if/1

theKashey:
Obsolete

PR #1122, opened 2018-05-26
Inline setProperty function

andrewiggins:
Inlining the setProperty function (which is only used once) saves 7 bytes. Thought I could try to pay back some of the bytes of new features that have gone into Preact recently.

### DIFF
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -80,7 +80,11 @@ export function setAccessor(node, name, old, value, isSvg) {
 		(node._listeners || (node._listeners = {}))[name] = value;
 	}
 	else if (name!=='list' && name!=='type' && !isSvg && name in node) {
-		setProperty(node, name, value==null ? '' : value);
+		// Attempt to set a DOM property to the given value.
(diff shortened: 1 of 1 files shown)


src/dom/index.js
```diff
@@ -80,7 +80,11 @@ export function setAccessor(node, name, old, value, isSvg) {
 		(node._listeners || (node._listeners = {}))[name] = value;
 	}
 	else if (name!=='list' && name!=='type' && !isSvg && name in node) {
-		setProperty(node, name, value==null ? '' : value);
+		// Attempt to set a DOM property to the given value.
+		// IE & FF throw for certain property-value combinations.
+		try {
+			node[name] =  value==null ? '' : value;
```
marvinhagemeister: Two spaces or a tab after the equals sign :)

andrewiggins: Whoops - amended my commit with the fix

marvinhagemeister:
Thanks again for your contribution 🎉 💯

PR #1123, opened 2018-05-26
Specify that gDSFP is not called if neither props nor state have changed

andrewiggins:
Adds an additional unit test for getDerivedStateFromProps that mirrors [a unit test that React added recently](https://github.com/facebook/react/pull/12802)


andrewiggins:
Cool! If I come across more behaviors that would be useful to specify in tests, I'll open PRs for them.

PR #1124, opened 2018-05-29
chore: add Node.js 10

DanielRuf:


mkxml:
@DanielRuf Thank you for this PR.

Since this will launch two simultaneous builds on Travis every commit we need to check Sauce Labs behavior.

Maybe we can go with 10 only.

@marvinhagemeister What's your view on this?

DanielRuf:
> Since this will launch two simultaneous builds on Travis every commit we need to check Sauce Labs behavior.
> 
> Maybe we can go with 10 only.

We can also use the build matrix feature with different env settings and run Saubelabs just on one of both.

PR #1128, opened 2018-05-30
fix #1052: removing multiply keyed children results in unnecessary DOM events

### LINKED ISSUE #1052: Diff behaviour not as expected - preact unmounting then remounting the same DOM nodes even when there is a key
I have found that preact is unecessarily tearing down and rebuilding `DOM` nodes even when a key is used ([example codepen](https://codepen.io/HuntingLamas/pen/WzJNxN?editors=0110))

In this case even though the key is supplied almost all of the DOM nodes are being removed and then added again even though only three are changed

**EDIT**
TO be clear it is expected that the components are being rerendered, but the issue is that all of the `DOM` nodes (not virtual dom) are being removed and then re-added again

(before changing the document):

screen shot 2018-03-31 at 9 31 36 pm

(after changing the document):

screen shot 2018-03-31 at 9 31 38 pm

After debugging I eventually found that each time the new document was selected the `span` elements (in the real DOM) inside the content were being removed and then added again, and the weird thing is that the `span`s that are being removed are exactly the same as the `span`s that are being added (as in they were the same instance)

**Expected Behaviour**
Only a few `DOM` nodes will be changed (as keys are supplied and most of the vdom remains the same)

**Actual Behaviour**
All of the `DOM` nodes are being removed and then re-added again

yaodingyd:
Currently removing multiply keyed children will cause subsequent children unmount and remount. This fix expand check on `child===f.nextSibling` to make sure `child` has a match if only removing children.

### DIFF
--- a/src/vdom/diff.js
+++ b/src/vdom/diff.js
@@ -191,7 +191,7 @@ function innerDiffNode(dom, vchildren, context, mountAll, isHydrating) {
 		len = originalChildren.length,
 		childrenLen = 0,
 		vlen = vchildren ? vchildren.length : 0,
-		j, c, f, vchild, child;
+		j, c, f, l, vchild, child;
 
 	// Build up a map of keyed children and an Array of unkeyed children:
 	if (len!==0) {
@@ -207,6 +207,7 @@ function innerDiffNode(dom, vchildren, context, mountAll, isHydrating) {
 				children[childrenLen++] = child;
 			}
 		}
+		l = keyedLen - vlen;
 	}
 
 	if (vlen!==0) {
@@ -244,7 +245,7 @@ function innerDiffNode(dom, vchildren, context, mountAll, isHydrating) {
 				if (f==null) {
 					dom.appendChild(child);
 				}
-				else if (child===f.nextSibling) {
+				else if (child===f.nextSibling || (l > 0 && child===originalChildren[i + l])) {
 					removeNode(f);
 				}
 				else {
--- a/src/vdom/diff.js
+++ b/src/vdom/diff.js
@@ -245,11 +245,14 @@ function innerDiffNode(dom, vchildren, context, mountAll, isHydrating) {
 				if (f==null) {
 					dom.appendChild(child);
 				}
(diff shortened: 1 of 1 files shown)


src/vdom/diff.js
```diff
@@ -244,7 +245,7 @@ function innerDiffNode(dom, vchildren, context, mountAll, isHydrating) {
 				if (f==null) {
 					dom.appendChild(child);
 				}
-				else if (child===f.nextSibling) {
+				else if (child===f.nextSibling || (l > 0 && child===originalChildren[i + l])) {
```
k1r0s: 1o1

Can we use another variable name?

yaodingyd: sure. Have you have suggestions?

k1r0s: `childLength` or so.. :\

I mean, this is not even important :D

yaodingyd:
@andrewiggins Thanks for pointing out my bug! I have updated my code and also a new [example](https://codesandbox.io/s/zrl8rwnypl) to showcase this fix. 

I want to explain a little more on my code and example: The bug is because current code can only detect removing one child in the middle of a list. By checking the before and after list length diff, we can extend the condition `child===f.nextSibling` so that if the to-be-checked node is `f`'s next next next...sibling, we can correctly defect this condition.

My fix is tended to solve only this scenario. That means the list length is changed only because there are node to be removed. However, there could be node being removed also node being added. So in my test example I have following cases:
1. insert and delete multiple children, insert is before delete
2. insert and delete multiple children, insert is after delete

If inserts are all done before delete, I can update my list length diff variable to find the correct node to match `child`, but if there are inserts done after delete, currently I don't have a way to correctly tell which original node to match `child`, so for `insert and delete multiple children, insert is after delete`, even though the rendering result is correct, there are extra removing and adding operation.

One more scenario is that there are in-place changes on list happening with node removing. Because again list length is changed only because of node removing, I can correct behavior. Please look at the last two options of my example.

So I think my fix is a non-perfect solution: it only fix some aspects of this issue.

PR #1130, opened 2018-05-31
Proposal to fix config for UglifyJS 3

mkxml:
I'm aware of #1077 and #1075.

This PR is a test I did which worked to update UglifyJS to version 3, maybe it's not ready to merge but I would like to see your thoughts on this.

Tests ran just fine and gzip size goes to `3481` from the `3498` bytes we currently have on `master`. 

One thing I noticed is that the behavior changed in `dist/preact.js`, now it mangles the variables internally. Is this going to be a problem?

I couldn't find a way to disable this and keep mangling props with `--mangle-props` enabled because `--mangle-props` activates default `--mangle` implicitly in version 3 apparently.

So, to mitigate that, I activated `keep_fnames` to keep function names, but the vars are being mangled no matter what (even when omitting `--mangle`).

I opened mishoo/UglifyJS2#3168 to see if we can get version 2 behavior back on this matter.

Nonetheless, it works. Exports look fine and props like `_component`, `__preactattr_`, etc too.

Please check this out when you have the time and share your thoughts.

### DIFF
--- a/config/properties.json
+++ b/config/properties.json
@@ -1,6 +1,5 @@
 {
   "props": {
-    "cname": 16,
     "props": {
       "$_dirty": "__d",
       "$_disable": "__x",
@@ -21,7 +20,6 @@
     }
   },
   "vars": {
-    "cname": -1,
     "props": {}
   }
 }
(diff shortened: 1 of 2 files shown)


developit:
Haven't had time to check the Uglify issue, but when I was migrating microbundle to Uglify 3 I noticed the file format for prop mangling changed to add a doubly-nested wrapping object with 2 levels of "props" keys.

marvinhagemeister:
Our build pipeline is abstracted via [microbundle](https://github.com/developit/microbundle) now :tada:

PR #1135, opened 2018-06-05
Drain entire render queue in a single microtask

rpetrich:
The previous implementation of the render queue would schedule any render operations that were enqueued as the queue was being processed in the next microtask. This proposed implementation will process even render operations that are enqueued while the queue is being drained as part of the same microtask.

_(this is a result of the discussion in #819, where it is common for additional render operations to be enqueued during the render process as a result of a component throwing an error in `render` and an ancestor component calling `setState` in its `componentDidCatch` method)_


src/render-queue.js
```diff
@@ -20,9 +20,8 @@ export function enqueueRender(component) {
 
 /** Rerender all enqueued dirty components */
 export function rerender() {
-	let p, list = items;
-	items = [];
-	while ( (p = list.pop()) ) {
+	let p;
+	while ( (p = items.pop()) ) {
```
rpetrich: Alternatively:
```JavaScript
for (let i = 0; i < items.length; i++) {
    if (items[i]._dirty) renderComponent(items[i]);
}
items = [];
```

developit:
I believe this was originally designed to push doubly-nested renders into the queue rather than firing initially because people were using setState() loops for animation.  With this change, that would become impossible - I wonder if people are still doing that?

rpetrich:
@developit Ran some tests and it appears that calling `setState` inside `render` just busy-loops the browser inside `Promise` microtasks unless `options.debounceRendering = requestAnimationFrame` (or similar) is set. Similarly, CSS transitions don't appear to be triggered by updating a transition-enabled CSS property as a result of `setState` inside `render`. Still, it makes more sense to preserve the existing behaviour if it was specifically chosen—would you prefer I close this PR and add tests to validate this?

developit:
Honestly I'm split here - when using Promise-based debouncing, this PR is a good improvement. It's when using custom debounce (rAF, rIC or setTimeout) that it seems to matter more.

That said, the "animation by looping setState" approach isn't something I would imagine anyone is relying on other than me, and I'm very willing to give that up haha.  I actually think we should probably work to land this PR.

rpetrich:
@developit let me know what I can do to help in the way of adding tests or similar.

PR #1138, opened 2018-06-09
Add support for createRef() and react-create-ref

developit:
Like the title says, this adds support for Object refs - the `createRef()` API (returns an empty object). It should also work with the `react-create-ref` package.

Worth noting, this is around a 38 byte size increase and puts us up to 3535b, which I'm not thrilled about.

---

It's possible we could punt support for createRef() into preact-compat, since it can be implemented atop function refs quite easily:

```js
function createRef() {
  return function ref(c) { ref.current = c; }
}
```

... however that would not be compatible with the `react-create-ref` package, so if we got that route we'll need to provide an alias for it that points to preact-compat's createRef.


src/preact.js
```diff
@@ -5,10 +5,13 @@ import { render } from './render';
 import { rerender } from './render-queue';
 import options from './options';
 
+const createRef = Object;
```
marvinhagemeister: Just curious: Is using `Object` faster than a literal `{}`?

developit: slower. I just benchmarked and while I'd originally thought it'd be worth using Object here, the performance hit is surprising (50%):
https://esbench.com/bench/5b1d4021f2949800a0f61d25

marvinhagemeister:
@developit Right, thanks for getting back to me. In that case I think we should not deviate too much from `react`.

marvinhagemeister:
ping @lukeed

developit:
Are we thinking the 38 byte hit is okay?

lukeed:
Sorry for the delay, crazy week.

The only shavings (lol) I can suggest is `typeof ref == 'function'` into `!!ref.call`. 

Other than that, I think I'm of the mindset that we should become react-16 compatible all at once, when that point comes, rather than piecemeal some 16 attributes. Will be harder for users to keep track of what we currently do / don't do while still intentionally being incomplete.

src/util.js
```diff
@@ -10,6 +10,17 @@ export function extend(obj, props) {
 	return obj;
 }
 
+/** Invoke or update a ref, depending on whether it is a function or object ref.
+ *  @param {object|function} [ref=null]
+ *  @param {any} [value]
+ */
+export function applyRef(ref, value) {
+	if (ref!=null) {
+		if (typeof ref=='function') ref(value);
```
marvinhagemeister: This line is here to be compatible with the current way `refs` work, right?

lukeed: Yep

developit:
I think I tried the `ref.call` bit but it ended up producing a larger output since we already do typeof checks in the codebase but don't currently have any `.call`.

alexparish:
> Other than that, I think I'm of the mindset that we should become react-16 compatible all at once, when that point comes, rather than piecemeal some 16 attributes. Will be harder for users to keep track of what we currently do / don't do while still intentionally being incomplete.

I tend to disagree. Lack of support for react-16 features is holding back adoption of Preact due to incompatibility with libraries (eg. https://github.com/reach/router/issues/9). Rolling out support as soon as a feature is ready could help adoption. I'm really keen to use Preact in production but am held back through issues like this.

marvinhagemeister:
Merging as everybody is happy with the change 🎉

GordyGordy:
Sorry in advance if this is a dumb question.
Is there a way to get createRef to work with preact.min.js? Like is there a version of that file floating around somewhere that has react.createRef() support?

marvinhagemeister:
@GordyGordy No, we haven't made a new release yet. If absolutely need the latest unreleased code from `master` (note that it may be unstable), you can clone the repo and just run `npm run build` to create the same files as we ship on npm.

GordyGordy:
I think I solved it, i just pasted this into the bottom of the preact.min.js file and now everything works :)

function createRef() {return function ref(c) { ref.current = c; }}

marvinhagemeister:
@MathieuLoutre should work right out of the box with `react` aliased to `preact`

MathieuLoutre:
I've tried it and it seems to work brilliantly. Just wanted to make sure that it was supposed to do that ;) Thank you!

Issue #1143, opened 2018-06-13
Refs should not be allowed to leak into props


Pyrolistical:
When `ref` is set as an attribute it should be handled separately from props. It should never be allowed to leak into props.

I found a bug where preact-emotion (https://github.com/emotion-js/emotion/issues/718) set ref=undefined, which is passed along to reach router and caused https://github.com/reach/router/issues/73

Repo case in preact: https://codesandbox.io/s/n3mpmk2054
Works fine in react: https://codesandbox.io/s/887o6v4p9j

Kanaye:
The reason is that ref (and key) are only `delete`d when their value is `truthy`.
https://github.com/developit/preact/blob/86361f020ffb85cea9aac6af5a07f2682da425ed/src/vdom/component.js#L22-L23

@developit optimized it 2 years ago to delete `ref` conditional 
https://github.com/developit/preact/commit/d1548a5f2283282c48be40b758f60ac9a3efc351

I'm not sure how major this optimization is and would either roll back to a straight `delete` or change the condition to 
```javascript
if ('ref' in props) delete props.ref;
```

What do you think @developit ?

developit:
Ah interesting. We could use `!=null` in those checks to fix the issue without inflating the size at all:

```js
 if ((component.__ref = props.ref)!=null) delete props.ref; 
 if ((component.__key = props.key)!=null) delete props.key; 
```

Thanks for the investigation @Kanaye!

Kanaye:
You 're welcome. 
Just spending some time on the preact codebase to get more familiar with it.
It's always good to have some understanding on how frameworks work under the hood. 
And I guess I can still learn a lot about minimizing code size 😄

onstash:
I believe this issue can be closed as a related pull request has been merged. Ping @developit @lukeed @reznor.

PR #1145, opened 2018-06-14
Fixes refs from leaking into props

onstash:
Fixes issue mentioned in #1143 .

### DIFF
--- a/src/vdom/component.js
+++ b/src/vdom/component.js
@@ -19,8 +19,8 @@ export function setComponentProps(component, props, renderMode, context, mountAl
 	if (component._disable) return;
 	component._disable = true;
 
-	if ((component.__ref = props.ref)) delete props.ref;
-	if ((component.__key = props.key)) delete props.key;
+	if ((component.__ref = props.ref)!=null) delete props.ref;
+	if ((component.__key = props.key)!=null) delete props.key;
 
 	if (typeof component.constructor.getDerivedStateFromProps === 'undefined') {
 		if (!component.base || mountAll) {
--- a/src/vdom/component.js
+++ b/src/vdom/component.js
@@ -20,7 +20,7 @@ export function setComponentProps(component, props, renderMode, context, mountAl
 	component._disable = true;
 
 	if ((component.__ref = props.ref)!=null) delete props.ref;
-	if ((component.__key = props.key)!=null) delete props.key;
+	if ((component.__key = props.key)!=null || props.key!=undefined) delete props.key;
 
 	if (typeof component.constructor.getDerivedStateFromProps === 'undefined') {
 		if (!component.base || mountAll) {


src/vdom/component.js
```diff
@@ -19,8 +19,8 @@ export function setComponentProps(component, props, renderMode, context, mountAl
 	if (component._disable) return;
 	component._disable = true;
 
-	if ((component.__ref = props.ref)) delete props.ref;
-	if ((component.__key = props.key)) delete props.key;
+	if ((component.__ref = props.ref)!=null) delete props.ref;
```
Pyrolistical: I don't get how this is suppose to work when props.ref is undefined.

Tried it in node cli
```js
> props = {}
{}
> props.ref = undefined
undefined
> component = {}
{}
> if((component.__ref = props.ref) != null) delete props.ref
undefined
> props
{ ref: undefined }
```

Pyrolistical: @developit

onstash: Thanks for pointing it out, @Pyrolistical. Have added an additional check for `undefined` as well. Not sure if it is optimal enough. @developit knows more about it, from my understanding.

Kanaye: A check for `!= undefined` won't help.
The problem occurs if ref *is* `undefined`.
Also in javascript a condition with "normal" equality operators aren't type bound.
That means `undefined == null` is `true`.
Which means the second `!= undefined` is completly pointless.
For more informations on equality in js you could read the [MDN article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators).

lukeed: Not exactly ideal, but this would work, handling either `null` or `undef`:

```js
if (!!(component.__ref = props.ref)) delete props.ref;
```

Pyrolistical: `!!` won't work either.

onstash: How about this?

```
props = {};
props.ref = undefined;
component = {};
console.log("before");
console.log({ component });
console.log({ props });
if ((component.__ref = props.ref && !!props.ref)|| props.ref==undefined || props.ref==null) delete props.ref;
console.log("after");
console.log({ component });
console.log({ props });
```
gives the output
```
node v9.7.1 linux/amd64
before
{ component: {} }
{ props: { ref: undefined } }
after
{ component: { __ref: undefined } }
{ props: {} }
=> undefined
```

It works similarly for `ref == null` as well.

Source: https://repl.it/repls/CalmFrivolousScience

Kanaye: Works, but in that case the condition is completly pointless because it will *always* resolve to `true`.
No matter if `ref` is defined or not.
That means in that case it we could just omit the codition and roll back to a straight `delete props.ref;`

I'd need to profile if the performance gain would be worth the extra size when changing the condition to `('ref' in props) delete props.ref;` instead of a straight `delete props.ref;

onstash: Thanks for the inputs, @Kanaye. In my humble opinion, the condition will _always_ resolve to `true` because `ref` must be deleted from `props` before further processing.

developit:
I opened #1150 with my performance testing findings. It doesn't look like the guard was helping us much.

onstash:
Closing this PR in favor of #1150. Thanks for the opportunity guys. :smile:

PR #1150, opened 2018-06-17
Ref clearing perf

developit:
This is my attempt to benchmark a few options for #1143 / #1145.
I tested a few variants of the deletion guard (they're in the interim commit), and haven't found there to be much of a different with/without the guard.  All of the guard options except the existing combined approach added around 10 bytes.

@onstash @Pyrolistical @Kanaye @lukeed - y'all had a good discussion on the other PR. Think this is the right compromise?

### DIFF
--- a/src/vdom/component.js
+++ b/src/vdom/component.js
@@ -19,8 +19,22 @@ export function setComponentProps(component, props, renderMode, context, mountAl
 	if (component._disable) return;
 	component._disable = true;
 
-	if ((component.__ref = props.ref)) delete props.ref;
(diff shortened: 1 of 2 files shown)


Kanaye:
Haven't had time to profile it.
But assuming the impact isn't huge, it looks good to me 👍

onstash:
I think always deleting props.key & props.ref is the correct move. :thumpsup:

developit:
@onstash you should be getting the credit for this PR btw :P

onstash:
@developit The changeset matters more than who does it, IMHO. :smile:

PR #1156, opened 2018-06-24
Add `playsInline` attribute to HTMLAttributes

andybons:
It is used by the `<video>` element.

From https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video

> `playsinline`: A Boolean attribute which indicates that the video is to be played "inline", that is within the element's playback area. Note that the absence of this attribute does not imply that the video will always be played in fullscreen.


andybons:
You're very welcome. Thank you for the quick merge!

PR #1160, opened 2018-06-30
Add ShadowRoot as parent option for render()

cromefire:
Add ShadowRoot as type option for "parent" parameter of render()


developit:
I think we can add DocumentFragment as well!

cromefire:
Found this one again, but it seem that I have deleted the source branch, so I will do it and open a new one

PR #1166, opened 2018-07-10
Change dist module extension to .mjs and strip comments.

### LINKED ISSUE #1165: "module" export should by an .mjs file
This is important because right now, `dist/preact.esm.js` is actually not treated as a strict Module by Webpack 4. You can try this out:

```js
// name this file foo.mjs
```

Running the above through a vanilla Webpack config will resolve `preact` to `node_modules/preact/dist/preact.esm.js`, but will then throw an exception since Webpack does not infer strict ESM validity from the package.json "module" fields - it's only inferred from the file extension.

developit:
Fixes #1165.


developit:
That was fast :D

lukeed:
Critical issue, no?

edmorley:
Hi! Switching to an `.mjs` extension is likely the best choice long term - however this is actually a breaking change (for example neutrinojs/neutrino#1078) but the preact major version was not bumped.

Issue #1168, opened 2018-07-15
`setState` mutates state instead of replacing it


Mitranim:
Currently in Preact `8.2.9`, `Component.prototype.setState` mutates the previous state instead of replacing it with a patched version. This is unnecessarily inconsistent with React, can break user intuitions, can break things ported from React, and makes the second argument to `shouldComponentUpdate` useless and misleading.

I discovered this the hard way when sharing an initial state constant between components, assuming it would be treated as immutable. As previously mentioned, this can break user intuitions when coming from React.

Can we fix this? I would be willing to do a PR.

developit:
Definitely definitely happy to see a PR for this if you have the time. In the meantime do you have a concise test we could use to reproduce the issue?

Mitranim:
Here's a simple reproduction (run in Node):

```js
'use strict'

const preact = require('preact')
const render = require('preact-render-to-string')

const STATE = {}

class Component extends preact.Component {
  constructor() {
    super(...arguments)
    this.state = STATE
  }

  componentWillMount() {
    this.setState({key: 'value'})
  }
}

console.info(STATE)
// {}

render(preact.h(Component))

console.info(STATE)
// { key: 'value' }

// Expected behavior: STATE remains {}
```

KevinDoughty:
Hello I've fixed this bug but have never contributed before. I'm willing to send a PR or describe the fix for someone else. I'm a bit hesitant since I'm new to the codebase. I have a test and two modifications.

```js
	it("issue 1168, state should not be mutated", () => {
		const STATE = {};
		class C1168 extends Component {
			constructor() {
				super(...arguments);
				this.state = STATE;
			}
			componentWillMount() {
				this.setState({key: 'value'});
			}
		}
		render(<C1168 />, scratch);
		expect(STATE).to.deep.equal({});
	});
```

src/component.js Component.prototype:
```js
	setState(state, callback) {
		if (!this.prevState) this.prevState = this.state;
		this.state = extend(extend({},this.state), typeof state==='function' ? state(this.state, this.props) : state);
		if (callback) this._renderCallbacks.push(callback);
		enqueueRender(this);
	},
```

src/vdom/component.js renderComponent:
```js
	if (component.constructor.getDerivedStateFromProps) {
		state = extend({},state);
		component.state = extend(state, component.constructor.getDerivedStateFromProps(props, state));
	}
```

[...]

PR #1170, opened 2018-07-22
`setState` now creates new versions instead of mutating state

Mitranim:
Addresses #1168.

### DIFF
--- a/test/browser/lifecycle.js
+++ b/test/browser/lifecycle.js
@@ -1186,7 +1186,7 @@ describe('Lifecycle methods', () => {
 	});
 
 
-	describe('shouldComponentUpdate', () => {
+	describe('#shouldComponentUpdate', () => {
 		let setState;
 
 		class Should extends Component {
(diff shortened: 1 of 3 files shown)


test/browser/lifecycle.js
```diff
@@ -1325,6 +1325,38 @@ describe('Lifecycle methods', () => {
 	});
 
 
+	describe('#setState', () => {
```
gamebox: Could I suggest also including a test of the function parameter version of the function.  It looks like you are passing `prev` by reference here, so wouldn't any changes to it also mutate prev?  It would be interesting to explore the effects of that.

KevinDoughty:
getDerivedStateFromProps in src/vdom/component.js renderComponent() also mutates state, you might want to create a test that implements it.

KevinDoughty:
This test still fails:
```js
	it('getDerivedStateFromProps should not mutate state', () => {
		const stateConstant = {};
		class Stateful extends Component {
			constructor() {
				super(...arguments);
				this.state = stateConstant;
			}
			static getDerivedStateFromProps(props,state) {
				return { key: 'value' };
			}
			componentDidMount() {
				expect(stateConstant).to.deep.equal({});
			}
		}
		render(<Stateful />, scratch);
	});
```

test/browser/lifecycle.js
Mitranim: Good suggestion, I just feel like a full `setState` test suite is somewhat out of scope for the PR. Strange that it's missing though.

Yes it passes the previous state by reference, exactly as it should. Treating the previous state as immutable is the responsibility of the user.

Mitranim:
@KevinDoughty Good catch! Added a similar change for `getDerivedStateFromProps`, with a test. Now it also creates new versions instead of mutating the state.

dandv:
Should also add tests for nested props, given [this linkState bug](https://github.com/developit/linkstate/issues/17).

dandv:
@Mitranim: the [linkState bug](https://github.com/developit/linkstate/issues/17)? I haven't tried to repro it with this PR because linkState works with React as well, and I've just confirmed that the bug reproduces with React too. I was just suggesting adding tests for nested props since it's a tricky enough situation that linkState was tripped by it.

developit:
I think we're good to merge this.

marvinhagemeister:
@Mitranim Awesome, thanks for your PR 🎉

PR #1175, opened 2018-08-04
Add BitMidi; add link to Play.cash repo

feross:
https://bitmidi.com is built with Preact! ✨ Also, I've open sourced the Play.cash repo too.


developit:
Huzzah! :)

PR #1179, opened 2018-08-09
Remove ElementChildrenAttributes

surma:
The definition of `ElementChildrenAttributes` broke JSX typings for me. `children` was said to be expected to be `HTMLCollection` but was `Elementp[]`.

Gonna be honest, not 100% clear how this change caused _that_ error, so please make sure this is the right way to fix it — but it _does_ fix it for me.

cc @marvinhagemeister

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -136,10 +136,6 @@ declare global {
 			props: any;
 		}
 
-		interface ElementChildrenAttribute {
-			children: any;
-		}
-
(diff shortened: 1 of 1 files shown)


surma:
Hm... I should have run the tests, shouldn’t I? Investigating...

surma:
Closing this. I have isolated the issue, but I don’t know the fix. Will open an issue.

surma:
Openend #1180

PR #1183, opened 2018-08-13
Added more parents as option to render()

cromefire:
Added `ShadowRoot` and `DocumentFragment` as parent option to render()
Old: #1160


marvinhagemeister:
Awesome, thanks for your PR 👍🎉

PR #1187, opened 2018-08-15
Fix SVG foreignObject, Issue #773

jasonpecor:
The fix proposed by @developit on Jan 24 is incomplete, and causes issues when diffing attributes on the `foreignObject` node itself, including a Read Only exception when the node has a `class` or `className` attribute.  This fix switches back to SVG mode before diffing attributes if the vnodeName is "foreignObject".

### DIFF
--- a/src/vdom/diff.js
+++ b/src/vdom/diff.js
@@ -120,7 +120,7 @@ function idiff(dom, vnode, context, mountAll, componentRoot) {
 
 
 	// Tracks entering and exiting SVG namespace when descending through the tree.
-	isSvgMode = vnodeName==='svg' ? true : vnodeName==='foreignObject' ? false : isSvgMode;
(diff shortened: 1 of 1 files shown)


developit:
Good catch! I'd like to compare the size of this solution with passing `isSvg` through `diff()` and `innerDiffNode()`.

leonardodino:
can confirm, this fixed it for me on `preact@8.3.1`.

i'm using [`package-patch`](https://www.npmjs.com/package/patch-package) as a workaround. here's the [patch](https://gist.github.com/leonardodino/6ffe3969568de5a3ad87bab8d260ae79).

developit:
Note: Size increases from 3512b -> 3521b.

marvinhagemeister:
A change similar to this is already in master ✨

PR #1188, opened 2018-08-15
Fix prevState pointing to interim state value when batching (fixes #1186)

### LINKED ISSUE #1186: batched calls to setState: invalid previousState passed as this.state to shouldComponentUpdate
problem
--------

```javascript
constructor(...args) {
  super(...args);
  this.state.foo = 'notbar';
}

componentDidMount() {
  // two calls with same value
  this.setState({ foo: 'bar' });
  this.setState({ foo: 'bar' });
}

  return this.state.foo !== nextState.foo;
}

```
according to

https://github.com/developit/preact/blob/master/src/component.js#L55

1.  the first ```setState``` call updates ```this.prevState``` correctly, sets the new value to ```this.state```, and enqueues a render.

2. the second ```setState``` call will update ```this.prevState``` again, **but this time also merging in the updated** ```this.state``` value from the **first** ```setState``` call.

https://github.com/developit/preact/blob/master/src/vdom/component.js#L98

3. the queued render happens and ```shouldComponentUpdate``` is invoked, with ```component.state``` set to ```component.prevState```. but ```prevState``` already is set to the value of the first ```setState``` call.

4. hence, ```shouldComponentUpdate``` returns false and the queued render never happens - **although the state changed.**

expected behaviour
--------------------

```this.state``` in ```shouldComponentUpdate``` should be the state value from the **last** render run. even if batching multiple ```setState``` calls.

test
----

https://github.com/carlhopf/preact/commit/8695663573a7ea2d4ab2f847a56e833fe8e179fb

developit:
This should avoid resetting `prevState` for already-pending state updates, so that lifecycle methods reflect the original state value instead of the most recent previous value.

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -53,9 +53,11 @@ extend(Component.prototype, {
 	 * 	updated
 	 */
 	setState(state, callback) {
-		const prev = this.prevState = this.state;
-		if (typeof state === 'function') state = state(prev, this.props);
-		this.state = extend(extend({}, prev), state);
+		if (!this.prevState) this.prevState = this.state;
+		this.state = extend(
+			extend({}, this.state),
+			typeof state === 'function' ? state(this.state, this.props) : state
+		);
 		if (callback) this._renderCallbacks.push(callback);
 		enqueueRender(this);
 	},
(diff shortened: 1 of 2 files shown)


carlhopf:
@developit cheers, glad to help :)

PR #1197, opened 2018-08-24
Fix cloneElement Typescript definition

hiddedejong:
This adds the optional third argument to the cloneElement function's Typescript definition.

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -114,7 +114,7 @@ declare namespace preact {
 
 	function render(node: ComponentChild, parent: Element | Document | ShadowRoot | DocumentFragment, mergeWith?: Element): Element;
 	function rerender(): void;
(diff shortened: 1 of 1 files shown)


src/preact.d.ts
```diff
@@ -114,7 +114,7 @@ declare namespace preact {
 
 	function render(node: ComponentChild, parent: Element | Document | ShadowRoot | DocumentFragment, mergeWith?: Element): Element;
 	function rerender(): void;
-	function cloneElement(element: JSX.Element, props: any): JSX.Element;
+	function cloneElement(element: JSX.Element, props: any, children?: ComponentChildren): JSX.Element;
```
marvinhagemeister: That should be a rest argument like `...children`

hiddedejong:
Thank you for your feedback and work on this great library!

PR #1199, opened 2018-08-25
Reorganize and improve render unit tests

andrewiggins:
Reorganize render unit tests to better represent the expectations that the test are validating. Also, improved a couple tests to better cover their scenario.

### DIFF
--- a/test/browser/components.js
+++ b/test/browser/components.js
@@ -151,12 +151,6 @@ describe('Components', () => {
 		expect(scratch.innerHTML).to.equal('');
 	});
 
-	// Test for #651
-	it('should set enumerable boolean attribute', () => {
-		render(<input spellcheck={false} />, scratch);
-		expect(scratch.firstChild.spellcheck).to.equal(false);
-	});
-
 	// Test for Issue #73
 	it('should remove orphaned elements replaced by Components', () => {
 		class Comp extends Component {
--- a/test/browser/render.js
+++ b/test/browser/render.js
@@ -417,61 +417,63 @@ describe('render()', () => {
 		});
 	});
 
-	it('should support dangerouslySetInnerHTML', () => {
-		let html = '<b>foo &amp; bar</b>';
-		// eslint-disable-next-line react/no-danger
-		let root = render(<div dangerouslySetInnerHTML={{ __html: html }} />, scratch);
+	describe('dangerouslySetInnerHTML', () => {
+		it('should support dangerouslySetInnerHTML', () => {
+			let html = '<b>foo &amp; bar</b>';
+			// eslint-disable-next-line react/no-danger
+			let root = render(<div dangerouslySetInnerHTML={{ __html: html }} />, scratch);
 
-		expect(scratch.firstChild, 'set').to.have.property('innerHTML', html);
-		expect(scratch.innerHTML).to.equal('<div>'+html+'</div>');
+			expect(scratch.firstChild, 'set').to.have.property('innerHTML', html);
+			expect(scratch.innerHTML).to.equal('<div>'+html+'</div>');
 
-		root = render(<div>a<strong>b</strong></div>, scratch, root);
+			root = render(<div>a<strong>b</strong></div>, scratch, root);
 
-		expect(scratch, 'unset').to.have.property('innerHTML', `<div>a<strong>b</strong></div>`);
+			expect(scratch, 'unset').to.have.property('innerHTML', `<div>a<strong>b</strong></div>`);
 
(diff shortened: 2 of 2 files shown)


test/browser/components.js
```diff
@@ -151,12 +151,6 @@ describe('Components', () => {
 		expect(scratch.innerHTML).to.equal('');
 	});
 
-	// Test for #651
-	it('should set enumerable boolean attribute', () => {
```
andrewiggins: I just moved this into the render.js UT file since this test doesn't actually use an components

test/browser/render.js
```diff
@@ -243,209 +284,279 @@ describe('render()', () => {
...
+
+			root = render((
+				<div style={{ backgroundColor: 'rgb(0, 255, 255)' }}>test</div>
+			), scratch, root);
+
+			expect(root.style.cssText).to.equal('background-color: rgb(0, 255, 255);');
+		});
+	});
+
+	describe('event handling', () => {
```
andrewiggins: I grouped all event handling under a common `describe` block

test/browser/render.js
```diff
@@ -243,209 +284,279 @@ describe('render()', () => {
 		expect(scratch.childNodes[0]).to.have.property('className', 'bar');
 	});
 
-	it('should apply style as String', () => {
-		render(<div style="top:5px; position:relative;" />, scratch);
-		expect(scratch.childNodes[0].style.cssText)
-			.that.matches(/top\s*:\s*5px\s*/)
-			.and.matches(/position\s*:\s*relative\s*/);
-	});
+	describe('style attribute', () => {
```
andrewiggins: Moved all style tests under a common `describe` block

test/browser/render.js
andrewiggins: I also broke up the event handling tests to better specify the various concerns they were testing

test/browser/render.js
```diff
@@ -243,209 +284,279 @@ describe('render()', () => {
...
 
-	it('should apply proper mutation for VNodes with dangerouslySetInnerHTML attr', () => {
-		class Thing extends Component {
-			constructor(props, context) {
-				super(props, context);
-				this.state.html = this.props.html;
-			}
-			render(props, { html }) {
-				return html ? <div dangerouslySetInnerHTML={{ __html: html }} /> : <div />;
+	describe('dangerouslySetInnerHTML', () => {
```
andrewiggins: Group all `dangerouslySetInnerHTML` tests under a common `describe` block

test/browser/render.js
```diff
@@ -588,44 +699,69 @@ describe('render()', () => {
...
 			}
 			render() {
-				const {todos, text} = this.state;
+				const { todos, text } = this.state;
 				return (
-					<div onKeyDown={ this.addTodo }>
-						{ todos.map( todo => (<div>{todo.text}</div> )) }
+					<div onKeyDown={this.addTodo}>
+						{ todos.map( todo => ([
+							<span>{todo.text}</span>,
```
andrewiggins: The original issue called out this example as a better test for this test case so I updated the test to cover this.

test/browser/render.js
```diff
@@ -243,209 +284,279 @@ describe('render()', () => {
...
+			render(<div style="top:5px; position:relative;" />, scratch);
+			expect(scratch.childNodes[0].style.cssText)
+				.that.matches(/top\s*:\s*5px\s*/)
+				.and.matches(/position\s*:\s*relative\s*/);
+		});
 
-	it('should only register on* functions as handlers', () => {
-		let click = () => {},
-			onclick = () => {};
+		it('should properly switch from string styles to object styles and back', () => {
```
developit: we'll want to revisit this if we decide to remove string style support (it's a preact-only feature)

test/browser/render.js
developit: Yay, meaningful failure messages!

test/browser/render.js
```diff
@@ -588,44 +699,69 @@ describe('render()', () => {
 		expect(sortAttributes(html)).to.equal(sortAttributes('<input type="range" min="0" max="100" list="steplist">'));
 	});
 
-	it('should not execute append operation when child is at last', (done) => {
+	it('should not execute append operation when child is at last', () => {
+		// See developit/preact#717 for discussion about the issue this addresses
```
developit: 👍 I wish I'd left more of these issue references around.

developit: side note: we should totally look into reusing this component + plumbing for a bunch of different tests.

developit:
Approved x1000

PR #1211, opened 2018-09-13
Chrome poor performance when calling array.unshift on a large array.

lowaa:
I'm really enjoying preact, thanks for developing the project!

I noticed when quickly adding and removing a large amount of nodes, Chrome 68 experiences poor performance. Using the profiler, most of the time is spent in `mounts.unshift()`. This PR changes `mounts.unshift()` to `mounts.push()` and reversing the order that mounts are flushed to maintain functionality.

This project demonstrates the issue: https://github.com/lowaa/preact-perf

With preact 8.3.1, Chrome 68 running on a macbook pro 2018:

image

After the change:

image

The scripting time is significantly decreased but maximum heap usage goes up by ~25%

Also tested on Firefox 62 and Edge 17. These browsers did not seem to have an issue with the use of `array.unshift()`. The change to `array.push` also seemed to have no effect.

### DIFF
--- a/src/vdom/component.js
+++ b/src/vdom/component.js
@@ -189,7 +189,7 @@ export function renderComponent(component, renderMode, mountAll, isChild) {
 	}
 
 	if (!isUpdate || mountAll) {
-		mounts.unshift(component);
+		mounts.push(component);
 	}
 	else if (!skip) {
 		// Ensure that pending componentDidMount() hooks of child components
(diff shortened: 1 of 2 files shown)


marvinhagemeister:
@lowaa Woa, that's dope 👍 👍 I love the detailed description of this PR and can confirm that this is noticeably faster due to not having to resize the array all the time 💯

PR #1219, opened 2018-09-24
Add boolean to ComponentChild in preact.d.ts and remove object from type

### LINKED ISSUE #1215: ComponentChild should also allow boolean in TypeScript
Is there a reason why a ComponentChild cannot be a boolean? https://github.com/developit/preact/blob/b5c4a36576b65623c2664d04c03a04e57dc1e06a/src/preact.d.ts#L7-L8

Due to this, I get the following error:
```
[ts]
Type '{ children: boolean; }' is not assignable to type 'HTMLAttributes'.
  Types of property 'children' are incompatible.
    Type 'boolean' is not assignable to type 'string | number | object | VNode<any> | ComponentChild[] | null | undefined'.
```

And is there a reason why ComponentChildren repeats many of the same types as ComponentChild?

If you want, I can make a pull request that replaces those two lines with
```
	type ComponentChild = VNode<any> | string | number | boolean | null;
	type ComponentChildren = ComponentChild[] | ComponentChild;
```
(which would also remove 'object' which seems strange to me anyway – including the fact that it only exists for ComponentChildren). Please note that I haven't checked yet whether still everything works with such a change.

KasparEtter:
Fixes #1215.

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -4,8 +4,8 @@ export as namespace preact;
 declare namespace preact {
 	type Key = string | number;
 	type Ref<T> = (instance: T) => void;
-	type ComponentChild = VNode<any> | string | number | null;
-	type ComponentChildren = ComponentChild[] | ComponentChild | object | string | number | null;
+	type ComponentChild = VNode<any> | object | string | number | boolean | null;
+	type ComponentChildren = ComponentChild[] | ComponentChild;
 
 	/**
 	 * @deprecated
--- a/test/ts/VNode-test.tsx
+++ b/test/ts/VNode-test.tsx
@@ -5,7 +5,6 @@ import {
 	Component,
 	FunctionalComponent,
 	ComponentConstructor,
-	VNode
 } from "../../src/preact";
 
 class SimpleComponent extends Component<{}, {}> {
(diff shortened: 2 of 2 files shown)


KasparEtter:
Please note that I couldn't test the change as the tests don't run for me on master. Happy to iterate based on your feedback, @marvinhagemeister. 🙂

KasparEtter:
Thanks to Travis, I see that removing `object` from `ComponentChildren` breaks https://github.com/developit/preact/blob/b5c4a36576b65623c2664d04c03a04e57dc1e06a/test/ts/VNode-test.tsx#L65

The other problem is the same as on master:
```
The command "if [ "$TRAVIS_NODE_VERSION" = "8" ]; then COVERAGE=false FLAKEY=false PERFORMANCE=false ALLOW_SAUCELABS=true npm run test:karma; ./node_modules/coveralls/bin/coveralls.js < ./coverage/lcov.info; fi" exited with 1.
```

Is there a use case for passing a function (or another object) as a child to a component? Since it's likely a mistake by the programmer, I would prefer to adapt the test instead of the type declaration. On the other hand, one could argue that it's up to the component what it does with its children (including whether to evaluate them with a function call). What do you think, shall I add `object` again (to `ComponentChild` instead of `ComponentChildren` this time)?

src/preact.d.ts
```diff
@@ -4,8 +4,8 @@ export as namespace preact;
 declare namespace preact {
 	type Key = string | number;
 	type Ref<T> = (instance: T) => void;
-	type ComponentChild = VNode<any> | string | number | null;
-	type ComponentChildren = ComponentChild[] | ComponentChild | object | string | number | null;
```
marvinhagemeister: The tests are failing because the `object` type is needed to get more complex children to typecheck properly with jsx. See #1116 .

marvinhagemeister:
@KasparEtter Passing a function as a child is a valid pattern and the way the new Context-API in React 16 works https://reactjs.org/docs/context.html#consumer

KasparEtter:
I updated my PR (by force-pushing an amended commit, I hope you don't mind that I squashed already but it was a tiny commit anyway). I hope you agree that the types belong to `ComponentChild` and not `ComponentChildren`. I also added test cases for the various types (I just realized that maybe I should also add one for `VNode<any>`). What I don't like about these tests is that they don't verify that the children are actually checked. However, the only thing I found after a quick search to add "negative" test cases is https://stackoverflow.com/questions/51194259/asserting-that-typescript-should-fail-to-type-check-some-example-code, which would add new dev dependencies and I don't know whether this is acceptable. (Adding at least `tslint` seems reasonable to me as the test file had an unused import and missing semicolons.) `npm run test:ts` works for me, I'm only having problems with `npm run test`.

Two more observation based on React's types:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/fbf888379b391b2a8519415cb1e66e26e0d0e87b/types/react/index.d.ts#L152
- They seem to be missing `object` if the idea is indeed to pass functions as children.
- They have added `undefined` as well but I think we don't need/want that because we add the children optionally if they are not defined by the user/programmer:
https://github.com/developit/preact/blob/1369ab358a735a67d33c1287e4c15c8a1c85ef58/src/preact.d.ts#L55-L57

PR #1221, opened 2018-09-26
Added "onInvalid" form event in TS

namankheterpal:
Added an event `onInvalid` form event which is of type `GenericEventHandler` in TS of preact

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -54,7 +54,7 @@ declare namespace preact {
 
 	type RenderableProps<P, RefType = any> = Readonly<
 		P & Attributes & { children?: ComponentChildren; ref?: Ref<RefType> }
-	>;
+		>;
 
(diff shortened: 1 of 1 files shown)


src/preact.d.ts
```diff
@@ -455,6 +455,7 @@ declare global {
 			onSearchCapture?: GenericEventHandler;
 			onSubmit?: GenericEventHandler;
 			onSubmitCapture?: GenericEventHandler;
+	                onInvalid?: GenericEventHandler;
```
marvinhagemeister: Seems like the indentation is off. I'm happy to merge it once that's fixed 👍

namankheterpal: @marvinhagemeister Done. Let me know if any other changes are required.

marvinhagemeister: Awesome, thanks for making these changes 👍

PR #1227, opened 2018-10-03
fix render test that relies on implementation detail

KevinDoughty:
setState is supposed to be asynchronous but in actuality is not. This pull request removes the reliance on that implementation detail in a single test, test/browser/render.js "should not execute append operation when child is at last".

setState should be fixed in the future


KevinDoughty:
setState is fixed to be asynchronous with https://github.com/developit/preact/pull/1228, and a related PR is https://github.com/developit/preact/pull/1229

KevinDoughty:
More specifically, this test sets state, then immediately reads it back. Passing requires this to be possible, but it should not be possible and should not pass. Thus I wrapped the remaining code in a setTimeout.

developit:
I'm good with removing reliance on synchronous state from this test.

We'll still need to figure out if we have the option to move away from setting state synchronously, but there's no reason to rely on it in this test.

KevinDoughty:
My first PR ever!
ｷﾀ━━━━｡ﾟ+.ヽ(´∀`*)ﾉ ﾟ+.ﾟ━━━━!!

PR #1228, opened 2018-10-03
setState should be asynchronous

KevinDoughty:
setState in React is asynchronous but is not in Preact. The current implementation actually sets state, then unsets it for lifecycle methods, then sets it again. This fixes that while reducing gzip size from 3516 to 3511, for a total of 5 bytes. The diff also shows related changes from test/browser/render.js from https://github.com/developit/preact/pull/1227


KevinDoughty:
Since one test relied on this erroneous implementation detail (the last one in test/browser/render.js, should not execute append operation when child is at last) and I'm sure people are relying on this out in the wild, this should probably coincide with a major semver increase, despite not technically breaking the public API.

KevinDoughty:
Related PRs are https://github.com/developit/preact/pull/1227 (fixing one old test to not rely on erroneous implementation detail) and https://github.com/developit/preact/pull/1229 (a larger in scope, complete vanity thrashing of your codebase)

KevinDoughty:
Please note that I do not have specific React implementation knowledge. My understanding of how it works and this change might be completely wrong. What do you think?
https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous
https://medium.com/@wereHamster/beware-react-setstate-is-asynchronous-ce87ef1a9cf3
https://medium.learnreact.com/setstate-is-asynchronous-52ead919a3f0

src/component.js
```diff
@@ -32,11 +32,13 @@ export function Component(props, context) {
 	 */
 	this.props = props;
 
+	this._nextState =
```
marvinhagemeister: @KevinDoughty Perhaps a bad push?

KevinDoughty: Actually no, Are those logged type declarations used by some automated system, perhaps for generating docs? Originally I wrote the line as the following:
```JS
	/**
	 * @public
	 * @type {object}
	 */
	this._nextState = this.state = this.state || {}
```
But that implies that `_nextState` is public. I understand how atrocious that looks, it only saves a single byte doing it that way. I could easily spend that byte for nicer looking code, and put it on a separate line:
```JS
	/**
	 * @public
	 * @type {object}
	 */
	this.state = this.state || {};

	this._nextState = this.state;
```

KevinDoughty:
These changes are better when paired with https://github.com/developit/preact/pull/1229 but that one requires a deeper dive into lifecycle order and related bugs, which I am looking at next. I hope you will accept https://github.com/developit/preact/pull/1227 for now.

KevinDoughty:
If setState is passed a function, that function is called synchronously, with or without this PR. I do not know if that is correct behavior, and so also requires a "deeper dive".

KevinDoughty:
Crickets? Well, I will delete the six tests involving the number of times private variables are set, and move to a different file the one about async state, in this PR and https://github.com/developit/preact/pull/1228. I will also move to a different file the three tests regarding applyRef, and try harder to make them fail. Sound good?

KevinDoughty:
Tests updated but the reduction in size is less than before, now only a mere 2 bytes from 3516 to 3514

KevinDoughty:
This now has a failing test, by request:
https://github.com/developit/preact/pull/1229#issuecomment-428083296

For tests to pass, this PR must be accepted first:
https://github.com/developit/preact/pull/1227

KevinDoughty:
This is technically not a bug, rather, synchronous setState is undefined behavior that should not be relied on. But I think we can fix it.

andrewiggins:
@KevinDoughty Preact X has similar behavior as described in this PR. Does that work for you? Can we close this PR if so?

KevinDoughty:
> @KevinDoughty Preact X has similar behavior as described in this PR. Does that work for you? Can we close this PR if so?

Glad to hear it.

PR #1229, opened 2018-10-03
should not set props and context, then unset for lifecycle methods, then set again

KevinDoughty:
This is a work in progress, not to be accepted yet.

It is a minor implementation detail, perhaps, but I do not like how props and context are set, then unset for lifecycle methods, then set again. I changed similar behavior with state for https://github.com/developit/preact/pull/1228

I know you're not supposed to subclass Component, but maybe for developer tools, debugging, or animation, someone might want to override props or context with an ES5 property. I did but won't bore you.

More important is clarity of implementation.

All tests pass but I suspect I can write some that fail. Specifically, at the end of src/vdom/component.js setComponentProps, there is a call to applyRef where props and context might not be correct with these changes.

I want to consider the possibility that applyRef should be moved elsewhere, but that would require discussion of the order of when refs are called, and maybe lifecycle order. Maybe lifecycle order could be changed to correspond with React 15 behavior? Or maybe just refs?

This pull request includes the changes from https://github.com/developit/preact/pull/1228 and https://github.com/developit/preact/pull/1227, but all combined, it reduces gzip size from 3516 to 3487, for a total of 29 bytes


KevinDoughty:
This comment suggests a similar sentiment, that applyRef might get called in the wrong place, but source is private: https://github.com/developit/preact/issues/1177#issuecomment-416054414

Not that I think React order is any better, just different, but this is the rundown:

REACT 15.4.2 https://codepen.io/kvndy/pen/zmqqRp

!Zeroth render
!First render
!Second render
!Third render
+Third mount
?Second ref:[object Object];
+Second mount
?First ref:[object Object];
+First mount
?Zeroth ref:[object Object];
+Zeroth mount
!Zeroth render
?Zeroth ref:null;
-First unmount
?First ref:null;
-Second unmount
?Second ref:null;
-Third unmount


PREACT 8.3.1 https://codepen.io/kvndy/pen/pwgGNy

!Zeroth render
?Zeroth ref:[object Object];
!First render
?First ref:[object Object];
!Second render
?Second ref:[object Object];
!Third render
+Third mount
+Second mount
+First mount
+Zeroth mount
!Zeroth render
-First unmount
-Second unmount
-Third unmount
?Second ref:null;
?First ref:null;
?Zeroth ref:null;

KevinDoughty:
I missed a line, gzip size is now reduced  from 3516 to 3483, for a total of 33 bytes.

KevinDoughty:
Sorry, I shouldn't keep posting about this, but it's too fun. I shaved off a few more unnecessary bytes and now it's down from 3516 to 3479, for a total of 37 bytes. I don't think there is much more reduction possible that is still related to the changes in this PR

src/vdom/component.js
```diff
@@ -33,26 +33,18 @@ export function setComponentProps(component, props, renderMode, context, mountAl
...
+			renderComponent(component, props, context, SYNC_RENDER, mountAll);
 		}
 		else {
-			enqueueRender(component);
+			enqueueRender(component, props, context);
 		}
 	}
 
-	applyRef(component.__ref, component);
+	applyRef(component.__ref, component); // kvndy TODO: props and context may not be correct here
```
KevinDoughty: With this PR, props and context would only be set in renderComponent. I find it fortuitous that the refactoring led me to the same conclusion as this comment: https://github.com/developit/preact/issues/1177#issuecomment-416054414 I haven't yet had to make any "decisions" so to speak, and seek recommendation.

KevinDoughty:
After this PR, `setComponentProps` should probably be renamed because it would no longer do that. It also does not to be exported. I did not do this to minimize noise in the diff.

KevinDoughty:
I have created tests that defy my expectations and pass inexplicably. I know that does not instill much confidence. Regardless, I dare anyone to deny that this is the most beautiful pull request in the history of computing, with the exception of the tests which might be a bit pedestrian and sub-standard for this project. Please advise.

KevinDoughty:
Crickets? Well, I will delete the six tests involving the number of times private variables are set, and move to a different file the one about async state, in this PR and https://github.com/developit/preact/pull/1228. I will also move to a different file the three tests regarding applyRef, and try harder to make them fail. Sound good?

KevinDoughty:
I deleted six tests involving the number of times private variables are set, they were not helpful. I created three passing tests which might do nothing other than highlight my lack of understanding about refs, although they were enough to convince me the refactoring is sound. The important tests are the one about async setState, in this PR and https://github.com/developit/preact/pull/1228, in test/browser/lifecycle.js, and the corrected test at the end of test/browser/render.js, which is in all three PRs including https://github.com/developit/preact/pull/1227. 

Should I open a general issue in case someone else wants to try and fix this?

marvinhagemeister:
Awesome 💯 it's great to see new ways to improve the existing code 👍👍 I must admit that I'm having a bit of trouble reviewing your PRs because they are not isolated changes, but contain each other.

KevinDoughty:
@marvinhagemeister I will fix that shortly!

src/vdom/component.js
```diff
@@ -61,38 +53,35 @@ export function setComponentProps(component, props, renderMode, context, mountAl
...
+		snapshot = component.context,
 		isUpdate = component.base,
 		nextBase = component.nextBase,
 		initialBase = isUpdate || nextBase,
 		initialChildComponent = component._component,
 		skip = false,
-		snapshot = previousContext,
-		rendered, inst, cbase;
+		rendered, inst, cbase,
+		getDerived = component.constructor.getDerivedStateFromProps;
```
developit: this will invoke `getDerivedStateFromProps` with a `this` value of `undefined`.

src/component.js
```diff
@@ -53,13 +55,12 @@ extend(Component.prototype, {
...
-			extend({}, this.state),
-			typeof state === 'function' ? state(this.state, this.props) : state
-		);
 		if (callback) this._renderCallbacks.push(callback);
-		enqueueRender(this);
+		this._nextState = extend(
+			extend({},this._nextState),
+			typeof state==='function' ? state(this._nextState, this.props) : state
+		);
+		enqueueRender(this, this.props, this.context);
```
developit: I believe this breaks when enqueuing multiple setState() updates? Struggling to think of a way to demonstrate but it seems like the value of `this.props` being pushed into the render queue will result in it being incorrect for intermixed prop+state driven updates to the same component.

src/vdom/component.js
KevinDoughty: That was one of my byte-shavings, sad but true.

KevinDoughty: The really sad part is that is current behavior in 8.3.0
https://github.com/developit/preact/blob/e833786fb9543fd7becd561e59ae9132e7533dbd/src/vdom/component.js#L87
That line is a refactor that does not change behavior. This is a bug from the master branch, not this PR.

KevinDoughty: Funny, I thought getDerivedStateFromProps was supposed to be a static method, and I don't know what the value of `this` should be if there is no instance.

https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

KevinDoughty: I don't think it's a bug after all, in my PR or master!

src/component.js
KevinDoughty: I would be surprised if what you describe is not already covered by current tests. I understand the concern, especially with the recent regression involving the mutable setState fix, in issue https://github.com/developit/preact/issues/1168 and https://github.com/developit/preact/issues/1170

(Had I made a PR instead of just commenting that regression might have been avoided.)

I hope you see that these changes will make similar regressions less likely to happen because there will be less conditional gymnastics needed to maintain prevState and prevProps.

It will take time but I will look for specific tests that refute your concerns, and if I cannot find any I will write more tests.

Since you mentioned it, my concern now is what happens if `options.syncComponentUpdates` is `true` because there aren't any tests for that, and that is the only situation where what you describe might possibly happen, at least according to my admittedly limited understanding.

src/vdom/component.js
developit: The current `master` will invoke getDerivedStateFromProps() with `this` pointing to the constructor. I am not sure if that's something people use, but it's what React does.

KevinDoughty:
I assume this PR is untenable after Preact X but will verify before closing it.

PR #1231, opened 2018-10-04
fix #1225

### LINKED ISSUE #1225: ref() called only once with a DOM element on conditional rendering
I'm trying to get a reference to an input displayed in a modal dialog in an app that looks more or less like this:

```jsx
class App extends Component {
  render() {
    const { open } = this.state
    return (
      <div>
        { // ... more code }
        {
          open ? (
            <Overlay>
              <input type="file" ref={this.refInput} />
            </Overlay>
          ) : null
        }
        { // ... more code }
       </div>
    )
  }

  refInput = (el) => {
    console.log('refInput', el)
  }
}
```

When the `open` flag is set to `true` for the first time, the `refInput` is called with a DOM element as expected. 
Now, when the `open` flag is set to `false`, the `refInput` is called with `null` - as expected.

The problem appears when we set the `open` flag to `true` again. `refInput` is never called with a DOM element again. 

Interestingly, when we replace `<Overlay/>` with a `<div/>` - `refInput` gets called as expected.

Here is an example illustrating the issue: https://jsfiddle.net/3hvru6ex/2990/
And here the same example in React, working correctly: https://jsfiddle.net/n5u2wwjg/206396/

38elements:
When the HTML Element in `recyclerComponents` is used, 
`ref` which is owned by the HTML Element is same as `ref` which is owned by VNode.
Therefore, `ref` is not reseted.

https://github.com/developit/preact/blob/329d7c14140624d8b194f04f6c093f7f55e19d97/src/vdom/diff.js#L335

### DIFF
--- a/src/vdom/diff.js
+++ b/src/vdom/diff.js
@@ -270,6 +270,17 @@ function innerDiffNode(dom, vchildren, context, mountAll, isHydrating) {
 }
 
 
+/**
+ * Remove ref if ref exists in props.
+ * @param {import('../dom').PreactElement} node DOM node
+ */
+export function removeRef(node) {
+	if (node[ATTR_KEY]!=null) {
+		applyRef(node[ATTR_KEY].ref, null);
+		delete node[ATTR_KEY].ref;
+	}
+}
+
 
 /**
  * Recursively recycle (or just unmount) a node and its descendants.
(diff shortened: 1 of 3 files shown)


src/vdom/diff.js
```diff
@@ -287,7 +287,10 @@ export function recollectNodeTree(node, unmountOnly) {
 	else {
 		// If the node's VNode had a ref function, invoke it with null here.
 		// (this is part of the React spec, and smart for unsetting references)
-		if (node[ATTR_KEY]!=null) applyRef(node[ATTR_KEY].ref, null);
+		if (node[ATTR_KEY]!=null) {
+			applyRef(node[ATTR_KEY].ref, null);
+			delete node[ATTR_KEY].ref;
```
developit: This just always calls refs. I think we need to either hoist ref invocation back out of `setProperty()` (it used to happen during rendering), or wait for component recycling to go away.

developit:
Awesome - thanks for putting this together.  I think we'll need a more durable fix than always re-invoking ref() even if it didn't change, but this test is useful regardless of what we end up choosing there.

38elements:
@developit
Thank you for reviewing.
Since I can not come up with a better fix,
Would you please resolve this issue.

PR #1234, opened 2018-10-09
Adds in forwardRef

tbranyen:
- [ ] Verify solution is correct
- [ ] Write unit tests
- [ ] Get @developit's approval

### DIFF
--- a/src/preact.js
+++ b/src/preact.js
@@ -9,11 +9,16 @@ function createRef() {
 	return {};
 }
 
+function forwardRef(render) {
+	return { $$typeof: Symbol.for('react.forward_ref'), render };
+}
+
 export default {
 	h,
 	createElement,
 	cloneElement,
 	createRef,
+	forwardRef,
 	Component,
 	render,
 	rerender,
@@ -25,6 +30,7 @@ export {
 	createElement,
 	cloneElement,
 	createRef,
+	forwardRef,
 	Component,
 	render,
 	rerender,
--- a/src/vdom/diff.js
+++ b/src/vdom/diff.js
@@ -38,7 +38,7 @@ export function flushMounts() {
  * Apply differences in a given vnode (and it's deep children) to a real DOM Node.
  * @param {import('../dom').PreactElement} dom A DOM node to mutate into the shape of a `vnode`
  * @param {import('../vnode').VNode} vnode A VNode (with descendants forming a tree) representing
- *  the desired DOM structure
+ *	the desired DOM structure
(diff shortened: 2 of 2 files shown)


developit:
This actually seems pretty clever.

src/vdom/diff.js
```diff
@@ -114,6 +114,12 @@ function idiff(dom, vnode, context, mountAll, componentRoot) {
 	}
 
 
+	// Support finding forwarded refs and render them as the next children.
+	if (vnode.nodeName.$$typeof === Symbol.for('react.forward_ref')) {
+		vnode = vnode.nodeName.render(vnode.props, vnode.ref);
```
tbranyen: This should be attributes, not props.

src/preact.js
```diff
@@ -9,11 +9,16 @@ function createRef() {
 	return {};
 }
 
+function forwardRef(render) {
+	return { $$typeof: Symbol.for('react.forward_ref'), render };
```
developit: No need to match React's format here - we don't use Symbol for browser compat and performance reasons. I think this could be:


```
// in src/constants.js
const FORWARD_REF = function() {};

function forwardRef(render) {
  return h(FORWARD_REF, null, render);
}

// processing step (from below):
	if (vnode.nodeName===FORWARD_REF) {
		const props = getNodeProps(vnode)
		vnode = vnode.children[0](props, vnode.ref);

```

borovik96:
Guys, are there any updates here?

rshtg:
We are extremely need to use it :)

mabels:
I also stuck on this, would be nice to get this integrated soon.

marvinhagemeister:
ping @tbranyen :)

jessicabyrne:
@tbranyen and @developit any plans to get this in? We're blocked on a preact project due to forward refs from another project we integrate with. Thanks!

marvinhagemeister:
@jessicabyrne Note that `forwardRef` are part of our next major release titled `Preact X`. The alpha is scheduled to land on March 4th. See https://twitter.com/marvinhagemeist/status/1097973028426788864 :+1:

marvinhagemeister:
Closing, we just merged `forwardRef` into master via #1302 .

tbranyen:
@marvinhagemeister awesome!

marvinhagemeister:
@jessicabyrne It's out: https://github.com/developit/preact/releases 🎉 You can install it via `npm install preact@next` :)

PR #1239, opened 2018-10-17
Touch Events to support Passive option

aminshoman:
Having blocking touch events affect the scrolling of the page.
https://alligator.io/js/speed-up-scroll-events/

You can reproduce this case by having an image galleru carsoel on a page, and try to swipe the page vertically, you will notice that the is unable to scroll!
see this popular npm libarry not working correctly with preact: https://github.com/akiran/react-slick


andrewiggins:
This PR has been open for a while without much activity so gonna close this. Definitely still interested in supporting this scenario if we can reach a consensus with people about it. If you'd still like to pursue this, open a new discussion where we can participate in the design.

PR #1243, opened 2018-10-20
Directly add logo to README

reviewher:
removing dependency on the soon-to-be-sunsetted rawgit service


developit:
I might move this to imgur or something to keep the repo root clean, or hotlink it from preactjs.com

reviewher:
It can be moved to a subdirectory.  In any case, most projects include the logo in the project directly.

PR #1246, opened 2018-10-24
[types] Add more type tests for h function

gpoitch:
Fixes & tests type errors for the `h` function when passing a component instead of a string.

Without this, the following fails type checking 2x:
```ts
h(MyComponent, { foo: 'bar' })
```

1. `Argument of type 'typeof MyComponent' is not assignable to parameter of type 'string'`
Solution: component h function needs to overload the string version (see #1214)
2. `Object literal may only specify know properties, and 'foo' does not exist in type 'Attributes & DummyProps'`
Solution: Added missing `Record<string, any>` to params

--- 

Edit: This was fixed by #1214.  This PR just adds more tests.

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -101,14 +101,15 @@ declare namespace preact {
 		abstract render(props?: RenderableProps<P>, state?: Readonly<S>, context?: any): ComponentChild;
 	}
 
-	function h<P>(
-		node: ComponentFactory<P>,
-		params: Attributes & P | null,
-		...children: ComponentChildren[]
-	): VNode<any>;
 	function h(
 		node: string,
 		params: JSX.HTMLAttributes & JSX.SVGAttributes & Record<string, any> | null,
+		...children: ComponentChildren[]
+	): VNode<any>;
+
+	function h<P>(
+		node: ComponentFactory<P>,
+		params: Attributes & P & Record<string, any> | null,
 		...children: ComponentChildren[]
 	): VNode<any>;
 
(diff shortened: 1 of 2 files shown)


src/preact.d.ts
```diff
@@ -101,17 +101,18 @@ declare namespace preact {
...
-	): VNode<any>;
 	function h(
 		node: string,
 		params: JSX.HTMLAttributes & JSX.SVGAttributes & Record<string, any> | null,
 		...children: ComponentChildren[]
 	): VNode<any>;
 
+	function h<P>(
+		node: ComponentFactory<P>,
+		params: Attributes & P & Record<string, any> | null,
```
marvinhagemeister: Were you running into an issue with the `props` typings with the `Records<string, any>`?

test/ts/preact.tsx
```diff
@@ -39,6 +39,8 @@ function DummerComponent({ input, initialInput }: DummerComponentProps) {
 	return <div>Input: {input}, initial: {initialInput}</div>;
 }
 
+render(h('div', { key: "1" }), document);
```
marvinhagemeister: +1 for tests 🎉

src/preact.d.ts
gpoitch: Checking again, we actually don't need this change

gpoitch:
@marvinhagemeister after #1214 was merged, I actually don't need any other changes.  Therefore, this PR simply adds tests.  I'll update the title accordingly.

marvinhagemeister:
@gpoitch Awesome, thank you so much 👍💯

PR #1249, opened 2018-11-02
Fix AnyComponent with ComponentConstructor

scurker:
`AnyComponent` should be allowed to be a functional or class component.

Updated the typescript definition to allow for a component constructor and added a check to one of the typescript tests.


tassoevan:
@marvinhagemeister Why was it reverted?

PR #1251, opened 2018-11-05
Build properly when installing from git

garybernhardt:
Installing directly from a `git://` URL (or directly from GitHub) currently fails because the build script is never run. This PR adds a "prepare" NPM script that does the build. I've confirmed that my fork is directly installable from git.

(Motivation: I want to use #1214 ASAP. It makes the difference between frequent useless TypeScript errors and highly descriptive errors. Until a release happens, that means installing directly from git.)


developit:
Sorry for taking so long to merge this!

PR #1253, opened 2018-11-09
Pluggable recycling pr

staeke:
This PR is a way to allow a pluggable recycling via two new options: `recycle` and `reclaimRecycledBase`. I understand that the future holds a removal of this. But...this would help people in the meantime. And, possibly allow for recycling as an opt-in future wise. Default implementation could then be a no-op.

. With this PR:
- you can control when to do garbage collection
- you don't have to keep references to unused component instances (not HTML elements)
- you can use a more efficient Map structure
- you can allow components to opt out from recycling.

This addresses
https://github.com/developit/preact/issues/1208
https://github.com/developit/preact/issues/957
https://github.com/developit/preact/issues/716
...and possibly others

FYI: Our internal recycling plugs look like this:

```js
const recycled = new Map();
options.recycle = (component, base) => {
    if ('shouldRecycle' in component && !(component.shouldRecycle()))
        return;

    const arr = map.get(component.constructor);
    if (!arr) {
       arr = [];
       map.set(component.constructor, arr);
    }
    arr.push(base);
};
options.reclaimRecycledBase = (ctor) => {
    const arr = recycled.get(ctor);
    return arr && arr.pop();
};
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "preact",
-  "version": "8.3.0",
+  "version": "8.3.1",
   "description": "Fast 3kb React alternative with the same modern API. Components & Virtual DOM.",
   "main": "dist/preact.js",
   "jsnext:main": "dist/preact.mjs",
--- a/src/options.js
+++ b/src/options.js
@@ -14,6 +14,8 @@
  * @property {(component: Component) => void} [beforeUnmount] Hook invoked immediately before a component is unmounted.
  * @property {(rerender: function) => void} [debounceRendering] Hook invoked whenever a rerender is requested. Can be used to debounce rerenders.
  * @property {(event: Event) => Event | void} [event] Hook invoked before any Preact event listeners. The return value (if any) replaces the native browser event given to event listeners
+ * @property {(component: Component, base: HTMLElement) => void} [recycle] Stores a component's base (element) for later reuse. Default implementation reuses per component type.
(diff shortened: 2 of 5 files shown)


developit:
Hiya @staeke! This looks really interesting.  As you noted, it's difficult to know what to do with recycling since we've committed to its removal, but I'm wondering if an opt-in solution could be bolted on using hooks like these.

I can't find an email for you here or on Twitter, but would you be able to join our Slack? https://chat.preactjs.com

staeke:
Thanks for the reply @developit 

I wasn't able to follow that link. Just says "project not found" after proceeding on a bad ssl cert. My e-mail is staffan.eketorp@gmail.com if you want to e-mail or send me an invite.

Regarding the question at hand - yes I personally think it's a great hook for providing element recycling as an opt-in.

marvinhagemeister:
@staeke https://preact.slack.com/ should work 👍

staeke:
@marvinhagemeister that does seem to lead right, but I don't have an account in that workspace and "need an invite". Care to send me one?

marvinhagemeister:
@staeke yup, send an invite to your address 🎉

marvinhagemeister:
We completely removed the recycler in Preact X. It was introduced as a means to get more performance but turned out to be the source for a lot of bugs. With Preact X the recycler isn't that useful anymore and we find new ways to make `diff()` faster :tada:

PR #1254, opened 2018-11-14
document private vars to alert users if they're shadowing them

staeke:
This is to avoid bugs that can be caught by Typescript


src/preact.d.ts
```diff
@@ -99,6 +99,17 @@ declare namespace preact {
 		forceUpdate(callback?: () => void): void;
 
 		abstract render(props?: RenderableProps<P>, state?: Readonly<S>, context?: any): ComponentChild;
+
+		// Add these variables to avoid descendants shadowing them
+		private __key;
```
developit: In the Preact codebase, any properties with a leading underscore get minified using the mapping from here:
https://github.com/developit/preact/blob/master/config/properties.json

In addition, `next*` and `prev*` are also minified.  We might as well add them all if we're looking to warn folks.

staeke: Good catch. I'll add them as well

src/preact.d.ts
```diff
@@ -99,6 +99,32 @@ declare namespace preact {
...
+		private __h;
+		private __k;
+		private __r;
+		private __n;
+		private __b;
+		private __c;
+		private __p;
+		private __s;
+		private __u;
+		private _componentConstructor;
```
developit: nit: These last three are only ever used on HTML Elements.

developit:
@staeke want me to merge and remove the last 3?

staeke:
Sure. Pls

ForsakenHarmony:

Issue #1066, opened 2018-04-18
Potential Optimization: Set String style prop via setAttribute


developit:

marvinhagemeister:
Closing, in the end we all settled for `setProperty` for Preact X to be able to support CSS Custom Properties.

PR #1257, opened 2018-11-18
set string style prop via setAttribute ref: #1066

pdeona:
Changes node.style.cssText assignment to node.setAttribute call. added tests to confirm functionality was unchanged.

ref: #1066

### DIFF
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -35,8 +35,8 @@ export function setAccessor(node, name, old, value, isSvg) {
 		node.className = value || '';
 	}
 	else if (name==='style') {
-		if (!value || isString(value) || isString(old)) {
-			node.style.cssText = value || '';
+		if (!value || typeof value==='string' || typeof old==='string') {
+			node.setAttribute('style', value || '');
 		}
 		if (value && typeof value==='object') {
 			if (!isString(old)) {
--- a/src/dom/index.js
+++ b/src/dom/index.js
@@ -35,7 +35,7 @@ export function setAccessor(node, name, old, value, isSvg) {
 		node.className = value || '';
 	}
 	else if (name==='style') {
-		if (!value || typeof value==='string' || typeof old==='string') {
+		if (!value || isString(value) || isString(old)) {
 			node.setAttribute('style', value || '');
 		}
 		if (value && typeof value==='object') {
--- a/test/browser/render.js
+++ b/test/browser/render.js
@@ -194,7 +194,7 @@ describe('render()', () => {
 			.deep.equal(['style', style]);
 		expect(scratch.childNodes[0]).to.have.deep.property('style.cssText')
 			.that.equals(style);
-		spy.restore()
+		spy.restore();
 	});
 
 	it('object style prop should not call setAttribute', () => {
@@ -203,7 +203,7 @@ describe('render()', () => {
 		render(<div style={style} />, scratch);
 		
 		expect(spy).not.to.have.been.called;
-		spy.restore()
+		spy.restore();
 	});
 
 	it('should only register on* functions as handlers', () => {
--- a/test/browser/render.js
+++ b/test/browser/render.js
@@ -185,6 +185,27 @@ describe('render()', () => {
 			.and.matches(/position\s*:\s*relative\s*/);
 	});
 
+	it('string style should call setAttribute', () => {
+		const style = 'background-color: red;';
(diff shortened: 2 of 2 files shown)


src/dom/index.js
```diff
@@ -35,8 +35,8 @@ export function setAccessor(node, name, old, value, isSvg) {
 		node.className = value || '';
 	}
 	else if (name==='style') {
-		if (!value || isString(value) || isString(old)) {
-			node.style.cssText = value || '';
+		if (!value || typeof value==='string' || typeof old==='string') {
```
bekicot: Why replacing `isString` with `typeof ...` ?

bekicot: `isString` does exactly the same, and nicer to read, in my opinion 😄 .

pdeona: accidentally changed it when i forked from `master`, sorry! :D

test/browser/render.js
```diff
@@ -185,6 +185,27 @@ describe('render()', () => {
...
+			.that.equals(style);
+		spy.restore();
+	});
+
+	it('object style prop should not call setAttribute', () => {
+		const style = { backgroundColor: 'papayawhip' };
+		const spy = sinon.spy(Element.prototype, 'setAttribute');
+		render(<div style={style} />, scratch);
+		
+		expect(spy).not.to.have.been.called;
```
bekicot: Do you mean `expect(spy).not.to.have.been.called()` ?

pdeona: the type indicates that it is a property, not a function:

```
        interface Assertion {
            /**
             * true if the spy was called at least once.
             */
            called: Assertion;
```

is this incorrect?

test/browser/render.js
```diff
@@ -185,6 +185,27 @@ describe('render()', () => {
...
+		
+		expect(spy.firstCall.args).to
+			.deep.equal(['style', style]);
+		expect(scratch.childNodes[0]).to.have.deep.property('style.cssText')
+			.that.equals(style);
+		spy.restore();
+	});
+
+	it('object style prop should not call setAttribute', () => {
+		const style = { backgroundColor: 'papayawhip' };
```
bekicot: If it changes style attribute, as in your implementation, it should call the `setAttribute` right?

pdeona: no, `object` style props are passed to the object handler:

```
			if (!isString(old)) {
				for (let i in old) if (!(i in value)) node.style[i] = '';
			}
			for (let i in value) {
				node.style[i] = typeof value[i]==='number' && !NON_DIMENSION_PROPS[i] ? (value[i]+'px') : value[i];
			}
		}
```

test/browser/render.js
pdeona: changing to a function call throws an error. called is meant to be accessed as a property.

bekicot: Based on the example here `https://www.chaijs.com/plugins/chai-spies/` it is a func. Maybe you are right. I didn't know it works that way.

pdeona: the docs may be out of date. the types are directly from the repo

src/dom/index.js
```diff
@@ -36,7 +36,7 @@ export function setAccessor(node, name, old, value, isSvg) {
 	}
 	else if (name==='style') {
 		if (!value || isString(value) || isString(old)) {
-			node.style.cssText = value || '';
+			node.setAttribute('style', value || '');
```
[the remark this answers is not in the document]
pdeona: while it would be a smaller line of code, some (admittedly non-rigorous) benchmarking indicates setAttribute is about 30% faster:

http://jsben.ch/GJ0dD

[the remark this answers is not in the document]
developit: setAttribute ignores duplicate assignments, which will never happen here because this function is bypassed when values are equal.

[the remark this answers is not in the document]
developit: Alright I dropped your benchmark into ESBench to double-check, and it shows that both setAttribute and style= perform poorly compared to style.cssText:

https://esbench.com/bench/5bf437e74cd7e6009ef61ce5

<a href="https://esbench.com/bench/5bf437e74cd7e6009ef61ce5">screen shot 2018-11-20 at 11 46 23 am</a>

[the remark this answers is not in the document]
pdeona: can’t argue there, i’ll go ahead and close

PR #1272, opened 2018-12-10
Add `createRef` to Flow definition

ngyikp:
`createRef` was added in https://github.com/developit/preact/pull/1138 however the Flow definition wasn’t added, so using it on Flow codebases was inconvenient.


ForsakenHarmony:

PR #1276, opened 2018-12-17
Add browser field inside package.json for umd build

egdbear:


developit:
@egdbear - think `umd` is the best default here? I'm fine with it, but there's also a smaller version that is designed for browser loading and doesn't support AMD:
https://unpkg.com/preact@8.4.2/dist/preact.min.js

egdbear:
@developit i guess its by common standard to be UMD build on browser field. Tell me if i am wrong :)

developit:
Nah just figured I'd ask haha.

ForsakenHarmony:

developit:
@ForsakenHarmony  I think we may need to revert this. A few bundlers give precedence to `package.umd` over `package.module`, which means we've effectively disabled tree shaking by adding the umd field. Maybe it's fine to just stick with the unpkg field?

lukeed:
Agreed. I've actually been putting ESM inside the `browser` field; since the tooling that listens to that already handle ESM anyway.

PR #1286, opened 2019-01-09
Fix TypeScript type checking of setState with function argument.

### LINKED ISSUE #1284: TypeScript type definitions don't enforce setState argument type
**Steps to reproduce**:

1. In a fresh directory,
2. `yarn install -D preact typescript`
3. Save the code below as `test.ts`.
4. `tsc test.ts`

```typescript
import { Component } from 'preact'

// MyType is a type that requires a particular field
export interface MyType {
  myKey: string
}

// Our state contains a MyType
interface State {
  myObject: MyType
}

export default class MyComponent extends Component<{}, State> {
  initialize() {
    // We can set the field of type MyType to anything: the empty object, an
    // object with the wrong keys, or even values that aren't objects at all.
    // It seems like the type of the state isn't actually checked in any way.
    this.setState(() => {myObject: {thisIsWrong: "some-value"}})
    this.setState(() => {myObject: {}})
    this.setState(() => {myObject: 1})

    // We can also set keys that don't exist on the state at all. This *may* be
    // intentional; I'm not sure. But I include it here because it seems like
    // it might be related.
    this.setState(() => {thisIsWrong: "some-value"})
  }

  render() {
    return null
  }
}

// Just to have a very explicit comparison point: TypeScript correctly type
// checks the same MyType when used to declare an argument to a function. Both
// of the g() calls produce type errors because `myKey` is missing. (These are
// the two build errors that you'll see if you tsc this file.)
function g(arg: MyType) {
}
g({thisIsWrong: "some-value"})
g({})
```

**Expected result**:

The `setState` calls shouldn't type check. In the first three calls, the type of `myObject` is wrong. In the last one, the key being set doesn't exist on the state at all.

**Actual result**:

Any call to `setState` type checks regardless of its keys or the types of its values. No type checking of state seems to take place. Running `tsc tests.ts` shows no errors in the mistyped `setState` calls. Note that tsc **will** report two errors, but those don't involve `setState` and are included only as a comparison point.

**Notes**:

The above steps were verified with preact 8.4.2 and typescript 3.2.2, the latest versions as of today.

Read the comments in the reproduction source file, in order, to get a short narrative explaining in more detail.

I've searched the issue history and found no issues about this. Of course, it's possible that I'm missing something here. But there's no tsc configuration involved, so I'm at a loss for what it could be! This problem was originally found in a project where tsc is in strict mode, so I know that turning strict mode on doesn't help.

utkarshkukreti:
Fixes #1284.

See https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e836acc75a78cf0655b5dfdbe81d69fdd4d8a252/types/react/index.d.ts#L402
and https://github.com/DefinitelyTyped/DefinitelyTyped/issues/18365#issuecomment-351013257
for why this change is required.

Before:

    $ cat a.tsx
    import { Component } from './src/preact';

    class A extends Component<{}, { a: number }> {
      a() {
        this.setState(() => ({ a: 'a' }));
      }

      render() {
        return null;
      }
    }

    $ tsc --noEmit a.tsx && echo ok
    ok

After:

    $ tsc --noEmit a.tsx
    a.tsx:5:19 -     error TS2345: Argument of type '() => { a: string; }' is not assignable to parameter of type '{ a: number; } | Pick<{ a: number; }, "a"> | ((prevState: Readonly<{ a: number; }>, props: Readonly<{}>) => { a: number; } | Pick<{ a: number; }, "a">)'.
      Type '() => { a: string; }' is not assignable to type '(prevState: Readonly<{ a: number; }>, props: Readonly<{}>) => { a: number; } | Pick<{ a: number; }, "a">'.
        Type '{ a: string; }' is not assignable to type '{ a: number; } | Pick<{ a: number; }, "a">'.
          Type '{ a: string; }' is not assignable to type 'Pick<{ a: number; }, "a">'.
            Types of property 'a' are incompatible.
              Type 'string' is not assignable to type 'number'.

    5     this.setState(() => ({ a: 'a' }));

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -95,8 +95,14 @@ declare namespace preact {
 		context: any;
 		base?: HTMLElement;
 
-		setState<K extends keyof S>(state: Pick<S, K>, callback?: () => void): void;
-		setState<K extends keyof S>(fn: (prevState: S, props: P) => Pick<S, K>, callback?: () => void): void;
+		// From https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e836acc75a78cf0655b5dfdbe81d69fdd4d8a252/types/react/index.d.ts#L402
+		// // We MUST keep setState() as a unified signature because it allows proper checking of the method return type.
+		// // See: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/18365#issuecomment-351013257
+		// // Also, the ` | S` allows intellisense to not be dumbisense
+		setState<K extends keyof S>(
+			state: ((prevState: Readonly<S>, props: Readonly<P>) => (Pick<S, K> | S | null)) | (Pick<S, K> | S | null),
+			callback?: () => void
+		): void;
 
 		forceUpdate(callback?: () => void): void;
 


developit:
Looks great! Thanks for picking this one up, really appreciate it.

garybernhardt:
I reported the original issue #1284, so I pulled this down and tested it. It definitely seems to correct the problem! It found an actual type error in our system, where we were calling setState with an object that violated the declared type. Thanks for fixing this, @utkarshkukreti. I'm excited to see this merged so we can depend on the new version!

ForsakenHarmony:

garybernhardt:
@ForsakenHarmony Thanks for doing that release, Leah. I just upgraded us to 8.5.0 and immediately found one bug with our state management because of this fix!

PR #1292, opened 2019-01-22
Query recyclerComponents array length at its usage point

kuhe:
#### Question

Related commit: https://github.com/developit/preact/commit/7cbaa2ac0141ad7c98f6a1a960a4e208d25c166f#diff-a480bde1c2f4c6ab259cbb4d81e21199

It's possible to change the length of `recyclerComponents` between its read at the start of `createComponent` and the time iteration starts. 

I hear this is going away, but currently running into this on 8.3.1 and I see it's still in master for 8.x.

Perhaps this is mainly due to misuse on my part, but the current workaround I'm testing is to modify the distributable in a similar way as this PR.

** couldn't find any associated issues, PRs, or contribution guidelines
Is the label "The Wreckoning" about the recycler?

### DIFF
--- a/src/vdom/component-recycler.js
+++ b/src/vdom/component-recycler.js
@@ -17,7 +17,7 @@ export const recyclerComponents = [];
  * @returns {import('../component').Component}
  */
 export function createComponent(Ctor, props, context) {
-	let inst, i = recyclerComponents.length;
+	let inst;
 
 	if (Ctor.prototype && Ctor.prototype.render) {
 		inst = new Ctor(props, context);
@@ -29,6 +29,7 @@ export function createComponent(Ctor, props, context) {
 		inst.render = doRender;
 	}
 
+	let i = recyclerComponents.length;
 
 	while (i--) {
 		if (recyclerComponents[i].constructor===Ctor) {


kuhe:
Here is an example: https://codesandbox.io/s/xoy39rjzmo
```tsx
// code from the example:
import { h, Component, render } from "preact";

class A extends Component {
  constructor() {
    super();
    render(h(B, {}), document.body); // preact.render, not this.render.
  }
  render() {
    return h("a", {});
  }
}
class B extends Component {
  render() {
    return h("b", {}, this.props.children);
  }
}
let merge;
merge = render(h(B, {}, [h(B, {}), h(B, {}), h(B, {})]), document.body);
merge = render(h(B, {}), document.body, merge);
merge = render(h(A, {}), document.body, merge);
```

The example does the following:

- create some `B` components, then discard them
- create an `A` component whose constructor recycles one of the `B` components.
- the recycling action mutates the array before A's code path through `createComponent` completes, and crashes.

I know, you wouldn't write components like this, but on the balance I still suggest the fix be made in the Preact repo.

- fix is small in scope
- intent of this code is array iteration, not to stop you from attempting to call render from a ctor
- nothing in the API suggests this is illegal (?) 
- and unknown scope of real world applicability. I don't want to excessively outline the details of the whole thing but I feel my use case in which I ran into this problem is valid.

marvinhagemeister:
FYI: We completely removed the recycler from our next major version. An alpha is scheduled to be released on March 4th. See https://twitter.com/marvinhagemeist/status/1097973028426788864

marvinhagemeister:
We just merged the code for Preact X into `master` :tada: Closing, because the recycler is no more :+1:

PR #1297, opened 2019-02-08
FIX: Allow desc element for SVGs

programbo:
Add `<desc>` elements to the Typescript typings for accessibility. (https://developer.mozilla.org/en-US/docs/Web/SVG/Element/desc)

Fixes https://github.com/developit/preact/issues/1296


ForsakenHarmony:

PR #1305, opened 2019-03-04
Add back createFactory to compat

marvinhagemeister:
Missed that when porting `compat` to `ceviche`. Thanks for catching this @pl12133 :tada:

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -19,6 +19,14 @@ options.event = e => {
 };
 
 /**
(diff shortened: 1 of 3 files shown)


marvinhagemeister:
Looks like coveralls is overloaded and it looks like I have no permission to retry the build.

developit:
I kicked the build. Not sure Travis has a way to let builds restart, annoyingly.

PR #1308, opened 2019-03-04
Sort DOM attributes before assertion

marvinhagemeister:
Browsers do sort DOM attributes differently. Ported over from the `8` branch.


marvinhagemeister:
Replaced by #1311

PR #1310, opened 2019-03-04
Fix mangle.json config

andrewiggins:
Add a missing mangle prop in hooks, and remove a property config that overrode an existing one (`__c` used for `_context` overrode `__c` used for `_constructor`).

### DIFF
--- a/hooks/mangle.json
+++ b/hooks/mangle.json
@@ -7,6 +7,7 @@
     "cname": 6,
     "props": {
       "$__hooks": "__H",
+      "$_component": "__c",
       "$_defaultValue": "__p",
       "$_id": "__c"
     }
--- a/mangle.json
+++ b/mangle.json
@@ -18,7 +18,6 @@
       "$__html": "__html",
       "$_ancestorComponent": "__a",
       "$_processingException": "__p",
-      "$_context": "__c",
       "$_defaultValue": "__p",
       "$_id": "__c"
     }
--- a/mangle.json
+++ b/mangle.json
@@ -18,6 +18,7 @@
       "$__html": "__html",
       "$_ancestorComponent": "__a",
       "$_processingException": "__p",
+      "$_context": "__n",
       "$_defaultValue": "__p",
       "$_id": "__c"
     }


mangle.json
```diff
@@ -18,7 +18,6 @@
       "$__html": "__html",
       "$_ancestorComponent": "__a",
       "$_processingException": "__p",
-      "$_context": "__c",
```
andrewiggins: I didn't see other code using `_context` so I thought I'd just remove it and let terser decide what the best minification of this prop is.

marvinhagemeister: I think we added that at some point when working on the new context api. An early iteration had a `_context` property that was shared.

developit: It's used. FWIW we could probably have Terser leave this as `_context` - I'd imagine that's fewer bytes than `__c`.

https://github.com/developit/preact/blob/ed4de76051256f093ca6611c84495f94e68ef6df/src/component.js#L75

https://github.com/developit/preact/blob/ed4de76051256f093ca6611c84495f94e68ef6df/src/diff/index.js#L82

marvinhagemeister: But it's only used in core. Because of that it shouldn't need to have a consistent property, or maybe I miss something.

developit: Definitely less important, true. The consistent property mapping does two things we still want though: it helps to avoid it changing on every build, and it ensures the minified property has leading underscores instead of making it something easily collided with like `component.a`.

marvinhagemeister: You're right, didn't think of possible name clashes :+1:

hooks/mangle.json
```diff
@@ -7,6 +7,7 @@
     "cname": 6,
     "props": {
       "$__hooks": "__H",
+      "$_component": "__c",
```
developit: we can delete this now that #1313 is merged.

```suggestion
```

mangle.json
andrewiggins: Got it, I'll undo and play with this to get it consistent, correct, and small

andrewiggins: Updated! `_context` turns out to be bigger (probably because there are already so many `__` props). `__n` didn't change the size from the original so I went with it. Most other shortcuts I tried (e.g. `__ctxt`, `__ct`) were bigger.

PR #1311, opened 2019-03-04
Tests: deterministic innerHTML

developit:
This subs in a deterministic innerHTML implementation during tests.

### DIFF
--- a/test/_util/helpers.js
+++ b/test/_util/helpers.js
@@ -4,6 +4,77 @@ import { clearLog, getLog } from './logCall';
 
 /** @jsx h */
 
+
+const VOID_ELEMENTS = /^(area|base|br|col|embed|hr|img|input|link|meta|param|source|track|wbr)$/;
+
+const el = document.createElement('div');
+const innerHtmlDef = Object.getOwnPropertyDescriptor(el.ownerDocument.defaultView.Element.prototype, 'innerHTML');
+function encodeEntities(str) {
+	el.textContent = str;
+	return innerHtmlDef.get.call(el);
+}
+
+/**
+ * Serialize a DOM tree.
+ * Uses deterministic sorting where necessary to ensure consistent tests.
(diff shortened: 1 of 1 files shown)


developit:
implementation is basically ripped from preact-render-to-string. though I did the determinism bits and then only realized after that `preact-render-to-string/jsx` has this already lol

test/_util/helpers.js
```diff
@@ -4,6 +4,77 @@ import { clearLog, getLog } from './logCall';
...
+		return '<!--' + encodeEntities(node.data) + '-->';
+	}
+	else if (node.nodeType === 1 || node.nodeType === 9) {
+		let str = '<' + node.localName;
+		const attrs = [];
+		for (let i=0; i<node.attributes.length; i++) {
+			attrs.push(node.attributes[i].name);
+		}
+		attrs.sort();
+		for (let i=0; i<attrs.length; i++) {
```
cristianbote: Do you think we could merge these two loops, into one? Would it make sense?

```js
for (let i=0; i<node.attributes.length; i++) {
    const name = attrs[i];
    const value = node.getAttribute(name);
    if (value == null) continue;
    if (!value && name==='class') continue;
    attrs.push(node.attributes[i].name + '="' + encodeEntities(value) + '"');
}
// This will sort the _entire_ list _compiled_ already
attrs.sort();

// Join them with an empty space
str += attrs.join(" ");
```

cristianbote: Pff, I was too late 😄

marvinhagemeister: Oh sorry, I was too quick with the merge I guess :S

developit:
yup yup. for tests I'm not super worried, I didn't want to end up in the situation where we're dealing with sort() for an array like `['a="b"', 'a', 'a$="b"']` lol

cristianbote:
Indeed 🙂 gonna check the render to string package, as you mentioned, and see if this could apply there?

PR #1317, opened 2019-03-04
Inline vnode children

developit:
Decreases size. Further decrease with developit/microbundle#335.


andrewiggins:
We can do this now or wait for my Fragments PR. Doesn't matter too much to me. Doing this was a to-do in my branch :)

marvinhagemeister:
@andrewiggins yeah, we didn't want to interfere with your indices based ceviche PR. You can do the call to merge here :)

PR #1320, opened 2019-03-05
Add missing _parentDom mangle prop for hooks

andrewiggins:
`_parentDom` is used by `useEffect` to ensure a component is mounted before invoking its effect. So we need to consistently mangle the `_parentDom` property so it can be shared between them.

### DIFF
--- a/hooks/mangle.json
+++ b/hooks/mangle.json
@@ -9,7 +9,8 @@
       "$__hooks": "__H",
       "$_component": "__c",
       "$_defaultValue": "__p",
-      "$_id": "__c"
+      "$_id": "__c",
+      "$_parentDom": "__P"
     }
   }
 }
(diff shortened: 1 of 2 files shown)


ForsakenHarmony:
well that explains why I never ran into this issue, always used the unminified version 🙃

timpur:
@developit Wondering if its possible to run Coverage against mangled code ? May this would catch some of these simple bugs ?

marvinhagemeister:
@timpur We have the same idea 👍 I'm currently working on making our whole test suite run against the compiled production bundles.

timpur:
@marvinhagemeister :)

Think there are still issues with DevTools, similar to https://github.com/developit/preact/issues/1330, hopefully this can help pick them up :) (haven't created a issue yet though).

marvinhagemeister:
@timpur good idea👍 I've been focussing on other bugs first but I'm planning to check the devtools next week. If you have any links to repos or something for me to look at, feel free to ping me 🙂

PR #1336, opened 2019-03-05
Prettier

kristoferbaxter:
Enables `prettier` using `lint-staged`. This is not ready to commit, since there are conflicts between the rules from @developit's eslint config and `prettier` settings.


package.json
```diff
@@ -48,7 +47,104 @@
       "react/sort-comp": 0,
       "jest/valid-expect": 0,
       "jest/no-disabled-tests": 0,
-      "react/no-find-dom-node": 0
+      "react/no-find-dom-node": 0,
```
ForsakenHarmony: please extract this

also see https://github.com/developit/preact-cli/blob/next/.eslintrc#L4

.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "printWidth": 120,
+  "trailingComma": "all",
+  "singleQuote": true,
+  "useTabs": false
```
developit: are we going to 2-space? :(

marvinhagemeister:
Based on this branch we merged prettier in #2065 🎉

PR #1338, opened 2019-03-05
Fix saucelabs cross browser test assertions

marvinhagemeister:
- [x] sort DOM attributes
- [x] sort `cssText`
- [x] make svg tests deterministic
- [x] make devtools tests deterministic
- [x] ~~Skip devtools tests in non-chrome browsers (maybe keep Firefox?)~~
- [x] Skip css-grid tests in browsers that don't support it
- [x] Sort css with CSS Custom Properties
- [x] IE11: Fix `parentNode` is undefined or null error
- [x] IE11: Fix return value of `Object.getOwnPropertyDescriptor` being undefined
- [x] IE11: Add `Promise` polyfill
- [x] IE11: Add polyfill for `Function.name`
- [x] Normalize event creation across all browsers in tests
- [x] Safari: Fails on `<input list>` test
- [x] Browsers throwing on falsy attribute values that are not valid per spec
- [x] Drop commit with failed `Object.getOwnPropertyDescriptor` polyfill
- [x] Drop saucelabs hack

### DIFF
--- a/compat/test/browser/Children.test.js
+++ b/compat/test/browser/Children.test.js
@@ -1,4 +1,4 @@
-import { setupScratch, teardown } from '../../../test/_util/helpers';
+import { setupScratch, teardown, serializeHtml } from '../../../test/_util/helpers';
 // eslint-disable-next-line no-unused-vars
 import React, { Children, render } from '../../src';
 import { div, span } from '../../../test/_util/dom';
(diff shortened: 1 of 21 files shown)


marvinhagemeister:
It's coming along, but IE11 is still not easy to work with. I'm sorry for the email spam, but I haven't been able to connect to saucelabs locally even though the credentials are correct...

marvinhagemeister:
Only 2 tests remaining :muscle:

marvinhagemeister:
All tests passing :tada: I'll take a quick break and grab some groceries before cleaning up the PR and making it ready to merge :+1:

PR #1340, opened 2019-03-05
Add hgroup to typings

ForsakenHarmony:


hbroer:
hgroup is obsolet and should not be used anymore. Why is this added?
https://caniuse.com/#search=hgroup

ForsakenHarmony:
Emotion uses it in it's typings

marvinhagemeister:
@hbroer it was removed from the W3C HTML spec, but it is still part of the WHATWG HTML spec and implemented in all browser. Use of it is discouraged, but it's still there and will probably remain with us for a while.

See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hgroup#Usage_notes

hbroer:
@marvinhagemeister yes i know, but adding it means that it is just now usable here for new projects. And new projects should not use it. It feels not good to add something obsolete.

Not shure what happens if browsers remove the element. Now i think it is a display:block, but default maybe will later switch to display:inline because that is the default for unknown elements AFAIK.

If that element was in that typings since a while it would be a other thing. If other frameworks supporting it, that does not count as an argument. I would not wonder if font or marquee are supported by some libs too.

My point is clear: Do not add obsolete types, force the developer to use up to date specs. If he want to use obsolete tech, then he can add his own workaround (in this case adding the typings for the obsolete element to his project locally).

Btw: My IDE tells me that hgroup is depricated when i write plain HTML. Not shure if my IDE is doing this when i write JSX. In this case you are hiding a important notice away from the developer.

ForsakenHarmony:
This was added for compatibility (emotions typings had an error because react has it but preact didn't), not just to add it

It's in no way intended to encourage people to use it, but we don't want to break existing code

Issue #1341, opened 2019-03-06
Preact X: Issue with hooks code when running on jest


armand1m:
Disclaimer: I am not using preact hooks yet nor react hooks here =)

When trying to render a component in Jest while using `preact@10.0.0-alpha.0` and`preact-render-to-string@^4.1.0`:

```js
/* eslint-env node, jest */

import { h } from "preact"
import { render } from "preact-render-to-string"

import Widget from "./"

describe("Widget Snapshot", () => {
  it("should render", () => {
    const tree = render(<Widget />);
    expect(tree).toMatchSnapshot();
  });
});
```

I got this error:

```
  ● Test suite failed to run

    ReferenceError: MessageChannel is not defined
      at Object.component (node_modules/preact/hooks/src/index.js:199:16)
      at Object.<anonymous> (node_modules/preact/compat/src/index.js:8:45)
      at Object.<anonymous> (node_modules/rebass/dist/Provider.js:12:14)
      at Object.<anonymous> (src/components/Widget/index.tsx:4:1)
      at Object.<anonymous> (src/components/Widget/index.test.tsx:6:1)
```

It seems that running [these lines of code](https://github.com/developit/preact/blob/master/compat/src/index.js#L2~L3):

```js
import * as hooks from 'preact/hooks';
export * from 'preact/hooks';
```

is preparing the hooks environment making calls to the window even though I am not using hooks at all.

- https://github.com/developit/preact/blob/master/hooks/src/index.js#L196~L214

```js
if (typeof window !== 'undefined') {
	mc = new MessageChannel();

	afterPaint = (component) => {
		if (!component._afterPaintQueued && (component._afterPaintQueued = true) && afterPaintEffects.push(component) === 1) {
			requestAnimationFrame(onPaint);
		}
	};

	mc.port2.onmessage = () => {
		afterPaintEffects.forEach(component => {
			component._afterPaintQueued = false;
			if (!component._parentDom) return;
			component.__hooks._pendingEffects.forEach(invokeEffect);
			component.__hooks._pendingEffects = [];
		});
		afterPaintEffects = [];
	};
}
```

## Workaround

Mock `MessageChannel` in jest `setupFilesAfterEnv` hook:

```js
(function mockMessageChannel() {
  const MessageChannel = function () {
    return {
      port1: jest.fn(),
      port2: jest.fn(),
    };
  };

  global.MessageChannel = MessageChannel;
  global.onmessage = jest.fn();
})();
```

developit:
[...]

PR #1346, opened 2019-03-06
Switch to raf+setTimeout from MC

cristianbote:
This is addressing #1341.

Basically, go with raf + setTimeout solution. There's a possible clamping here, but let's discuss it.

cc @andrewiggins @developit @marvinhagemeister 

~`-33B`~
`-30B`

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -187,9 +187,9 @@ export function useContext(context) {
 let afterPaint = () => {};
 
 /**
- * Pending effects consumer.
+ * After paint effects consumer.
  */
-function onPaint() {
+function flushAfterPaintEffects() {
 	afterPaintEffects.forEach(component => {
 		component._afterPaintQueued = false;
 		if (!component._parentDom) return;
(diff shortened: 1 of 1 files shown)


hooks/src/index.js
```diff
@@ -186,31 +186,25 @@ export function useContext(context) {
...
+	afterPaintEffects = [];
 }
 
 if (typeof window !== 'undefined') {
-	mc = new MessageChannel();
-
 	afterPaint = (component) => {
 		if (!component._afterPaintQueued && (component._afterPaintQueued = true) && afterPaintEffects.push(component) === 1) {
-			requestAnimationFrame(onPaint);
+			requestAnimationFrame(() => setTimeout(onPaint, 0));
```
marvinhagemeister: For a tiny bit more perf we could create this function outside of `afterPaint`.

cristianbote: Good point! Changed it. Thank you!

hooks/src/index.js
```diff
@@ -186,31 +186,29 @@ export function useContext(context) {
  */
 let afterPaint = () => {};
 
-/** @type {MessageChannel} */
-let mc;
-
+/**
+ * Pending effects consumer.
+ */
 function onPaint() {
```
andrewiggins: Perhaps we can give this a more descriptive name? Something like `flushAfterPaintEffects()`

cristianbote: That's a fair point. That would give a bit more context on what it does. Renamed them.

PR #1348, opened 2019-03-06
Fix: componentWillReceiveProps must be called before shouldComponentUpdate

mochiya98:


src/diff/index.js
```diff
@@ -100,13 +100,14 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
 				if (c.componentDidMount!=null) mounts.push(c);
 			}
 			else {
+				if (newType.getDerivedStateFromProps==null && c._force==null && c.componentWillReceiveProps!=null) {
+					c.componentWillReceiveProps(newVNode.props, cctx);
+				}
```
marvinhagemeister: Can you add a test case that checks for the lifecycle order? They can be found in `/test/lifecycle.test.js` 👍

mochiya98:
sorry for delaying! I did it. :+1:

PR #1351, opened 2019-03-06
Fix stale props reference when sCU returns false

### LINKED ISSUE #1345: 10.0.0-alpha.0: shouldComponentUpdate blocks updating component as well as render.
Example:  
props is not updated permanently while shouldComponentUpdate returns false  
```jsx
import { render, Component } from "preact";

class A extends Component {
  state = { a: 0 };
  componentDidMount() {
    this.setState({ a: 1 }); // update props
  }
  render() {
    return <B a={this.state.a} />;
  }
}

class B extends Component {
  state = { now: "---" };
  timer = 0;
  updateState = () => {
    this.setState({ now: new Date().toString() });
  };
  componentDidMount() {
    this.timer = setInterval(this.updateState, 1000);
  }
  componentWillUnmount() {
    clearInterval(this.timer);
  }
    if (nextProps !== this.props) {
      console.log(
        "called infinitely!! it must be called only once.",
        new Date(),
        nextProps.a,
        this.props.a
      );
      // nothing do when props changed
      return false;
    }
    return true;
  }
  render() {
    return (
      <p>
        {this.state.now}, {this.props.a}
      </p>
    );
  }
}

render(<A />, document.getElementById("root"));

```
https://codesandbox.io/s/o57vp43o3q

--------------------
refs:  
https://github.com/developit/preact/blob/753d8849997612f2fa3d168a28ff4dfba65b39bc/src/diff/index.js#L103-L106

marvinhagemeister:
Adds ~~`+3`~~ `+2` bytes 🎉 Fixes #1345 .

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -106,6 +106,7 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
 
 				if (!c._force && c.shouldComponentUpdate!=null && c.shouldComponentUpdate(newVNode.props, s, cctx)===false) {
 					c.props = newVNode.props;
+					c.state = s;
 					c._dirty = false;
 					break outer;
 				}
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -105,6 +105,7 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
 				}
 
 				if (!c._force && c.shouldComponentUpdate!=null && c.shouldComponentUpdate(newVNode.props, s, cctx)===false) {
+					c.props = newVNode.props;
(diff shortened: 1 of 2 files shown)


mochiya98:
thanks! isn't it necessary to update state? maybe causes a similar problem.

marvinhagemeister:
Good point! I'll check if the same issue applies for `state`

developit:
Seems like we'll want state as well, yeah.

marvinhagemeister:
Can confirm: The state is updated but is never referentially equal https://codesandbox.io/s/j00kl29rv

marvinhagemeister:
hold on I somehow messed up the history.

PR #1358, opened 2019-03-07
Make all forwardRef tests pass

marvinhagemeister:
Ported this PR over from the `ceviche` repo.

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -283,8 +283,7 @@ function memo(c, comparer) {
 		let ref = this.props.ref;
 		let updateRef = ref==nextProps.ref;
 		if (!updateRef) {
-			if (typeof ref=='function') ref(null);
-			else ref.current = null;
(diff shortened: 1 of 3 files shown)


compat/src/index.js
```diff
@@ -288,16 +288,23 @@ Component.prototype.isReactComponent = {};
  */
 function memo(c, comparer) {
 	function shouldUpdate(nextProps) {
-		return !comparer(this.props, nextProps);
+		let ref = this.props.ref;
+		let updateRef = ref==nextProps.ref;
+		if (!updateRef) {
+			if (typeof ref=='function') ref(null);
```
cristianbote: Do you think, basically doing something like:

```js
if (!updateRef) {
  ref.call ? ref(null) : (ref.current = null);
}
```

Would shave some bytes?

marvinhagemeister: Ohh that's clever and saves `-2B`! I like your thinking :+1: :tada:

PR #1359, opened 2019-03-07
Release 10.0.0-alpha.1

marvinhagemeister:


hooks/package.json
```diff
@@ -18,7 +18,7 @@
     "test:karma:watch": "karma start test/karma.conf.js --no-single-run"
   },
   "peerDependencies": {
-    "preact": "10.0.0-alpha.0"
+    "preact": "10.0.0-alpha.1"
```
developit: idea: relax these to caret versions?

marvinhagemeister: Good idea, there is no way to install different versions anyway.

PR #1360, opened 2019-03-07
Add _renderers to reserved properties for devtools

calebeby:


calebeby:
I'm going to test the profiler and see if I need to un-mangle any other properties for that

debug/mangle.json
```diff
@@ -8,7 +8,12 @@
       "properties": {
         "regex": "^_",
         "reserved": [
-          "__REACT_DEVTOOLS_GLOBAL_HOOK__"
+          "__REACT_DEVTOOLS_GLOBAL_HOOK__",
+          "_renderers",
+          "emit",
```
developit: Only properties with a leading underscore get minified, so these might not be necessary.

calebeby: Good point! Fixed :heavy_check_mark:

PR #1361, opened 2019-03-07
remove double vnode coercion

developit:
`newChildren` comes from getVNodeChildren(), which coerces vnodes as it collects them.


andrewiggins:
Since we no longer double coerce anymore and after #2040 `coerceToVNode` won't exist anymore, I'mma close this PR. Let me know if you'd like to keep it open!

PR #1362, opened 2019-03-08
Run tests against production bundles

marvinhagemeister:
We've had a bit of trouble with mangling properties which shouldn't be mangled at all in our recent alpha releases. This PR runs the tests against the production bundles, hopefully catching these kinds of errors before we do our next release.

- [x] Optionally mangle test files
- [ ] Mangle and run each package's suite individually

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,13 +1,39 @@
 /*eslint no-var:0, object-shorthand:0 */
 
 var coverage = String(process.env.COVERAGE) === 'true',
+	prod = String(process.env.PRODUCTION) === 'true',
 	ci = String(process.env.CI).match(/^(1|true)$/gi),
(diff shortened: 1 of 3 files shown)


karma.conf.js
```diff
@@ -151,20 +177,43 @@ module.exports = function(config) {
 				// rather than referencing source files inside the module
 				// directly
 				alias: {
-					'preact/hooks': path.join(__dirname, './hooks/src'),
```
andrewiggins: If we just change these aliases to the directory that contains the package.json, does webpack pick up our build output? That way we don't have to minify again in the karma-webpack and are testing the same build output our users will consume.

developit: It'll break watch mode, but otherwise I think this might be good - it will account for microbundle issues if they exist.

marvinhagemeister: Need to push my latest changes. I have it close to working except an issue with our schedule bundle 👍

PR #1366, opened 2019-03-08
Add edge case for beforeinput event

### LINKED ISSUE #1364: beforeinput event listener not added
Took me a while to find this bug. 

`beforeinput` is often used by rich text editors (`contenteditable`), e.g Slate which I try to port to Preact. I could add the event listener manually but then I would probably need to maintain my own fork and Im not going to do that, especially considering that it's in beta and receives breaking updates all the time.

### [Codesandbox](https://codesandbox.io/s/l9x4xw67k9)

It seems that it would be a special case for Preact events logic and add few bytes because `'onbeforeinput' in el` doesn't work. Is this something you're willing to consider?

marvinhagemeister:
The event `onbeforeinput` fails the check for lowercase attributes, leading to it being attached as `BeforeInput` instead. This PR corrects that.

https://github.com/developit/preact/blob/15d4d07d93ff5846e825da4ea81a14fb03327bbd/src/diff/props.js#L74

 I'm not if there is a better approach. For now it seems fine but I'm worried that there will be more exceptions in the future. Would love to hear your thoughts @developit 

Adds `+13` bytes.
Fixes #1364 .

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -199,6 +199,10 @@ function applyEventNormalization({ type, props }) {
 		props.ondblclick = props[newProps.ondoubleclick];
 		delete props[newProps.ondoubleclick];
 	}
+	if (newProps.onbeforeinput) {
+		props.onbeforeinput = props[newProps.onbeforeinput];
+		delete props[newProps.onbeforeinput];
+	}
 	// for *textual inputs* (incl textarea), normalize `onChange` -> `onInput`:
 	if (newProps.onchange && (type==='textarea' || (type.toLowerCase()==='input' && !/^fil|che|rad/i.test(props.type)))) {
 		let normalized = newProps.oninput || 'oninput';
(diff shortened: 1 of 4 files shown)


developit:
I worry about this. There may be more events that fall into this category (newere DOM events), and whitelists/mappings will grow unbounded.

What about recommending folks use `onbeforeinput`?

marvinhagemeister:
Agree, this issue has only come up with react libraries. Using the lowercase version in react doesn't work. Maybe we should move this to `compat`?

marvinhagemeister:
On second thought moving it to compat is more appropriate in my opinion. In the end were trying to normalize react specific behaviour here. Moved it to `compat`. Let me know what you think 👍

marvinhagemeister:
Agree, it's something we should keep an eye on when there are more edge cases :+1:

Issue #1368, opened 2019-03-08
10.0.0-alpha.1: state is not updated when called setState in componentWillReceiveProps


mochiya98:
test case:

```jsx
class App extends Component {
  state = { dummy: 1 };
  componentDidMount() {
    // update child props
    this.setState({ dummy: 2 });
  }
  render() {
    return <Test dummy={this.state.dummy} />;
  }
}
class Test extends Component {
  state = { updated: false };
  componentWillReceiveProps(nextProps, nextContext) {
    this.setState({ updated: true });
  }
  render() {
    return (
      <p>
        should be true: <b>{this.state.updated.toString()}</b>
      </p>
    );
  }
}
```

react(-dom)?@16.8.3: https://codesandbox.io/s/qz6q14jp06  
preact@8.4.2: https://codesandbox.io/s/40vvkrpxo9  
preact@10.0.0-alpha.1: https://codesandbox.io/s/015lvkz3lv

PR #1369, opened 2019-03-08
Fix: state is not updated when called setState in componentWillReceiveProps

mochiya98:
refs: #1368


mochiya98:
thanks! 👍

developit:
@mochiya98 my issue ended up actually being different: `componentWillReceiveProps()` was not being called for top-level `render()` calls, and the fix is in #1377. 

Any chance you have a case you can test against that branch?

mochiya98:
probably it is not the same. still fails on that branch.

mochiya98:
ah, sorry. does it means request for add tests for that branch?  
sorry for my poor english.

developit:
Yeah I'm seeing that too. Interesting.

PR #1371, opened 2019-03-08
(Test) - testutils package with act for hooks

### LINKED ISSUE #1353: Investigate adding a test-util package
We've been getting reports of users trying to make Preact work with `react-dom/test-utils`. Problem is that this package relies on a lot of very react-specific internals that prevent it from working with preact. Nonetheless users will need something to flush update queues synchronously so that are easy to test.

Ideas:
- create a `preact/test-utils` package
- make `setupRerender` public
- ship with our own implementation for `act`

JoviDeCroock:
Please don't mind where it's located now, I've been trying to make it work.

Tests are still a success but the updates aren't synchronous yet, I feel like I am missing something extremely obvious.

Fixes #1353 .

**Edit:** Add link to relevant issue - @marvinhagemeister


hooks/src/index.js
```diff
@@ -199,14 +199,30 @@ function flushAfterPaintEffects() {
...
+	options.afterPaint = oldAfterPaint;
+	options.scheduleFlushAfterPaint = oldSchedule;
 }
 
 if (typeof window !== 'undefined') {
-	afterPaint = (component) => {
+	options.afterPaint = (component) => {
 		if (!component._afterPaintQueued && (component._afterPaintQueued = true) && afterPaintEffects.push(component) === 1) {
-			requestAnimationFrame(scheduleFlushAfterPaint);
+			requestAnimationFrame(options.scheduleFlushAfterPaint);
```
marvinhagemeister: `requestAnimationFrame` is async too, similar to `setTimeout` it has slightly different scheduling semantings, but it's something we'd need to make synchronous too nonetheless. Maybe instead of swapping out the callback to `requestAnimationFrame` we can swap out it itself:

```js
(options.afterPaint || requestanimationFrame)(scheduleFlushAfterPaint);
```

That way we don't need to touch `scheduleFlushAfterPaint` at all.

JoviDeCroock: Correct me if I am wrong but scheduleAfterPaint introduces a setTimeout so we would have to override this aswell, no?

hooks/src/index.js
```diff
@@ -1,4 +1,4 @@
-import { options } from 'preact';
+import { Component, options } from 'preact';
```
JoviDeCroock: TODO: remove this

hooks/src/index.js
marvinhagemeister: It appears so, but by letting the user call the flushing function itself we can sidestep `scheduleFlushAfterPaint` completely :+1:

JoviDeCroock:
Having inner conflicts on where to place the file though, since it needs to be accessible and well if the user wants typings we should also expose that.

JoviDeCroock:
npm-run-all really is not cooperating with me

test-utils/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "version": "0.1.0",
+  "private": true,
+  "description": "Test-utils for Preact",
+  "main": "dist/testUtils.js",
+  "module": "dist/testUtils.mjs",
+  "umd:main": "dist/testUtils.umd.js",
+  "source": "src/index.js",
+  "license": "MIT",
+  "types": "src/index.d.ts",
+  "scripts": {
```
marvinhagemeister: The scripts are not needed here. The ones in the root `package.json` are all that'S needed :+1:

marvinhagemeister:
@JoviDeCroock Oh wow you really made a lot of progress here :+1: Moving it to a dedicated packages even sooner is a great idea :+1: 

To get the tests to run we need to extend the 2 globs in `karma.conf`:

https://github.com/developit/preact/blob/79592076355df10c2a17482f27ef386f9a741476/karma.conf.js#L103

https://github.com/developit/preact/blob/79592076355df10c2a17482f27ef386f9a741476/karma.conf.js#L142

And then running `PERFORMANCE=false npm run test:karma` (or `test:karma:watch`) should be all that's needed :100:

JoviDeCroock:
I keep getting ```10 03 2019 14:55:21.034:ERROR [karma-server]: Error: Found 1 load error
    at Server.webServer.listen (C:\Users\jovid\Desktop\preact\node_modules\karma\lib\server.js:182:27)
    at Object.onceWrapper (events.js:273:13)
    at Server.emit (events.js:187:15)
    at emitListeningNT (net.js:1368:10)
    at process._tickCallback (internal/process/next_tick.js:63:19)``` even after adding it to the babel-loader

marvinhagemeister:
That's really strange. I don't have a windows machine to test on, so I'm not sure what's causing this. Maybe you have an idea @ForsakenHarmony ? Iirc you develop on windows too?

JoviDeCroock:
Works perfectly on my macbook... Probaby because I'm missing python to compile iltorb

test-utils/src/act.js
```diff
@@ -0,0 +1,12 @@
+import { options } from 'preact';
+import { setupRerender } from './setupRerender';
+
+export function act(cb) {
+	const rerender = setupRerender();
+	let flush;
+	options.afterPaint = (fc) => flush = fc;
```
marvinhagemeister: `options.afterPaint` will only be called when a component with a hook has been rendered. `flush` will be `undefined` when there is none. A check for `if (flush!=null) flush()` should be enough :+1:

JoviDeCroock:
@marvinhagemeister this feels finished to me, close this and make a real PR?

marvinhagemeister:
@JoviDeCroock No need to close this. GitHub has a `Ready for review` button above the CI status for exactly this reason :)

package.json
```diff
@@ -63,7 +65,9 @@
     "debug/package.json",
     "hooks/dist",
     "hooks/src",
-    "hooks/package.json"
+    "hooks/package.json",
+    "test-utils/src",
+    "test-utils/package.json"
```
marvinhagemeister: `test-utils/dist` should also be added to the files to include in pur `npm` package :+1:

test-utils/test/shared/act.js
```diff
@@ -0,0 +1,47 @@
...
+
+	beforeEach(() => {
+		scratch = setupScratch();
+	});
+
+	afterEach(() => {
+		teardown(scratch);
+	});
+
+	it('should reset options after act finishes', () => {
```
marvinhagemeister: Idea for another test case to check if previously set options are restored:

```js
it('should restore options.afterPaint', () => {
  const spy = sinon.spy();

  options.afterPaint = spy;
  act(() => null);

  expect(options.afterPaint).to.equal(spy);
  expect(spy).to.not.be.called;
});
```

test-utils/test/shared/act.js
```diff
@@ -0,0 +1,47 @@
+import { options, createElement as h, render } from 'preact';
```
marvinhagemeister: nit: Test files usually have the `.test.js` suffix in our suite

test-utils/test/shared/act.js
JoviDeCroock: Does this suffice, made two additional tests

JoviDeCroock:
> @JoviDeCroock No need to close this. GitHub has a `Ready for review` button above the CI status for exactly this reason :)

I can't click this since travis won't report to me (I don't belong to the repo)

test-utils/test/shared/setupRerender.test.js
```diff
@@ -0,0 +1,40 @@
...
+		expect(options.debounceRendering).to.equal(undefined);
+		rerender();
+		expect(options.debounceRendering).to.equal(undefined);
+	});
+
+	it('should restore options.debounceRendering', () => {
+		const spy = sinon.spy();
+
+		options.debounceRendering = spy;
+		rerender();
```
marvinhagemeister: I think the `setupRerender` call should happen here right after setting `options.debounceRendering`. Without that the test doesn't really test what it says it does.

```
1. act -> setupRerender sets options.debounceRendering
2. Test overwrites options.debounceRendering with a spy
```

What should be tested instead:

```
1. Set options.debounceRendering with a spy
2. act -> setupRerender overwrites options.debounceRendering
3. when act is done, options.debounceRendering should be a spy again.
```

test-utils/src/index.js
```diff
@@ -0,0 +1,2 @@
+export * from './act';
+export * from './setupRerender';
```
marvinhagemeister: Minor nit: Just inline both functions here. No need to make a file per function (that's a popular approach for some projects though, just not for us :) ).

marvinhagemeister:
@JoviDeCroock Oh I see, just clicked it :+1: Left a few follow-up comments. Let me know what you think :+1:

test-utils/test/shared/setupRerender.test.js
JoviDeCroock: As far as I know previous debouce was historically not being reset by `setupRerender` so I don't really get it sorry.

hooks/src/index.js
```diff
@@ -206,7 +206,12 @@ function scheduleFlushAfterPaint() {
 if (typeof window !== 'undefined') {
 	afterPaint = (component) => {
 		if (!component._afterPaintQueued && (component._afterPaintQueued = true) && afterPaintEffects.push(component) === 1) {
-			requestAnimationFrame(scheduleFlushAfterPaint);
+			if (options.afterPaint) {
+				options.afterPaint(flushAfterPaintEffects);
```
developit: Can we change this to `options.requestAnimationFrame`? It should compress better.

test-utils/src/index.js
```diff
@@ -0,0 +1,26 @@
+import { Component, options } from 'preact';
+
+/**
+ * Setup a rerender function that will drain the queue of pending renders
+ * @returns {() => void}
+ */
+export function setupRerender() {
+	Component.__test__previousDebounce = options.debounceRendering;
```
developit: I know this came from the existing test harness, but the naming is my fault and it's probably worth correcting now haha:

```js
export function setupRerender() {
	const prev = options.debounceRendering;
	options.debounceRendering = cb => prev = cb;
	return () => prev && prev();
}
```

JoviDeCroock: Hmm, this does not seem to work, I'll check what's up

JoviDeCroock: Doing this change makes me have to change the following test:

```	it('should restore options.debounceRendering after act', () => {
		const spy = sinon.spy();

		options.debounceRendering = spy;
		act(() => null);

		expect(options.debounceRendering).to.equal(spy);
		expect(spy).to.not.be.called;
	});```

To one where spy would get called

Comment (marvinhagemeister) [reply test-utils/src/index.js]: @developit We did support component-level scheduling for a good while in ceviche before settling for `options.debounceRendering`. Back then you could attach a different scheduler on `Component.debounce`. That's why the test harness still stores the old scheduler function on the component instance.

Comment (JoviDeCroock) [reply test-utils/src/index.js]: Just to note, the old one `setupRerender` is still in there. I was planning to remove that in another PR for the sake of keeping this one compact. Can still start replacing it in every test file if you prefer.

File: hooks/src/index.js [hunk 12]
```diff
@@ -206,7 +206,9 @@ function scheduleFlushAfterPaint() {
 if (typeof window !== 'undefined') {
 	afterPaint = (component) => {
 		if (!component._afterPaintQueued && (component._afterPaintQueued = true) && afterPaintEffects.push(component) === 1) {
+			/* istanbul ignore next */
 			if (options.requestAnimationFrame) {
+				/* istanbul ignore next */
```
marvinhagemeister: the comment above the `if`-statement should be enough. Istanbul works on the AST-level

JoviDeCroock: Does not seem to be the case here

test-utils/src/index.js
marvinhagemeister: You made the right call, better to do that in another PR 👍

JoviDeCroock:
Finally back to 100% coverage

marvinhagemeister:
Sounds good to me :+1: Let's do this :100:

Issue #1374, opened 2019-03-09
10.0.0-alpha.1: style is not updated correctly when using camelCase in style attr


mochiya98:
```jsx
import { Component, render } from "preact";

class App extends Component {
  state = { style: { fontSize: "70px", color: "red" } };
  componentDidMount() {
    // fontSize remains
    this.setState({ style: {} });
  }
  render() {
    return <span style={this.state.style}>Foo</span>;
  }
}

render(<App />, document.getElementById("root"));
```
preact@10.0.0-alpha.1: https://codesandbox.io/s/z25p6vlq7l

marvinhagemeister:
Fixed by #1375

PR #1375, opened 2019-03-09
Fix: style is not updated correctly when using camelCase in style attr

mochiya98:
refs: #1374

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -22,6 +22,8 @@ export function diffProps(dom, newProps, oldProps, isSvg) {
 	}
 }
 
+let CAMEL_REG = /-?(?=[A-Z])/g;
+
 /**
  * Set a property value on a DOM node
  * @param {import('../internal').PreactElement} dom The DOM node to modify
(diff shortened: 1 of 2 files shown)


src/diff/props.js
```diff
@@ -50,7 +50,7 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 			if (typeof oldValue==='string') s.cssText = '';
 			// remove values not in the new list
 			for (let i in oldValue) {
-				if (value==null || !(i in value)) s.setProperty(i, '');
+				if (value==null || !(i in value)) s.setProperty(i.replace(/-?(?=[A-Z])/g, '-'), '');
```
marvinhagemeister: This looks like the same regex like a few lines below:

https://github.com/developit/preact/blob/4f671e7fb99425f84434c0c8b56d830ade4ea792/src/diff/props.js#L60

Maybe the size will be smaller, when we hoist them out into a variable outside of `setProperty()`. We usually compare the number at the top when running `npm run build:core` :+1:

mochiya98: this solution is more smaller than outer function when gzipped.

marvinhagemeister: Are you sure? When hoisted the size is `5` bytes smaller. The PR currently sits at `3304 B` and when hoisted will be `3299 B`:

```js
let CAMEL_REG = /-?(?=[A-Z])/g;

  //...
  if (value==null || !(i in value)) s.setProperty(i.replace(CAMEL_REG, '-'), '');

  // ...
  s.setProperty(i.replace(CAMEL_REG, '-'), typeof v==='number' && IS_NON_DIMENSIONAL.test(i)===false ? (v + 'px') : v);
}
```

mochiya98: oh, nice!
i tried this way:
```js
function hoge(elem, k, v){
  elem.setProperty(k.replace(/-?(?=[A-Z])/g, '-'), v);
}
```
ok, i'll do that.

marvinhagemeister: Ah I see, creating an additional function is indeed more costly than just hoisting out a variable :)

mochiya98:
thank you for your help! 👍

PR #1376, opened 2019-03-09
Fix `useContext` double render

### LINKED ISSUE #1363: 10.0.0-alpha.1 - Children from context provider are rendered twice after state update
When you update state in a provider component and pass the value to it, the children will render twice instead of once (with the same data). It could impact performance quite a bit if the provider is at the top of the tree.

Here's a demo that compares the behaviour between React and Preact (same logic, different output in console): https://codesandbox.io/s/18lq5wmqjq

The example is using useEffect, useState and useContext, not sure if this has something to do with hooks. I could also make an example without if it would make the problem more clear!

marvinhagemeister:
This PR fixes an issue where `useContext` would lead to a component to be triggered twice when the `Provider` value changes. The problem is caused by the following line:

https://github.com/developit/preact/blob/47fdd6d37802a0219facb990f7dfeaeec04463fd/src/create-context.js#L33

There we check if we can skip the update by comparing the new value with `c.context`. The thing with hooks is that they can't bind all to the same `property`. On top of that multiple `useContext` hooks with different context objects all need to have their own way to check if they need updating.

I basically introduced another path for context updates when the current component has no `contextType`, but may have hooks present.

Marking this as a draft, because it has the unfortunate side effect of reference something hook specific in core. For now I don't know any better way to do this.

Adds `+41 B`.
Fixes #1363 .


hooks/src/index.js
```diff
@@ -170,11 +170,10 @@ export function useContext(context) {
 	const provider = currentComponent.context[context._id];
 	if (provider == null) return context._defaultValue;
 	const state = getHookState(currentIndex++);
-	if (state._value == null) {
-		state._value = true;
-		provider.sub(currentComponent);
-	}
-	return provider.props.value;
+	state.type = 1;
```
marvinhagemeister: Because I'm iterating over all hooks in `Provider` I need a way to find out if the hook at the current index is actually a `useContext` hook. We'll likely need something similar for the devtools soon.

developit: worth making each have a bit for this? `type` is a good property name since it's already elsewhere.

marvinhagemeister: Sounds good 👍

marvinhagemeister: I'd probably add the bit mask with #1383 . That change would be more appropriate there :)

marvinhagemeister:
Addressed the CI failure and I think it's ready for the next review round :tada:

src/create-context.js
```diff
@@ -30,7 +30,19 @@ export function createContext(defaultValue) {
 		};
 		comp.componentDidUpdate = () => {
 			let v = comp.props.value;
-			subs.map(c => v!==c.context && (c.context = v, enqueueRender(c)));
+			subs.map(c => {
+				// A component can have multiple `useContext` hooks
+				if (c.contextType==null && c.__hooks) {
+					c.__hooks._list.forEach(state => {
```
developit: (I'm trying to think of ways to avoid pulling the hooks-specific part of this into core.)

marvinhagemeister: Yeah, it doesn't feel right :/

JoviDeCroock:
This should be fixed by the other PR, or are there cases the other one does not cover yet?

marvinhagemeister:
This was solved in a much more elegant way in #2386  🎉

PR #1377, opened 2019-03-10
Fix force flag

### LINKED ISSUE #1404: 10.0.0-alpha.1 - Passing tagName as a (redundant) prop to an HTML node causes an error
Passing `tagName` as a prop like:

```js
export default function () {
  return <button tagName='button' />
};
```

Will result in an error like:

```
E-v1.js:formatted:10882 Uncaught (in promise) TypeError: Cannot assign to read only property 'tagName' of object '#<HTMLButtonElement>'
    at w (E-v1.js:formatted:10882)
    at g (E-v1.js:formatted:10859)
    at A (E-v1.js:formatted:10999)
    at _ (E-v1.js:formatted:10953)
    at _ (E-v1.js:formatted:10948)
    at _ (E-v1.js:formatted:10948)
    at d (E-v1.js:formatted:10816)
    at A (E-v1.js:formatted:11001)
    at _ (E-v1.js:formatted:10953)
    at _ (E-v1.js:formatted:10948)
```

I noticed this because I have a (sloppy but allowed) pattern in a couple places where we have custom elements (not shown in example), and specify tagName like so:

```js
class CustomElement extends Component {
  render () {
    const TagName = this.props.tagName;
    return <TagName {...this.props} />;
  }
}
render(<CustomElement tagName='button' class='the-class'>...</CustomElement>, parentEl);
```

To accommodate this, I need to explicitly remove `tagName` from props before passing it down. This can be solved like `const propsToPass = Object.assign({}, this.props); delete propsToPass.tagName`, but it's not a convenient pattern. Still, I think it might still be okay to leave it as is, as long as it's mentioned as a breaking change. Thoughts?

developit:
This fixes the `_force flag` (and actually removes the flag itself). Instead, I've plumbed the rendering mode indicating `setState|forceUpdate|render` using a parameter to diff, since it only ever needs to be indicated for the outermost diff() call following on from a setState() or forceUpdate().

I believe the issue before is that we were setting the `_force` flag within setState(), but state rendering is enqueued. This meant the flag could end up never being unset if the enqueued render was bypassed due to `component._dirty` being false (from a prop render).

Next steps here would be to switch to a bitmask for this info. Preact actually used one in 8.x (MODE_SYNC | MODE_ASYNC | MODE_FORCED or something like that). I think this would be a good first candidate.

`-10b` 🐕
Fixes #1404 - @marvinhagemeister


src/diff/index.js
```diff
@@ -21,7 +21,7 @@ import options from '../options';
  * @param {import('../internal').Component | null} ancestorComponent The direct
  * parent component
  */
-export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent) {
+export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, force) {
```
marvinhagemeister: The JSDoc block of `diff()` needs to be updated for the new parameter 🙂

marvinhagemeister:
Wondering if we should add a test case for this. That'd probably be a good idea for future refactoring.

developit:
@marvinhagemeister I might see if [the test](https://github.com/developit/preact/pull/1369/files#diff-eca8f3345fce0d36bdfecf6f6afc48d3R888) @mochiya98 added in #1369 fails before + passes after this, I have a feeling it might.

marvinhagemeister:
@developit iirc on slack you mentioned something like `render -> setState -> render` that was fixed via this issue? I can try to come up with a test case if you're short on time.

marvinhagemeister:
Well luckily this PR fixes a mangling issue by removing `this._force` completely :tada:

PR #1380, opened 2019-03-10
Allow native preact vnodes in isValidElement

### LINKED ISSUE #1349: 10.0.0-alpha.0 - compat cloneElement ignoring props and children
I believe this bug might have originated here: https://github.com/developit/preact/commit/d3c12b7ea3c16f0fa545842ac3bd2b79567e347c

marvinhagemeister:
This PR fixes an issue where `compat` would bail out on standard preact `vnodes` because they didn't have the `$$typeof` property. Instead patching the `isValidElement` function [like we did before](https://github.com/developit/preact-compat/blob/ae018abbefdfeb61a0e0f7a50ac0d7c9f12940ea/src/index.js#L336) I chose to add it in `options.vnode` instead.

Removes `111 B`. :tada: 
Fixes #1349 .

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -116,10 +116,6 @@ function upgradeToVNodes(arr, offset) {
 		if (Array.isArray(obj)) {
 			upgradeToVNodes(obj);
 		}
-		else if (obj && typeof obj==='object' && !isValidElement(obj) && ((obj.props && obj.type) || obj.text!=null)) {
-			if (obj.text) continue;
-			arr[i] = createElement(obj.type, obj.props, obj.props.children);
-		}
 	}
 }
 
@@ -133,7 +129,6 @@ function upgradeToVNodes(arr, offset) {
 function createElement(...args) {
 	upgradeToVNodes(args, 2);
 	let vnode = h(...args);
-	vnode.$$typeof = REACT_ELEMENT_TYPE;
 
 	let type = vnode.type, props = vnode.props;
 	if (typeof type!='function') {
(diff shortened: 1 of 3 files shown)


compat/src/index.js
```diff
@@ -182,7 +181,7 @@ function cloneElement(element) {
  * @returns {boolean}
  */
 function isValidElement(element) {
-	return element && element.$$typeof===REACT_ELEMENT_TYPE;
+	return element!=null && element.$$typeof===REACT_ELEMENT_TYPE;
```
marvinhagemeister: Added the null check here to ensure that `isValidElement` always returns a `boolean`.

compat/src/index.js
```diff
@@ -116,10 +116,6 @@ function upgradeToVNodes(arr, offset) {
 		if (Array.isArray(obj)) {
 			upgradeToVNodes(obj);
 		}
-		else if (obj && typeof obj==='object' && !isValidElement(obj) && ((obj.props && obj.type) || obj.text!=null)) {
```
marvinhagemeister: This branch is not needed anymore because every `vnode` that passes through `h` in compat or core will have the `$$typeof`-property now .

robertknight: The entire `upgradeToVNodes` function is a no-op now that this branch has been removed, all it does is call itself recursively.

marvinhagemeister: Ohh thanks for the pointer. Not sure why I didn't see that :joy:

marvinhagemeister:
@robertknight Awesome thanks for your review :100: I updated the PR with your suggestions :tada:

andrewiggins:
Wow, 100B! That's awesome!

PR #1383, opened 2019-03-10
Show hooks in devtools

### LINKED ISSUE #1381: Display hooks state in devtools

marvinhagemeister:
Getting started...

- [x] Make hooks panel show up for components with hooks
- [ ] Display hook name
- [ ] Display hook state
- [ ] Make hook state editable
- [ ] Show nested hooks

Fixes #1381 .


marvinhagemeister:
This is a lot easier with the next iteration of the `react-devtools`. It will probably be published soon which is why I don't think it's worth to spent more time with our existing integration.

Support for hooks will be added in #1549.

PR #1385, opened 2019-03-10
use `console.error` instead of `console.log` in catchErrors

yuqianma:
I saw the error messages with both red background and normal one. A bit weird.
Don't know why here's two forms of console. But I'd like the line to be`console.error` since it is an error object.


yuqianma:
Wow, you figure out the key 'UX'.
As frontend coders, we may forget that we are USERS when we use the lib.
Thanks for the preact and your awesome works!

PR #1386, opened 2019-03-10
`oldChildren.length` -> `oldChildrenLength` to save a few bytes

yuqianma:
As the length is saved in line:29, we can use it then.
The length of old children is fixed here right?


yuqianma:
👏

PR #1387, opened 2019-03-10
Add additional tests for other methods affected by #1369.

developit:
/cc @mochiya98


mochiya98:
good point! thanks.
but i think it is slightly different.
maybe we need test of the updated state with componentWillReceiveProps.
for example:
```jsx
it('should update state when called setState in componentWillReceiveProps', () => {
	let componentState;
	let spyShouldComponentUpdate = sinon.spy();
	let spyComponentWillUpdate = sinon.spy();

	class Foo extends Component {
		constructor(props) {
			super(props);
			this.state = {
				dummy: 0
			};
		}
		componentDidMount() {
			// eslint-disable-next-line react/no-did-mount-set-state
			this.setState({ dummy: 1 });
		}
		render() {
			return <Bar dummy={this.state.dummy} />;
		}
	}
	class Bar extends Component {
		constructor(props) {
			super(props);
			this.state = {
				value: 0
			};
		}
			spyShouldComponentUpdate(nextState);
			return true;
		}
		componentWillUpdate(nextProps, nextState) {
			spyComponentWillUpdate(nextState);
		}
		componentWillReceiveProps() {
			this.setState({ value: 1 });
		}
		render() {
			componentState = this.state;
			return <div />;
		}
	}

	render(<Foo />, scratch);
	rerender();

	expect(componentState).to.deep.equal({ value: 1 });
	expect(spyShouldComponentUpdate).to.be.calledWith({ value: 1 });
	expect(spyComponentWillUpdate).to.be.calledWith({ value: 1 });
});
```

PR #1393, opened 2019-03-11
add autocorrect html attribute to jsx types

garybernhardt:
This attribute is pretty widely used. E.g. it's described here: https://davidwalsh.name/disable-autocorrect

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -494,6 +494,8 @@ declare namespace JSX {
 		async?: boolean;
 		autocomplete?: string;
 		autoComplete?: string;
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
The coverage decrease by coveralls makes no sense. Somehow it didn't pick up the new changes, nothing to do with this PR.

garybernhardt:
Weird... coveralls' message doesn't even match the email I got for this thread, which says it stayed at 100%:

> Coverage remained the same at 100.0% when pulling fe6c857 on garybernhardt:add-autocorrect-html-attribute into dc2e810 on developit:master.

PR #1394, opened 2019-03-11
Don't call CSSStyleDeclaration.setProperty for style strings

choumx:
Avoids unnecessary calls to `setProperty` for style strings. 

- Browsers and JSDOM will ignore these but workerDOM (currently) doesn't since it lacks a property whitelist.
- Minor perf boost by avoiding loop of `value.length` and some regex ops.

/cc @kristoferbaxter


developit:
Wow, nice catch! This one was my bad 🥺

PR #1395, opened 2019-03-12
Fix Element.prototype.remove not available in IE11 

### LINKED ISSUE #624: Element.remove() is unsupported in Internet Explorer

marvinhagemeister:
This PR replaces calls to `Element.remove()` with `Element.removeChild()`. Reason being that the former is not available in IE11. I normalized the logger though because `<parent>.removeChild(foo)` is a lot harder on the eyes than `foo.remove()`. This is in particular makes it easier to see at one glance what our reconciler is doing.

I also tried if adding just the polyfill would be smaller, but it's in `6 B` bigger than the changes in this PR. It feels a lot cleaner though, so I'm wondering if we should just take the hit and keep our internals clean from any IE11 shenanigans. 

The downside would be that in IE11 every node will be removed recursively from a tree. `remove()` already detaches all sub-nodes from their parent automatically in the browser. So IE11 would be a lot slower with the polyfilled version.

This is the polyfill I tried:

```js
// IE11 Polyfill for Element.remove()
Element.prototype.remove || (Element.prototype.remove = function() {
	if (this.parentNode!=null) this.parentNode.removeChild(this);
});
```

Fixes #624 (again) 😬
Adds `+27 B`.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -306,7 +306,7 @@ export function unmount(vnode, ancestorComponent, skipRemove) {
 		applyRef(r, null, ancestorComponent);
 	}
 
-	if (!skipRemove && (skipRemove = ((r = vnode._dom)!=null))) removeNode(r);
+	if (!skipRemove && vnode._lastDomChild==null && (skipRemove = ((r = vnode._dom)!=null))) removeNode(r);
 
 	vnode._dom = vnode._lastDomChild = null;
 
@@ -325,7 +325,7 @@ export function unmount(vnode, ancestorComponent, skipRemove) {
 	}
 	else if (r = vnode._children) {
 		for (let i = 0; i < r.length; i++) {
-			unmount(r[i], ancestorComponent, skipRemove && vnode.type!==Fragment);
+			unmount(r[i], ancestorComponent, skipRemove);
 		}
 	}
 }
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -298,7 +298,7 @@ export function applyRef(ref, value, ancestorComponent) {
  * @param {import('../internal').Component} ancestorComponent The parent
  * component to this virtual node
  */
-export function unmount(vnode, ancestorComponent) {
+export function unmount(vnode, ancestorComponent, skipRemove) {
 	let r;
 	if (options.unmount) options.unmount(vnode);
 
@@ -306,7 +306,7 @@ export function unmount(vnode, ancestorComponent) {
 		applyRef(r, null, ancestorComponent);
 	}
 
-	if ((r = vnode._dom)!=null) removeNode(r);
+	if (!skipRemove && (skipRemove = ((r = vnode._dom)!=null))) removeNode(r);
 
 	vnode._dom = vnode._lastDomChild = null;
 
@@ -321,11 +321,11 @@ export function unmount(vnode, ancestorComponent) {
 		}
 
 		r.base = r._parentDom = null;
-		if (r = r._prevVNode) unmount(r, ancestorComponent);
+		if (r = r._prevVNode) unmount(r, ancestorComponent, skipRemove);
 	}
 	else if (r = vnode._children) {
 		for (let i = 0; i < r.length; i++) {
-			unmount(r[i], ancestorComponent);
+			unmount(r[i], ancestorComponent, skipRemove && vnode.type!==Fragment);
 		}
 	}
 }
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -297,6 +297,8 @@ export function applyRef(ref, value, ancestorComponent) {
  * @param {import('../internal').VNode} vnode The virtual node to unmount
  * @param {import('../internal').Component} ancestorComponent The parent
  * component to this virtual node
+ * @param {boolean} skipRemove Flag that indicates that a parent node of the
+ * current element is already detached from the DOM.
  */
 export function unmount(vnode, ancestorComponent, skipRemove) {
 	let r;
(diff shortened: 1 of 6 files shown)


src/diff/index.js
```diff
@@ -311,11 +311,11 @@ export function unmount(vnode, ancestorComponent) {
 		}
 
 		r.base = r._parentDom = null;
-		if (r = r._prevVNode) unmount(r, ancestorComponent);
+		if (r = r._prevVNode) unmount(r, ancestorComponent, skipRemove);
 	}
 	else if (r = vnode._children) {
 		for (let i = 0; i < r.length; i++) {
-			unmount(r[i], ancestorComponent);
+			unmount(r[i], ancestorComponent, skipRemove && vnode.type!==Fragment);
```
marvinhagemeister: The for `Fragment` is necessary, because `_dom` will point to the first child.

developit:
One way we could slim this down and keep things fast would be to avoid the helper function:  in a bunch of the callsites for it, we already have a reference to the parentNode we need to remove from.

src/diff/index.js
developit: In the case of a fragment, we could skip removing `_dom` and set `skipRemove=false`.

marvinhagemeister: Noticed that this isn't only true for `Fragments` but also for HOC-Components. They may have different `_dom` and `_lastDomChild` pointers. Went with checking for the latter property and only removing the dom when it is `null`.

marvinhagemeister:
@developit I had a go at trying to reuse existing parent pointers, but the trouble is that `parentDom` may not be the same parent when hydrating. I could only fix that by inlining the `element.parentNode!=null` check too, at which point the byte size got `+4 B` larger than what's in this PR.

robertknight:
> I also tried if adding just the polyfill would be smaller, but it's in 6 B bigger than the changes in this PR. It feels a lot cleaner though, so I'm wondering if we should just take the hit and keep our internals clean from any IE11 shenanigans.

I would be inclined to go this route if the performance hit isn't expected to be too great.

> So IE11 would be a lot slower with the polyfilled version.

Do you mean slower than the version in this PR? Do you have any idea, in quantitative terms, roughly how much slower?

src/util.js
```diff
@@ -9,3 +9,12 @@ export function assign(obj, props) {
 	for (let i in props) obj[i] = props[i];
 	return /** @type {O & P} */ (obj);
 }
+
+/**
+ * Remove a child node from its parent if attached.
+ * @param {Node} node The node to remove
+ */
+export function removeNode(node) {
```
robertknight: If you do go this route, I think it would be worth adding a note here that IE 11 is the only reason why we're not using the more obvious `node.remove()`.

src/diff/index.js
```diff
@@ -288,15 +288,15 @@ export function applyRef(ref, value, ancestorComponent) {
  * @param {import('../internal').Component} ancestorComponent The parent
  * component to this virtual node
  */
-export function unmount(vnode, ancestorComponent) {
+export function unmount(vnode, ancestorComponent, skipRemove) {
 	let r;
```
robertknight: In this function the `r` variable is used to refer to 5 different things (vnode ref, mounted DOM node, Component instance, last rendered output of component, array of child vnodes). Is this overloading for code golfing reasons?

src/diff/index.js
```diff
@@ -288,15 +288,15 @@ export function applyRef(ref, value, ancestorComponent) {
  * @param {import('../internal').Component} ancestorComponent The parent
  * component to this virtual node
  */
-export function unmount(vnode, ancestorComponent) {
+export function unmount(vnode, ancestorComponent, skipRemove) {
```
robertknight: So `skipRemove` here is a flag indicating that the mounted DOM node has already been removed because its parent was removed? It would be useful to have a JSDoc comment to that effect.

marvinhagemeister:
Thanks for the thorough review @robertknight 🎉 I agree with you on the polyfill. I'll make the necessary changes 👍 

Out of curiosity, I collected some data about the performance characteristics of the changes in this PR. To test this I compared both the `master` and `ie11_remove` branch against each other. The profile captures a full page transition where a bigger tree has to be destroyed and a new tree has to be created from scratch. 

preact-skipRemove

The biggest difference is that the `skipRemove` flag allows us to skip a whole bunch of work involving detaching nodes even if the parent is already detached from the DOM. I'm not sure how to profile this in IE11 but I haven't noticed any performance decrease when clicking around via the polyfill. The difference likely will only be noticeable in micro-benchmarks.

src/diff/index.js
marvinhagemeister: yup, overloading the variable saves a few bytes.

marvinhagemeister:
Played around a bit with the polyfill, but it will lead preact to throw in a node environment which doesn't have `Element`. This would cause a lot of friction over the version without the polyfill.

src/util.js
```diff
@@ -9,3 +9,14 @@ export function assign(obj, props) {
...
 	return /** @type {O & P} */ (obj);
 }
+
+/**
+ * Remove a child node from its parent if attached. This is a workaround for
+ * IE11 which doesn't support `Element.prototype.remove()`. Using this function
+ * is smaller than including a dedicated polyfill.
+ * @param {Node} node The node to remove
+ */
+export function removeNode(node) {
```
kristoferbaxter: This is an opportunity with a module build to remove the additional method via build system transform.

marvinhagemeister: We're all rooting for your proposal :+1: :+1:

PR #1396, opened 2019-03-12
(refactor): use test-utils package

JoviDeCroock:
Use the newly created test-utils package in our tests.
This also adds a fix for the typings in test-utils


JoviDeCroock:
I was wondering if it’d be worth it to refactor hooks tests with act aswell?

marvinhagemeister:
Good idea👍

JoviDeCroock:
> Good idea👍

Also in a seperate PR, I assume? Just so reviewers have an easy overview about changes etc.

PR #1397, opened 2019-03-12
(fix) - select value not set

JoviDeCroock:
This is not the ideal fix yet, but it shows what @marvinhagemeister sais, the diffing of props going first makes the value unset.
This solution is not ideal but I don't know what is changed will continue working on it but will leave this as a draft so it can be referenced from the issue to see this works.

I assume this happens because preact strips properties from the native html element and puts them into props, that way we lose the native working of value. Should we propagate value to the native element when it is a `select`? Slowly discovering the internals so pardon the naive solution atm

EDIT: tried changing to this in diffProps:

```
else if (name==='value' && isSelect) {
  console.log('setting', name, value);
  dom.setAttribute(name, value);
}
```

Turns out I was on the wrong track with that one, so it's really the children that have to be evaluated.

Fixes https://github.com/developit/preact/issues/1392

Adds `+24 B` - @marvinhagemeister

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -218,6 +218,7 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
 
 	// Tracks entering and exiting SVG namespace when descending through the tree.
 	isSvg = newVNode.type==='svg' || isSvg;
+	const isSelect = newVNode.type==='select';
 
 	if (dom==null && excessDomChildren!=null) {
 		for (let i=0; i<excessDomChildren.length; i++) {
@@ -258,10 +259,14 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
 					}
 				}
 			}
+			if (isSelect) {
+				diffChildren(dom, newVNode, oldVNode, context, newVNode.type==='foreignObject' ? false : isSvg, excessDomChildren, mounts, ancestorComponent);
+			}
 			diffProps(dom, newVNode.props, oldProps, isSvg);
 		}
-
-		diffChildren(dom, newVNode, oldVNode, context, newVNode.type==='foreignObject' ? false : isSvg, excessDomChildren, mounts, ancestorComponent);
+		if (!isSelect) {
+			diffChildren(dom, newVNode, oldVNode, context, newVNode.type==='foreignObject' ? false : isSvg, excessDomChildren, mounts, ancestorComponent);
+		}
 	}
 
 	return dom;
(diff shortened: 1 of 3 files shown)


JoviDeCroock:
I haven't quite figured out yet why just those two tests are failing but I am getting there.

JoviDeCroock:
This line: https://github.com/developit/preact/pull/1397/files#diff-44070032789f10581d1b70740a2aaacfR271

Is probably redundant now, no?

src/diff/index.js
```diff
@@ -258,9 +258,16 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
 					}
 				}
 			}
+			if (newVNode.props.dangerouslySetInnerHTML || oldProps.dangerouslySetInnerHTML) {
```
cristianbote: I'm not sure if this would be the right path - to extract the `innerHTML` part from props. Have you tried to do the _hydration_ mode, @developit suggested on slack?

src/diff/props.js
```diff
@@ -63,10 +63,7 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 		}
 	}
 	else if (name==='dangerouslySetInnerHTML') {
```
cristianbote: Have you tried to remove this `else` entirely?

src/diff/index.js
JoviDeCroock: That would mean that we'd somehow have to get into line 246 when old was set and new is null right?

marvinhagemeister: Storing `dangerouslySetInnerHTML` in variables is `5 B` smaller than `vnode.props.dangerouslySetInnerHTML` 🎉 

```js
let newHtml = newVNode.props.dangerouslySetInnerHTML;
let oldHtml = oldProps.dangerouslySetInnerHTML;
if (newHtml || oldHtml) {
  // Avoid re-applying the same '__html' if it did not changed between re-render
  if (!newHtml || !oldHtml || newHtml.__html!=oldHtml.__html) {
  }
}
```

JoviDeCroock: @marvinhagemeister is this solution acceptable because as stated by @cristianbote, developit also mentioned one about entering hydration, personally I'm not following that one all too well to be honest.

marvinhagemeister:
@JoviDeCroock This is awesome 👍💯 Only had one minor golf idea, but apart from that it's ready to be merged imo 🎉

JoviDeCroock:
It includes your PR remark now @marvinhagemeister, still curious about the hydration though.

src/diff/index.js
marvinhagemeister: @JoviDeCroock The hydration optimisation @developit talked about in slack can and should happen in a separate PR. This one is just for fixing #446 which it does wonderfully 👍

PR #1401, opened 2019-03-13
(chore) - allow debug to be tree-shaken in production

JoviDeCroock:
As proposed in the following issue static import make the dependencies be included by default. This is why this resolves the issue. 
I saw @developit mention that it would be better not to export an esm in this scenario, is this still the case?

Fixes: https://github.com/developit/preact/issues/1355


debug/src/index.js
```diff
@@ -1,7 +1,4 @@
-import { initDebug } from './debug';
-import { initDevTools } from './devtools';
-
 if (process.env.NODE_ENV==='development') {
-	initDebug();
-	initDevTools();
+	require('./debug').initDebug();
+	require('./devtools').initDevTools();
```
marvinhagemeister: `microbundle` seems to have trouble with `require` statements. They seem to break the `umd` bundle.

/cc @ForsakenHarmony @Andarist

JoviDeCroock: Hmm maybe the terser is made to already elim this if it's set for production?

JoviDeCroock: Maybe microbundle is atm only focussed on using import/export?

JoviDeCroock:
So I have made a small temporairy fix for now, the main issue is that when I make the index like this:

```
if (__dev__) require('./debug').initDebug()
```

Microbundle bundles it to this:

```
"development"===process.env.NODE_ENV&&(require("./debug").initDebug(),require("./devtools").initDevTools());
```

Completely ignoring the the required stuff.
This is just a temp fix so `prop-types` does not get included in production.

marvinhagemeister:
I'll likely be the minority here: One thing I quite liked about having a proper umd bundle (with `prop-types`) is that we can just drop it into a script tag and have it work. This is even true for the devtools, which is kinda awesome :tada: 

But I can see why many prefer just import something and have it be tree-shaken out for production.

JoviDeCroock:
You mean a bundle where `prop-types` is included in the bundle? It would also make it safer since then you can't import/require something that isn't there. In all honesty the prop-types import should be a resolveWeak in this tempfix but I don't know if that works in web.

debug/src/index.js
Andarist: **I think** this is a problem with rollup-plugin-commonjs. This whole file is written as module (it has import/export statements) and thus require calls stay untouched - there is no reason (from the plugin's POV) to handle this as `require` is a valid function name, it's not special cased.

In general mixing ESM & CJS doesn't sound quite right, I understand the problem at hand - CJS can be sync & lazy and that would be helpful here.

JoviDeCroock: @Andarist Do you think this can be solved by doing a require.resolveWeak here:

https://github.com/developit/preact/pull/1401/files#diff-5092cf2353750c5deb78d54517684aa5R5

Andarist: Not quite, `resoleWeak` is a webpack thing and I assume that you want to support wider range of bundlers etc

README.md
```diff
@@ -476,7 +476,9 @@ import { h, Component, render } from 'preact';
 // Enable debug mode. You can reduce the size of your app by only including this
 // module in development builds. eg. In Webpack, wrap this with an `if (module.hot) {...}`
 // check.
-require('preact/debug');
+if (process.env.NODE_ENV !== 'production') {
```
Andarist: if webpack is set  correctly this check should be sufficient, no need to wrap with `module.hot` check

debug/src/index.js
JoviDeCroock: Well rollup also supports it, but it isn't a solution. As a temporairy fix I added it in the README to brace the import with a `__dev__` check. Another solution could be to or make our own rollup for debug or look for a solution in `microbundle`

debug/package.json
```diff
@@ -4,8 +4,8 @@
   "version": "1.0.0",
   "private": true,
   "description": "Preact extensions for development",
-  "main": "dist/debug.js",
   "module": "dist/debug.mjs",
+  "main": "dist/debug.js",
```
marvinhagemeister: This change seems unnecessary :upside_down_face:

JoviDeCroock: Oh, sorry that's a residual from me eliminating module :)

developit:
Could we drop the module build for `/debug` and only ship `main` + `unpkg`/`umd:main` (umd)?

JoviDeCroock:
> Could we drop the module build for `/debug` and only ship `main` + `unpkg`/`umd:main` (umd)?

Like this: https://github.com/developit/preact/pull/1401/commits/835853cc480521537097afb1bda9beae17e8e3b2?

developit:
I guess without the inline `require()` statements though it'll still include those modules in the bundle.

JoviDeCroock:
Yes, the main point was to show it in the readme since it's not possible with `rollup-plugin-commonjs` to do inline require statements.

debug/index.js
```diff
@@ -0,0 +1,3 @@
+if (process.env!=='production') {
```
marvinhagemeister: I think it should be `process.env.NODE_ENV` 👍

JoviDeCroock: Oh yeah, still have to test it aswell so that would’ve showed ^^

JoviDeCroock:
Should I remove the entry from readme, since it’s not relevant?

debug/package.json
```diff
@@ -5,7 +5,6 @@
   "private": true,
   "description": "Preact extensions for development",
   "main": "dist/debug.js",
-  "module": "dist/debug.mjs",
```
marvinhagemeister: This change can probably be removed when the new approach works.

developit: @marvinhagemeister you mean after the switch to `.module.js`?

marvinhagemeister: @developit No, I'm not sure if the change here is needed for this PR in particular.

JoviDeCroock:
It works now, the name has changed to preact-debug though in compiled filename... According to pkg json

JoviDeCroock:
So this works,

Downsides being that it always uses the CJS bundle and the end bundle changed name to preact-debug instead of debug.

I personally don't think that missing a module in here is a problem since it's debug and only dev anyway.

debug/package.json
```diff
@@ -4,8 +4,7 @@
   "version": "1.0.0",
   "private": true,
   "description": "Preact extensions for development",
-  "main": "dist/debug.js",
-  "module": "dist/debug.module.js",
+  "main": "index.js",
   "umd:main": "dist/debug.umd.js",
```
JoviDeCroock: this is still incorrect if we go through with the nameChange.

debug/index.js
```diff
@@ -0,0 +1,3 @@
+if (process.env.NODE_ENV!=='production') {
```
andrewiggins: Should we remove the `env` check in `src/index.js` in favor of this one?

andrewiggins: Might also fix the UMD build of `preact/debug` which has a `'development' === process.env.NODE_ENV` check in it which would fail in the browser

JoviDeCroock: Yes probably

PR #1405, opened 2019-03-14
(feature) - multi select support

JoviDeCroock:
We did not support having multiple `selected` attributes on `<option>` tags.
After evaluating this issue I saw that our compat missed the option to pass an array of values to `<select>`. This was made possible by the first fix.

Fixes: https://github.com/developit/preact/issues/761
Adds ~~`+24B`~~ `+17B` - @marvinhagemeister

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -125,8 +125,8 @@ function createElement(...args) {
 		}
 
 		if (Array.isArray(props.value) && props.multiple && type==='select') {
-			props.children.forEach((child) => {
-				if (props.value.includes(child.props.value)) {
+			toChildArray(props.children).forEach((child) => {
+				if (props.value.indexOf(child.props.value)!==-1) {
 					child.props.selected = true;
 				}
 			});
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -114,6 +114,7 @@ let Children = {
  */
 function createElement(...args) {
 	let vnode = h(...args);
+
 	let type = vnode.type, props = vnode.props;
 	if (typeof type!='function') {
 		if (props.defaultValue) {
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -114,7 +114,6 @@ let Children = {
  */
 function createElement(...args) {
 	let vnode = h(...args);
-
 	let type = vnode.type, props = vnode.props;
 	if (typeof type!='function') {
 		if (props.defaultValue) {
@@ -124,6 +123,14 @@ function createElement(...args) {
 			delete props.defaultValue;
 		}
 
+		if (Array.isArray(props.value) && props.multiple && type==='select') {
+			props.children.forEach((child) => {
+				if (props.value.includes(child.props.value)) {
+					child.props.selected = true;
+				}
+			});
+			delete props.value;
+		}
 		handleElementVNode(vnode, props);
 	}
 
--- a/compat/test/browser/select.test.js
+++ b/compat/test/browser/select.test.js
@@ -3,7 +3,7 @@ import { render, createElement as h } from '../../src';
 
 /** @jsx h */
 
-describe.only('Select', () => {
+describe('Select', () => {
 	let scratch;
 
 	beforeEach(() => {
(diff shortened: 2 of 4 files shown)


compat/test/browser/select.test.js
```diff
@@ -0,0 +1,32 @@
...
+				<select multiple value={['B', 'C']}>
+					<option value="A">A</option>
+					<option value="B">B</option>
+					<option value="C">C</option>
+				</select>
+			);
+		}
+
+		render(<App />, scratch);
+		expect(scratch.firstChild.value).to.equal('B');
```
marvinhagemeister: Let's check that both `options B` and `options A` is selected here to make sure it works :+1:

test/browser/select.test.js
```diff
@@ -28,4 +28,34 @@ describe('Select', () => {
...
+				<select multiple>
+					<option value="A">A</option>
+					<option selected value="B">B</option>
+					<option selected value="C">C</option>
+				</select>
+			);
+		}
+
+		render(<App />, scratch);
+		expect(scratch.firstChild.value).to.equal('B');
```
marvinhagemeister: Same here: Let's check that both values are selected :)

compat/src/index.js
```diff
@@ -124,6 +124,14 @@ function createElement(...args) {
 			delete props.defaultValue;
 		}
 
+		if (Array.isArray(props.value) && props.multiple && type==='select') {
+			props.children.forEach((child) => {
```
marvinhagemeister: `props.children` is not always guaranteed to be an array. When only one child is passed this will break. We can use `toChildArray()` to turn it into an array :tada:

compat/src/index.js
```diff
@@ -124,6 +124,14 @@ function createElement(...args) {
...
+				if (props.value.includes(child.props.value)) {
```
marvinhagemeister: Let's use `indexOf` instead. We officially support browsers all the way down to IE11 and unofficially even older. They have no support for `Array.prototyp.includes`

test/browser/select.test.js
JoviDeCroock: Can we check this with secondChild or? Couldn't figure this out while testing

src/diff/index.js
```diff
@@ -266,6 +266,9 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
 					dom.innerHTML = newHtml && newHtml.__html || '';
 				}
 			}
+			if (newVNode.props.multiple && newVNode.type==='select') {
```
marvinhagemeister: At this point in `diff()` we should know if we're dealing with a `component` or a native `html` element. We can reuse that information in this if-statement here :+1:

test/browser/select.test.js
marvinhagemeister: These are standard DOM methods. The DOM doesn't have a `secondChild` property but you can turn all children (= `dom.childNodes`) into an array:

https://stackoverflow.com/a/3199627/755391

JoviDeCroock:
I notice you are always adding the added bytesize, do you have an addon for this? Or you just check by hand? If you just check by hand I'll add it myself next time.

Allthough I don't think added bytesize is too important when it revolves around added support (correct me if I'm wrong)

marvinhagemeister:
@JoviDeCroock We don't have an automatic process for that. The number is taken from our build output. There we just compare it to the result you get in `master` and post the delta. We always use the first number at the top:

preact-build

It's more of a mental game. You'd be surprised how much all these little savings add up. It also helps being very aware of the size implications. There are cases where a PR may introduce a cool new feature, but the size increase is way more than it should be. In that case the team usually comes together and finds ways to rewrite the code to bring it down again. Having that culture is the secret sauce as to why we're still at 3kB despite adding tons of new features over time :tada:

JoviDeCroock:
Allright, will start paying attention to it in my next fixes/features. You all are very aware of the transpilation costs of things, that's so nice. Hope I can learn the same reflex, since sometimes I really can't see it. Trying to do it really pragmatic in my own OSS libraries but still got a lot to learn as you can see.

jerry153fish:
hello @JoviDeCroock is this feature going to preact as well in the future ?

Issue #1301, opened 2019-03-01
Numeric grid style props being suffixed with 'px'


brandonaaron:
I'm using recal which uses [CSS Grids](https://github.com/rubencodes/recal/blob/5b1a2a7419ea6d16f9978d067fe056e447ece7eb/src/lib/Utils.js#L60-L65). It seems that Preact ends up suffixing these numeric values with a unit due to [not matching the property name here](https://github.com/developit/preact/blob/8aa7ec9e87e34596d0ec292d12fb9f79135e382f/src/constants.js#L16).

PR #1407, opened 2019-03-14
(feat) - support css grid

JoviDeCroock:
Fixes: #1301

One thing I can't figure out is if I should add the mentioned msGridRow because this is as far as I know a measure to support Internet explorer in css and does not belong in our style object.

Seems like my master was not up to date will remake this sec.

File size diff: +7B


JoviDeCroock:
On slack I saw someone "complain" about stutter https://codesandbox.io/s/7m0wjvvok6

Should I add this aswell?

marvinhagemeister:
@JoviDeCroock Mh looks like it's not about the variable name `stutter` in particular but more about variables in general. We should be able to skip the `px` normalization for all variables. Let's work on that in a follow-up PR :+1:

JoviDeCroock:
> @JoviDeCroock Mh looks like it's not about the variable name `stutter` in particular but more about variables in general. We should be able to skip the `px` normalization for all variables. Let's work on that in a follow-up PR 👍

Do you mean css variables? How would we make a distinction between those?

marvinhagemeister:
They always start with two leading dashes `--foo`. We could add a check for that in our regex: `^--`

PR #1409, opened 2019-03-14
(fix) - don't execute non function events and debug them

JoviDeCroock:
Fixes: https://github.com/developit/preact/issues/1350

This adds a props check to debug that throws an error when a prop starting with "on" is no function. In all honesty the other change in preact actually prevents this behavior, but yes... Didn't know what one was most desired.

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -27,6 +27,16 @@ export function initDebug() {
 			);
 		}
 
+		Object.keys(vnode.props).forEach((key) => {
+			if (key[0]==='o' && key[1]==='n' && vnode.props[key]!=='function') {
+				throw new Error(
+					`Component's "${key}" property should be a function,
+					but got [${typeof vnode.props[key]}] instead\n` +
+					serializeVNode(vnode)
+				);
+			}
+		});
+
 		// Check prop-types if available
 		if (typeof vnode.type==='function' && vnode.type.propTypes) {
 			checkPropTypes(vnode.type.propTypes, vnode.props, getDisplayName(vnode), serializeVNode(vnode));
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -27,7 +27,7 @@ export function initDebug() {
 			);
 		}
 
-		Object.keys(vnode.props).forEach((key) => {
+		for (const key in vnode.props) {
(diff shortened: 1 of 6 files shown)


debug/src/debug.js
```diff
@@ -27,6 +27,16 @@ export function initDebug() {
 			);
 		}
 
+		Object.keys(vnode.props).forEach((key) => {
```
JoviDeCroock: Kind off afraid about the performance impact of this, since `Object.keys` iterates over the object and then we iterate over that created array. If this happens on every rerender...

cristianbote: Well maybe you could do a inline `for(let key in vnode.props)` instead?

src/diff/props.js
```diff
@@ -66,7 +66,7 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 		return;
 	}
 	// Benchmark for comparison: https://esbench.com/bench/574c954bdb965b9a00965ac6
-	else if (name[0]==='o' && name[1]==='n') {
+	else if (name[0]==='o' && name[1]==='n' && (typeof value==='function' || typeof oldValue==='function')) {
```
cristianbote: Maybe we should add some tests for this specific handler? Maybe there are others?

debug/src/debug.js
JoviDeCroock: Good point, will update this pr tomorrow after work

src/diff/props.js
JoviDeCroock: Yes, was planning on it but drafted this to see if this was the way to go, like this we could allow people to insert props starting with on

JoviDeCroock: Drafted some tests for this, thing is that the preact changes excludes the need for this change, I just don't know which is better personally.

debug/src/debug.js
JoviDeCroock: It has been adjusted

src/diff/props.js
marvinhagemeister: I don't think these checks belong into core here. They're not useful in production and meant for development purposes only. The code above in `preact/debug` already does that :+1:

JoviDeCroock: So just limit it to debug then?

marvinhagemeister: yup :+1:

Issue #1321, opened 2019-03-05
Build fails when using hooks


flppv:
Hey!

First of all, thank you for a great work on updating Preact, was waiting for it and totally satisfied!

I have a weird issue when trying to build my app, which is already full of hooks:

```
Unable to read file: /Users/.../src/node_modules/preact/hooks/dist/hooks.mjs

TypeError: Cannot read property '__H' of undefined
```

When in a dev mode, everything works like a charm. 

I tried to build without prerender, it didn't help :(

Any ideas on what it could be? Or maybe anyone have built preact@10 app successfully?

developit:
Hi @flppv! Any chance you could set up a reproduction? I'll be trying, but it'd be great if you had something handy.

TimPetricola:
Hey, I had the exact same issue (in a CRA app), and the plugin suggested by @developit fixes the error :)

danieldiekmeier:
If #1331 is the same error, then it has a repro at https://github.com/weijarz/preact-x-bug-r-is-undefined

flppv:
>It might be worth trying out this webpack plugin to see if that fixes the production build.
I'd like not to eject this project if possible.

> Any chance you could set up a reproduction?
Sure, just:
1) Use hooks somewhere
2) run `npm run build` or `preact build`
3) Error

38elements:
Using Rollup can build.

marvinhagemeister:
Fixed with #1425 :tada:

westtrade:
Same bug with Parcel build
preact - 10.0.0-rc.0

Chrome
```
index.js:63 Uncaught TypeError: Cannot read property '__H' of undefined
    at c (index.js:63)
    at v (index.js:105)
    at p.App [as constructor] (App.jsx:26)
    at p.z [as render] (options.js:2)
    at P (component.js:63)
    at g (index.js:260)
    at P (component.js:63)
    at g (index.js:260)
    at P (component.js:63)
    at g (index.js:260)
```

Firefox

```
TypeError: r is undefined
```

marvinhagemeister:
@westtrade Do you happen to have an incorrect import like `preact/hooks/src` instead of `preact/hooks` somewhere in your application?

braco:
This was fixed for me by installing `preact-cli@rc`

didymu5:
I'm having the same issue as @westtrade building with preact-cli widget template.
I initially reported it in https://github.com/preactjs/preact/issues/1373

```
"devDependencies": {
[...]

Issue #1285, opened 2019-01-06
Providing a minified mjs bundle


rmacklin:
It's recommended to use `preact.min.js` instead of `preact.js` in production builds due to the file size reduction from the property mangling: https://github.com/developit/preact/blob/87a4ebe99dc068eaeb8503644c60ffe8ad735771/config/properties.json#L1-L27

This works well if you are using a regular script to pull preact from the CDN and relying on the `preact` global or if you are bundling with webpack. However, it doesn't work if you are importing preact as a native ES module.

In that case, you can successfully import the unminified `preact.mjs` bundle, as seen here:
http://jsfiddle.net/tpck6Lf4/
But since `preact.min.js` is not a native ES module, you cannot change it like this:
```diff
- import { Component, h, render } from 'https://npmcdn.com/preact@8.4.2/dist/preact.mjs'
+ import { Component, h, render } from 'https://npmcdn.com/preact@8.4.2/dist/preact.min.js'
```
(you'd get `SyntaxError: The requested module 'https://npmcdn.com/preact@8.4.2/dist/preact.min.js' does not provide an export named 'h'`)

It'd be awesome if there was a corresponding `preact.min.mjs` that we could import from in production builds to take advantage of the reduced file size.

One way to do this might be to switch from uglify to terser where the minifier can understand ES2015 syntax. Alternatively perhaps we can minify the code with `export` stripped out, and then add it back. Thoughts?

developit:
This is something already fixed in Preact X 💃 
I think we made the call not to backport all the build system changes. Once the repos are merged though it's probably something worth considering (9.x updates that move to microbundle).

marvinhagemeister:
We published an alpha version for Preact X just a few hours ago. It's available on npm via the `preact@next` tag and ships with a minified `mjs` bundle 💯

PR #1425, opened 2019-03-19
Switch from *.mjs -> *.module.js

### LINKED ISSUE #1424: 10.0.0-alpha.2 - Hooks - Uncaught TypeError: Cannot read property 'render' of undefined
Using hooks like this.

```
import {useState} from 'preact/hooks';
```
getting this error
```
hooks.mjs:1 
Uncaught TypeError: Cannot read property 'render' of undefined
    at Module../node_modules/preact/hooks/dist/hooks.mjs (hooks.mjs:1)
    at __webpack_require__ (bootstrap:63)
    at Module../src/index.js (index.js:2)
    at __webpack_require__ (bootstrap:63)
    at Object.2 (rx-preact-app.bundle.js:20462)
    at __webpack_require__ (bootstrap:63)
    at bootstrap:195
    at bootstrap:195
```

marvinhagemeister:
This PR basically reverts #1285 and moves back to `.js` files. It reduces friction with webpack which doesn't resolve `.mjs` the same way as `.js` thus breaking imports in submodules. We've been pointing users to the excellent [webpack-modules](https://github.com/lukeed/webpack-modules) plugin which corrects that as a current workaround.

This PR should remove the need for that plugin making preact work right out of the box with any webpack configuration.

Fixes #1424 , #1321 and lots of questions on slack :+1:

### DIFF
--- a/compat/package.json
+++ b/compat/package.json
@@ -5,7 +5,7 @@
   "private": true,
   "description": "A React compatibility layer for Preact",
   "main": "dist/compat.js",
-  "module": "dist/compat.mjs",
+  "module": "dist/compat.module.js",
   "umd:main": "dist/compat.umd.js",
   "source": "src/index.js",
   "license": "MIT",
--- a/debug/package.json
+++ b/debug/package.json
@@ -5,7 +5,7 @@
   "private": true,
   "description": "Preact extensions for development",
   "main": "dist/debug.js",
-  "module": "dist/debug.mjs",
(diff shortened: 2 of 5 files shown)


steffenmllr:
@marvinhagemeister thanks for the fix - any change this gets published as `10.0.0-alpha.3` ?

marvinhagemeister:
@steffenmllr yup this will be part of the next alpha. We're aiming for the end of the week because we want to get #1440 in too

PR #1428, opened 2019-03-20
Fix safari failing on calc with css variables in SauceLabs

marvinhagemeister:
This PR fixes the failing CI again.

### DIFF
--- a/test/browser/render.test.js
+++ b/test/browser/render.test.js
@@ -427,8 +427,8 @@ describe('render()', () => {
 			});
 
 			it('should not add "px" suffix for custom properties', () => {
-				render(<div style={{ '--foo': 2, animationDelay: 'calc(var(--foo) * 20ms)' }}>test</div>, scratch);
-				expect(sortCss(scratch.firstChild.style.cssText)).to.equal('--foo: 2; animation-delay: calc(var(--foo) * 20ms);');
+				render(<div style={{ '--foo': '100px', width: 'var(--foo)' }}>test</div>, scratch);
+				expect(sortCss(scratch.firstChild.style.cssText)).to.equal('--foo: 100px; width: var(--foo);');
 			});
 		}
 	});


marvinhagemeister:
Just tested this with vanilla js in Safari 10.1 via a browserstack account and I can't reproduce it there. Will need to investigate further.

marvinhagemeister:
Found what's causing this: Safari will add the property twice when the name has mixed casing like `animation-Delay`. Both `animationDelay` and `animation-delay` work correctly, but `animation-Delay` doesn't 🤷‍♂️ 

Simple test-case:

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <title>Page Title</title>
  </head>
  <body>
    <h1>Hello world</h1>
    <div id="root">
      inspect me
    </div>
    <script>
      var el = document.getElementById("root");
      root.style.setProperty("animation-Delay", "calc(var(--foo) * 20ms)");
    </script>
  </body>
</html>
```

marvinhagemeister:
Aha! It looks like their vendor-prefix detection is causing issues. Whenever you use `calc()` with a variable it will add the property twice. When you only supply simple numbers to `calc()` it will work and add both the unprefixed version + one with a `-webkit` prefix.

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
    <title>Page Title</title>
    <meta name="viewport" content="width=device-width, initial-scale=1" />
  </head>
  <body>
    <h1>Hello world</h1>
    <div id="root">
      inspect me ^=
    </div>
    <script>
      var el = document.getElementById("root");
      root.style.setProperty("--foo", "100ms");
      // This will apply `animation-delay` twice
      root.style.setProperty("animation-Delay", "calc(var(--foo) / 2)");
      // Without a variable in calc it will work
      // root.style.setProperty("animation-Delay", "calc(20ms / 2)");
    </script>
  </body>
</html>
```

PR #1429, opened 2019-03-20
Don't iterate over old css string

marvinhagemeister:
This PR fixes an issue where we did iterate over each character of the previous style value. This only happened when the previous value was of type `string`. See the debug logs here: https://travis-ci.org/developit/preact/builds/50882293

Related to #1394

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -50,9 +50,11 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 		}
 		else {
 			if (typeof oldValue==='string') s.cssText = '';
-			// remove values not in the new list
-			for (let i in oldValue) {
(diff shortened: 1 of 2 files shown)


src/diff/props.js
```diff
@@ -50,9 +50,11 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 		}
 		else {
 			if (typeof oldValue==='string') s.cssText = '';
-			// remove values not in the new list
-			for (let i in oldValue) {
-				if (value==null || !(i in value)) s.setProperty(i.replace(CAMEL_REG, '-'), '');
+			else {
```
cristianbote: Nice! 🎉

PR #1433, opened 2019-03-20
Replace Object.assign with custom assign function

### LINKED ISSUE #1430: Remove Object.assign from compat
Our `compat` bundle includes calls to `Object.assign` which is not supported in IE11. This is likely caused by the spread operator and should be replaced with our own `assign()` function.

marvinhagemeister:
Fixes #1430 .
Adds `+8 B` to compat.


marvinhagemeister:
That's the great thing about working with enterprise clients :+1: :100:

PR #1435, opened 2019-03-21
Fix input value not in specified range

### LINKED ISSUE #850: Range Input: Order of Attributes Bug

marvinhagemeister:
Fixes #850 .
Adds ~~`+11`~~ ~~`+10`~~ ~~`+21`~~ `+19 B`

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -19,7 +19,7 @@ export function diffProps(dom, newProps, oldProps, isSvg) {
 	}
 
 	for (let i in oldProps) {
-		if (i!=='children' && i!=='key' && !(i in newProps)) {
+		if (i!=='children' && i!=='key' && (!newProps || !(i in newProps))) {
 			setProperty(dom, i, null, oldProps[i], isSvg);
 		}
 	}
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -12,9 +12,8 @@ import options from '../options';
 export function diffProps(dom, newProps, oldProps, isSvg) {
 	let keys = Object.keys(newProps).sort();
 	for (let i = 0; i < keys.length; i++) {
-		let k = keys[i];
-		if (k!=='children' && k!=='key' && (!oldProps || ((k==='value' || k==='checked') ? dom : oldProps)[k]!==newProps[k])) {
-			setProperty(dom, k, newProps[k], oldProps[k], isSvg);
+		if (keys[i]!=='children' && keys[i]!=='key' && (!oldProps || ((keys[i]==='value' || keys[i]==='checked') ? dom : oldProps)[keys[i]]!==newProps[keys[i]])) {
+			setProperty(dom, keys[i], newProps[keys[i]], oldProps[keys[i]], isSvg);
 		}
 	}
 
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -10,15 +10,13 @@ import options from '../options';
  * @param {boolean} isSvg Whether or not this node is an SVG node
  */
 export function diffProps(dom, newProps, oldProps, isSvg) {
-	for (let i in newProps) {
-		if (i!=='children' && i!=='key' && (!oldProps || ((i==='value' || i==='checked') ? dom : oldProps)[i]!==newProps[i])) {
-			setProperty(dom, i, newProps[i], oldProps[i], isSvg);
+	let keys = Object.keys(newProps).sort();
+	for (let i = 0; i < keys.length; i++) {
+		let k = keys[i];
+		if (k!=='children' && k!=='key' && (!oldProps || ((k==='value' || k==='checked') ? dom : oldProps)[k]!==newProps[k])) {
+			setProperty(dom, k, newProps[k], oldProps[k], isSvg);
 		}
 	}
-
-	// `value` needs to be applied after an input's `min`, `max` and `step`
-	// attribute are set. See preact/#850
-	if (newProps.value!=null) dom.value = newProps.value;
 
 	for (let i in oldProps) {
 		if (i!=='children' && i!=='key' && !(i in newProps)) {
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -15,8 +15,13 @@ export function diffProps(dom, newProps, oldProps, isSvg) {
 			setProperty(dom, i, newProps[i], oldProps[i], isSvg);
 		}
 	}
+
+	// `value` needs to be applied after an input's `min`, `max` and `step`
+	// attribute are set. See preact/#850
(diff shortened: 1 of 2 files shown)


src/diff/props.js
```diff
@@ -11,12 +11,17 @@ import options from '../options';
  */
 export function diffProps(dom, newProps, oldProps, isSvg) {
```
robertknight: Can `newProps` or `oldProps` be null or undefined here? If so it would help to clarify that in the function signature.

src/diff/props.js
```diff
@@ -11,12 +11,17 @@ import options from '../options';
...
 		}
 	}
+
+	// `value` needs to be applied after an input's `min`, `max` and `step`
+	// attribute are set. See preact/#850
+	if (newProps.value!=null) dom.value = newProps.value;
+
 	for (let i in oldProps) {
-		if (i!=='children' && i!=='key' && (!newProps || !(i in newProps))) {
+		if (i!=='children' && i!=='key' && !(i in newProps)) {
```
robertknight: Is the removal of `!newProps` check because `newProps` is guaranteed to be non-null here?

test/browser/render.test.js
```diff
@@ -241,6 +241,14 @@ describe('render()', () => {
 		expect(root.children[3]).to.have.property('value', '');
 	});
 
+	it('should set value inside the specified range', () => {
+		render(
+			<input type="range" value={0.5} min="0" max="1" step="0.05" />,
+			scratch
+		);
```
robertknight: _(Note for my understanding)_ - So in this test case, if the `type` property is set to `range` and then `value` is set before `min`, `max` or `step` then `value` will be rounded to "1" because the [default values for min, max and step](https://html.spec.whatwg.org/multipage/input.html#range-state-(type=range)) are 0, 100 and 1 respectively.

src/diff/props.js
```diff
@@ -11,12 +11,17 @@ import options from '../options';
...
 export function diffProps(dom, newProps, oldProps, isSvg) {
 	for (let i in newProps) {
-		if (i!=='children' && i!=='key' && (!oldProps || oldProps[i]!=newProps[i])) {
+		if (i!=='children' && i!=='key' && i!=='value' && (!oldProps || oldProps[i]!=newProps[i])) {
 			setProperty(dom, i, newProps[i], oldProps[i], isSvg);
 		}
 	}
+
+	// `value` needs to be applied after an input's `min`, `max` and `step`
+	// attribute are set. See preact/#850
```
robertknight: I wonder if there are any other dependencies here? I presume there is one between `type` and `value`, which is also covered by this change.

robertknight: Do you mean after an input's `min`, `max` and `step` _properties_ are set?

src/diff/props.js
marvinhagemeister: yup, we always set at least an empty `object` https://github.com/developit/preact/blob/0baf368ca803e0deb6b3d7edd2e09f6bbd564505/src/create-element.js#L12

test/browser/render.test.js
marvinhagemeister: That's exactly what happens 👍 The browser won't update the value when the range boundaries change.

src/diff/props.js
marvinhagemeister: ohh you got me haha

marvinhagemeister: Not sure but we didn't receive any bug reports for these so far, even during the 8.x release line. So I guess browsers deal with `type` themselves.

src/diff/props.js
marvinhagemeister: Afaik they are never `null` here. When I remove the `!oldProps` check the byte-size goes up by `+7B` 🤷‍♂️

marvinhagemeister:
This PR also fixes the CI error on the `master` branch 👍

src/diff/index.js
```diff
@@ -270,9 +270,9 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
 					dom.innerHTML = newHtml && newHtml.__html || '';
 				}
 			}
-			if (newProps.multiple) {
-				dom.multiple = newProps.multiple;
-			}
+			// Needs to be set before diffing children for <select>'s with multiple
+			// options.
+			if (newProps.multiple) dom.multiple = newProps.multiple;
```
developit: Is this still needed with children now being diffed before props?

src/diff/props.js
developit: Any chance this would trip for a vnode that swaps from Text to Element?

```
render(<div>foo</div>, root)
render(<div><span a /></div>, root)
```

developit:
So one weird thing about this is that the issue only shows up when the `value` isn't specified last. At the time #850 was filed I'd wondered if this would be easier to implement via Object.keys.

src/diff/index.js
marvinhagemeister: Unfortunately not. This issue is caused when the value is set before the range boundaries are applied. At the time when value is set, the browser will default to the fallback range of 0-100 and won't update `value` when `min/max/step` are added later.

marvinhagemeister: yeah the multiple property must be set before `children` are diffed, otherwise there will be only one `<option>` with a selected value. That's the only exception we've found so far :/

src/diff/props.js
marvinhagemeister: Good call, we need to keep that check :+1:

marvinhagemeister:
Went with the `Object.keys()` route as @developit suggested. Added back the `!newProps` null check and with it even the previous solution would be the same in term of byte size.

PR #1436, opened 2019-03-21
Chore/support compat testutils

JoviDeCroock:
Adds a script to copy test-utils to /compat after build, this introduces no breaking changes for people depending on /test-utils

Closes https://github.com/developit/preact/issues/1431

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,6 @@
     "build:hooks": "microbundle build --raw --cwd hooks",
     "build:test-utils": "microbundle build --raw --cwd test-utils",
     "build:compat": "microbundle build --raw --cwd compat --globals 'preact/hooks=preactHooks'",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -17,6 +17,7 @@
     "build:hooks": "microbundle build --raw --cwd hooks",
     "build:test-utils": "microbundle build --raw --cwd test-utils",
     "build:compat": "microbundle build --raw --cwd compat --globals 'preact/hooks=preactHooks'",
+    "postbuild": "cp -R test-utils/ compat/test-utils",
```
marvinhagemeister: Maybe we can make it easier for us by just creating a file at `compat/test-utils.js` and populate it with:

```js
module.exports = require("preact/test-utils");
```

developit: +1 to this - I'd be worried about breaking import paths when copying things around.

PR #1437, opened 2019-03-21
Fix/inconsistent act

JoviDeCroock:
Act would not flush effects introduced by the changing of a state, this would result in inconsistent test behavior

### DIFF
--- a/test-utils/src/index.js
+++ b/test-utils/src/index.js
@@ -11,16 +11,24 @@ export function setupRerender() {
 }
 
 export function act(cb) {
-	const previousDebounce = options.debounceRendering;
 	const previousRequestAnimationFrame = options.requestAnimationFrame;
 	const rerender = setupRerender();
 	let flush;
+	// Override requestAnimationFrame so we can flush pending hooks.
 	options.requestAnimationFrame = (fc) => flush = fc;
+	// Execute the callback we were passed.
 	cb();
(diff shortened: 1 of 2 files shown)


test-utils/src/index.js
```diff
@@ -11,16 +11,24 @@ export function setupRerender() {
...
 	let flush;
+	// Override requestAnimationFrame so we can flush pending hooks.
 	options.requestAnimationFrame = (fc) => flush = fc;
+	// Execute the callback we were passed.
 	cb();
+	// State COULD be built up flush it.
 	if (flush) {
 		flush();
 	}
+	// Rerender
```
marvinhagemeister: Nit: We can skip this comment. It doesn't add any new information to the code :)

test-utils/test/shared/act.test.js
```diff
@@ -1,5 +1,6 @@
 import { options, createElement as h, render } from 'preact';
 import { useEffect, useState } from 'preact/hooks';
+import sinon from 'sinon';
```
marvinhagemeister: This breaks tests on IE because this will import an `esm` bundle. For that reason `sinon` is available as a global in our test suite and never imported. The global is aliased to the proper `es5` file.

JoviDeCroock:
How did my coverage decrease.... I haven't even touched src

PR #1442, opened 2019-03-24
fix: missing mangle config for _prevState

mochiya98:
it is used in:
https://github.com/developit/preact/blob/750189b1978a032dfcd8e255325a32580d2c32b8/debug/src/devtools/index.js#L158-L166


mochiya98:
thanks!

PR #1443, opened 2019-03-24
Update Component in internal.d.ts

38elements:
https://github.com/developit/preact/pull/1313/files

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -13,6 +13,7 @@ import { Fragment } from './create-element';
 export function Component(props, context) {
 	this.props = props;
 	this.context = context;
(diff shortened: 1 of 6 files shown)


src/component.js
```diff
@@ -13,6 +13,7 @@ import { Fragment } from './create-element';
 export function Component(props, context) {
 	this.props = props;
 	this.context = context;
+	// this.constructor // When component is functional component, this is reseted to functional component
```
marvinhagemeister: Nit: Past tense of `reset` is also `reset`.

38elements: Thank you for pointing out.
I fixed it.

PR #1446, opened 2019-03-24
Fix: changes in getDerivedStateFromProps is not reflected to _nextState

mochiya98:
```jsx
let afterSetState;

class App extends Component {
  constructor(props) {
    super(props);
    this.state = { a: 0 };
    afterSetState = async () => {
      await new Promise(r => this.setState({ a: 1 }, r));
      this.setState({ a: 2 });
    };
  }
  render() {
    return <Tester a={this.state.a} />;
  }
}

class Tester extends Component {
  constructor(props) {
    super(props);
    this.state = { n: 0 };
  }
  static getDerivedStateFromProps(_, prevState) {
    return { n: prevState.n + 1 };
  }
  componentDidMount() {
    this.setState({ dummy: "hoge" }, afterSetState);
  }
  render() {
    return <div className="App">result: {String(this.state.n)}</div>;
  }
}
// 1st - mount 0 -> 1 (_nextState: null, state: { n: 1 })
// 2nd - setState 1 -> 2 (_nextState: { n: 2 }, state: { n: 2 })(_nextState===state)
// 3rd - updateProps 2 -> 3 (_nextState: { n: 2 }, state: { n: 3 }) <- incorrect!
// (_nextState should be { n: 3 } or null)
// 4th - updateProps 2 -> 3 (_nextState: { n: 2 }, state: { n: 3 }) <- :(
```

There are two ways to solve this:

- make `c._nextState` and `s` the same reference to reflect the changes
- assign null to `c._nextState` after applying the changes

In the former way, it is possible to avoid unnecessary copies in setState.

React: https://codesandbox.io/s/rroz85x61q
Preact@10.0.0-alpha.2: https://codesandbox.io/s/k3ov5xl2w5


src/diff/index.js
```diff
@@ -90,7 +90,7 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
 			let s = c._nextState || c.state;
 			if (newType.getDerivedStateFromProps!=null) {
 				oldState = assign({}, c.state);
-				if (s===c.state) s = assign({}, s);
+				if (s===c.state) s = c._nextState = assign({}, s);
```
marvinhagemeister: Awesome :tada: I'd love to have a test case for this so that we don't regress on this feature in any future refactorings :+1:

test/browser/lifecycle.test.js
```diff
@@ -510,6 +510,22 @@ describe('Lifecycle methods', () => {
...
+			render(<Foo foo="baz" />, scratch);
+			expect(element.textContent).to.be.equal('5');
+			expect(stateArg).to.deep.equal({
+				value: 4
+			});
+
+			// New Props (see #1446)
+			// 5 -> 6 in gDSFP
+			render(<Foo foo="qux" />, scratch);
+			expect(element.textContent).to.be.equal('6');
```
mochiya98: test will fail here before the patch

src/diff/index.js
mochiya98: I did it. 👍

PR #1447, opened 2019-03-24
Fix type definition

38elements:
I think that `text`, `key`, `ref`, `_children`, `_dom`, `_lastDomChild`,  and `_component` always exist.

https://github.com/developit/preact/blob/66aa9e24bbf116fa30a5fddce7ebc21d8ea3199f/src/create-element.js#L54-L64

`props` is set to null.

https://github.com/developit/preact/blob/66aa9e24bbf116fa30a5fddce7ebc21d8ea3199f/src/create-element.js#L87

May I ask.
This is another issue.
Is the reason why setting null to `key` and `ref`  for V8?
I don't know why it is `return createVNode(null, null, possibleVNode, null, null);` instead of `return createVNode(null, null, possibleVNode);`.
https://github.com/developit/preact/blob/66aa9e24bbf116fa30a5fddce7ebc21d8ea3199f/src/create-element.js#L87

### DIFF
--- a/src/internal.d.ts
+++ b/src/internal.d.ts
@@ -25,16 +25,16 @@ export interface PreactElement extends HTMLElement {
 export interface VNode<P = {}> extends preact.VNode<P> {
 	// Redefine type here using our internal ComponentFactory type
 	type: string | ComponentFactory<P> | null;
-	_children?: Array<VNode> | null;
+	_children: Array<VNode> | null;
 	/**
 	 * The [first (for Fragments)] DOM child of a VNode
 	 */
-	_dom?: PreactElement | Text | null;
+	_dom: PreactElement | Text | null;
 	/**
 	 * The last dom child of a Fragment, or components that return a Fragment
 	 */
-	_lastDomChild?: PreactElement | Text | null;
-	_component?: Component | null;
+	_lastDomChild: PreactElement | Text | null;
+	_component: Component | null;
 }
 
 export interface Component<P = {}, S = {}> extends preact.Component<P, S> {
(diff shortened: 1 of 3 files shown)


src/internal.d.ts
```diff
@@ -25,16 +25,16 @@ export interface PreactElement extends HTMLElement {
 export interface VNode<P = {}> extends preact.VNode<P> {
 	// Redefine type here using our internal ComponentFactory type
 	type: string | ComponentFactory<P> | null;
-	_children?: Array<VNode> | null;
+	_children: Array<VNode> | null;
```
andrewiggins: I think these are marked as "optional" with the `?` because these properites are lazily set in the diff, and so are initially undefined until their value is set. Specifically for `_children`, I don't think we ever set this to `null` so a more specific type might be `Array<VNode | null> | undefined`.

38elements: Thank you for reply.
I may be misunderstanding.
I think that VNode is produced only by `createVNode()`.
Since there is always `_children` key, I think that `?` is unnecessary.
https://github.com/developit/preact/blob/66aa9e24bbf116fa30a5fddce7ebc21d8ea3199f/src/create-element.js#L54-L64
  
  
This is another issue.
I don't know why it is `return createVNode(null, null, possibleVNode, null, null);` instead of `return createVNode(null, null, possibleVNode);`.
https://github.com/developit/preact/blob/66aa9e24bbf116fa30a5fddce7ebc21d8ea3199f/src/create-element.js#L87

andrewiggins: 🤦‍♂️ Oh, I'm completely wrong. I misread your description which includes the link to the source demonstrating how they are always defined (just like you did again here). My mistake. And yes, all VNode creation should go through `createVNode` for this exact reason. Thanks for the thoroughness and kind reply 😊

PR #1448, opened 2019-03-24
move JSX namespace into preact one

### LINKED ISSUE #1036: TS: Move JSX namespace to preact one
Just read the changelog for TypeScript 2.8 and they finally allow mixing different JSX namespaces. This improves interoperability when preact is used alongside react components.

Note that we should not release this until TS 2.8 is out.

TS 2.8 Changelog: https://blogs.msdn.microsoft.com/typescript/2018/03/15/announcing-typescript-2-8-rc/

just-boris:
closes #1036. Still requires some work, but open for initial feedback.

This change fixes issues when preact is used together with `@types/react` in dependencies. Conflicting JSX namespace definitions confuse typescript. Here is a demo project, where I tested integration with react typings: https://github.com/just-boris/preact-mix 

Implementing the approach described in the related typescript issue: https://github.com/Microsoft/TypeScript/issues/23762#issuecomment-385444169

Not yet ready to merge, because it requires the fix in typescript to be released https://github.com/Microsoft/TypeScript/pull/30160.

### DIFF
--- a/test/ts/jsx-namespacce-test.tsx
+++ b/test/ts/jsx-namespacce-test.tsx
@@ -1,0 +1,16 @@
+import { createElement, Component } from "../../src";
+
+// declare global JSX types that should not be mixed with preact's internal types
+declare global {
+	namespace JSX {
+		interface Element {
+			unknownProperty: string
+		}
+	}
+}
+
+class SimpleComponent extends Component {
+	render() {
+		return <div>It works</div>;
+	}
+}
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,9 +1,11 @@
 export = preact;
 export as namespace preact;
 
-import "./jsx";
+import { JSXInternal } from "./jsx";
 
 declare namespace preact {
+	export import JSX = JSXInternal;
+
 	//
 	// Preact Virtual DOM
 	// -----------------------------------
@@ -134,7 +136,7 @@ declare namespace preact {
 
 	function createElement(
 		type: string,
-		props: JSX.HTMLAttributes & JSX.SVGAttributes & Record<string, any> | null,
+		props: JSXInternal.HTMLAttributes & JSXInternal.SVGAttributes & Record<string, any> | null,
 		...children: ComponentChildren[]
 	): VNode<any>;
 	function createElement<P>(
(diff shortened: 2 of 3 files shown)


developit:
@just-boris It seems like it might be a little while before we could release this since it breaks TS <3.4, right?

marvinhagemeister:
It's not a good sample size but from my experience TS users don't take long to upgrade.

marvinhagemeister:
FYI: TypeScript 3.4 was released just a few days ago :tada:

just-boris:
> It seems like it might be a little while before we could release this since it breaks TS

I will do extra testing to see how exactly it breaks older TS and if there are any possible workarounds (for example, various options in `tsconfig.json`).

> It's not a good sample size but from my experience TS users don't take long to upgrade.

Typescript minor releases contain breaking changes, so users are advised to lock version in dependencies. So, onboarding on the new version would require an effort from developers.

So, it is worth to consider possible impact of shipping this change. I will check if this change fixes more things than breaks and update the PR.

just-boris:
Updated the PR, added an extra test for this scenario.

I have also checked how this works under different versions of Typescript. The above-mentioned error (Microsoft/TypeScript#30160) is happening only when you are using Preact and JSX directly. But if you have `preact` is a transitive dependency and you do not use it directly, you will be fine.

That said, I believe this change brings more benefits than opens potential issues. Moreover, as soon as change gets released as an alpha version, it should be fine to require the newest typescript to make it work.

test/ts/jsx-namespacce-test.tsx
```diff
@@ -0,0 +1,16 @@
+import { createElement, Component } from "../../src";
+
+// declare global JSX types that should not be mixed with preact's internal types
+declare global {
+	namespace JSX {
+		interface Element {
+			unknownProperty: string
+		}
+	}
+}
```
just-boris: This snippet simulates what `@types/react` does. Will you be interested in having the real React typings as a dev-dependency instead of this mock?

jeremy-coleman:
I recently ran into a case where some types are actually different, onDoubleClick in react is onDblClick in preact, if i was using react types i would have never found the cause

JoviDeCroock:
> I recently ran into a case where some types are actually different, onDoubleClick in react is onDblClick in preact, if i was using react types i would have never found the cause

Yes this is normalized in preact/compat but in preact the native events are used.

just-boris:
@marvinhagemeister @developit do you need anything else from me to get this PR merged?

marvinhagemeister:
@just-boris The PR is awesome and will be merged soon :+1: We were just more focused on getting `alpha 3` out first :slightly_smiling_face:

PR #1449, opened 2019-03-24
Render bailout on equal state transition

JoviDeCroock:
Fixes: https://github.com/developit/preact/issues/1441

Size hooks: 741B -> 757B (+16)

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -87,8 +87,11 @@ export function useReducer(reducer, initialState, init) {
 			init == null ? invokeOrReturn(null, initialState) : init(initialState),
 
 			action => {
-				hookState._value[0] = reducer(hookState._value[0], action);
-				hookState._component.setState({});
+				const nextValue = reducer(hookState._value[0], action);
+				if (!Object.is(hookState._value[0], nextValue)) {
+					hookState._value[0] = nextValue;
+					hookState._component.setState({});
+				}
 			}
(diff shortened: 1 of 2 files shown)


hooks/src/index.js
```diff
@@ -87,8 +87,11 @@ export function useReducer(reducer, initialState, init) {
 			init == null ? invokeOrReturn(null, initialState) : init(initialState),
 
 			action => {
-				hookState._value[0] = reducer(hookState._value[0], action);
-				hookState._component.setState({});
+				const nextValue = reducer(hookState._value[0], action);
+				if (!Object.is(hookState._value[0], nextValue)) {
```
andrewiggins: `Object.is` is an ES6 feature of JS so I don't think we can use it here (or we have to change our browser support matrix or specify that an Object.is polyfill is pre-req of `preact/hooks`). Should we just do an `===` check in `preact/hooks` and provide a `Object.is` polyfill and version of `useReducer` in `preact/compat`?

JoviDeCroock: It could be a feature to make it so that we have options.stateBailoutAlgo or something along those lines.

marvinhagemeister: Just to amend: The difference between `===` and `Object.is` only seems to be that the latter can compare `NaN` values and differentiate between `+0` and `-0` :man_shrugging:

PR #1450, opened 2019-03-24
apply normalisation for every vnode

JoviDeCroock:
fixes: https://github.com/developit/preact/issues/1445

### DIFF
--- a/test/browser/compat.test.js
+++ b/test/browser/compat.test.js
@@ -1,0 +1,54 @@
+import { createElement, render } from 'preact';
+import { setupScratch, teardown } from '../../test/_util/helpers';
+
+import React from 'preact/compat'; // eslint-disable-line
+
+describe('imported compat', () => {
+
+	let scratch;
+
+	beforeEach(() => {
+		scratch = setupScratch();
+	});
+
+	afterEach(() => {
+		teardown(scratch);
+	});
+
+	it('should patch events', () => {
+		let spy = sinon.spy();
+		render(<div onClick={spy} />, scratch);
+		scratch.firstChild.click();
+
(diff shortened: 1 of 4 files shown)


JoviDeCroock:
Coverall seems to bug on my PR's sometimes, also still getting spammed quite hard by travis.

JoviDeCroock:
It could be a feature to make it so that we have options.stateBailoutAlgo or something along those lines.

JoviDeCroock:
we have a test for that first case where we check if onDoubleClick and onBeforeInput are normalized. Am I misunderstanding this or?

test/browser/compat.test.js
```diff
@@ -0,0 +1,54 @@
+import { createElement, render } from 'preact';
+import { setupScratch, teardown } from '../../test/_util/helpers';
+
+import React from 'preact/compat'; // eslint-disable-line
+
+describe('imported compat', () => {
```
marvinhagemeister: Since these tests are for compat, we should move them to the compat test suite :+1:

marvinhagemeister:
That's strange: Looking at the output in our CI it seems like it doesn't run all tests and aborts somewhere inside Lifecylce tests. Past builds seem to halt at the same position, but no error code is thrown or anything. The test suite returns a proper `exit 0` code. Because a lot of tests are not run the coverage will fail. Not sure what's causing this.

test/browser/compat.test.js
JoviDeCroock: moved file

JoviDeCroock:
Yes saw so aswell :/ Don't understand why though

PR #1451, opened 2019-03-24
support svg attributes

JoviDeCroock:
Fixes: https://github.com/developit/preact/issues/1444

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -11,7 +11,7 @@ import options from '../options';
  */
 export function diffProps(dom, newProps, oldProps, isSvg) {
 	for (let i in newProps) {
-		if (i!=='children' && i!=='key' && (!oldProps || oldProps[i]!=newProps[i])) {
+		if (i!=='children' && i!=='key' && (!oldProps || ((i==='value' || i==='checked') ? dom : oldProps)[i]!==newProps[i])) {
 			setProperty(dom, i, newProps[i], oldProps[i], isSvg);
 		}
 	}
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -89,7 +89,6 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 		else dom.removeAttribute(name);
 	}
 	else if (typeof value!=='function') {
-		console.log(isSvg, typeof value, name, name.replace(/^xlink:?/, ''), name !== name.replace(/^xlink:?/, '')); // eslint-disable-line
 		if (name!==(name = name.replace(/^xlink:?/, ''))) dom.setAttributeNS('http://www.w3.org/1999/xlink', name.toLowerCase(), value);
 		else dom.setAttribute(name, value);
 	}
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -85,11 +85,12 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 		dom[name] = value==null ? '' : value;
 	}
 	else if (value==null || value===false) {
-		if (name !== (name = name.replace(/^xlink:?/, ''))) dom.removeAttributeNS('http://www.w3.org/1999/xlink', name.toLowerCase());
+		if (name!==(name = name.replace(/^xlink:?/, ''))) dom.removeAttributeNS('http://www.w3.org/1999/xlink', name.toLowerCase());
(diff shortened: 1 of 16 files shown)


JoviDeCroock:
This test does fail locally now, any clue?

JoviDeCroock:
Old tests did not seem to try these cases https://github.com/developit/preact/blob/3aaa578270c0c6cd1fb7f5cf992737eec2934433/test/browser/svg.js

src/diff/props.js
```diff
@@ -35,6 +35,7 @@ let CAMEL_REG = /-?(?=[A-Z])/g;
 function setProperty(dom, name, value, oldValue, isSvg) {
 	let v;
 	if (name==='class' || name==='className') name = isSvg ? 'class' : 'className';
+	const ns = isSvg && (name !== (name = name.replace(/^xlink:?/, '')));
```
andrewiggins: Do we have to restrict this to SVG elements? I think setting namespaced attributes on normal dom elements (while not as common) is still valid, so perhaps we can save some bytes by removing the `isSvg` check?

developit: Should be able to leave this enabled for any element. I believe it was guarded by isSvg in Preact 8 just to avoid the cost of doing a comparison and `.replace()` call in the more common non-svg case.

developit: @JoviDeCroock I'm wondering:  is it possible to move this line (38) down into the attribute set/remove clauses on 88 and 91?  It's duplication, but it would mean the vast majority of setProperty() calls never have to run this code.

andrewiggins: >  avoid the cost of doing a comparison and .replace() call in the more common non-svg case

Oh, true. That make sense :)

test/browser/svg.test.js
```diff
@@ -29,6 +29,22 @@ describe('svg', () => {
 		`.replace(/[\n\t]+/g,'')));
 	});
 
+	it('should support svg attributes', () => {
```
marvinhagemeister: This test passes even without using any of the `*NS` functions. Maybe we should add a test with an attribute that has a namespace. Something like `xlink:href` which I think was the why @developit found out that we lack the `*NS` functions.

_Note: mdn lists [xlink:href](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href) as being deprecated in SVG2 in favour of just `href`, but trouble is that no browser seem to really support SVG2 at this point._

JoviDeCroock:
> Interesting... I'm having a little trouble following. The test is failing because its rendering a different year than we expect? Are browsers inconsistent in their handling of rendering namespaces?

Yes, locally I get the 1999 tag injected and travis injects the 2000 one

JoviDeCroock:
I've added the test that fails in travis again.

andrewiggins:
Wow, that is bizarre...

marvinhagemeister:
Ok I just cross checked with a real IE11 instance and it looks like it always adds the `xmlns:xlink` attribute as soon as an `xlink:*` attribute is present. The same is true for `xmlns="http://www.w3.org/2000/svg"`.

There doesn't even have to be any JavaScript involved, the behaviour is observable with plain HTML:

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8" />
    <title>foo</title>
  </head>
  <body>
    <svg viewBox="0 0 360 360" xlink:href="www.preact.com"></svg>
    <script>
      console.log(document.body.innerHTML);
    </script>
  </body>
</html>
```

Screenshot 2019-03-25 at 20 53 41

To make our CI green again I'd filter out both `xmlns` and `xmlns:xlink` attributes. Looking at the test case we're already filtering out the former, but not the latter.

JoviDeCroock:
Wow @marvinhagemeister, the more you know. Why does it execute in another browser though (travis vs local)? Changed it now

marvinhagemeister:
Travis runs all tests via SauceLabs across all browsers in parallel. Karma then just consolidates the logs and sometimes they are confusing. The bug is weird no doubt, but so are lots of things in IE11 🤷‍♂️ 

One last thing left to do and then it's ready to go in 🎉 It looks like we need a second test for removing a namespaced attribute:

Screenshot 2019-03-25 at 21 11 03

andrewiggins:
Woohoo! Nice find :D

Issue #1343, opened 2019-03-06
10.0.0-alpha.0 - Returning undefined redraws the component tree


jpodwys:
Thank you for the exciting major release! I've noticed a behavior difference between Preact 8 and X.

In Preact 8, returning nothing (undefined) from a render method bails with no noticeable negatives.
https://codesandbox.io/s/z3155xpy1m

In Preact X, doing the same thing redraws the component tree (made visible with a CSS animation).
https://codesandbox.io/s/llkv0l9qx7

Strangely, moving the component that returns undefined to the bottom of the markup in Preact X hides the issue.
https://codesandbox.io/s/j3x8pk299w

marvinhagemeister:
Related #1327

jpodwys:
Thanks a bunch for your work on this @marvinhagemeister and @Almo7aya!

PR #1452, opened 2019-03-24
Jump to the next childDom if its `_dom` is null

Almo7aya:
I think this should fix https://github.com/developit/preact/issues/1343, but not other reconciliation issues like https://github.com/developit/preact/issues/1326

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -29,7 +29,15 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 
 	let oldChildrenLength = oldChildren.length;
 
-	childDom = oldChildrenLength ? oldChildren[0] && oldChildren[0]._dom : null;
+	if (oldChildrenLength) {
+		for (i = 0; i< oldChildrenLength; i++) {
+			if (oldChildren[i] && oldChildren[i]._dom) {
+				childDom = oldChildren[i]._dom;
+				break;
+			}
+		}
+	}
+
 	if (excessDomChildren!=null) {
 		for (i = 0; i < excessDomChildren.length; i++) {
 			if (excessDomChildren[i]!=null) {
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -29,12 +29,10 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 
 	let oldChildrenLength = oldChildren.length;
 
-	if (oldChildrenLength) {
-		for (i = 0; i< oldChildrenLength; i++) {
-			if (oldChildren[i] && oldChildren[i]._dom) {
-				childDom = oldChildren[i]._dom;
-				break;
-			}
+	for (i = 0; i < oldChildrenLength; i++) {
+		if (oldChildren[i] && oldChildren[i]._dom) {
+			childDom = oldChildren[i]._dom;
+			break;
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


andrewiggins:
Hi @Almo7aya! This is great! Thanks for taking the time to contribute a fix 🎉 

Could you add test case so we don't regress this in the future?

marvinhagemeister:
Related: #1440

Almo7aya:
@andrewiggins I'm stuck with writing a test case for this, the saveFocus function does not let me check for moving children, it saves the focused child and refocused it after rendering, so I can't detect if the children moved or not. Have any idea how to do this? :)

marvinhagemeister:
@Almo7aya Let's use the code from #1343 as a test :tada: 

```jsx
it('should not re-render when a component returns undefined', () => {
  let Dialog = () => undefined;
  let updateState;
  class App extends Component {
    constructor(props) {
      super(props);
      this.state = { name: '' };
      updateState = () => this.setState({ name: ', friend' });
    }

    render(props, { name }) {
      return (
        <div>
          <Dialog />
          <h1 class="fade-down">Hi{name}</h1>
        </div>
      );
    }
  }

  render(<App />, scratch);
  clearLog();

  updateState();
  rerender();

  // We don't log text updates
  expect(getLog()).to.deep.equal([]);
});
```

Almo7aya:
@marvinhagemeister Sorry!, but even this doesn't work, it always passes. The problem is children reconciler append children at the end of the parent not in their correct position this cause starting the animation, I cannot detect if the children were appended at the end or at their correct position using `insertBefore` 🤦‍♂️!!

marvinhagemeister:
@Almo7aya are you sure? The test only passes with the changes from your PR. It fails in `master`.

Almo7aya:
@marvinhagemeister I think I have some problems with my testing environment if you are sure that this fails in master and passes with this PR I'll commit your test case anyway?!

marvinhagemeister:
@Almo7aya Good idea, we can always check if it runs properly in Travis here :+1:

Almo7aya:
Just check it with travis it did fail with the previous changes
https://travis-ci.com/Almo7aya/preact/builds/105919357#L1317
https://github.com/Almo7aya/preact/blob/test-tesing/src/diff/children.js#L32

marvinhagemeister:
@Almo7aya Oh right, I found why it always passes. Our test logger needs to be set up for it to work. We usually do it in the test runners `before` hook:

```js
import { logCall } from '../_util/logCall';

describe("render()", () => {
  // ...

  before(() => {
    logCall(Element.prototype, 'appendChild');
    logCall(Element.prototype, 'insertBefore');
    logCall(Element.prototype, 'removeChild');
    logCall(Element.prototype, 'remove');
  });

  // ...
});
```

Only once it's attached to a prototype it will start logging DOM operations.

Almo7aya:
@marvinhagemeister That's great, thank you 🎉!

src/diff/children.js
```diff
@@ -29,7 +29,15 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 
 	let oldChildrenLength = oldChildren.length;
 
-	childDom = oldChildrenLength ? oldChildren[0] && oldChildren[0]._dom : null;
+	if (oldChildrenLength) {
```
marvinhagemeister: Removing the length check around the `for` loop saves 2 bytes and all tests keep passing 🎉

src/diff/children.js
```diff
@@ -29,7 +29,15 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
...
+		for (i = 0; i< oldChildrenLength; i++) {
```
marvinhagemeister: nit: spacing: `i< oldChildrenLength` -> `i < oldChildrenLength`

Almo7aya:
@marvinhagemeister  I think it's ready now. Thank you for your help, I didn't know about loggers functions in the test utils. Again thank you!🎉.

Issue #1388, opened 2019-03-10
Handling keys


phaux:
This happens since forever. 

- Single elements/components aren't remounted when their key changes
- Sometimes elements/components are remounted when they should be just reordered

Example GIF: every element has an entering animation and transition on `top` css property and this is what happens when I just reorder them. (data from randomuser.me)

![Peek 2019-03-10 20-45](https://user-images.githubusercontent.com/1270987/54092905-341b4780-4392-11e9-80fa-a2d3dfec8bae.gif)

Example GIF 2: User details page's key changes upon navigation, but it doesn't get remounted. You can tell by the lack of entering animation.

![Peek 2019-03-11 00-17](https://user-images.githubusercontent.com/1270987/54093046-988ad680-4393-11e9-91f8-e7bc06db136c.gif)

I'm using preact@10.0.0-alpha.1

[Repro repo](https://github.com/phaux/web-app-boilerplate)

marvinhagemeister:
Oh wow that easily the most beautiful test case that'd we've received :100:

With our rewritten reconciler in Preact X we're finally in a position to address these kind of issues. We already started working on it before cutting our first alpha release and it's coming along greatly. It still needs time until it can be merged, but rest assured that it's something that's on our radar and we definitely want to see fixed for the final release :+1: 

/cc @andrewiggins

developit:
Definitely voting to use this demo to test keys, it's very pleasant :D

marvinhagemeister:
#1440 fixes the second point from the list :tada:

marvinhagemeister:
Just checked the demo again and we're not doing any unnecessary unmounting anymore :+1:

Preact X alpha 4:

![people-demo-preact](https://user-images.githubusercontent.com/1062408/55667520-9d8d5980-585d-11e9-9678-63d808307a5d.gif)

Although we don't do any unmounts anymore the visual output probably looks not like you intended. My guess is that the list items are only meant to move around instead of "zooming in" when the list is reordered. So I was curious, because we only call `dom.appendChild` and `dom.insertBefore()` during that phase to reorder the elements. 

After a bit of digging it seems like these cause the zoom animation to trigger. To be able to smoothly move the list items around one would have to prevent any methods that somehow change the the dom ordering and only move the nodes around via css. The dom ordering happens because each row is tagged with the person's `id` as the `key`. Note that this is correct the correct way, just difficult for animations.

[...]

PR #1453, opened 2019-03-25
Add people demo from #1388

andrewiggins:
Add the amazing people demo @phaux linked to in #1388 to our repo to help us catch errors and inefficiencies in children diffing sooner.

The original in written in TypeScript. I just copied it over as is and am using the new babel-typescript preset to remove the types.

I did convert the original css files to scss in order to scope the styles to only the people demo HTML and not the entire app.


demo/webpack.config.js
```diff
@@ -35,12 +60,12 @@ module.exports = {
 							modules: false,
 							loose: true
 						}],
-						require.resolve('@babel/preset-react'),
+						[require.resolve('@babel/preset-react')],
```
andrewiggins: I added these here just cuz I found it easier to read when al the plugins and presets lined up in a nice column. yay alignment!

demo/package.json
```diff
@@ -24,6 +27,10 @@
   "dependencies": {
     "d3-scale": "^1.0.7",
     "d3-selection": "^1.2.0",
-    "preact-render-to-string": "^4.1.0"
+    "mobx": "^5.9.0",
+    "mobx-preact": "^3.0.0",
+    "mobx-state-tree": "^3.12.2",
+    "preact-render-to-string": "^4.1.0",
```
marvinhagemeister: We have a `next` tag for `preact-render-to-string` which points to 5.x and supports ceviche :+1:

demo/index.js
```diff
@@ -21,6 +22,9 @@ if (!isBenchmark) {
 	initDevTools();
 }
 
+// mobx-state-tree fix
+window.setImmediate = setTimeout;
```
marvinhagemeister: @mweststrate Are there any particular reasons to use `setImmediate` over `setTimeout`?

developit: Also interested to hear why this was used. Looks like it's here:
https://github.com/mobxjs/mobx-state-tree/blob/d3e24ab5f89e0e197e2b8594a6de774f3e739351/packages/mobx-state-tree/src/core/flow.ts#L117

marvinhagemeister:
FYI: Had to upgrade `webpack` and remove the `babel-pgramatic-jsx` plugin to get it to work for me.

PR #1455, opened 2019-03-25
Fix componentDidUpdate(prevProps, prevState, snapshot)

38elements:
If `getSnapshotBeforeUpdate()` is not implemented, `snapshot` is `undefined`.

https://reactjs.org/docs/react-component.html#componentdidupdate

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -33,7 +33,7 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
 
 	if (options.diff) options.diff(newVNode);
 
-	let c, p, isNew = false, oldProps, oldState, oldContext,
+	let c, p, isNew = false, oldProps, oldState, snapshot,
 		newType = newVNode.type;
 
 	/** @type {import('../internal').Component | null} */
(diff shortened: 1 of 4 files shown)


andrewiggins:
@marvinhagemeister Note, this is a new breaking change we should call out in our release notes. Anyone who was depending on the Preact specific behavior of receiving the context as the 3rd argument to `cDU` will now be broken.

marvinhagemeister:
@andrewiggins you're right, didn't saw that. We even had it in the `8.x` release line. Should we revert the merge?

38elements:
Since this change is breaking change, I think it is necessary that this is described in release note.
Since there is not the description of the third argument in Preact document, I think this behavior is not a feature.
Therefore, I think that it is better not to revert. 
https://preactjs.com/guide/differences-to-react

38elements:
@andrewiggins @marvinhagemeister 
IMHO
Since this change is breaking change, I think it is necessary that this is described in release note.
Since there is not the description of the third argument in Preact document, I think the behavior is not a feature.
Therefore, I think that it is better not to revert.
https://preactjs.com/guide/differences-to-react

PR #1458, opened 2019-03-26
(chore) - add teardown to test-utils

JoviDeCroock:
This offers a full testability from the `test-utils` package. This helps document an easy way to test preact components.

In a later PR, I'll refactor the existing tests to work with these methods.

### DIFF
--- a/test-utils/src/index.js
+++ b/test-utils/src/index.js
@@ -28,30 +28,13 @@ export function act(cb) {
 	if (flush) {
 		flush();
 	}
-	options.debounceRendering = Component.__test__previousDebounce;
 	options.requestAnimationFrame = previousRequestAnimationFrame;
 }
 
 /**
- * Setup the test environment
- * @returns {HTMLDivElement}
- */
-export function setupScratch() {
-	const scratch = document.createElement('div');
-	scratch.id = 'scratch';
-	(document.body || document.documentElement).appendChild(scratch);
-	return scratch;
-}
-
-/**
  * Teardown test environment and reset preact's internal state
- * @param {HTMLDivElement} scratch
  */
-export function teardown(scratch) {
-	if (scratch) {
-		scratch.parentNode.removeChild(scratch);
-	}
-
+export function teardown() {
 	if (Component.__test__drainQueue) {
 		// Flush any pending updates leftover by test
 		Component.__test__drainQueue();
--- a/test-utils/src/index.js
+++ b/test-utils/src/index.js
@@ -52,11 +52,6 @@ export function teardown(scratch) {
 		scratch.parentNode.removeChild(scratch);
 	}
 
-	if (oldOptions != null) {
-		assign(options, oldOptions);
-		oldOptions = null;
-	}
-
 	if (Component.__test__drainQueue) {
 		// Flush any pending updates leftover by test
 		Component.__test__drainQueue();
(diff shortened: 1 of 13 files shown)


JoviDeCroock:
I really do wonder why this decreases coverage in hooks :/

JoviDeCroock:
I really think coverall hates me

test-utils/src/index.js
```diff
@@ -31,3 +31,35 @@ export function act(cb) {
 	options.debounceRendering = Component.__test__previousDebounce;
 	options.requestAnimationFrame = previousRequestAnimationFrame;
 }
+
+/**
+ * Setup the test environment
+ * @returns {HTMLDivElement}
+ */
+export function setupScratch() {
```
andrewiggins: Hmm I think I'd prefer we don't export a `setupScratch` helper in test-utils. I don't think it adds much value specifically for Preact testing, of which there are many approaches only some of which include adding a DOM. While it is very useful for how we do testing internally in the Preact repo, it isn't necessarily useful for users of Preact.

In other words, I don't think this function adds something unique about testing in Preact to justify its maintenance as a public api.

(Note: Personally, I like to keep exposed APIs to a minimum to reduce long term baggage, but perhaps I'm overreacting here 😅)

test-utils/src/index.js
```diff
@@ -31,3 +31,35 @@ export function act(cb) {
 	options.debounceRendering = Component.__test__previousDebounce;
```
andrewiggins: Just a random thought, not something to necessarily address in this PR: How do `setupRerender()` and `act()` interact? If I have a test that calls `setupRerender` in a `beforeEach` and then a test that uses `act`, do they behave well together?

andrewiggins:
Hahah yea I don't know what's up with coveralls lol

test-utils/src/index.js
JoviDeCroock: Yes, I don't think this still belongs here.

test-utils/src/index.js
JoviDeCroock: Well in all honesty, I don't really mind it was more or less to give a clearcut way for people to test basic stuff.
That's also one thing we can leave out of teardown then

test-utils/src/index.js
```diff
@@ -28,6 +28,21 @@ export function act(cb) {
...
 	}
-	options.debounceRendering = Component.__test__previousDebounce;
 	options.requestAnimationFrame = previousRequestAnimationFrame;
 }
+
+/**
+ * Teardown test environment and reset preact's internal state
+ */
+export function teardown() {
+	if (Component.__test__drainQueue) {
```
developit: I believe this one is on me - do you think we could store this property elsewhere?  It could be changed to `options.__test__drainQueue`, or potentially just be a module-level variable.

test-utils/src/index.js
```diff
@@ -28,6 +28,21 @@ export function act(cb) {
...
+		// Flush any pending updates leftover by test
+		Component.__test__drainQueue();
+		delete Component.__test__drainQueue;
+	}
+
+	if (typeof Component.__test__previousDebounce !== 'undefined') {
+		options.debounceRendering = Component.__test__previousDebounce;
```
developit: Same deal for `Component.__test__previousDebounce` - seems like these belong on `options`, the Component thing was inherited from when `options.debounceRendering` got renamed to `Component.debounce` in ceviche.

JoviDeCroock:
I added it to options now, if you want it to be a module level variable let me know

marvinhagemeister:
Let's do it :+1: :tada:

PR #1464, opened 2019-03-27
Fix: state is not updated when called setState in componentWillMount 

BengBu-YueZhang:
https://github.com/developit/preact/issues/1461


JoviDeCroock:
Hey, looks good. Mind adding a test for your scenario?

BengBu-YueZhang:
@JoviDeCroock Thank you, I added unit tests

developit:
Note: #1465 might make this work in a more generic way.

BengBu-YueZhang:
> Note: #1465 might make this work in a more generic way.

Does this pr need close?

PR #1465, opened 2019-03-27
refactor: ward off evil `s`

### LINKED ISSUE #1461: [10.0.0-alpha.2] - setState() doesn't apply state when called within componentWillMount
In 10.0.0-alpha.2, state does not set properly from `componentWillMount`. It seems this has been [experienced by others in the past](https://github.com/developit/preact/issues/1368) in `componentWillReceiveProps`. Here's a simple test case:

```javascript
export class TestComponent extends Component {
    
    //This does not work:
    componentWillMount() {
        this.setState({message:"componentWillMount"});
    }

    /*
    //This works:
    componentDidMount() {
        this.setState({message:"componentDidMount"});
    }
    */
    
    render() {
        return <p>State came through from {this.state.message || 'nothing =('}</p>;
    }
}
```

mochiya98:
booyah!!  :sparkles:  
fixes #1461


test/browser/lifecycle.test.js
```diff
@@ -901,6 +901,33 @@ describe('Lifecycle methods', () => {
...
+					this.setState({ value: 1 });
+				}
+				render() {
+					componentState = this.state;
+					return <div />;
+				}
+			}
+
+			render(<Foo />, scratch);
+			rerender();
```
JoviDeCroock: This `rerender` should not be needed

src/diff/index.js
```diff
@@ -87,11 +87,11 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
...
 			// Invoke getDerivedStateFromProps
-			let s = c._nextState || c.state;
+			if (c._nextState==null) {
+				c._nextState = c.state;
+			}
 			if (newType.getDerivedStateFromProps!=null) {
-				oldState = assign({}, c.state);
-				if (s===c.state) s = c._nextState = assign({}, s);
-				assign(s, newType.getDerivedStateFromProps(newVNode.props, s));
+				c.setState(newType.getDerivedStateFromProps(newVNode.props, c._nextState));
```
JoviDeCroock: doesn't this enqueue a double render or is that safeguarded somehow

mochiya98:
Thank you for the review, I will reconsider. 🚀

developit:
@mochiya98 this looks promising!  Should be able to swap out the `setState()` call for just setting `.state=`.

src/diff/index.js
mochiya98: I checked it. maybe no problem.

test/browser/lifecycle.test.js
mochiya98: thanks! I removed that.

mochiya98:
I fixed a mistake.
@developit thanks! I'm using `setState()` to process `_nextState`.

src/diff/index.js
developit: It's likely inert because `component._dirty` is `true`. I'm still a little wary of using setState() here though, it's not an API that was supposed to be used by the diff.

@mochiya98 - any chance you could explain why this works better than setting `c._nextState` directly? Sorry for asking!

mochiya98: that is to reduce the bundle size.
I thought it is okay because `setState()` is working when using in lifecycle.
but dependencies will be complicated that i see.

marvinhagemeister: Calling `setState` here scares me a bit to be honest. This would introduce a hard dependency in our diff code with the way components are scheduled to render.

mochiya98:
```jsx
assign(c._nextState===c.state ? (c._nextState = assign({}, c._nextState)) : c._nextState, newType.getDerivedStateFromProps(newVNode.props, s));
```

`setState`: 3440(gz) / 3137(br)
`in diff`: 3462(+22) / 3156(+19)

choose size or dependency...
do you think it is better than?

mochiya98: I took that into consideration and decided to write directly.

src/diff/index.js
```diff
@@ -87,11 +87,11 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
...
 			// Invoke getDerivedStateFromProps
-			let s = c._nextState || c.state;
+			if (c._nextState==null) {
+				c._nextState = c.state;
+			}
 			if (newType.getDerivedStateFromProps!=null) {
-				oldState = assign({}, c.state);
-				if (s===c.state) s = c._nextState = assign({}, s);
-				assign(s, newType.getDerivedStateFromProps(newVNode.props, s));
+				assign(c._nextState==c.state ? (c._nextState = assign({}, c._nextState)) : c._nextState, newType.getDerivedStateFromProps(newVNode.props, c._nextState));
```
JoviDeCroock: The true clause of your condition, can't this just be c._nextState since `assign({}, c._nextState)` is equal to returning c._nextState, or am I misunderstanding something here?

marvinhagemeister: The copy is necessary to make sure that the new result from `gDSFP` is not copied onto `c.state`. Without this both `c.state` and `c._nextState` would be referentially equal.

JoviDeCroock: Aha, thanks for the explanation. Nevermind my comment ^^

mochiya98:
Any updates on this?

mochiya98:
Is there any problem? I can help if you need.

mochiya98:
Great work, thanks for all reviewers! 👍

PR #1468, opened 2019-03-27
(refactor) move context update logic to diffing

### LINKED ISSUE #1363: 10.0.0-alpha.1 - Children from context provider are rendered twice after state update

JoviDeCroock:
I've made a first draft since it seems to solve the issue adressed by @marvinhagemeister with useContext.
Added the test in his branch here, and it works.

Will now look at how we can implement the others.

Size change: ~~+26B~~ ~~+17B~~ ~~-3B~~ -10B
Fixes #1363


src/diff/index.js
```diff
@@ -110,6 +110,11 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
 					break outer;
 				}
 
+				if (isProvider && c.props.value!==newVNode.props.value) {
```
andrewiggins: Is this basically the same as changing the original `componentDidUpdate` on Provider to `componentWillUpdate`?

JoviDeCroock: Hmm, yes that's a good point. Not fully sure about this approach yet personally but shared it already since it seems to solve an issue that is open at this point.

JoviDeCroock:
Added some tests to see if this would introduce bugs in Context.Consumer but those come out positive aswell.

src/diff/index.js
```diff
@@ -110,6 +109,11 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
 					break outer;
 				}
 
+				if (newType._id==='_P' && c.props.value!==newVNode.props.value) {
+					const v = newVNode.props.value;
+					newVNode._component.subs.map(sub => v!==sub.context && (sub.context = v, enqueueRender(sub)));
```
JoviDeCroock: Something I was wondering, can these two ever be equal? I don't really think so.

JoviDeCroock: I removed it and nothing in test results changed. 

We are comparing v which is a value to a Context Object so it always evaluates to true. This saves 9Bytes

JoviDeCroock:
Another approach could be to instead of use componentDidUpdate to use shouldComponentUpdate to notify subs, how are you standing on this?

marvinhagemeister:
@JoviDeCroock can you check your IDE settings? There are lots of mixed spaces and tabs as indentation in the same file :slightly_smiling_face:

src/create-context.js
```diff
@@ -23,14 +22,21 @@ export function createContext(defaultValue) {
...
 		};
-		comp.componentDidUpdate = () => {
-			let v = comp.props.value;
-			subs.map(c => v!==c.context && (c.context = v, enqueueRender(c)));
+		comp.shouldComponentUpdate = props => {
+			if (props.value!==comp.value) {
+				subs.map(c => {
+					// Check if still mounted
+					if (c._parentDom) {
+						c.context = props.value;
```
marvinhagemeister: Removing both the `if`-statement and the `c.context` assignment has no effect on the test outcomes. Maybe we're missing something?

src/diff/index.js
```diff
@@ -55,7 +55,6 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
 			let cxType = newType.contextType;
 			let provider = cxType && context[cxType._id];
 			let cctx = cxType != null ? (provider ? provider.props.value : cxType._defaultValue) : context;
-
```
marvinhagemeister: Nit: Change is not needed for this PR :slightly_smiling_face:

src/create-context.js
```diff
@@ -1,5 +1,4 @@
 import { enqueueRender } from './component';
-
```
marvinhagemeister: Nit: It may be just me, but I feel like having at least one empty line after imports is easier to scan.

src/create-context.js
JoviDeCroock: Parent -> context -> noUpdate-> consumer could have skewed results without it, no?

JoviDeCroock: I tested this and well it does not fail.... No tests fails + my additional test for that scenario also succeeds on top that saves an additional 7B

PR #1471, opened 2019-03-28
Fix DOM removed before `componentWillUnmount`

### LINKED ISSUE #732: ComponentWillUnmount throwing error, when refering to a DOM node

marvinhagemeister:
Fixes #732 .
Adds `+3 B`


marvinhagemeister:
@robertknight Thanks for your review :tada: The behavior is the same as before: Only the parent dom node will be removed from the document. This works because `skipRemove` ensures that the `dom` variable is only ever non-null for the root of the tree that should be dropped.

PR #1474, opened 2019-03-29
Remove class attribute when it is null or false

shirotech:
```jsx
<div class={null}>text</div>
```

will render

```html
<div class>text</div>
```

This patch will remove it so it becomes:

```html
<div>text</div>
```


andrewiggins:
Thanks for the PR! I'm gonna go ahead and close this PR since it is open against Preact 8. If you are still interested in this feature, please re-open it against master and we'll discuss further. Cheers!

PR #1478, opened 2019-03-29
Add test for fragment ordering

JoviDeCroock:


test/browser/fragments.test.js
```diff
@@ -88,6 +88,38 @@ describe('Fragment', () => {
...
+					<div>
+						<X key={2}>2</X>
+						<X key={1}>1</X>
+					</div>
+				);
+			}
+		}
+
+		render(<App i={0} />, scratch);
+		expect(scratch.firstChild.innerHTML).to.equal('12');
```
marvinhagemeister: We could also use `scratch.textContent` :upside_down_face:

test/browser/fragments.test.js
```diff
@@ -88,6 +88,38 @@ describe('Fragment', () => {
...
+		render(<App i={1} />, scratch);
+		expect(scratch.firstChild.innerHTML).to.equal('21');
+	})
```
marvinhagemeister: Travis fails because of a missing `;` here

PR #1484, opened 2019-03-30
Fix `_dom` being `null` whenever sCU returns `false`

### LINKED ISSUE #1475: Detail view not unmounted correctly in People demo
I think the newly added demo is not working correctly, the remove contact button doesn't remove the contact profile!?

marvinhagemeister:
The issue is caused by a mismatch of `component.base` with `component._vnode._dom`. We use the latter for diffing and it lead to leaving old DOM nodes around whenever `shouldComponentUpdate` returns `false`.

I can't take credit alone for it, because the fix wouldn't be possible without all the debugging help from @JoviDeCroock, @developit and @andrewiggins :+1: :100: 

Adds ~~`+3 B`~~ ~~`+11 B`~~ ~~`+17 B`~~ `+13 B` :tada:

### DIFF
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -93,7 +93,9 @@ export function coerceToVNode(possibleVNode) {
 
 	// Clone vnode if it has already been used. ceviche/#57
 	if (possibleVNode._dom!=null) {
-		return createVNode(possibleVNode.type, possibleVNode.props, possibleVNode.text, possibleVNode.key, null);
+		let vnode = createVNode(possibleVNode.type, possibleVNode.props, possibleVNode.text, possibleVNode.key, null);
+		vnode._dom = possibleVNode._dom;
+		return vnode;
 	}
 
 	return possibleVNode;
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -108,6 +108,7 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
 				}
 
 				if (!force && c.shouldComponentUpdate!=null && c.shouldComponentUpdate(newVNode.props, s, cctx)===false) {
+					dom = oldVNode._dom;
 					c.props = newVNode.props;
 					c.state = s;
 					c._dirty = false;
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -64,6 +64,7 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
 			if (oldVNode._component) {
 				c = newVNode._component = oldVNode._component;
 				clearProcessingException = c._processingException;
+				newVNode._dom = oldVNode._dom;
 			}
 			else {
 				// Instantiate the new component
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -108,7 +108,7 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
 				}
 
 				if (!force && c.shouldComponentUpdate!=null && c.shouldComponentUpdate(newVNode.props, s, cctx)===false) {
-					dom = oldVNode._dom;
+					dom = newVNode._dom;
 					c.props = newVNode.props;
 					c.state = s;
 					c._dirty = false;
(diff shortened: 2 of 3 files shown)


Almo7aya:
I was about to commit this 👍 🎉

marvinhagemeister:
Shoot, just discovered a second case in the demo where a similar issue pops up. I think they're related.

Almo7aya:
The issue still exists, Yes it removes the person's details view in people demo but only ones. And also a new empty div get added every time I click a person!

marvinhagemeister:
@Almao7aya yeah maybe there are multiple issues at play. Marked the PR as WIP for now

Almo7aya:
@marvinhagemeister Finally 🎉. I found it, the lose of `_dom` was in `diffElementNodes` function, it works perfectly now even without these changes in this PR, Should I open a new PR?

marvinhagemeister:
@Almo7aya go for it :+1:

marvinhagemeister:
Spent a few hours today debugging this. The issue is caused by 2 separate bugs. One is the one this PR addresses already (although it needs to be `newVNode._dom` instead of `oldVNode._dom` and the other issue is caused by a double coercion of `vnodes` in `diffChildren`.

When the second render comes through `coerceVNode` will see that there is already a non-null `_dom` value and will clone the `vnode`. At this point we're losing the reference to the dom. Because the `newVNode._dom===null` our reconciler correctly concludes that nothing needs to be done, not knowing that the DOM doesn't match the vnode tree anymore.

@developit stumbled upon the same issue in #1361

Almo7aya:
We can restore the lost `_dom` in `diffElementNodes` function we can add this `!dom && oldVNode && (dom = oldVNode._dom);` before `if (dom==null) {`

marvinhagemeister:
@Almo7aya That's great for `dom` nodes :tada: I found that the same issue exists with Components, especially when there are multiple renders and `shouldComponentUpdate -> false` working together.

The shortest fix I could come up with was to make sure that we don't get into the situation where we'd lose the reference to `_dom`. The only to places where I found that this could happen was in `coerceToVNode` and when `shouldComponentUpdate` returns `false` :100:

Need to come up with better tests though. The one that's in this PR doesn't suffice and pushes the code through all scenarios like the demo app does...

marvinhagemeister:
@andrewiggins I still need to add more test cases to this PR, but I was wondering if you can have a look at this. I vaguely remember you having dealt with `vnode` coercion during the ceviche times.

Almo7aya:
@marvinhagemeister The people demo still falling, Adding my change in `diffElementNodes` fixes it, Is there another lose 🤷‍♂️?!

marvinhagemeister:
@Almo7aya Just checked, the people demo works fine with this PR. I can't find further unmounting issues or any problems with stale DOM nodes. Did you restart webpack? Sometimes the hot-reloading caches to aggressively when switching branches.

Almo7aya:
@marvinhagemeister It shows and removes a person correctly, But when clicking on many people in the nav when the person view is viewing, a new empty `div` get created. Maybe we lost the `_dom` again🤦‍♂️!

marvinhagemeister:
@Almo7aya Just pushed a fix. Fingers crossed that this was the last one :crossed_fingers:

src/create-element.js
```diff
@@ -93,7 +93,9 @@ export function coerceToVNode(possibleVNode) {
 
 	// Clone vnode if it has already been used. ceviche/#57
 	if (possibleVNode._dom!=null) {
-		return createVNode(possibleVNode.type, possibleVNode.props, possibleVNode.text, possibleVNode.key, null);
+		let vnode = createVNode(possibleVNode.type, possibleVNode.props, possibleVNode.text, possibleVNode.key, null);
```
developit: Can we add `_dom` as an argument to createVNode here? I think it might be shorter (could totally be wrong!)

marvinhagemeister: Tried that and adding the argument to `createVNode` makes the PR bigger by `+3 B` :S

PR #1486, opened 2019-03-30
fix: support IE9

JoviDeCroock:
Recovers support for IE9 and maintains supporting custom variables.

Support for `!important` still does not work but I wonder if this is even needed, it does however if people use it disable the full propagation of the prop.

`style={{ background: 'blue !important' }}` --> evaluates to `style={{}}`. This is pretty akward for devs relying on it. If we would want to support !important we would have to use `setProperty` and neglect the bug with IE9. setProperty can set !important with its thirth argument.

I do feel like that maybe we support the majority of IE9 and then we're fine but this could also open up a rabbit hole against future features. (this is when we'd officially declare IE9 support)

ADDS: ~~12~~ ~~10B~~ 9B

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -59,7 +59,12 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 			for (let i in value) {
 				v = value[i];
 				if (oldValue==null || v!==oldValue[i]) {
-					s.setProperty(i.replace(CAMEL_REG, '-'), typeof v==='number' && IS_NON_DIMENSIONAL.test(i)===false ? (v + 'px') : v);
+					if (i.indexOf('--')===0) {
+						s.setProperty(i.replace(CAMEL_REG, '-'), v);
+					}
+					else {
+						s[i] = typeof v==='number' && IS_NON_DIMENSIONAL.test(i)===false ? (v + 'px') : v;
+					}
 				}
 			}
 		}
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -59,7 +59,7 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 			for (let i in value) {
 				v = value[i];
 				if (oldValue==null || v!==oldValue[i]) {
-					if (!i.indexOf('--')) {
+					if (i[0]=='-') {
 						s.setProperty(i.replace(CAMEL_REG, '-'), v);
 					}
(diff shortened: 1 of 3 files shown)


src/diff/props.js
```diff
@@ -59,7 +59,12 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 			for (let i in value) {
 				v = value[i];
 				if (oldValue==null || v!==oldValue[i]) {
-					s.setProperty(i.replace(CAMEL_REG, '-'), typeof v==='number' && IS_NON_DIMENSIONAL.test(i)===false ? (v + 'px') : v);
+					if (i.indexOf('--')===0) {
```
marvinhagemeister: Just using `!i.indexOf('--')` saves 2 bytes :tada:

marvinhagemeister:
I'd be happy to add it just for supporting `!important` alone :tada: Thanks for diving in to this :100:

JoviDeCroock:
I don't think we can support !import even with this, I still have to explore that part.

marvinhagemeister:
Oh ok, if that's the case I'd vote against adding IE9 specific code.

JoviDeCroock:
So rather have it catch and replace important to the thirth argument?

src/diff/props.js
```diff
@@ -59,7 +59,12 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 			for (let i in value) {
 				v = value[i];
 				if (oldValue==null || v!==oldValue[i]) {
-					s.setProperty(i.replace(CAMEL_REG, '-'), typeof v==='number' && IS_NON_DIMENSIONAL.test(i)===false ? (v + 'px') : v);
+					if (!i.indexOf('--')) {
```
developit: I believe this would be smaller and faster:

```suggestion
					if (i[0]=='-') {
```

cloudkite:
Hey 👋Just chiming in as I feel like my comment in https://github.com/developit/preact/issues/1482 may have been misleading. I was under the impression that React supported !important in inline styles, was curious so I tried it today in a React project and it does not work.

JoviDeCroock:
Closing this since the decision made involves:
- style[key] = value is slow, this punishes all browsers except IE9
- !important can be supported through style="color: red !important"

developit:
Bench from the above:

PR #1487, opened 2019-03-31
(feat) - debug render

JoviDeCroock:
I'm personally quite conflicted at this point... It solves a LOT of developer experience bugs but the 27 added bytes are a lot. We could probably get this down to 15-20 by golfing with it a bit but still...

However it does provide a fallback to several bug reports we're getting.

Example: https://github.com/developit/preact/issues/1485

ADDS: ~~27B~~ ~~24B~~ 11B

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -6,7 +6,7 @@ export function initDebug() {
 	/* eslint-disable no-console */
 	let oldBeforeDiff = options.diff;
 
-	options.checkRenderParent = parentNode => {
+	options.checkParent = parentNode => {
 		if (!parentNode) {
 			throw new Error('Undefined parent passed to render(), this is the second argument.\nCheck if the element is available in the DOM/has the correct id.');
 		}
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -18,10 +18,9 @@ export function initDebug() {
 			case DOCUMENT_NODE: isValid = true; break;
 			default: isValid = false;
 		}
-		const value = typeof parentNode === 'object' ? JSON.stringify(parentNode) : parentNode;
 		if (!isValid) throw new Error(`
 			Expected a valid HTML node as a second argument to render.
-			Received ${value} instead: render(<${vnode.type.name || vnode.type} />, ${value});
+			Received ${parentNode} instead: render(<${vnode.type.name || vnode.type} />, ${parentNode});
 		`);
 		return parentNode._prevVNode
 	}
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -8,8 +8,18 @@ export function initDebug() {
 
 	options.checkRenderParent = parentNode => {
 		if (!parentNode) {
-			throw new Error('Undefined parent passed to render(), this is the second argument.\nCheck if the element is available in the DOM/has the correct id.')
+			throw new Error('Undefined parent passed to render(), this is the second argument.\nCheck if the element is available in the DOM/has the correct id.');
 		}
+		let isValid;
+		switch (parentNode.nodeType) {
+			case 1:
+			case 3:
+			case 8:
+			case 9:
+			case 11: isValid = true; break;
+			default: isValid = false;
+		}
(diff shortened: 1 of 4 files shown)


debug/src/debug.js
```diff
@@ -6,6 +6,23 @@ export function initDebug() {
 	/* eslint-disable no-console */
 	let oldBeforeDiff = options.diff;
 
+	options.checkParent = parentNode => {
```
marvinhagemeister: Renaming `checkParent` to `root` saves 3 bytes because the latter is already in our gzip dictionary :tada:

src/render.js
```diff
@@ -10,7 +11,7 @@ import { createElement, Fragment } from './create-element';
  * render into
  */
 export function render(vnode, parentDom) {
-	let oldVNode = parentDom._prevVNode;
+	let oldVNode = options.checkParent ? options.checkParent(parentDom) : parentDom._prevVNode;
```
marvinhagemeister: I think for now we should just add the hook without using it's return value. I can't imagine a scenario where it would make sense to return a completely different dom node.

```js
if (options.root) options.root(parentDom);
let oldVNode = parentDom._prevVNode;
```

marvinhagemeister:
With the suggestions we can bring it down to just `+11 B` :+1:

debug/src/debug.js
```diff
@@ -6,6 +6,23 @@ export function initDebug() {
...
+		if (!parentNode) {
+			throw new Error('Undefined parent passed to render(), this is the second argument.\nCheck if the element is available in the DOM/has the correct id.');
+		}
+		let isValid;
+		switch (parentNode.nodeType) {
+			case 1:
```
marvinhagemeister: Maybe we can add constants here. I never can remember which number stands for which `nodeType` :slightly_smiling_face:

debug/src/debug.js
```diff
@@ -6,6 +6,23 @@ export function initDebug() {
...
+			case 3:
+			case 8:
+			case 9:
+			case 11: isValid = true; break;
+			default: isValid = false;
+		}
+		if (!isValid) throw new Error('Expected a valid HTML node as a second argument to render.');
```
marvinhagemeister: I could imagine new users being confused about which `render` the message is referring to. Perhaps just amending a second line where we print the component name would help out.

```js
`Expected a valid HTML node as a second argument to render. Received ${parentDom} instead: render(<MyComponent />, ${parentDom});`
```

Not sure if that is easily doable.

debug/src/debug.js
```diff
@@ -1,11 +1,29 @@
...
+		}
+		let isValid;
+		switch (parentNode.nodeType) {
+			case ELEMENT_NODE:
+			case DOCUMENT_NODE: isValid = true; break;
+			default: isValid = false;
+		}
+		if (!isValid) throw new Error(`
+			Expected a valid HTML node as a second argument to render.
+			Received VALUE instead: render(<${vnode.type.name || vnode.type} />, VALUE);
```
marvinhagemeister: We could replace `VALUE` with the actual value of the passed `parentNode` and cast it to a string :+1:

JoviDeCroock: Like this typeof parentNode === 'object' ? JSON.stringify(parentNode) : parentNode?

marvinhagemeister: Just inserting `parentNode` should be fine as it would be automatically cast to a string. It may even be advantageous for us if it's just `[object Object]` as that looks really wrong and like something must be up :D

JoviDeCroock: Allright, fixed it :)

marvinhagemeister:
This is so dope :+1: :tada:

ForsakenHarmony:
Seems like this is missing the typescript definition?

JoviDeCroock:
@ForsakenHarmony, do you mean for options?

ForsakenHarmony:
yup

JoviDeCroock:
Thanks for the heads up, fixed it :)

PR #1488, opened 2019-03-31
Fix remove button adding todos in demo app

marvinhagemeister:
This PR fixes an issue where the `x` buttons in the todo demo would actually add a todo instead of removing it. Note that there is a second issue at play that needs to fixed separately which causes stale dom nodes when wrapped in a `Fragment`.

The issue is caused by weird browser behavior which treat any `<buttton>` that doesn't have a `type` attribute as a `submit` button if it is inside a `<form>` :man_shrugging:

In our case it would immediately submit the form which always adds a todo.
```jsx
<form onSubmit={this.addTodo} action="javascript:">...</form>
```

### DIFF
--- a/demo/todo.js
+++ b/demo/todo.js
@@ -38,7 +38,7 @@ class TodoItems extends Component {
 	render({ todos, removeTodo }) {
 		return todos.map( todo => (
 			<li key={todo.id}>
-				<button onClick={removeTodo} data-id={todo.id}>&times;</button>
+				<button type="button" onClick={removeTodo} data-id={todo.id}>&times;</button>
 				{' '}
 				{todo.text}
 			</li>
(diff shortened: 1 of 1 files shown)


demo/todo.js
```diff
@@ -38,7 +38,7 @@ class TodoItems extends Component {
 	render({ todos, removeTodo }) {
 		return todos.map( todo => (
 			<li key={todo.id}>
-				<button onClick={removeTodo} data-id={todo.id}>&times;</button>
+				<button type="button" onClick={removeTodo} data-id={todo.id}>&times;</button>
```
JoviDeCroock: Oh I actually thought this only happened inside of `<form>`

marvinhagemeister: You're right it does! The `TodoList` component adds a `<form>` element around the items https://github.com/developit/preact/blob/efb56776618d7fa62929776421c68dfd6681a54d/demo/todo.js#L25-L31

PR #1489, opened 2019-03-31
Fix stale DOM caused by empty Fragments

marvinhagemeister:
This PR fixes an issue where the DOM for `Fragments` wasn't cleared when they previously had children. This was most noticeable in the `todo` demo where the first added todo item could never be removed.

Adds `+2 B` :tada:


JoviDeCroock:
This probably fixes: https://github.com/developit/preact/issues/1475?

marvinhagemeister:
Only partially, in that issue there seem to be more problems causing havok.

PR #1494, opened 2019-04-01
Add more createElement warnings in debug

marvinhagemeister:
- Add new warnings for various invalid input passed to `createElement`.
- Enable debug messages in demo app.

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -29,7 +29,17 @@ export function initDebug() {
 		let children = props && props.children;
 
 		if (type===undefined) {
-			throw new Error('Undefined component passed to createElement()\n'+serializeVNode(vnode));
+			throw new Error('Undefined component passed to createElement()\n\n'+
+			'You likely forgot to export your component or might have mixed up default and named imports'+
+			serializeVNode(vnode));
+		}
+		else if (type!=null && typeof type==='object') {
+			if (type._lastDomChild!==undefined && type._dom!==undefined) {
+				let info = 'Did you accidentilly passed a JSX Literal as JSX twice?';
+				throw new Error('Invalid type passed to createElement(): '+type+'\n\n'+info+'\n\n'+serializeVNode(type));
+			}
+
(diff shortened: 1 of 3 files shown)


debug/src/debug.js
```diff
@@ -29,7 +29,17 @@ export function initDebug() {
...
 
 		if (type===undefined) {
-			throw new Error('Undefined component passed to createElement()\n'+serializeVNode(vnode));
+			throw new Error('Undefined component passed to createElement()\n\n'+
+			'You likely forgot to export your component or might have mixed up default and named imports'+
+			serializeVNode(vnode));
+		}
+		else if (type!=null && typeof type==='object') {
+			if (type._lastDomChild!==undefined && type._dom!==undefined) {
+				let info = 'Did you accidentilly passed a JSX Literal as JSX twice?';
```
JoviDeCroock: I think this should be `accidentally`

JoviDeCroock: I also think we would need an example here like <<App /> /> to be really verbose

marvinhagemeister: Something like `<<Foo />/>`?

JoviDeCroock: Yes, this explanation won't hit on the people who make this mistake which are often really new to jsx.

ForsakenHarmony:
```suggestion
				let info = 'Did you accidentally pass a JSX literal as JSX twice? (i.e. `let Bar = <Foo />; return <Bar/>`)';
```

marvinhagemeister: @ForsakenHarmony awesome, the idea to put the example in writing is really great! Changed the formatting a bit:

```
main.js:669 Uncaught Error: Invalid type passed to createElement(): [object Object]

Did you accidentally pass a JSX literal as JSX twice?

  let MyHome = <Home />;
  let vnode = <MyHome />;

This usually happens when you export a JSX literal and not the component.

[stack trace]
```

PR #1495, opened 2019-04-01
(feat) - debugging effect hooks

JoviDeCroock:
This debugs the passing of an arguments array to hooks that need it. This avoids nasty endless calling bugs. This throws an error to breaking that endless calling.

Or is this actually a pattern that people want calling an effect on every render, personally it always bites me...

This could potentially be extended to useMemo and useCallback, warning about not providing an arguments array makes it a useless function call.

Warn about hooks in classes, I don't know if this is possible?

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -6,6 +6,7 @@ import { ELEMENT_NODE, DOCUMENT_NODE, DOCUMENT_FRAGMENT_NODE } from './constants
 export function initDebug() {
 	/* eslint-disable no-console */
 	let oldBeforeDiff = options.diff;
+	let oldDiffed = options.diffed;
 
 	options.root = (vnode, parentNode) => {
 		if (!parentNode) {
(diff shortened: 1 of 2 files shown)


debug/src/debug.js
```diff
@@ -83,6 +84,32 @@ export function initDebug() {
 
 		if (oldBeforeDiff) oldBeforeDiff(vnode);
 	};
+
+	options.diffed = (vnode) => {
+		if (vnode._component && vnode._component.__hooks && vnode._component.__hooks._pendingEffects.length > 0) {
+			vnode._component.__hooks._pendingEffects.forEach((x) => {
```
marvinhagemeister: Maybe we can simplify this check and combine it with the one a few lines down:

```js
if (vnode._component && vnode._component._hooks) {
  let hooks = vnode._component._hooks;

  if (hooks._pendingEffects.length > 0) {
    // ...
  }
  else if (hooks._pendingLayoutEffects.length > 0) {
    // ...
  }
}

```

debug/src/debug.js
```diff
@@ -83,6 +84,32 @@ export function initDebug() {
...
+				if (!x._args || !Array.isArray(x._args)) {
+					throw new Error(`
+						You should provide an array of arguments as the second argument to the "useEffect" hook.
```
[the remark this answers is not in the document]
JoviDeCroock: Haha, Oops... I like to write it like that... I'll fix my other ones aswell then.

debug/src/debug.js
marvinhagemeister: That's a good question. Personally I'd add it. If there really is a valid reason to trigger `useEffect` on each render one could just pass an always changing dependency:

```jsx
let i = 0;
useEffect(() => x, [i++]);
```

JoviDeCroock:
I wonder if it’d be worth giving useMemo and useCb an ‘id’ prop for debugging purposes

marvinhagemeister:
@JoviDeCroock In the long term we'd need a way to tag hooks either way for the devtools integration. We've been pondering how to do that for a while in slack but haven't moved forward with it yet. So a big +1 from me on tagging hooks :+1:

PR #1498, opened 2019-04-01
(fix) - event.persist should be a function, not an object

JoviDeCroock:
According to: https://reactjs.org/docs/events.html#event-pooling
e.persist should be a function and not an object.

Fixes: https://github.com/developit/preact/issues/1497


jessicabyrne:
Awesome thanks for this fix! Any idea when it might be available in a release?

PR #1499, opened 2019-04-02
Chore/dx use cb use memo

JoviDeCroock:
Debugging useMemo and useCB. Personally I would like to add an id of the hook in question to the _hooks array.

Also improved some naming, for future edits.

### DIFF
--- a/debug/test/browser/debug.test.js
+++ b/debug/test/browser/debug.test.js
@@ -126,7 +126,7 @@ describe('debug', () => {
 
 	it('should warn for useless useMemo calls', () => {
 		const App = () => {
-			const [people] = useState([40, 20,60, 80]);
+			const [people] = useState([40, 20, 60, 80]);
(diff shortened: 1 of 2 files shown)


debug/test/browser/debug.test.js
```diff
@@ -120,6 +124,17 @@ describe('debug', () => {
 		expect(fn).to.throw(/createElement/);
 	});
 
+	it('should warn for useless useMemo calls', () => {
+		const App = () => {
+			const [people] = useState([40, 20,60, 80]);
```
marvinhagemeister: nit, spacing `[40, 20,60, 80] -> [40, 20, 60, 80]` 👀

JoviDeCroock:
> This is awesome! Let's talk about tagging hooks once alpha 3 is out :)

Yes, was thinking the same thing so we can have a full release dedicated to DX

PR #1501, opened 2019-04-02
(fix) - effect execution and cleanup order

JoviDeCroock:
Fixes: https://github.com/developit/preact/issues/1500

Effects were being cleaned up and then invoked which is not the right order if we are conforming to react, where all effects first get cleaned up before new ones get invoked.

Edit:
`+13B`

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -17,8 +17,7 @@ options.render = vnode => {
 	currentIndex = 0;
 
 	if (!currentComponent.__hooks) return;
-	currentComponent.__hooks._pendingEffects.forEach(invokeCleanup);
-	currentComponent.__hooks._pendingEffects.forEach(invokeEffect);
+	handleEffects(currentComponent.__hooks._pendingEffects);
 	currentComponent.__hooks._pendingEffects = [];
 };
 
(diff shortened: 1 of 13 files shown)


hooks/src/index.js
```diff
@@ -196,6 +198,7 @@ function flushAfterPaintEffects() {
 	afterPaintEffects.forEach(component => {
 		component._afterPaintQueued = false;
 		if (!component._parentDom) return;
+		component.__hooks._pendingEffects.forEach(invokeCleanup);
 		component.__hooks._pendingEffects.forEach(invokeEffect);
 		component.__hooks._pendingEffects = [];
```
ForsakenHarmony: could maybe move these to a function

JoviDeCroock: Good spot, will do so! Thank you

hooks/src/index.js
```diff
@@ -220,12 +217,21 @@ if (typeof window !== 'undefined') {
 	};
 }
 
+function handleEffects(effects) {
+	effects.forEach(invokeCleanup);
+	effects.forEach(invokeEffect);
+	return [];
```
cristianbote: Instead of returning maybe we could, just set the `effects = []`? Then we could call `handleEffects` directly with the `effects` list. That would shave `8B` from the current implementation.

cristianbote: Actually, scratch that. These are not references.

JoviDeCroock: My first implementation did this aswell

marvinhagemeister: @cristianbote seems like the suggestion has been addressed. Can you unblock the PR?

cristianbote: Sure thing 😊

PR #1505, opened 2019-04-02
(fix) - debug message should not throw for undefined and null

JoviDeCroock:
Fixes: https://github.com/developit/preact/issues/1504

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -60,7 +60,7 @@ export function initDebug() {
 		}
 
 		for (const key in vnode.props) {
-			if (key[0]==='o' && key[1]==='n' && typeof vnode.props[key]!=='function') {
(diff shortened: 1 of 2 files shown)


debug/src/debug.js
```diff
@@ -60,7 +60,7 @@ export function initDebug() {
 		}
 
 		for (const key in vnode.props) {
-			if (key[0]==='o' && key[1]==='n' && typeof vnode.props[key]!=='function') {
+			if (key[0]==='o' && key[1]==='n' && typeof vnode.props[key]!=='function' && vnode.props[key]!==null && vnode.props[key]!==undefined) {
```
marvinhagemeister: We can use a lose equality check here to combine checking for both `undefined` and `null`:

```js
false == null // false
null == null // true
undefined == null // true
```

So the condition can effectively be shortened to:

```js
vnode.props[key]!=null
```

PR #1506, opened 2019-04-02
pass through props on debugger

jeremy-coleman:
returning the props allows ux to continue rendering if the props are temporarily reset


JoviDeCroock:
It used to be so that preact itself would throw. That being said I think this PR: https://github.com/developit/preact/pull/1505 also solves your case.

JoviDeCroock:
Hey, would you mind testing  the issue you had here against the new alpha release (alpha.4)? It should be resolved.

jeremy-coleman:
@JoviDeCroock seems to be fixed, I updated and no more ux crash , but still get the error (which is warranted/useful in my case)

PR #1509, opened 2019-04-03
(feat) - warn about hooks outside render and/or components

JoviDeCroock:
Adds: 11B to main package (this seems a bit much for one if check ?)

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -101,7 +101,7 @@ export function initDebug() {
 		if (oldBeforeDiff) oldBeforeDiff(vnode);
 	};
 
-	options.preHook = (comp) => {
+	options.hook = (comp) => {
 		if (!comp) {
 			throw new Error('Hook can only be invoked from render methods.');
 		}
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -1,4 +1,4 @@
-import { checkPropTypes } from 'prop-types';
+import { checkPropTypes } from './check-props';
 import { getDisplayName } from './devtools/custom';
 import { options, toChildArray } from 'preact';
 import { ELEMENT_NODE, DOCUMENT_NODE, DOCUMENT_FRAGMENT_NODE } from './constants';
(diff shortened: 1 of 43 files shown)


JoviDeCroock:
coverall is seeing ghosts

src/diff/index.js
```diff
@@ -134,6 +134,8 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
 			let vnode = c._prevVNode = coerceToVNode(c.render(c.props, c.state, c.context));
 			c._dirty = false;
 
+			if (options.afterRender) options.afterRender();
```
marvinhagemeister: Wondering if we could use `options.diffed` instead. It's callsite is just a couple lines later at the end of `diff()`.

hooks/src/index.js
```diff
@@ -21,6 +21,12 @@ options.render = vnode => {
 };
 
 
+let oldDiffed = options.diffed;
+options.diffed = () => {
+	if (oldDiffed) oldDiffed();
+	currentComponent = undefined;
+};
+
```
robertknight: There is another `options.diffed` hook just below. Any reason not to fold the `currentComponent` assignment into that?

debug/test/browser/debug.test.js
```diff
@@ -84,7 +104,7 @@ describe('debug', () => {
 				<p>{state}</p>
 			);
 		};
-		const fn = () => render(<App />, scratch);
+		const fn = () => act(() => render(<App />, scratch));
```
robertknight: What is the effect of adding the `act` call here?

src/index.d.ts
```diff
@@ -191,6 +191,8 @@ declare namespace preact {
 		diff?(vnode: VNode): void;
 		/** Attach a hook that is invoked before a vnode has rendered. */
 		render?(vnode: VNode): void;
+		/** Attach a hook that is invoked before a hook's state is queried. */
+		preHook?(component: VNode): void;
```
robertknight:
```suggestion
		preHook?(component: Component): void;
```

The type of the argument here isn't a `VNode`.

debug/test/browser/debug.test.js
JoviDeCroock: Hooks inside the render method of App will be flushed

hooks/src/index.js
JoviDeCroock: Thought it would improve readability a bit, I'll move it

debug/src/debug.js
```diff
@@ -101,6 +101,12 @@ export function initDebug() {
 		if (oldBeforeDiff) oldBeforeDiff(vnode);
 	};
 
+	options.preHook = (comp) => {
```
marvinhagemeister: Changing `preHook` to just `hook` saves `3 B` :tada:

hooks/src/index.js
```diff
@@ -34,6 +33,7 @@ options.diffed = vnode => {
 	// TODO: Consider moving to a global queue. May need to move
 	// this to the `commit` option
 	hooks._pendingLayoutEffects = handleEffects(hooks._pendingLayoutEffects);
+	currentComponent = undefined;
```
marvinhagemeister: Removing this line has no effect on the test outcome :+1:

PR #1510, opened 2019-04-03
(fix) - event bubbling

JoviDeCroock:
Fixes: https://github.com/developit/preact/issues/1290

Removes: 24Bytes from the main package.

This is no definite change yet, we should try and weigh what is being lost by moving from Promise.resolve -> setTimeout

Explanation by @developit: https://github.com/developit/preact/issues/1290#issuecomment-479656134

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -97,12 +97,6 @@ Component.prototype.render = Fragment;
  */
 let q = [];
 
-/**
- * Asynchronously schedule a callback
- * @type {(cb) => void}
- */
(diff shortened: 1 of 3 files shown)


developit:
Personally I think this represents a nice compromise. setTimeout clamping is avoided in most cases, and only kicks in when folks trigger setState loops. However, there may be a drawback when it comes to many nested setState() calls triggered by rendering a large tree. It would be good to put a demo/test around that case to see how long such a render would take.

PR #1512, opened 2019-04-04
WIP: (feat) - debugging tables

JoviDeCroock:
This is a first try, I don't know all use cases for tables so I would really appreciate a second/... pair of eyes on this.

Thanks in advance!

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -2,7 +2,6 @@ import { checkPropTypes } from 'prop-types';
 import { getDisplayName } from './devtools/custom';
 import { options, toChildArray } from 'preact';
 import { ELEMENT_NODE, DOCUMENT_NODE, DOCUMENT_FRAGMENT_NODE } from './constants';
-import debugTable from './debugTable';
 
 export function initDebug() {
 	/* eslint-disable no-console */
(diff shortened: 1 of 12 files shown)


developit:
I'm thinking this might not work if there are components in the tree, right?

```js
const TableRow = ({ a, b, c }) => <tr><td>{a}</td><td>{b}</td><td>{c}</td></tr>
render(<table>{data.map(row => <TableRow {...row} />)}</table>, document.body)
```

Just a thought as I read through:  what about triggering validation when encountering a child that can only be nested in a given parent (rather than when encountering a parent that can only accept certain children)?  Something like:

```js
function getParentType(vnode) {
  return vnode._dom && vnode._dom.parentNode && vnode._dom.parentNode.localName;
}
options.diffed = vnode => {
  if (vnode.type == 'tr') {
    const parentType = getParentType(vnode);
    if (parentType !== 'tbody' && parentType !== 'thead') {
      console.warn(`Warning: improper nesting of table row\n<tr> can only be placed inside <tbody> or <thead>.\n    ${serializeVnode(vnode)}`);
    }
  }
};
```

JoviDeCroock:
Good point, I'll rewrite it for this case and add tests for it aswell

andrewiggins:
FWIW, now that we have the `_parent` pointer, checking that the first parent DOM vnode is of type `tr`, etc. shouldn't be too hard

JoviDeCroock:
@andrewiggins I'll try to pick this up asap

PR #1513, opened 2019-04-04
Fix reading children from vnode without props in debug.serializeVNode

d3x7r0:
When using the [redux-zero](https://github.com/redux-zero/redux-zero) preact bindings I came across an issue where the debug tools were blowing up on a node without the props object defined.

Looking at the code for `serializeVNode` I can see that most of the code is wrapped around an if to check if props exist except for the one line that tries to get the list of children.

I did a very quick fix for this but I would very much like to write a test for it but I'm not sure what causes a vnode to be created without props being set.


d3x7r0:
Hi @JoviDeCroock I'd very much love to do it but I have no idea what causes a vnode to be created without having the `props` property set. I can see the code handles it and I can trigger it locally with a modified copy of `redux-zero` (the current version doesn't work with Preact X thanks to the changes to the `children` prop)

I have some ideas so I'll try to see if I can do it tonight but if anyone can shed some light on how this condition can come about I would be very much in their debt. All I know is that the current code handles that scenario for the entire `serializeVNode` aside from this one line:

https://github.com/developit/preact/blob/16ac6a4d042e6e211e42e5a3a47e0a93647e5349/debug/src/debug.js#L143-L150

JoviDeCroock:
I don't know where the issue lays but if it is located in Provider I would mimic a Provider without a value or something since normally we do this check in `createElement`:

`if (props==null) props = {};`

d3x7r0:
Actually it seems to be the Connector instead. I think it might have something to do with the usage of a function as children but I can't seem to replicate it.

The code where this is happening seems to be this one: 
```jsx
class ConnectUntyped extends Component {
    ...

  render({ children }, state, { store }) {
    return children[0]({ store, ...state, ...this.actions });
  }
}

...

<ConnectUntyped {...props} mapToProps={mapToProps} actions={actions}>
    {mappedProps => <Child {...mappedProps} {...props} />}
</ConnectUntyped>
```

I already have a PR for the lib open to work around the fact that children isn't always an array now. Hence why It's a bit tricky to setup a codesandbox or something with the error.

Anyway what do you mean by "Provider without a value"? You mean not passing the store to the provider? Or not passing any children? Sorry if it sounds a bit of a stupid question but I'm still a bit new on all this component biz 😅

developit:
Hmm - it seems like this is caused by 2 versions of Preact?

Otherwise this PR should probably be opened against the [8 branch](https://github.com/developit/preact/tree/8).

debug/test/browser/debug.test.js
```diff
@@ -275,6 +275,15 @@ describe('debug', () => {
 			expect(serializeVNode(<Foo />)).to.equal('<Foo />');
 		});
 
+		it('should serialize vnode without props', () => {
+			const vnode = {
```
marvinhagemeister: This kind of `vnode` doesn't exist. Only `vnode`'s representing a text node have `props` set to `null`. That's very likely why the error occurred.

```suggestion
const vnode = {
  type: null,
  props: null,
  text: 'hello world',
  key: null,
  ref: null
};
```

d3x7r0: Thanks! I was really struggling to figure out how the vnode could end up with null props. A text node didn't really cross my mind. 👍

debug/test/browser/debug.test.js
```diff
@@ -275,6 +275,21 @@ describe('debug', () => {
...
+				type: null,
+				props: null,
+				text: 'hello world',
+				key: null,
+				ref: null
+			};
+				type: 'string',
```
JoviDeCroock: I think this is a syntax error

d3x7r0: You're right. I trusted github to apply the change but shouldn't have. :)

Issue #1415, opened 2019-03-17
10.0.0-alpha.4 shouldComponentUpdate keeps Fragment's other children


mischnic:
The issue happens when conditionally rendering a Component that renders multiple elements in a Fragment and also has a `shouldComponentUpdate` method. The state now changes to not render that component anymore, but the second and third children aren't removed (the text that was rendered as well does get removed).
Using no `Fragment`s or removing `shouldComponentUpdate` fixes it.

The only two correct "page states" are (`textarea` + "Preview" + "This should never..." + "... neither this...") and ("Error!"):

![bug](https://user-images.githubusercontent.com/4586894/54870271-a2a1d180-4da4-11e9-93ed-f6d7d8bfd8d3.gif)

```jsx
if (process.env.NODE_ENV === "development") {
	require("preact/debug");
}

import { h, render, Component, Fragment } from "preact";

class Preview extends Component {
	shouldComponentUpdate(nextProps, nextState) {
		return nextProps.output !== this.props.output;
	}

	render() {
		const { input, output } = this.props;
		return (
			<Fragment>
				Preview:
				<div>
					This should never be displayed together with "Error"
				</div>
				<div>... and either this third element</div>
			</Fragment>
		);
	}
}

class App extends Component {
	constructor(props) {
		super(props);

		this.state = {
			input: 'console.log("A");',
			output: null,
			error: null
		};
	}

	run() {
		if ((this.state.input.split('"').length - 1) % 2 !== 0) {
			this.setState({
				error: true,
				output: null
			});
			return;
		}

		const output = "console.log('Hello');";
		this.setState({
			error: null,
			output
		});
	}

	render() {
		return (
			<div>
				<textarea
					onInput={e => this.setState({ input: e.target.value })}
				>
					{this.state.input}
				</textarea>
				<br />
				<button onClick={() => this.run()}>Run!</button>
				<hr />
				{this.state.error ? (
					<div>Error!</div>
				) : (
					this.state.output && (
						<div>
							<textarea>{this.state.output}</textarea>
							<br />
							<Preview
								input={this.state.input}
								output={this.state.output}
							/>
						</div>
					)
				)}
			</div>
		);
	}
}
render(<App />, document.getElementById("root"));
```

Codesandbox: https://codesandbox.io/s/mq65lo49kx

mischnic:
[...]

Issue #1326, opened 2019-03-05
Replacing component root node with node that has a different type should insert new node into the correct position


localvoid:
It is probably another edge case with fragments.

Preact X: https://codesandbox.io/s/6w342oj2y3
React: https://codesandbox.io/s/01nznm190v

marvinhagemeister:
Fixed by #1515 :tada:

PR #1515, opened 2019-04-05
Keep track of childDom between `diffChildren` calls

### LINKED ISSUE #1325: Fragments should handle reordering
Preact X: https://codesandbox.io/s/8ko4or0x42
React: https://codesandbox.io/s/q9o1wlz0rw

andrewiggins:
Improve Fragment correctness by keeping track of the current DOM position being diff'ed between calls to `diffChildren`.

Note: while this does fix a lot of Fragment correctness issues, in some cases it doesn't optimally operate on the DOM and bails out to re-appending all Fragment children. This is a known issue (I'll open that issue) that I'd like to address in a future PR (as it will likely require more/larger changes to `diff`/`diffChildren`).

~~b8bd7b4: +72 B 😬~~
26c359b: +73 B 😬 

Fixes #1325 #1326 #1415

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -18,12 +18,12 @@ import { removeNode } from '../util';
  * which have mounted
  * @param {import('../internal').Component} ancestorComponent The direct parent
  * component to the ones being diffed
- * @param {Node | Text} childDom The current attached DOM
+ * @param {Node | Text} oldDom The current attached DOM
  * element any new dom elements should be placed around. Likely `null` on first
  * render (except when hydrating). Can be a sibling DOM element when diffing
  * Fragments that have siblings. In most cases, it starts out as `oldChildren[0]._dom`.
  */
-export function diffChildren(parentDom, newParentVNode, oldParentVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, childDom) {
+export function diffChildren(parentDom, newParentVNode, oldParentVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, oldDom) {
 	let childVNode, i, j, p, index, oldVNode, newDom,
 		nextDom, sibDom, focus;
 
@@ -62,10 +62,10 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 			oldChildren[index] = null;
 		}
 
-		nextDom = childDom!=null && childDom.nextSibling;
+		nextDom = oldDom!=null && oldDom.nextSibling;
 
 		// Morph the old element into the new one, but don't append it to the dom yet
-		newDom = diff(oldVNode==null ? null : oldVNode._dom, parentDom, childVNode, oldVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, null, childDom);
+		newDom = diff(oldVNode==null ? null : oldVNode._dom, parentDom, childVNode, oldVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, null, oldDom);
 
 		// Only proceed if the vnode has not been unmounted by `diff()` above.
 		if (childVNode!=null && newDom !=null) {
@@ -81,23 +81,23 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 				// this Fragment's DOM tree.
 				newDom = childVNode._lastDomChild;
 			}
-			else if (excessDomChildren==oldVNode || newDom!=childDom || newDom.parentNode==null) {
+			else if (excessDomChildren==oldVNode || newDom!=oldDom || newDom.parentNode==null) {
 				// NOTE: excessDomChildren==oldVNode above:
 				// This is a compression of excessDomChildren==null && oldVNode==null!
 				// The values only have the same type when `null`.
 
-				outer: if (childDom==null || childDom.parentNode!==parentDom) {
+				outer: if (oldDom==null || oldDom.parentNode!==parentDom) {
 					parentDom.appendChild(newDom);
 				}
 				else {
-					sibDom = childDom;
+					sibDom = oldDom;
 					j = 0;
 					while ((sibDom=sibDom.nextSibling) && j++<oldChildrenLength/2) {
 						if (sibDom===newDom) {
 							break outer;
 						}
 					}
-					parentDom.insertBefore(newDom, childDom);
+					parentDom.insertBefore(newDom, oldDom);
 				}
 			}
 
@@ -106,7 +106,7 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 				focus.focus();
 			}
 
-			childDom = newDom!=null ? newDom.nextSibling : nextDom;
+			oldDom = newDom!=null ? newDom.nextSibling : nextDom;
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


src/diff/children.js
```diff
@@ -18,33 +18,20 @@ import { removeNode } from '../util';
  * which have mounted
  * @param {import('../internal').Component} ancestorComponent The direct parent
  * component to the ones being diffed
+ * @param {Node | Text} childDom The current attached DOM
+ * element any new dom elements should be placed around. Likely `null` on first
+ * render (except when hydrating). Can be a sibling DOM element when diffing
+ * Fragments that have siblings. In most cases, it starts out as `oldChildren[0]._dom`.
  */
-export function diffChildren(parentDom, newParentVNode, oldParentVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent) {
+export function diffChildren(parentDom, newParentVNode, oldParentVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, childDom) {
```
andrewiggins: Wondering if `childDom` is better named `oldDom` since it often comes from `oldChildren[i]._dom`...

src/diff/index.js
```diff
@@ -376,3 +385,34 @@ function catchErrorInComponent(error, component) {
...
+/**
+ * Determine which currently attached DOM node to use when beginning to
+ * diff the children of a VNode
+ * @param {import('../internal').VNode} oldVNode The old VNode whose children
+ * are about to be diffed
+ * @param {import('../internal').PreactElement[]} excessDomChildren If hydrating,
+ * the currently attached DOM elements that are being hydrated
+ * @returns {import('../internal').PreactElement | Text | undefined}
+ */
+export function getFirstChildDom(oldVNode, excessDomChildren) {
```
andrewiggins: Looking to inline this function to save some bytes

test/browser/fragments.test.js
```diff
@@ -1172,29 +1237,22 @@ describe('Fragment', () => {
...
 		expect(scratch.innerHTML).to.equal(htmlForFalse, 'rendering from true to false');
 		expectDomLogToBe([
-			// see issue #193 - re-appends all children after Fragment
-			'<ol>002345.appendChild(<li>0)',
-			'<ol>013450.appendChild(<li>1)',
-			'<ol>024501.appendChild(<li>2)',
-			'<ol>345012.appendChild(<li>3)'
+			'<ol>012345.insertBefore(<li>4, <li>0)',
+			'<ol>401235.insertBefore(<li>5, <li>0)',
+			// TODO: see issue #193 - Hmmm why does this extra append happen?
```
andrewiggins: TODO: Migrate issue 193 from old repo and integrate with issue mentioned in description

src/diff/children.js
marvinhagemeister: Agree, `oldDom` is a more fitting name. `childDom` implies that it would be the `_dom` property of `childVNode` but that's not the case and always confuses me a bit tbh :+1:

src/diff/index.js
```diff
@@ -41,15 +45,18 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
...
 			// we'll set `dom` to the correct value just a few lines later.
 			dom = null;
 
 			if (newVNode._children.length) {
 				dom = newVNode._children[0]._dom;
-				newVNode._lastDomChild = newVNode._children[newVNode._children.length - 1]._dom;
+
+				// If lastChild is a Fragment, use _lastDomChild, else use _dom
+				let lastChild = newVNode._children[newVNode._children.length - 1];
+				newVNode._lastDomChild = lastChild._lastDomChild || lastChild._dom;
```
marvinhagemeister: I was initially not sure why we need to set `_lastDomChild` here. After a few minutes of playing around I discovered that it is necessary to set that property when `newVNode` represents a `Fragment` node. Maybe we can mention that in the comment?

src/diff/index.js
```diff
@@ -111,6 +118,7 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
 					c.props = newVNode.props;
 					c.state = s;
 					c._dirty = false;
+					newVNode._lastDomChild = oldVNode._lastDomChild;
```
marvinhagemeister: Good catch :+1: :tada:

src/diff/index.js
```diff
@@ -303,7 +313,7 @@ export function applyRef(ref, value, ancestorComponent) {
  * @param {import('../internal').VNode} vnode The virtual node to unmount
  * @param {import('../internal').Component} ancestorComponent The parent
  * component to this virtual node
- * @param {boolean} skipRemove Flag that indicates that a parent node of the
+ * @param {boolean} [skipRemove] Flag that indicates that a parent node of the
```
marvinhagemeister: :+1:

src/diff/index.js
marvinhagemeister: nvm, the `if`-statement around it makes this obvious. Should've payed more attention :sweat_smile:

src/diff/children.js
marvinhagemeister: Went ahead and renamed it :+1:

test/browser/fragments.test.js
```diff
@@ -120,6 +130,79 @@ describe('Fragment', () => {
...
+
+		expect(scratch.innerHTML).to.equal(div([
+			div(1),
+			span(2),
+			span(2)
+		].join('')));
+		expectDomLogToBe([
+			'<span>1.remove()',
+			// TODO: Why is it re-appending all the children 😢
+			'<div>.appendChild(#text)',
```
marvinhagemeister: Somehow `oldDom` (previously `childDom`) is `null` here. On top of that the index of the old vnode and the new vnode are the same, but when I add a check for that the resulting dom will be incorrect :thinking:

src/diff/index.js
marvinhagemeister: I had a go at this and while it should be possible I'm having a hard time doing that. It breaks all kinds of tests for me.

natevw:
I wonder if this is related to my #1524, maybe the hooks part is something of a red-herring?

src/diff/index.js
andrewiggins: Lol same! That's where I'm at too. Haven't looked at it in a day or two so maybe my fresh eyes will see what I'm missing now

test/browser/fragments.test.js
andrewiggins: Interesting... Yeah, need to figure out why it goes to null

Yup, that was my plan. I'll come back to this after this PR goes through.

andrewiggins:
@natevw No worries! Thanks for taking the time to see if this fixes your issue. We'll have to keep looking to find the fix.

src/diff/children.js
andrewiggins: Thanks!

src/diff/index.js
andrewiggins: Got it working, but it's kinda weird. I'll open it as a separate PR for discussion after we merge this one.

JoviDeCroock:
I've been looking into it, I've gotten pretty far but it'll be for this weekend probably.

PR #1519, opened 2019-04-07
TS: Use same type naming as @types/react

### LINKED ISSUE #1518: Align exported type names with React
I often copy typescript code between preact and react apps and it's slightly annoying when some types have different names :( especially when they don't have to.

Here's a list from top of my head

- `Preact.ComponentConstructor` → `React.ComponentClass`
- `Preact.FunctionalComponent` → `React.FunctionComponent`
- `Preact.ComponentFactory` → `React.ComponentType`
- `Preact.Preact{Context,Provider,Consumer}` → `React.{Context,Provider,Consumer}`

marvinhagemeister:
This makes copy & pasting react components that are written in TypeScript easier.

Fixes #1518 .


JoviDeCroock:
Maybe we need to export them as an alias, so that we stay backwards compatible for people relying on the old names?

marvinhagemeister:
True, this is a breaking change and we should make the transition as smooth as possible :+1:

PR #1520, opened 2019-04-07
(tests) - act should be able to track and flush the full queue and all subsequent queues

JoviDeCroock:
While reasoning about our current testing approach I found a critical flaw. What happens when for example you type in an input -> this triggers a validation -> this sets state. A series of hooks get executed and well our current approach in act has no way of following these up.

This is why I introduced a global array in options, named effects. Every queued effect will be pushed and will be removed when we're invoking/cleaning up. This makes for a consistent testing method.

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -220,12 +220,14 @@ if (typeof window !== 'undefined') {
 function handleEffects(effects) {
 	effects.forEach(invokeCleanup);
 	effects.forEach(invokeEffect);
+	if (options.effects) {
+		effects.forEach(hook => options.effects = options.effects.filter(h => h!==hook));
+	}
 	return [];
 }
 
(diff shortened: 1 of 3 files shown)


JoviDeCroock:
The weird thing seems to be that when I call `rerender` the render method of my Functional Component does not seem to be executed.

robertknight:
Great. I ran into this recently and had to implement [a utility](https://github.com/hypothesis/client/blob/2587a17f202406eb87f9b604aeaca403cef703f2/src/sidebar/components/test/render-utils.js#L18) as a workaround. I'll try to test this out locally and see if it resolves the issue in my case.

JoviDeCroock:
@robertknight that would be awesome, please do! It works for the case I ran into, if you find another one just hit me up and we can add it in tests.

PR #1525, opened 2019-04-09
Provide simplified checkPropTypes implementation

### LINKED ISSUE #1496: Preact package should not have a hard dependency on prop-types
In https://github.com/developit/preact/commit/069bc8f8a9ea4f514df034d77cadd8160ebb4baa#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R63 a new top-level dependency was added on `prop-types@^15.6.2`. This package in turn has three dependencies, one of which has an additional dependency of its own. So as a user I end up with six Other People's Codes on my machine (`preact`, `prop-types`, `loose-envify`, `js-tokens`, `object-assign`, `react-is`) even though I'm only using/needing/wanting one (`preact`).

The `prop-types` package itself is only needed by a "debug" sub-package hosted in this same repo, but the simple cleanup at #1460 was problematic and had to be reverted. Can a solution be found for the `debug` folder's usage that still gets us back to a place where `preact` itself has no dependencies?

natevw:
This is intended to fix #1496.

I've once again removed the top-level dependency on `prop-types`. In its place, I've provided a fairly simplified implementation of `checkPropTypes` based on just the core of the original.

Questions/disclosures:

- [x] I don't use propTypes myself so I haven't so much as kicked the tires on this in the real world, just ran `npm rm -r node_modules && npm install && npm test` when done. [Resolved: now more tests/familiarity]
- [x] The file "debug/test/browser/debug.test.js" still needs the real 'prop-types'; I'm assuming that it *will* have that via "debug/package.json" somehow when the tests are run? If not, then I imagine the simplest fix would be to make `prop-types` a top-level **dev**Dependency. [Resolved: added as top-level devDependency]
- [x] The original version was ± ES3. I've replaced a `for (var typeSpecName in typeSpecs) if (typeSpecs.hasOwnProperty(typeSpecName)) {` loop with `Object.keys(typeSpecs).forEach((typeSpecName) => {` which is ES7 (but could trivially become ES5) [Resolved: handled by dist build]
- [x] …there's still some `vars` though to keep the `process.env.NODE_ENV` code stripping stuff simple. [Resolved: didn't need NODE_ENV here.]
- [x] The original version does its own validation on both the typeSpec function and its return value, essentially validating the validators and printing helpful messages if it finds them wanting. I've removed that and just let the raw fallout get logged instead. [Resolved: won't fix]
- [x] I wasn't sure whether the implementation belonged in `debug/src` (next to e.g. "constants.js") or in `debug/src/devtools` (next to e.g. "custom.js") so I just went with the parent folder. [Resolved: debug/ was correct]

### DIFF
--- a/debug/src/check-props.js
+++ b/debug/src/check-props.js
@@ -1,28 +1,24 @@
 
-if (process.env.NODE_ENV !== 'production') {
-  var ReactPropTypesSecret = 'SECRET_DO_NOT_PASS_THIS_OR_YOU_WILL_BE_FIRED',
-      loggedTypeFailures = {};
-}
+var ReactPropTypesSecret = 'SECRET_DO_NOT_PASS_THIS_OR_YOU_WILL_BE_FIRED',
+    loggedTypeFailures = {};
 
 export function checkPropTypes(typeSpecs, values, location, componentName, getStack) {
-  if (process.env.NODE_ENV !== 'production') {
-    Object.keys(typeSpecs).forEach((typeSpecName) => {
-      let error;
-      try {
-        error = typeSpecs[typeSpecName](values, typeSpecName, componentName, location, null, ReactPropTypesSecret);
-      } catch (e) {
-        error = e;
+  Object.keys(typeSpecs).forEach((typeSpecName) => {
+    let error;
+    try {
+      error = typeSpecs[typeSpecName](values, typeSpecName, componentName, location, null, ReactPropTypesSecret);
+    } catch (e) {
+      error = e;
+    }
+    if (error && !(error.message in loggedTypeFailures)) {
+      loggedTypeFailures[error.message] = true;
+      
+      let errorReport = `Failed ${location} type: ${error.message}${getStack && getStack() || ''}`;
+      if (typeof console !== 'undefined') {
+        console.error(errorReport);
+      } else {
+        throw new Error(errorReport);
       }
-      if (error && !(error.message in loggedTypeFailures)) {
-        loggedTypeFailures[error.message] = true;
-        
-        let errorReport = `Failed ${location} type: ${error.message}${getStack && getStack() || ''}`;
-        if (typeof console !== 'undefined') {
-          console.error(errorReport);
-        } else {
-          throw new Error(errorReport);
-        }
-      }
-    });
-  }
+    }
+  });
 }
--- a/debug/src/check-props.js
+++ b/debug/src/check-props.js
@@ -12,13 +12,7 @@ export function checkPropTypes(typeSpecs, values, location, componentName, getSt
     }
     if (error && !(error.message in loggedTypeFailures)) {
       loggedTypeFailures[error.message] = true;
-      
-      let errorReport = `Failed ${location} type: ${error.message}${getStack && getStack() || ''}`;
-      if (typeof console !== 'undefined') {
-        console.error(errorReport);
-      } else {
-        throw new Error(errorReport);
-      }
+      console.error(`Failed ${location} type: ${error.message}${getStack && getStack() || ''}`);
     }
   });
 }
--- a/debug/src/check-props.js
+++ b/debug/src/check-props.js
@@ -1,0 +1,28 @@
+
+if (process.env.NODE_ENV !== 'production') {
+  var ReactPropTypesSecret = 'SECRET_DO_NOT_PASS_THIS_OR_YOU_WILL_BE_FIRED',
+      loggedTypeFailures = {};
+}
+
+export function checkPropTypes(typeSpecs, values, location, componentName, getStack) {
+  if (process.env.NODE_ENV !== 'production') {
+    Object.keys(typeSpecs).forEach((typeSpecName) => {
+      let error;
+      try {
+        error = typeSpecs[typeSpecName](values, typeSpecName, componentName, location, null, ReactPropTypesSecret);
+      } catch (e) {
+        error = e;
+      }
+      if (error && !(error.message in loggedTypeFailures)) {
+        loggedTypeFailures[error.message] = true;
+        
+        let errorReport = `Failed ${location} type: ${error.message}${getStack && getStack() || ''}`;
+        if (typeof console !== 'undefined') {
+          console.error(errorReport);
+        } else {
+          throw new Error(errorReport);
+        }
+      }
+    });
+  }
+}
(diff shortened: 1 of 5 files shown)


debug/src/check-props.js
```diff
@@ -0,0 +1,28 @@
+
+if (process.env.NODE_ENV !== 'production') {
+  var ReactPropTypesSecret = 'SECRET_DO_NOT_PASS_THIS_OR_YOU_WILL_BE_FIRED',
```
marvinhagemeister: Let's use newer keywords like `const` or `let` here :+1:

debug/src/check-props.js
```diff
@@ -0,0 +1,28 @@
+
+if (process.env.NODE_ENV !== 'production') {
```
marvinhagemeister: The whole `preact/debug` import has this check around all code in the entry file. We can safely omit it here :+1:

debug/src/check-props.js
```diff
@@ -0,0 +1,28 @@
...
+  var ReactPropTypesSecret = 'SECRET_DO_NOT_PASS_THIS_OR_YOU_WILL_BE_FIRED',
+      loggedTypeFailures = {};
+}
+
+export function checkPropTypes(typeSpecs, values, location, componentName, getStack) {
+  if (process.env.NODE_ENV !== 'production') {
```
marvinhagemeister: Same as above, this check can be omitted here :+1:

debug/src/check-props.js
```diff
@@ -0,0 +1,28 @@
...
+      try {
+        error = typeSpecs[typeSpecName](values, typeSpecName, componentName, location, null, ReactPropTypesSecret);
+      } catch (e) {
+        error = e;
+      }
+      if (error && !(error.message in loggedTypeFailures)) {
+        loggedTypeFailures[error.message] = true;
+        
+        let errorReport = `Failed ${location} type: ${error.message}${getStack && getStack() || ''}`;
+        if (typeof console !== 'undefined') {
```
marvinhagemeister: We can safely assume that `console` is available in both `node` and `browser` environments :100:

marvinhagemeister:
Awesome, that's an excellent start :+1: A few minor changes + some tests cases are missing and then I think it will be ready to go in :tada: 

Regarding your questions:

Questions/disclosures:

> The file "debug/test/browser/debug.test.js" still needs the real 'prop-types'; I'm assuming that it will have that via "debug/package.json" somehow when the tests are run? If not, then I imagine the simplest fix would be to make prop-types a top-level devDependency.

yup, to be able to test against `prop-types` we need to declare it as a `devDependency`. Note that from npm's point of view there is only one package which is why it must be added to the main `package.json` file, not the one in `debug/package.json`.

> The original version was ± ES3. (...)

Our build tool [microbundle](https://github.com/developit/microbundle) will automatically transpile it down to ES5 :tada: 

> The original version does its own validation on both the typeSpec function and its return value, essentially validating the validators and printing helpful messages if it finds them wanting. I've removed that and just let the raw fallout get logged instead.

Good call! Agree, I don't think we need to support that use case. There was a time where devs could write their own prop-type checkers but I haven't seen any of those in the real world so far :+1: 

> Oh, also I wasn't sure whether the implementation belonged in debug/src (next to e.g. "constants.js") or in debug/src/devtools (next to e.g. "custom.js") so I just went with the parent folder.

It belongs in `debug/src`. The code in `debug/src/devtools` is specifically for the [react-devtools](https://github.com/facebook/react-devtools) extension. So you made the right choice by intuition :+1: 

Let me know what you think :slightly_smiling_face:

natevw:
Thanks, appreciate the feedback. I can make some of the tweaks right now, but it will take me a bit longer to get up to speed with the test/coverage improvements (likely not until *after* [April 15](https://www.irs.gov/filing)…).

debug/src/check-props.js
natevw: Cool, simplified this in a8655d6.

debug/src/check-props.js
natevw: Yep, without the NODE_ENV stuff we don't need `var`. Done in bd70cf3 with a bit of judgment call (`ReactPropTypesSecret` definitely const, but I left the `loggedTypeFailures` as a `let` even though technically the reference itself is `const`…)

debug/src/check-props.js
natevw: This was cleaned up in 9bcfb14

debug/src/check-props.js
natevw: in 9bcfb14 as well

debug/src/check-props.js
marvinhagemeister: That's perfectly fine we have lots of places where a `let` declaration could be a `const` :slightly_smiling_face:

natevw:
Okay, inline feedback addressed but there's I think two remaining threads from https://github.com/developit/preact/pull/1525#issuecomment-481133751 here:

> yup, to be able to test against `prop-types` we need to declare it as a `devDependency`

Right now `prop-types` is *not* a `devDependency` of the top-level package but my own local `npm test` and the travis-ci integration here seem to be happy 😬 

If this is just some caching side-effect (I did `rm -r node_modules debug/node_modules` before starting to test earlier) then we should probably add it as a top-level `devDependency` to be safe? But that seems a bit questionable if it already works fine without?

Otherwise the followup here would be to *move* `prop-types` in "debug/package.json" to be a `devDependency` since even there it is only needed for tests.

---

As far as coverage, it's a lot better after the recent commits, -0.1% instead of -2.04% but especially since there's already tests for this should be relatively easy to trigger the [missing line](https://coveralls.io/builds/22729772/source?filename=debug%2Fsrc%2Fcheck-props.js#L11):

>    PropTypes
>      ✔ should fail if props don't match prop-types
>      ✔ should not print to console when types are correct

Maybe I'll just look into that while I'm on a roll anyway!

natevw:
Ah, figured out the "why do the tests work" bit too. `prop-types` is currently getting installed anyway via:

```
├─┬ eslint-config-developit@1.1.1
│ └─┬ eslint-plugin-react@7.12.4
│   ├─┬ prop-types@15.7.2
```

Which is only coincidental.

natevw:
I think all the concerns have been addressed; I'm happy with where this is at and it's is ready for review again.

PR #1527, opened 2019-04-09
Add support for arrays in class prop

safrankamate:
A small convenience feature to make it easier to dynamically add CSS classes.
Example:

```
const LAYOUT_CLASS = 'layout-flex-column';
const isMobile = window.matchMedia('(max-width: 640px)').matches;

<div class={[
  'panel',
  LAYOUT_CLASS,
  isMobile && 'mobile',
]}></div>
```

Will be rendered as:

```
On desktop:
<div class="panel layout-flex-column"></div>

On mobile:
<div class="panel layout-flex-column mobile"></div>
```

**Adds: 37B to core**

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -34,7 +34,10 @@ let CAMEL_REG = /-?(?=[A-Z])/g;
  */
 function setProperty(dom, name, value, oldValue, isSvg) {
 	let v;
-	if (name==='class' || name==='className') name = isSvg ? 'class' : 'className';
+	if (name==='class' || name==='className') {
+		name = isSvg ? 'class' : 'className';
+		value = Array.isArray(value) ? value.filter(i => !!i).join(' ') : value;
+	}
 
 	if (name==='style') {
 


marvinhagemeister:
That's awesome, thanks for the PR! We actually had something similar in Preact a few versions back, but later removed it as most of our users weren't using it. This was especially true for users coming from React who relied on `classnames` or a similar library. Given our strict size constraints, we were happy to use the saved bytes for other features.

That said you can easily extend Preact in your project to allow passing arrays for `class/className`. This can be done via our `option` hooks, see https://marvinh.dev/blog/preact-options/ . It would probably look something like this:

```jsx
import { options } from "preact";

options.vnode = vnode => {
  const { className } = vnode.props;
  if (className && Array.isArray(className)) {
    vnode.props.className = className.filter(Boolean).join(' ');
  }
};

// Somewhere else
<div className={["foo", "bar", undefined]} />
// renders: <div class="foo bar"></div>
```

safrankamate:
Thanks for the quick and thorough response! The option hooks feature looks really useful, I'll definitely check them out. I also appreciate the principled attitude to size limits :) Cheers!

PR #1528, opened 2019-04-09
(feat) - prevent json injection

JoviDeCroock:
Some tests are now skipped since functions won't pass through the new `_` assignment. I opted for this above __proto__ since proto is in essence deprecated and would probably require a lot of changing in how we work in /debug.

This adds a property on vnode that tracks itself, noone can insert an object with the same ref so this will prevent JSON Injection.

TEST2: __proto__ overrides component.constructor implying that catchErrorInComponent stops working. I could really be approaching this wrong but this is what I conclude after testing it with __proto__ instead of _ with obj ref.

ADDS: 12B to core

### DIFF
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -62,7 +62,7 @@ export function createVNode(type, props, text, key, ref) {
 		_lastDomChild: null,
 		_component: null
 	};
-	vnode._ = vnode;
+	vnode._self = vnode;
 
 	if (options.vnode) options.vnode(vnode);
 
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -62,6 +62,7 @@ export function createVNode(type, props, text, key, ref) {
 		_lastDomChild: null,
 		_component: null
 	};
+	vnode._ = vnode;
 
 	if (options.vnode) options.vnode(vnode);
 
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -22,7 +22,6 @@ import options from '../options';
  * parent component
  */
 export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, force) {
-
 	// If the previous type doesn't match the new type we drop the whole subtree
 	if (oldVNode==null || newVNode==null || oldVNode.type!==newVNode.type) {
 		if (oldVNode!=null) unmount(oldVNode, ancestorComponent);
@@ -30,6 +29,11 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
 		dom = null;
 		oldVNode = EMPTY_OBJ;
 	}
+
+	// When passing through createElement it assignes the object
+	// ref on _, to prevent JSON Injection we check if this attribute
+	// is equal.
+	if (newVNode._!==newVNode) return null;
 
 	if (options.diff) options.diff(newVNode);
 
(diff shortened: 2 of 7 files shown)


test/browser/render.test.js
```diff
@@ -173,7 +179,7 @@ describe('render()', () => {
 		expect(scratch.innerHTML).to.equal('');
 	});
 
-	it('should throw an error on function children', () => {
+	it.skip('should throw an error on function children', () => {
```
JoviDeCroock: function children don't seem to pass createElement making them victim to the _ check

test/browser/toChildArray.test.js
```diff
@@ -59,7 +59,7 @@ describe('props.children', () => {
 		expect(scratch.innerHTML).to.equal('<div></div>');
 	});
 
-	it('throws an error if given a function child', () => {
+	it.skip('throws an error if given a function child', () => {
```
JoviDeCroock: Same story here

src/diff/index.js
```diff
@@ -31,6 +30,11 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
 		oldVNode = EMPTY_OBJ;
 	}
 
+	// When passing through createElement it assignes the object
```
robertknight:
```suggestion
	// When passing through createElement it assigns the object
```

src/index.d.ts
```diff
@@ -28,6 +28,7 @@ declare namespace preact {
 		 * Default value: `-1`
 		 */
 		endTime?: number;
+		_: VNode;
```
robertknight: No other private properties are documented here. Is there a reason to document this one?

test/browser/render.test.js
robertknight: How does React handle this case?

src/create-element.js
```diff
@@ -62,6 +62,7 @@ export function createVNode(type, props, text, key, ref) {
 		_lastDomChild: null,
 		_component: null
 	};
+	vnode._ = vnode;
```
robertknight: Could this be given a more self-documenting name, which is then shortened in the build process like the other private properties are?

test/browser/render.test.js
JoviDeCroock: Well are functions as children allowed? Normally it should throw and now they won't get passed. Don't really know what we should do

src/create-element.js
JoviDeCroock: Is this better, it's _self now. I must admit the name was a bit odd of me to do, I wanted to contain bundle size increase but you're right that's what we have our mangle file for :D thanks for pointing it out

test/browser/render.test.js
robertknight: React produces this detailed error if you try to do this:

Screenshot 2019-04-14 23 13 57

This logic could be moved to the `vnode` validation hook in `preact/debug` perhaps? The current behaviour on this branch of rendering nothing if a function child is passed to a DOM node seems reasonable to me, although that should have a test if it is agreed upon.

JoviDeCroock: Is this in scope of this pr?

JoviDeCroock:
I'll make a seperate PR one of these days to add it in debug

test/browser/toChildArray.test.js
robertknight: Does this test cover anything the other one to do with function children doesn't? If not I would suggest to just remove it.

JoviDeCroock:
> I have a query over whether there is any reason to keep the remaining skipped test. Overall LGTM though.

Good point, I must've looked over it apologies for that. Will fix it asap

PR #1530, opened 2019-04-11
fix: components having attributes that start with on

JoviDeCroock:
Fixes: https://github.com/developit/preact/issues/1504#issuecomment-482220610

There were scenario's that components had properties starting with 'on' for example online.

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -61,7 +61,7 @@ export function initDebug() {
 		}
 
 		for (const key in vnode.props) {
-			if (key[0]==='o' && key[1]==='n' && typeof vnode.props[key]!=='function' && vnode.props[key]!=null) {
+			if (key[0]==='o' && key[1]==='n' && typeof vnode.props[key]!=='function' && vnode.props[key]!=null && typeof vnode.type!=='function') {
 				throw new Error(
 					`Component's "${key}" property should be a function, ` +
 					`but got [${typeof vnode.props[key]}] instead\n` +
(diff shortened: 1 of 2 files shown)


debug/src/debug.js
```diff
@@ -61,7 +61,7 @@ export function initDebug() {
 		}
 
 		for (const key in vnode.props) {
-			if (key[0]==='o' && key[1]==='n' && typeof vnode.props[key]!=='function' && vnode.props[key]!=null) {
+			if (key[0]==='o' && key[1]==='n' && typeof vnode.props[key]!=='function' && vnode.props[key]!=null && typeof vnode.type!=='function') {
```
marvinhagemeister: Let's move the check outside the `for`-loop so that we don't have to do it for every prop. If we change it to check for `vnode.type === 'string'` we can automatically skip checking text nodes which don't have props 👍

JoviDeCroock: Good idea wasn't thinking straight there

debug/src/debug.js
```diff
@@ -60,13 +60,15 @@ export function initDebug() {
...
 		}
 
-		for (const key in vnode.props) {
-			if (key[0]==='o' && key[1]==='n' && typeof vnode.props[key]!=='function' && vnode.props[key]!=null) {
-				throw new Error(
-					`Component's "${key}" property should be a function, ` +
-					`but got [${typeof vnode.props[key]}] instead\n` +
-					serializeVNode(vnode)
-				);
+		if (typeof vnode.type!=='function') {
```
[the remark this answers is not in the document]
marvinhagemeister: Went ahead and changed it, hope you don't mind :eyes:

PR #1531, opened 2019-04-12
Remove getFirstOldDom (-42 B)

andrewiggins:
Move the implementation of `getOldFirstDom` into `diffChildren` and only invoke it when necessary.

See comments inside of the PR for a discussion about this implementation.

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -32,6 +32,26 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 
 	let oldChildrenLength = oldChildren.length;
 
+	if (oldDom == EMPTY_OBJ) {
+		oldDom = null;
+		if (excessDomChildren!=null) {
+			for (i = 0; i < excessDomChildren.length; i++) {
+				if (excessDomChildren[i]!=null) {
+					oldDom = excessDomChildren[i];
+					break;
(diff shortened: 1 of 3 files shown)


src/diff/children.js
```diff
@@ -32,6 +32,26 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 
 	let oldChildrenLength = oldChildren.length;
 
+	if (oldDom == EMPTY_OBJ) {
```
andrewiggins: Only in very specific places should this logic be invoked (top level `render` and `diffElementNodes`). I'm using `EMPTY_OBJ` to signal when `diffChildren` is invoked in these situations. I can't use `null` for this purpose, because `null` is a valid value for `oldDom` which can mean to skip to this logic (e.g. if mounting a new tree in which the old DOM should be ignored (usually for Fragments).

I tried using `=== undefined`  and `== EMPTY_ARR` but both ended up being a couple bytes larger.

andrewiggins: I'll add this as a comment in the code

marvinhagemeister: That's really clever, I was way too focused on finding any differences in the `vnodes` themselves rather than just checking function arguments :+1:

PR #1534, opened 2019-04-12
(fix) - consistent order of state updates

### LINKED ISSUE #874: RenderQueue doesn't appear to be ordered by depth
Ideally you render top down to avoid doing wasted work on child components that a parent component is going to remove or pass different props to. It currently appears that the queue is in the order of the setState calls which means a child component might rerender and build up new dom and then a parent would render and do it again and possibly remove that child entirely or pass down different props.

https://github.com/developit/preact/blob/80ad4367e5352cbf6720f973a4488f6504498eb9/src/component.js#L58
https://github.com/developit/preact/blob/576dda310a8e31e524fa2ecc803d6d406597c41e/src/render-queue.js#L10

Instead I think you want to walk up the tree of components until you find the highest one with `_dirty === true`, render that and then repeat until every component is clean.

JoviDeCroock:
This fixes the problem that got added in demo, where the order of setting state mattered.

In one case where we first altered the list and then altered the key some things wouldn't properly unmount and caused a loss of focus.

To test: The performance benchmarks of `yarn test:karma:watch` seem to have improved quite a lot on my machine. Could someone else test this aswell?

Note that I am still looking to make a test for this.

Fixes: https://github.com/developit/preact/issues/1524 , Fixes #874
Adds: 34B

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -125,7 +125,7 @@ export function enqueueRender(c) {
 /** Flush the render queue by rerendering all queued components */
 function process() {
 	let p;
-	let toRender = q.splice(0).sort((a, b) => a._depth - b._depth);
+	let toRender = q.sort((a, b) => a._depth - b._depth);
 	while ((p=toRender.pop())) {
 		// forceUpdate's callback argument is reused here to indicate a non-forced update.
 		if (p._dirty) p.forceUpdate(false);
--- a/src/component.js
+++ b/src/component.js
@@ -125,8 +125,7 @@ export function enqueueRender(c) {
 /** Flush the render queue by rerendering all queued components */
 function process() {
 	let p;
-	let toRender = q.sort((a, b) => a._depth - b._depth);
-	while ((p=toRender.pop())) {
+	while ((p=q.pop())) {
 		// forceUpdate's callback argument is reused here to indicate a non-forced update.
 		if (p._dirty) p.forceUpdate(false);
 	}
--- a/src/component.js
+++ b/src/component.js
@@ -125,6 +125,7 @@ export function enqueueRender(c) {
 /** Flush the render queue by rerendering all queued components */
 function process() {
 	let p;
+	q.sort((a, b) => a._depth - b._depth);
 	while ((p=q.pop())) {
 		// forceUpdate's callback argument is reused here to indicate a non-forced update.
 		if (p._dirty) p.forceUpdate(false);
--- a/src/component.js
+++ b/src/component.js
@@ -70,7 +70,7 @@ Component.prototype.forceUpdate = function(callback) {
 		const force = callback!==false;
 
 		let mounts = [];
-		dom = diff(dom, parentDom, vnode, vnode, this._context, parentDom.ownerSVGElement!==undefined, null, mounts, this._ancestorComponent, force, dom, this._depth);
+		dom = diff(dom, parentDom, vnode, vnode, this._context, parentDom.ownerSVGElement!==undefined, null, mounts, this._ancestorComponent, force, dom);
 		if (dom!=null && dom.parentNode!==parentDom) {
 			parentDom.appendChild(dom);
 		}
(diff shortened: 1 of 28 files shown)


src/component.js
```diff
@@ -125,6 +125,7 @@ export function enqueueRender(c) {
 /** Flush the render queue by rerendering all queued components */
 function process() {
 	let p;
+	q.sort((a, b) => a._depth - b._depth);
```
andrewiggins: Hmm we should consider splicing `q` (e.g. `q.splice(0)` or something similar) so that any `setState` calls queued up in `forceUpdate` don't get added to `q` after it has been sorted. I believe that any `setState` calls invoked while processing a previous `setState` should be rerendered after the current `setState` finished (i.e. should be added at the "end" of the queue, which for us means processed in another pass through the queue).

src/diff/index.js
```diff
@@ -25,8 +25,8 @@ import options from '../options';
  * render (except when hydrating). Can be a sibling DOM element when diffing
  * Fragments that have siblings. In most cases, it starts out as `oldChildren[0]._dom`.
  */
-export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, force, oldDom) {
-
+export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, force, oldDom, depth) {
+	if (!depth) depth = 0;
```
andrewiggins: Hmm would the size be any smaller if instead of adding an argument to `diff` we just set it directly on each vnode as we loop through them in `diffChildren`?

Something like: (pseudo-code of `diffChildren`)
```js
function diffChildren(newParentVNode) {
    for (let child of newParentVNode._children) {
        // The initial VNode is always a Fragment that won't have `_depth` set
        child._depth = (newParentVNode._depth | 0) + 1;
    }
}
```
It's one benefit of `diffChildren` being the entry point into the diff algorithm 😁

src/component.js
andrewiggins: Gonna see what the react behavior is here and perhaps that can become a test?

src/diff/index.js
JoviDeCroock: Problem here being that you won't be able to set your root, or we should set it there aswell.

src/component.js
JoviDeCroock: I have added this change, good spot. Makes me wonder if we should propagate depth in diffChildren aswell....

JoviDeCroock: With splice it seems to fail a bunch of tests

src/diff/index.js
marvinhagemeister: Just tried that and it saves `5 B`. All tests are passing too.

JoviDeCroock: Yes but demo will still fail, can’t repro a test for this atm

marvinhagemeister: True, I stand corrected! Thanks for clearing this up :+1:

JoviDeCroock:
Thank you for the extra pair of eyes @marvinhagemeister, I think this is ready for review now I removed the residual _depth

src/component.js
marvinhagemeister: I had a go at this, but it seems to be a bit more involved. As far as I understand splicing the queue would demand a way to know the cutoff point at which the first item that was not part of the original queue is processed, so that we can sort the second update individually.

Another way could be to append components to the queue as as usual, but to resort it before the next call to `pop()` in case there were any new items added.

The third way I can come up with is to use a separate array all together. That is my least preferred option, because it can get really tricky when a component is in both queues and so on.

But then again I'm wondering if we don't already support that. Whenever we trigger an update and something calls `forceUpdate`, like during a lifecycle, we immediately update the component's children. Those are then flagged via `c._dirty` so that the queue will skip processing them. I'm probably missing something, but to me it seems to work out well (at least in my head).

Either way I feel like this requires more thought and would be better to work on in another PR. It will likely brew some discussions and that would leave the issue open for longer.

Issue #1322, opened 2019-03-05
Strictly equal VNodes should short-circuit diffing


localvoid:
Libraries like react-redux are relying on this behavior for optimizations.

Preact X: https://codesandbox.io/s/4q307zk7o9
React: https://codesandbox.io/s/m3o2qy7398

asilvas:
Surprised v10 is being rolled out without resolving this first as some apps will experience massive slowdowns by upgrading. More specifically the issue linked to this: https://github.com/preactjs/preact/pull/1538

JoviDeCroock:
No apps should be faster by upgrading normally, are you experiencing something else @asilvas

marvinhagemeister:
Yup we didn't have that optimization in 8.x either iirc.

asilvas:
I see, wasn't aware this was a pre-existing issue. In my testing it resulted in significant slowness compared to React, and must have assumed it was related to v10.

marvinhagemeister:
This was added earlier this year with #2386 :tada: Updating the codesandbox to the latest version of Preact resolves it :+1:

PR #1538, opened 2019-04-13
(feat) - short-circuit diffing

JoviDeCroock:
Libraries like redux and other state management libraries rely on this to propagate state changes.

Initial problem being that we pass the current vnode as old and new vnode on setState, this is logical but limits us in certain aspects for this case.

tests: https://github.com/developit/preact/issues/1322
Fixes: https://github.com/developit/preact/issues/1398 (This works now)

Adds:  ~~15~~ 14Bytes to the main package

### DIFF
--- a/hooks/test/browser/useMemo.test.js
+++ b/hooks/test/browser/useMemo.test.js
@@ -73,8 +73,8 @@ describe('useMemo', () => {
 		expect(spy2).to.be.calledOnce;
 		expect(scratch.innerHTML).to.equal('<div><span>0</span><p>0</p></div>');
 		render(<App x={1} />, scratch);
-		expect(spy2).to.be.calledTwice;
 		expect(spy).to.be.calledThrice;
+		expect(spy2).to.be.calledTwice;
 		expect(scratch.innerHTML).to.equal('<div><span>1</span><p>1</p></div>');
 	});
 });
--- a/hooks/test/browser/useMemo.test.js
+++ b/hooks/test/browser/useMemo.test.js
@@ -41,7 +41,7 @@ describe('useMemo', () => {
 		expect(memoFunction).to.have.been.calledTwice;
 	});
 
-	it.skip('short circuits diffing for memoized components', () => {
+	it('short circuits diffing for memoized components', () => {
 		let spy = sinon.spy();
 		let spy2 = sinon.spy();
 		const X = ({ count }) => {
--- a/hooks/test/browser/useMemo.test.js
+++ b/hooks/test/browser/useMemo.test.js
@@ -1,7 +1,7 @@
 import { setupRerender } from 'preact/test-utils';
 import { createElement as h, render } from 'preact';
 import { setupScratch, teardown } from '../../../test/_util/helpers';
-import { useMemo } from '../../src';
+import { useMemo, useState } from '../../src';
 
 /** @jsx h */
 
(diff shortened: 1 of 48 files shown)


test/browser/render.test.js
```diff
@@ -815,6 +815,42 @@ describe('render()', () => {
...
+			}
+
+			render(props) {
+				return <div>{A}</div>;
+			}
+		}
+
+		render(<App />, scratch);
+		expect(spy).to.be.calledOnce;
+		expect(scratch.firstChild.firstChild.innerHTML).to.equal('1');
```
marvinhagemeister: Maybe `scratch.textContent` does suffice here :+1:

JoviDeCroock:
Can't seem to find the right approach for #1322 since the wrapper should bail out of rendering but it might send out something to Providers or? 

The hard part is that when we are setting state we call diff with old and new set on the same entity which makes it kind off hard to do checks on this.

I've put this up to review with the test for 1322 skipped and 1398 being a success. I will handle 1322 in a seperate PR when I've gained more insight in this.

JoviDeCroock:
reused vnodes seem to trigger short-circuiting pretty agressively

hooks/test/browser/useMemo.test.js
```diff
@@ -42,4 +41,40 @@ describe('useMemo', () => {
...
+					{y}
+				</div>
+			);
+		};
+
+		render(<App x={0} />, scratch);
+		expect(spy).to.be.calledOnce;
+		expect(spy2).to.be.calledOnce;
+		expect(scratch.innerHTML).to.equal('<div><span>0</span><p>0</p></div>');
+		render(<App x={0} />, scratch);
```
marvinhagemeister: super tiny nit: Personally I prefer a blank line before the next update so that it's easier to scan for the eyes to see what's happening :)

test/browser/render.test.js
```diff
@@ -815,6 +815,42 @@ describe('render()', () => {
 		expect(scratch.innerHTML).to.contain(`<span>${todoText}</span>`);
 	});
 
+	it.skip('equal vnodes should short-circuit diffing', () => {
```
marvinhagemeister: Is this skipped by accident or does the test fail?

JoviDeCroock: This is the test I prepared for the setState one, this fails for now as I can't find out how to properly tackle this yet.

JoviDeCroock:
Will try and debug this tomorrow... It seems that something broke because of the merge, _original and current seem to lose equality again

src/create-element.js
```diff
@@ -95,6 +95,7 @@ export function coerceToVNode(possibleVNode) {
 	if (possibleVNode._dom!=null) {
 		let vnode = createVNode(possibleVNode.type, possibleVNode.props, possibleVNode.text, possibleVNode.key, null);
 		vnode._dom = possibleVNode._dom;
+		vnode._original = possibleVNode._original || possibleVNode;
```
JoviDeCroock: This had to be added since we now possibly coerce two times, once when newChildren get added and once before diffing.

developit: We could add another argument for `original` to `createVNode`. That would also ensure the object's shape is unchanged, which is needed for performance.

JoviDeCroock:
I'm really struggling to find a reasoning when to short-circuit on setState and when not to. We could do a check if arguments are supplied or its just an empty object but it feels odd to even instantiate diffing for that. Is it just to activate a trickle-effect or? Someone that can shed more light on the use case of: https://github.com/developit/preact/issues/1322 ?

JoviDeCroock:
Hmm again errors after updating the branch, I'll leave this feature as it is until we get back to it.

JoviDeCroock:
So apparently something changed and when it bubbles it's considered equal now, this gets fixed by changing this line `(possibleVNode._dom!=null || possibleVNode._component!=null)` to not take _component in account.

PR #1540, opened 2019-04-14
Fix setting % width and height on img

### LINKED ISSUE #492: simple property assignment fails to set img width

marvinhagemeister:
Fixes #492 .
Adds `+13 B`.

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -81,7 +81,7 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 		}
 		(dom._listeners || (dom._listeners = {}))[name] = value;
 	}
-	else if (name!=='list' && name!=='tagName' && !isSvg && (name in dom)) {
+	else if (name!=='list' && name!=='tagName' && name!=='width' && name!=='height' && !isSvg && (name in dom)) {
 		dom[name] = value==null ? '' : value;
 	}
 	else if (value==null || value===false) {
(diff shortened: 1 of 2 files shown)


robertknight:
The [HTML spec](https://html.spec.whatwg.org/multipage/embedded-content-other.html#attr-dim-width) states that the `width` and `height` properties must be pixels and [Chrome code](https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/html/html_element.cc?sq=package:chromium&g=0&l=1179) clearly indicates that this is the case.

I wonder if we could simply say that percentage values for these attributes are not supported and that this needs to be done with inline styles instead, or is the difference to React behaviour big enough of a deal to justify the exception? I don't have a strong opinion, just wondering how compat vs size trade-offs get weighed up.

marvinhagemeister:
You're right. It's not worth the cost to add non-spec behaviour. Thanks for the hint, should have checked the spec myself before making the PR.

PR #1542, opened 2019-04-17
Add _depth to mangle.json

38elements:
This adds `_depth` to `mangle.json`.

https://github.com/developit/preact/blob/1cddbec5be4d11938ba0f430604edf2e2aeddd0d/src/diff/index.js#L151-L153

### DIFF
--- a/mangle.json
+++ b/mangle.json
@@ -17,6 +17,7 @@
   "props": {
     "cname": 6,
     "props": {
+      "$_depth": "__D",
       "$_dirty": "__d",
       "$_nextState": "__s",
       "$_renderCallbacks": "__h",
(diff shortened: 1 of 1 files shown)


JoviDeCroock:
This seems to add 2 bytes to core

38elements:
Thank you for comment.
I thought this change is necessary to keep consistent property name.
I think this change is unnecessary.

marvinhagemeister:
@38elements Thanks for the PR, you're right that we should add it to our `mangle` file to avoid any potential name clashes of class properties :+1: 

@JoviDeCroock it probably adds 2 bytes because the name is a bit longer than a single char.

38elements:
Thank you for comment.
Since Changing `__b` is +1B, `__b` is better than `__D`.

PR #1548, opened 2019-04-17
(chore) - remove focusElement

JoviDeCroock:
Adding items in react: https://codesandbox.io/s/748j0p3jpj
Swapping in react: https://codesandbox.io/s/olq48z68r5

So the first we should support, which also feels obviosu to me personally but the second not so much.

The odd thing is that it only  fails when you insert a second thing above

So add one before --> works
Add one after --> works
Add one after previous --> works
Add one after the first step --> fails

Size of core stays the same
Seems to give a significant performance increase

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -44,7 +44,6 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 		if (excessDomChildren!=null) {
 			for (i = 0; oldDom==null && i < excessDomChildren.length; i++) {
 				oldDom = excessDomChildren[i];
-				oldChild = excessDomChildren[i];
 			}
 		}
 		else {
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -25,7 +25,7 @@ import { removeNode } from '../util';
  */
 export function diffChildren(parentDom, newParentVNode, oldParentVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, oldDom) {
 	let childVNode, i, j, p, index, oldVNode, newDom,
-		nextDom, sibDom, focus;
+		nextDom, sibDom;
 
 	let newChildren = newParentVNode._children || toChildArray(newParentVNode.props.children, newParentVNode._children=[], coerceToVNode, true);
 	let oldChildren = oldParentVNode!=null && oldParentVNode!=EMPTY_OBJ && oldParentVNode._children || EMPTY_ARR;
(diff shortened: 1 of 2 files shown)


JoviDeCroock:
Don't take the Passing as it succeeds I have adjusted this in tests:
https://github.com/developit/preact/pull/1548/files#diff-4860f971b6d2f661624c4d35fbba8cc3L106

Now it effectively tests with keys.

Without keys it loses focus when the last insertion happens so insert 0 before 1.
Then we treat input like a new element, logging output:

```
LOG: 'parent before', <div><span>0</span><input type="text"><span>1</span><span>3</span></div>, <input type="text">, <span>3</span>
LOG: 'parent append', <span></span>, 3
LOG: 'parent append', <div><span>0</span><span>1</span><input type="text"><span>2</span></div>, <span>3</span>
```

'parent before' is first the existing parentDom and then adding a newDom before given oldDom
'parent append' adds the second argument to the first

src/diff/children.js
```diff
@@ -42,6 +43,7 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 			for (i = 0; i < excessDomChildren.length; i++) {
 				if (excessDomChildren[i]!=null) {
 					oldDom = excessDomChildren[i];
+					oldChild = excessDomChildren[i];
```
JoviDeCroock: This can be removed.

test/browser/focus.test.js
```diff
@@ -54,7 +54,7 @@ describe('focus', () => {
 		teardown(scratch);
 	});
 
-	it('should maintain focus when swapping elements', () => {
+	it.skip('should maintain focus when swapping elements', () => {
```
ForsakenHarmony: should this be skipped?

JoviDeCroock: Yes, Swapping in react: https://codesandbox.io/s/olq48z68r5 this is not supported in most frameworks and can lead to unwanted behavior imo

marvinhagemeister: @JoviDeCroock The linked sandbox just appends new `h1` elements for me

JoviDeCroock: It swaps the input from before to after the h1 and that loses focus on the input or am I misunderstanding it? @marvinhagemeister

ForsakenHarmony: wait am I missing something there? what does it swap?

JoviDeCroock: it swaps the input to bbefore the H1 same behavior as the test that is skipped?

ForsakenHarmony: The h1 stays?

marvinhagemeister: Maybe you forgot to save the sandbox. Re-created it where the elements are swapped all the time: https://codesandbox.io/s/0yo8kxqjzn . Can confirm that `react` loses focus there, too 👍

ForsakenHarmony:
🎉

Issue #1068, opened 2018-04-20
[Chrome devtool] User Timing api with Preact


Mathieu-R:
React 16 uses `user timing api` to provide relevants informations in the `performance` panel of `chrome devtool`.

Is there any way to use it with `Preact` ? When I record a performance trace, I do not see the user timing panel.

Thanks for the help.

developit:
That would be something we'd need to add to `preact/debug`.

lorenzomigliorero:
+1

PR #1549, opened 2019-04-18
Update devtools bridge for devtools v4

### LINKED ISSUE #1381: Display hooks state in devtools

marvinhagemeister:
This PR will update our devtools adapter for the next major release of the [React-devtools extension](https://github.com/bvaughn/react-devtools-experimental).

There is no need to review it for now, as it's under heavy development and I'm still poking the devtools to see how things work.

What works so far:

- [x] Get new devtools panel to show up
- [x] Display correct bundle type
- [x] Display mounted components
- [x] Properly encode `displayName`
- [x] Display `state`
- [x] Display `props`
- [x] Pretty format complex values
- [x] Child reordering
- [x] Updating
- [x] Unmounting
- [x] Suspense
- [x] DOM Selection
- [x] DOM Inspection
- [x] Override Props
- [x] Override State
- [x] Profile + Reload - blocked by https://github.com/bvaughn/react-devtools-experimental/pull/342
- [x] Profiler
- [x] Display hooks state
- [x] Edit `useState` value
- [x] Display recursive hooks
- [x] Support `useDebugValue`
- [x] Display Context
- [x] Clear temporary variables
- [x] Display owner stack (via parents)
- [ ] Use true owners instead of parents
- [x] Owner tree view
- [x] Component Filtering
- [x] "What caused a render" panel
- [x] Add back support for legacy devtools by switching the adapter based on a check if the new `__REACT_DEVTOOLS_ATTACH__` or `__REACT_DEVTOOLS_COMPONENT_FILTERS__` variables are present

**Bugs**

- [x] Properly detect changes in hooks (see `didHooksChange()`)
- [ ] On multiple profiling runs the parents may have a shorter duration than the children (check how react deals with this)
- [x] Late binding of extension ~~(wrap `connect()` in a `setInterval` until the extension is loaded~~ - found a better solution by using queues and flushing them once the extension is connected.
- [x] Filters sometimes not applied at boot time (not sure how to reproduce)
- [x] Find a better way to deal with Profiler state. It's currently a bit confusing
- [x] Tests, tests, tests (obviously)
- [x] Selection is restored wrongly when devtools connects later

Fixes #1381


ForsakenHarmony:
#1068

might be worth considering

marvinhagemeister:
Unflagged some todo items because the devtools API has changed a bit and some features are broken in this PR because of that. Nothing too grave, just need the time to fix them one by one.

marvinhagemeister:
Updating and child reordering is pretty stable so far. It somehow trips up with `Suspense + lazy` but I'll probably find a fix for that.

marvinhagemeister:
Alright, @sventschui poked me a bit about `Suspense` support so I added that too. Now I really need to work on backporting all this to 8.x

ForsakenHarmony:
Not even sure if anyone can really review this given the knowledge you aquired 🙃

debug/src/devtools/hooks.js
```diff
@@ -0,0 +1,181 @@
+/* istanbul ignore file */
+import { options } from 'preact';
+import { USE_STATE, USE_CALLBACK } from 'preact/hooks/constants';
```
calebeby: This import is failing for me :thinking: 

Should it be `preact/hooks/src/constants` or `preact/hooks/dist/constants`?

calebeby: Also it's unclear to me why this is a separate import from the one below :point_down:

ForsakenHarmony: probably

calebeby:
@marvinhagemeister I'm seeing this error when I test out this branch with a very simple test:

```js
const App = () => {
  return <div>Hi</div>
}
```

screenshot

screenshot

debug/src/devtools/hooks.js
marvinhagemeister: Good catch, that was just an oversight and it should be merged with the other import :+1:

marvinhagemeister:
@calebeby Are you sure that you're checking the code in this branch? The code in the screenshot shows the legacy bridge which is not used anymore with this PR :thinking:

marvinhagemeister:
@ForsakenHarmony very true. I also feel somewhat guilty as this PR is much larger than I'd hoped.

Seeing that I can't find any further issues anymore in our demo app, plus the fact that the devtools are currently broken for users, I'll probably go ahead and merge this.

sventschui:
Pretty sure @marvinhagemeister did a rock solid job on this PR. I guess the only possibly helpful review before releasing this in a new RC is if anybody with a bigger preact app could try out this PR. Unfortunately I do not have a preact codebase at hand. The Suspense issue is still preventing us to migrate :/

calebeby:
I'll be able to re-test this in ~3 hours. I'm pretty sure I was testing on this branch because I was seeing the import error until I fixed it locally

calebeby:
Actually here's the repo if anyone else wants to try out the new devtools on it: https://github.com/Pigmice2733/peregrine-frontend

marvinhagemeister:
@calebeby The import error is fixed in this PR, thanks for pointing it out. I simply missed it :slightly_smiling_face:

calebeby:
> Are you sure that you're checking the code in this branch?

> The import error is fixed in this PR,

I know it's fixed in this pr. I'm just saying that when I was testing it yesterday, I ran into the import error before you had fixed it, so then I fixed it locally, but then I ran into the error in my screenshot. So I'm pretty sure I was on this branch

marvinhagemeister:
@calebeby Thanks for the repo link, it makes everything so much easier :+1: I'll have a look :slightly_smiling_face:

calebeby:
In case anyone is wondering the status of this, @marvinhagemeister posted a blog post about this: https://marvinh.dev/blog/state-of-devtools/

marvinhagemeister:
This PR is not necessary anymore as we now have our own extension 🎉 

See: https://preactjs.github.io/preact-devtools/

PR #1552, opened 2019-04-19
(chore) - tree-shakeable preact/utils

JoviDeCroock:
Provides a solution for: https://github.com/developit/preact/issues/1529

I'm going to make a dedicated test-suite for preact/src then because now we only test if it works in /compat

Another way of going about this could be to make the side-effects in compat part of createElement. I still think that webpack would have issues tree-shaking it though. Not quite sure.

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -259,7 +259,6 @@ export {
 	render,
 	render as hydrate,
 	unmountComponentAtNode,
-	createPortal,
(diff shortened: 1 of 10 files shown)


JoviDeCroock:
Removed forwardRef since this relies on an internal in memo where we replace vnode.ref --> vnode.props.ref.

PR #1557, opened 2019-04-21
(feat) - replaceNode parameter

LukasBombach:
This implementation brings back the third _optional_ parameter `replaceNode` from `8.x`

> `render(component, containerNode, [replaceNode])`

https://preactjs.com/guide/api-reference#preact-render-

With this parameter you can provide child node in your parent DOM that will be used as render root. It does not replace other children inside the parent and it will hydrate the provided node and not just replace it, the DOM nodes remain the same.

I made a more comprehensive infographic to describe why I see this as useful

https://gist.github.com/LukasBombach/884319d5430a3fb85f3b4385d7a31c89

You can try out a working usage in this repo: https://github.com/spring-media/pool-attendant (See [index.partial.hydration.js](https://github.com/spring-media/pool-attendant/blob/master/packages/app/src/index.partial.hydration.js))

ADDS: 16B to core

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -10,13 +10,27 @@ import options from './options';
  * @param {import('./internal').PreactElement} parentDom The DOM element to
  * render into
  */
-export function render(vnode, parentDom) {
+export function render(vnode, parentDom, replaceNode) {
 	if (options.root) options.root(vnode, parentDom);
 	let oldVNode = parentDom._prevVNode;
 	vnode = createElement(Fragment, null, [vnode]);
 
 	let mounts = [];
-	diffChildren(parentDom, parentDom._prevVNode = vnode, oldVNode, EMPTY_OBJ, parentDom.ownerSVGElement!==undefined, oldVNode ? null : EMPTY_ARR.slice.call(parentDom.childNodes), mounts, vnode, EMPTY_OBJ);
+	diffChildren(
+		parentDom,
+		replaceNode ? vnode : (parentDom._prevVNode = vnode),
+		replaceNode ? undefined : oldVNode,
+		EMPTY_OBJ,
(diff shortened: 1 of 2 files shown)


src/render.js
```diff
@@ -10,13 +10,27 @@ import options from './options';
  * @param {import('./internal').PreactElement} parentDom The DOM element to
```
JoviDeCroock: Hey, nice work one little nitty remark. Could you add the JSDoc types for `replaceNode`

LukasBombach: Ahh right. Done. I used the comment from `8.x`

test/browser/render.test.js
```diff
@@ -966,4 +966,44 @@ describe('render()', () => {
...
+			expect(childA.innerHTML).to.equal('contents');
+		});
+
+		it('should not remove siblings of replaceNode', () => {
+			const childA = scratch.querySelector('#a');
+			render(<div id="a" />, scratch, childA);
+			expect(scratch.innerHTML).to.equal('<div id="a"></div><div id="b"></div><div id="c"></div>');
+		});
+		
+		it('should render multiple render roots in one parentDom', () => {
```
marvinhagemeister: This test is so good :+1: :100:

marvinhagemeister:
@LukasBombach Congrats on getting your first PR merged :tada:

test/browser/render.test.js
LukasBombach: ❤️

LukasBombach:
Thank you so much! I am really happy and proud! And this feature will be so useful to us!

Thank you!

prateekbh:
@developit this should fix our problem with SSR in preact-cli, right?

PR #1568, opened 2019-04-24
(fix) - prevent crash when there is no last DOM child node

JoviDeCroock:
Fixes: https://github.com/developit/preact/issues/1567

Adds 1B to core

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -60,7 +60,7 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
 
 				// If the last child is a Fragment, use _lastDomChild, else use _dom
 				p = newVNode._children[newVNode._children.length - 1];
-				newVNode._lastDomChild = p._lastDomChild || p._dom;
+				newVNode._lastDomChild = p && (p._lastDomChild || p._dom);
 			}
 		}
 		else if (typeof newType==='function') {
--- a/test/browser/fragments.test.js
+++ b/test/browser/fragments.test.js
@@ -98,7 +98,7 @@ describe('Fragment', () => {
 		expect(scratch.innerHTML).to.equal('hello <span>world</span>');
 	});
 
-	it('should not crash with null as last child', () => {
+	it.only('should not crash with null as last child', () => {
 		let fn = () => {
 			render((
 				<Fragment>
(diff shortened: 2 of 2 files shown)


src/diff/index.js
```diff
@@ -60,7 +60,7 @@ export function diff(dom, parentDom, newVNode, oldVNode, context, isSvg, excessD
 
 				// If the last child is a Fragment, use _lastDomChild, else use _dom
 				p = newVNode._children[newVNode._children.length - 1];
-				newVNode._lastDomChild = p._lastDomChild || p._dom;
+				newVNode._lastDomChild = p && (p._lastDomChild || p._dom);
```
cristianbote: Would this mean that, if my last vnode is falsy, it's gonna get skipped?

JoviDeCroock: We are targetting a vnode so only vnodes that evaluate to null are skipped. This is introduced in the `holes` PR. So '' --> gets coerced to vnode but null can be kept that's why this crashes null._lastDomChild or null._dom

cristianbote: Thank you 🙂

JoviDeCroock: Thanks for this heads up, reminds me that I should always elaborate in my PR description instead of assuming that people also debugged it (which is a really strange thinking process I sometimes have)

cristianbote: No worries 👍! I would've checked my assumption, first, but I was on my phone 😟

robertknight:
Can someone explain how `_lastDomChild` is used when diffing fragments?

My expectation just from the name is that it would refer to "the last child of a fragment which is a DOM node" rather than "the last child of a fragment _if_ it is a DOM node". Is that correct? Will the result of the diff still be correct if there _is_ a last DOM child of the fragment, but it isn't the last child. For example:

```js
<Fragment>
  <span/>
  {null}
</Fragment>
```

In this case there _is_ a DOM child, it just isn't `children[children.length-1]`. Does that matter?

JoviDeCroock:
@robertknight afaik `_lastDomChild` is only populated on Fragment VNodes so if the last child is a Fragment we actually point to the lastChild of that Fragment.

In the holes PR the support was added for arrays including null values, to prevent skipping and bad reuse. These nulls are practically placeholders (https://github.com/developit/preact/pull/1440).

marvinhagemeister:
yup, before the holes PR the `_lastDomChild`-property always referred to the last child of a `Fragment`. If there is only a single child present it will be the same as `_dom`. With the holes PR a child can either be `null` or a DOM node.

For diffing we need to have the right reference, so I'm wondering if setting `_lastDomChild` to `null` even though the `Fragment` has children could lead to errors.

JoviDeCroock:
@marvinhagemeister I tried breaking it in tests, if you have other ones feel free to point me in the right direction.

PR #1572, opened 2019-04-24
(fix) - Error boundary not applying in array cases

JoviDeCroock:
In cases where a series of components would throw the ErrorBoundary would get skipped since `_processingException` would be filled. This check made for it to be skipped and thus thrown resulting in faulty behavior.

This change seems to break testing in some way... :/ wondering if this could be something to do with cleaning up tests or related?

Tests seem to run really inconsistent, anyone has any clue as to why this could happen...

Like I know it probably loops or something somewhere but I can't tell where since I have no error output, I'm also having difficulties seeing how _processingException is bypassing that.

Fixes: https://github.com/developit/preact/issues/1570
~~Removes 23B from core~~
Adds 8B to core


JoviDeCroock:
Okay.... This is extremely weird tests succeed here but locally I get:

```
FAILED TESTS:
  performance
    × should rerender without changes fast
      HeadlessChrome 73.0.3683 (Windows 10.0.0)
    Error: Uncaught AssertionError: Inner after componentWillUnmount: expected <div id="InnerDiv"></div> to not exist (node_modules/chai/chai.js:239)
```

JoviDeCroock:
All checks are successfull in travis, but the performance things are still failing locally with `yarn test:karma:watch` so don't merge!

Issue at hand if anyone knows more:

```
FAILED TESTS:
  performance
    × should rerender without changes fast
      HeadlessChrome 73.0.3683 (Windows 10.0.0)
    Error: Uncaught AssertionError: Inner after componentWillUnmount: expected <div id="InnerDiv"></div> to not exist (node_modules/chai/chai.js:239)
```

No clue how to debug this perf build at this moment.

JoviDeCroock:
Okay, I think all is fixed now! :D

src/internal.d.ts
```diff
@@ -58,6 +58,7 @@ export interface Component<P = {}, S = {}> extends preact.Component<P, S> {
 	_prevVNode?: VNode | null;
 	_ancestorComponent?: Component<any, any>;
 	_processingException?: Component<any, any> | null;
+	_pendingError?: Component<any, any> | null;
```
marvinhagemeister: Nit: We need to add this to our mangling properties to avoid potential name clashes.

JoviDeCroock: Does it matter what name I give it? Was wondering that since I see clashing names in mangle and that worries me to pick one.

JoviDeCroock: Mangling this to __E actually adds 3B

developit: The default is to mangle `_` properties to single-letter variables, so that makes sense. However, we want to avoid single-letter property names since they're likely to collide.

PR #1573, opened 2019-04-25
Add support for multiple select values in core

### LINKED ISSUE #1562: Move multiple select from compat to core
In #1481 we proposed using the `value` attribute of a `select` element instead of setting the `selected` value on all options manually. But this means that setting multiple selected values is not possible. How should we solve this? In compat we added support for passing an array to `value` similar to how react handles it.

marvinhagemeister:
Fixes #1562 .
Adds `+40 B` :S


andrewiggins:
In the spirit of being "closer to the metal",. I think I'd prefer this stay in compat. If I understand correctly, the DOM doesn't support arrays for value attr of select so I don't think we should in core.

Just my 2 cents.

PR #1578, opened 2019-04-25
Reduce size by golfing down some pieces 🏌️‍♂️ (-39 B)

andrewiggins:
Summary:
9f6f0fa Move all component diff'ing logic into component diff condition (-6 B)
b547990 Remove `dom` parameter from `diff()` (-11 B)
8928b2f Golf createContext (-6 B)
1f7002b Remove unnecessary code from `diffChildren` (-16 B)
~~1baa6bb Simplify originalDom check (-4 B)~~ (broke some hydration tests)

~~Total: -23 B~~
~~Total: -43 B~~
Total: -39 B

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -242,7 +242,7 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
 	}
 
 	if (newVNode.type===null) {
-		if (dom===originalDom && newVNode.text!==oldVNode.text) {
+		if ((originalDom==null || dom===originalDom) && newVNode.text!==oldVNode.text) {
 			dom.data = newVNode.text;
 		}
 	}
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -242,7 +242,7 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
 	}
 
 	if (newVNode.type===null) {
-		if ((originalDom==null || dom===originalDom) && newVNode.text!==oldVNode.text) {
+		if (dom===originalDom && newVNode.text!==oldVNode.text) {
 			dom.data = newVNode.text;
 		}
 	}
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -8,8 +8,6 @@ import options from '../options';
 
 /**
  * Diff two virtual nodes and apply proper changes to the DOM
- * @param {import('../internal').PreactElement | Text} dom The DOM element representing
- * the virtual nodes under diff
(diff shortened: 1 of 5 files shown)


src/diff/index.js
```diff
@@ -228,7 +219,7 @@ export function commitRoot(mounts, root) {
  * @returns {import('../internal').PreactElement}
  */
 function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent) {
-	let d = dom;
+	let originalDom = dom;
```
marvinhagemeister: +1 for the new name :+1:

src/diff/index.js
```diff
@@ -216,8 +207,6 @@ export function commitRoot(mounts, root) {
 
 /**
  * Diff two virtual nodes representing DOM element
- * @param {import('../internal').PreactElement} dom The DOM element representing
- * the virtual nodes being diffed
```
filoozom: Shouldn't this one stay?

marvinhagemeister: Good catch! Yup, this should stay 👍

src/diff/index.js
```diff
@@ -252,10 +241,9 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
 		// we created a new parent, so none of the previously attached children can be reused:
 		excessDomChildren = null;
 	}
-	newVNode._dom = dom;
 
 	if (newVNode.type===null) {
-		if ((d===null || dom===d) && newVNode.text!==oldVNode.text) {
+		if ((originalDom==null || dom===originalDom) && newVNode.text!==oldVNode.text) {
```
developit: since `dom` can never be `null` here, checking `dom===originalDom` implies `originalDom` is also non-null, so we can drop that condition entirely:

```suggestion
		if (dom===originalDom && newVNode.text!==oldVNode.text) {
```

src/diff/index.js
andrewiggins: Whoops! Good catch, thanks!

marvinhagemeister:
The last commit broke two tests regarding `hydration` behaviour in compat. They test that text nodes are changed on mismatch during hydration. That behaviour is listed in the docs: https://reactjs.org/docs/react-dom.html#hydrate . If we can't find a way around that I'm more than happy to land this PR in `master` without the last commit :)

andrewiggins:
I'm looking at it now

andrewiggins:
Reverted the commit and merged in master. Travis should be green now.

marvinhagemeister:
FYI: Just rebased the PR against `master`

PR #1579, opened 2019-04-26
Better Integration for TypeScript + Preact + Unistore

yokotaso:
I write PoC program with TypeScript, Preact, Unistore(Preact & Unistore is great!)

I happened TypeScript error. Because type definition of onClick is like below(in jsx.d.ts):
```typescript
onClick?: MouseEventHandler;
type MouseEventHandler = EventHandler<MouseEvent>;
interface EventHandler<E extends Event> {
		(event: E): void;
}
```
So onClick accepts function that has 1 argument, and the type is MouseEvent.
But Integrate with unistore, first argument of function is State object.

TypeScript compile error:
jsxerror

Sample code is here: https://github.com/yokotaso/preact-unistore-typescript-integration

I suggest to add UnistoreEventHandler for unistore.

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -273,24 +273,30 @@ export namespace JSXInternal {
 	interface PathAttributes {
 		d: string;
 	}
+	
+	type EventHandlers<E extends Event> = EventHandler<E> | UnistoreEventHandler<any, E>;
 
 	interface EventHandler<E extends Event> {
 		(event: E): void;
 	}
 
-	type ClipboardEventHandler = EventHandler<ClipboardEvent>;
-	type CompositionEventHandler = EventHandler<CompositionEvent>;
-	type DragEventHandler = EventHandler<DragEvent>;
-	type FocusEventHandler = EventHandler<FocusEvent>;
-	type KeyboardEventHandler = EventHandler<KeyboardEvent>;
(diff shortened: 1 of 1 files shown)


src/jsx.d.ts
```diff
@@ -273,24 +273,30 @@ export namespace JSXInternal {
 	interface PathAttributes {
 		d: string;
 	}
+	
+	type EventHandlers<E extends Event> = EventHandler<E> | UnistoreEventHandler<any, E>;
```
marvinhagemeister: Thanks for your PR 🎉 I'm honestly uneasy with adding library specific code to `preact`. If we start with it soon others will follow and we'll be forced to maintain it. Maybe there is a way to fix this in the typings for `unistore`? That would be a lot cleaner 🙂

ForsakenHarmony:
As Marvin said, this is the wrong place, we can't add library specific typings

One solution to the problem would be to use `props.increment.bind(null, null)`

But your typings for the props are incorrect as well, the function doesn't take the state as an argument, that gets added by unistore

the correct type would be `increment(): void;`, as it doesn't return the update either

yokotaso:
@marvinhagemeister @ForsakenHarmony 
Thank you. I readlized that I misunderstood type of unistore

PR #1580, opened 2019-04-26
Code golf diffChildren size down by 33 B ⛳️

jviide:
This pull request reduces the gzipped code size by 33 bytes by progressively refactoring `diffChildren`. The commits are roughly in an increasing order of weirdness, so just some of them can be cherry-picked if needed.

 * 65dfda6, d676eba Do pretty run-of-the mill refactoring to avoid comparisons (-9 B)
 * 5a9ad93 Avoids two `break` statements (-7 B)
 * f9bc8c5 Refactors out local variables `p` and `index`, with hopefully sufficient comments explaining the logic (-13 B)
 * 7b3606d Compresses code by assuming that `EMPTY_OBJ._children` is `undefined` (-4 B)

These changes conflict a bit with @andrewiggins's #1578, sorry about that.

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -94,7 +94,7 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 			nextDom = oldDom!=null && oldDom.nextSibling;
 
 			// Morph the old element into the new one, but don't append it to the dom yet
-			newDom = diff(oldVNode==null ? null : oldVNode._dom, parentDom, childVNode, oldVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, null, oldDom);
+			newDom = diff(oldVNode && oldVNode._dom, parentDom, childVNode, oldVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, null, oldDom);
 
 			// Only proceed if the vnode has not been unmounted by `diff()` above.
 			if (newDom!=null) {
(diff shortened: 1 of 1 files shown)


src/diff/children.js
```diff
@@ -24,11 +24,13 @@ import { removeNode } from '../util';
...
 export function diffChildren(parentDom, newParentVNode, oldParentVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, oldDom) {
-	let childVNode, i, j, p, index, oldVNode, newDom,
+	let childVNode, i, j, oldVNode, newDom,
 		nextDom, sibDom, focus;
 
 	let newChildren = newParentVNode._children || toChildArray(newParentVNode.props.children, newParentVNode._children=[], coerceToVNode, true);
-	let oldChildren = oldParentVNode!=null && oldParentVNode!=EMPTY_OBJ && oldParentVNode._children || EMPTY_ARR;
+	// This is a compression of oldParentVNode!=null && oldParentVNode != EMPTY_OBJ && oldParentVNode._children || EMPTY_ARR
+	// as EMPTY_OBJ._children should be `undefined`.
+	let oldChildren = oldParentVNode!=null && oldParentVNode._children || EMPTY_ARR;
```
developit: This seems safe to me. My guess is the original checks were a combination of before + after `oldParentVNode` became reliably a `(vnode | null)`.

PR #1581, opened 2019-04-27
Update type definition

38elements:
This updates type definition.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -117,7 +117,7 @@ declare namespace preact {
 		state: Readonly<S>;
 		props: RenderableProps<P>;
 		context: any;
-		base?: HTMLElement;
+		base?: Element | Text;
 
 		// From https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e836acc75a78cf0655b5dfdbe81d69fdd4d8a252/types/react/index.d.ts#L402
 		// // We MUST keep setState() as a unified signature because it allows proper checking of the method return type.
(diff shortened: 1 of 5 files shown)


src/index.d.ts
```diff
@@ -160,8 +160,12 @@ declare namespace preact {
 	// Preact render
 	// -----------------------------------
 
-	function render(vnode: ComponentChild, parent: Element | Document | ShadowRoot | DocumentFragment): void
-	function hydrate(vnode: ComponentChild, parent: Element | Document | ShadowRoot | DocumentFragment): void
+	function render(
```
38elements: https://github.com/developit/preact/blob/5cac157a6ce351246c6b57dfb0c788f0209abc88/src/render.js#L14-L16

src/index.d.ts
```diff
@@ -160,8 +160,12 @@ declare namespace preact {
...
+		vnode: ComponentChild,
+		parent: Element | Document | ShadowRoot | DocumentFragment,
+		replaceNode?: Element | Text
```
38elements: https://github.com/developit/preact/blob/5cac157a6ce351246c6b57dfb0c788f0209abc88/src/render.js#L14-L16

src/internal.d.ts
```diff
@@ -25,6 +25,7 @@ export interface PreactElement extends HTMLElement {
 export interface VNode<P = {}> extends preact.VNode<P> {
 	// Redefine type here using our internal ComponentFactory type
 	type: string | ComponentFactory<P> | null;
+	_self: this;
```
38elements: https://github.com/developit/preact/blob/5cac157a6ce351246c6b57dfb0c788f0209abc88/src/create-element.js#L56-L67

PR #1590, opened 2019-05-02
(fix) Add support for global event handlers like animation and transition

### LINKED ISSUE #1589: onAnimationEnd event not firing
Perhaps a regression with preact 10, or perhaps intended behavior?

Prior to 10 `onAnimationEnd` would fire the end event when an animation ends... which no longer happens. 

Test case of it failing to alert here https://codesandbox.io/s/71p06y27xj ~ temp (or real) fix was to use lowercase onanimationend

cristianbote:
Instead for looking into the `dom` reference widen the lookup on the window instead. That means, supporting Global Event Handlers as well. Like `animation` or `transition`.

Fixes #1589 
~~Adds +5B~~
Adds +3B

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -73,7 +73,7 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 	else if (name[0]==='o' && name[1]==='n') {
 		let useCapture = name !== (name=name.replace(/Capture$/, ''));
 		let nameLower = name.toLowerCase();
-		name = (nameLower in dom ? nameLower : name).substring(2);
+		name = (nameLower in window ? nameLower : name).substring(2);
 
 		if (value) {
 			if (!oldValue) dom.addEventListener(name, eventProxy, useCapture);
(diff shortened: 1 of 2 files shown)


src/diff/props.js
```diff
@@ -73,7 +73,7 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 	else if (name[0]==='o' && name[1]==='n') {
 		let useCapture = name !== (name=name.replace(/Capture$/, ''));
 		let nameLower = name.toLowerCase();
-		name = (nameLower in dom ? nameLower : name).substring(2);
+		name = (nameLower in window ? nameLower : name).substring(2);
```
marvinhagemeister: Does this work for Custom Elements? iirc we're laking test cases for them. /cc @andrewiggins @developit

developit: It does! This is better than what I had here.

One nit:  this breaks in Workers. Maybe we should replace `window` with `self`?

```suggestion
		name = (nameLower in self ? nameLower : name).substring(2);
```

developit: Another option we could consider is using a sigil:

```js
<div onAnimationEnd={..} />  // gets lowercase'd
<div on-animationEnd={..} />  // addEventListener('animationEnd', ..)
```

Thoughts?

cristianbote: Uh, nice suggestion! :) using `self` is a tad more _decoupled_ 😄

Not sure about the sigil though. It seems that custom elements does/should support events with dash in them, so one could actually _name_ the event with sigil as well, and this worries me that would lead to much more confusion than clarity? It's another different char than dash, but looks like a dash? 😄

PR #1592, opened 2019-05-02
Debug component thrashing

developit:
Proposed Error detail page on preactjs.com _(hit edit on this PR for source)_:

```
---
url: /errors/1
title: Preact Error 1: Component Thrashing
---
```

## Error 1: Component Thrashing

This problem occurs when components are redefined every time Preact renders. 

This means that the component and all of its descendants will be destroyed and recreated on every render.

### Background

Virtual DOM works by building up a new description of your application's structure every time something changes. To make this work with acceptable performance, Virtual DOM libraries implement a diffing algorithm that compares this new structure to the previous one (the structure of the previous render).

The first step of this process is to go through the tree of nodes created by your JSX - called VNodes - and match them up with the previous ones. In order to do this, it must enforce a definition of what makes two equivalent. This is done by comparing the VNode's type:

```js
<meter value="50">home</a>
  ^-type  ^-props  ^-children
type === "meter"

<Avatar user="1" />
  ^-type  ^-props
type === Avatar
```

Two VNodes are considered equal when their types are the same. When a VNode's type changes from one render to the next, any components and DOM that has been created for that VNode and all of its descendants must be destroyed in order to ensure output doesn't "bleed" across renders.

### The Problem

Now think about how where we define components. When a component (class or function) is defined inside of a function, each time that outer function is called we'll get a new copy of that component:

```js
function outer() {
  const Avatar = ({ src }) => <img src={src} />

  return Avatar
}

outer() !== outer()  // a new Avatar function is created every time
```

This is the same as what the Virtual DOM library sees when looking at the JSX returned from your render functions or functional components. Take a look at what happens when we define Avatar from inside of another component's render:

```js
function Profile(props) {
    const Avatar = ({ src }) => <img src={src} />

    return <Avatar src={props.src} />
}

const render1 = Profile({ src: '/1.gif' })
const render2 = Profile({ src: '/1.gif' })

render1.type !== render2.type
// They're not equal!
// This means we have to destroy and recreate the image.
```

This creation of new components on each render is how "Component Thrashing" comes to be.

## How to Fix Component Thrashing

Let's take our Profile example from above:

```js
function Profile(props) {
    const Avatar = ({ src }) => <img src={src} />   // 🚨

    return <Avatar src={props.src} />
}
```

We now know that `Avatar` is going to be created every time `Profile` is called by Preact.

To fix this, move the component's definition out of the outer function:

```js
function Profile(props) {
    return <Avatar src={props.src} />
}

const Avatar = ({ src }) => <img src={src} />
```

Now, every time `Profile()` is called, it will return the same reference to `Avatar`! It's generally best to write nested components this way so they can't rely on scoped variables, as this makes them "pure".

This will correct the issue and allow Preact to diff properly once again.


debug/src/debug.js
```diff
@@ -98,9 +116,18 @@ export function initDebug() {
...
 
-		if (oldBeforeDiff) oldBeforeDiff(vnode);
+		if (oldBeforeDiff) oldBeforeDiff(vnode, oldVnode);
 	};
 
+	// If the constructor source text and prototype properties match,
+	// two different component references are likely to be the same.
+	function componentsAreEquivalent(a, b) {
+		return (
+			Function.prototype.toString.call(a) === Function.prototype.toString.call(b) &&
```
robertknight: I did run into an issue in the past where `Function.prototype.toString` was slow (relative to Chrome) in [old versions of Firefox](https://github.com/angular/angular.js/issues/14268). I believe this has been resolved a while back, I'm just flagging it as a possible issue with this approach.

developit: Yeah I was iffy about this one. Also it'll break when browsers start to roll out toString censoring for functions, and JS needs the optimizations that'll unlock.

We could cache it:

```
const FNSTR = new WeakMap();
function toString(fn) {
  let str = FNSTR.get(fn);
  if (!str) {
    FNSTR.set(fn, str = Function.prototype.toString.call(fn));
  }
  return str;
}
```

Other than that, I'm not sure there's another way around this... the premise of this error is that we get a new reference to something that is roughly the same every render.

Maybe another approach would be to monitor component diffing (by WeakMap of constructors) and check if any of the children contain in-place type changes?

```
const shape = ctor => ctor.prototype && Object.keys(ctor.prototype).join();
for (let i=0; i<vnode._children.length; i++) {
  let v = vnode._children[i];
  let old = oldVnode._children[i];
  if (typeof v.type==='function' && typeof old.type==='function' && v.type!==old.type) {
    if (shape(v.type)==shape(v.type)) {
      // warn
    }
  }
}
```

developit: @robertknight @marvinhagemeister actually come to think of it, we could probably justify using the same caching technique for `serializeVNode()`, especially if we had it look up the source via `vnode._original` and cache based on that.

kristoferbaxter: Like this idea better in the long run, but it doesn't necessarily need to block shipping this feature as it stands.

kristoferbaxter (approved):

JoviDeCroock (approved):
LGTM :D it's really following the idea of the project to be perf-oriented so can't be any more supportive of this

debug/src/debug.js
marvinhagemeister: Shall we ship this PR as is and later do the `WeakMap` optimization?

marvinhagemeister (approved):

developit:
Rebased, ready for re-review + merge.

andrewiggins (approved):
Let's roll! 🚲

developit:
took a stab at adding caching in WeakMap-supporting browsers.

cristianbote (approved):
🎉 let's ship it! 😄

developit:
Also fixed an issue where `options._diff` wasn't chained for `null` vnodes.

Thompson1985 (approved):

JoviDeCroock:
Closing this as stale

PR #1593, opened 2019-05-02
Introduce Suspense and lazy

sventschui:
This introduces Suspense and lazy. There are some open things tough.

**EDIT**: After the excellent feedback on this PR things seem to be close to finish 🎉   Currently this PR adds 187 bytes gzipped to core.

Open things to do / disucss

- [x] Fix issue of `Suspense` inside `Fragment` messing up the order of the DOM (#1605)
  - Skipping this test for now
- [x] Should `Suspense` / `lazy` remain part of the core or live in `preact/suspense` or `preact/compat`? This would save 137 bytes from core what would only add 50 bytes to core for somebody not using suspense
  - Keeping them in core for now. If you wish I can move them...
- [x] Shall we drop `isSuspense = true` inside `catchErrorInComponent` resulting in 3B saving but also make errors thrown inside `_childDidSuspend` result in a `Missing Suspense` error?
  - I'd not do so, marking as complete
- [x] Warn about `propTypes` on `lazy()` in `preact/debug`
- [x] Warn about missing `Suspense` in `preact/debug`
  - Not possible without some more changes. A `Missing Suspense` error must do the job for now
- [x] Golf some bytes
- [x] Create a demo

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -74,6 +74,22 @@ export function initDebug() {
 
 		// Check prop-types if available
 		if (typeof vnode.type==='function' && vnode.type.propTypes) {
+			if (vnode.type.name === 'Lazy') {
+				const m = 'PropTypes are not supported on lazy(). Use propTypes on the wrapped component itself. ';
+				try {
+					const lazied = vnode.type();
+					console.warn(m + 'Component wrapped in lazy() is ' + lazied.then.displayName || lazied.then.name);
+				}
+				catch (promise) {
+					console.warn(m + 'We will log the wrapped component\'s name once it is loaded.');
+					if (promise.then) {
+						promise.then((exports) => {
+							console.warn('Component wrapped in lazy() is ' + (exports.default.displayName || exports.default.name));
+						});
+					}
+
+				}
+			}
 			checkPropTypes(vnode.type.propTypes, vnode.props, getDisplayName(vnode), serializeVNode(vnode));
 		}
 
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -77,8 +77,8 @@ export function initDebug() {
 			if (vnode.type.displayName === 'Lazy') {
 				const m = 'PropTypes are not supported on lazy(). Use propTypes on the wrapped component itself. ';
 				try {
-					const lazied = vnode.type();
-					console.warn(m + 'Component wrapped in lazy() is ' + lazied.then.displayName || lazied.then.name);
+					const lazyVNode = vnode.type();
+					console.warn(m + 'Component wrapped in lazy() is ' + (lazyVNode.type.displayName || lazyVNode.type.name));
 				}
 				catch (promise) {
 					console.warn(m + 'We will log the wrapped component\'s name once it is loaded.');
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -74,7 +74,7 @@ export function initDebug() {
 
 		// Check prop-types if available
 		if (typeof vnode.type==='function' && vnode.type.propTypes) {
-			if (vnode.type.name === 'Lazy') {
+			if (vnode.type.displayName === 'Lazy') {
 				const m = 'PropTypes are not supported on lazy(). Use propTypes on the wrapped component itself. ';
 				try {
 					const lazied = vnode.type();
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -5,6 +5,7 @@ import { diffChildren } from './children';
 import { diffProps } from './props';
 import { assign, removeNode } from '../util';
 import options from '../options';
+import { sym as suspenseSymbol } from '../suspense';
 
 /**
  * Diff two virtual nodes and apply proper changes to the DOM
@@ -362,10 +363,24 @@ function doRender(props, state, context) {
  * component check for error boundary behaviors
  */
 function catchErrorInComponent(error, component) {
+	// thrown Promises are meant to suspend...
+	let isSuspend = typeof error.then === 'function';
+	let suspendingComponent = component;
+
 	for (; component; component = component._ancestorComponent) {
 		if (!component._processingException) {
 			try {
-				if (component.constructor.getDerivedStateFromError!=null) {
+				if (isSuspend) {
+					// console.log('catchErrorInComponent component[suspenseSymbol]', component[suspenseSymbol]);
+					if (component[suspenseSymbol] === suspenseSymbol && component.componentDidCatch!=null) {
+						// console.log('hitting suspense...');
+						component.componentDidCatch(error);
+					}
+					else {
+						continue;
+					}
+				}
+				else if (component.constructor.getDerivedStateFromError!=null) {
 					component.setState(component.constructor.getDerivedStateFromError(error));
 				}
 				else if (component.componentDidCatch!=null) {
@@ -378,8 +393,21 @@ function catchErrorInComponent(error, component) {
 			}
 			catch (e) {
 				error = e;
+				isSuspend = typeof error.then === 'function';
+				suspendingComponent = component;
 			}
 		}
 	}
+
+	if (isSuspend && suspendingComponent) {
+		// TODO: what is the preact way to determine the component name?
+		const componentName = suspendingComponent.displayName || (suspendingComponent._vnode && suspendingComponent._vnode.type
+			&& suspendingComponent._vnode.type.name);
+
+		return catchErrorInComponent(new Error(`${componentName} suspended while rendering, but no fallback UI was specified.
+
+Add a <Suspense fallback=...> component higher in the tree to provide a loading indicator or placeholder to display.`), suspendingComponent);
+	}
+
 	throw error;
 }
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -5,7 +5,6 @@ import { diffChildren } from './children';
 import { diffProps } from './props';
 import { assign, removeNode } from '../util';
 import options from '../options';
-import { sym as suspenseSymbol } from '../suspense';
 
 /**
  * Diff two virtual nodes and apply proper changes to the DOM
@@ -27,19 +26,7 @@ import { sym as suspenseSymbol } from '../suspense';
 export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, force, oldDom) {
 	// If the previous type doesn't match the new type we drop the whole subtree
 	if (oldVNode==null || newVNode==null || oldVNode.type!==newVNode.type || oldVNode.key!==newVNode.key) {
-		if (oldVNode!=null) {
-			const ancestor = oldVNode._component && oldVNode._component._ancestorComponent;
-			if (ancestor && ancestor[suspenseSymbol]) {
-				ancestor._vnode.suspendedContent = oldVNode;
-				// TODO: check whether parent was already removed from DOM...
-				if (oldVNode._dom) {
-					removeNode(oldVNode._dom);
-				}
-			}
-			else {
-				unmount(oldVNode, ancestorComponent);
-			}
-		}
+		if (oldVNode!=null) unmount(oldVNode, ancestorComponent);
 		if (newVNode==null) return null;
 		oldVNode = EMPTY_OBJ;
 	}
@@ -56,6 +43,9 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 
 	try {
 		outer: if (oldVNode.type===Fragment || newType===Fragment) {
+			// Passing the ancestorComponent here is needed for catchErrorInComponent to properly traverse upwards through fragments
+			// to find a parent Suspense
+			c = { _ancestorComponent: ancestorComponent };
 			diffChildren(parentDom, newVNode, oldVNode, context, isSvg, excessDomChildren, mounts, c, oldDom);
 
 			// Mark dom as empty in case `_children` is any empty array. If it isn't
@@ -153,11 +143,6 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 
 			let prev = c._prevVNode || null;
 			c._dirty = false;
-
-			if (c[suspenseSymbol] === suspenseSymbol && c._vnode.suspendedContent) {
-				if (prev) { unmount(prev); }
-				prev = c._vnode.suspendedContent;
-			}
 			let vnode = c._prevVNode = coerceToVNode(c.render(c.props, c.state, c.context));
 
 			if (c.getChildContext!=null) {
@@ -169,7 +154,6 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 			}
 
 			c._depth = ancestorComponent ? (ancestorComponent._depth || 0) + 1 : 0;
-			// this creates a new WrapperOne / new CustomSuspense
 			c.base = newVNode._dom = diff(parentDom, vnode, prev, context, isSvg, excessDomChildren, mounts, c, null, oldDom);
 
 			if (vnode!=null) {
@@ -389,10 +373,8 @@ function catchErrorInComponent(error, component) {
 		if (!component._processingException) {
 			try {
 				if (isSuspend) {
-					// console.log('catchErrorInComponent component[suspenseSymbol]', component[suspenseSymbol]);
-					if (component[suspenseSymbol] === suspenseSymbol && component.componentDidCatch!=null) {
-						// console.log('hitting suspense...');
-						component.componentDidCatch(error);
+					if (component.__s) {
+						component.__s(error);
 					}
 					else {
 						continue;
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -373,8 +373,8 @@ function catchErrorInComponent(error, component) {
 		if (!component._processingException) {
 			try {
 				if (isSuspend) {
-					if (component.__s) {
-						component.__s(error);
+					if (component._childDidSuspend) {
+						component._childDidSuspend(error);
 					}
 					else {
 						continue;
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -400,12 +400,6 @@ function catchErrorInComponent(error, component) {
 		}
 	}
 
-	// TODO: Add a react-like error message to preact/debug
-	/*
-		[componentName] suspended while rendering, but no fallback UI was specified.
-
-		Add a <Suspense fallback=...> component higher in the tree to provide a loading indicator or placeholder to display.
- 	*/
 	if (isSuspend) {
 		return catchErrorInComponent(new Error('Missing Suspense'), suspendingComponent);
 	}
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -43,10 +43,9 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 
 	try {
 		outer: if (oldVNode.type===Fragment || newType===Fragment) {
-			// Passing the ancestorComponent here is needed for catchErrorInComponent to properly traverse upwards through fragments
-			// to find a parent Suspense
-			c = { _ancestorComponent: ancestorComponent };
-			diffChildren(parentDom, newVNode, oldVNode, context, isSvg, excessDomChildren, mounts, c, oldDom);
+			// Passing the ancestorComponent instead of c here is needed for catchErrorInComponent
+			// to properly traverse upwards through fragments to find a parent Suspense
+			diffChildren(parentDom, newVNode, oldVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, oldDom);
 
 			// Mark dom as empty in case `_children` is any empty array. If it isn't
 			// we'll set `dom` to the correct value just a few lines later.
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -393,8 +393,7 @@ function catchErrorInComponent(error, component) {
 			}
 			catch (e) {
 				error = e;
-				isSuspend = typeof error.then === 'function';
-				suspendingComponent = component;
+				isSuspend = false;
 			}
 		}
 	}
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -27,7 +27,19 @@ import { sym as suspenseSymbol } from '../suspense';
 export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, force, oldDom) {
 	// If the previous type doesn't match the new type we drop the whole subtree
 	if (oldVNode==null || newVNode==null || oldVNode.type!==newVNode.type || oldVNode.key!==newVNode.key) {
-		if (oldVNode!=null) unmount(oldVNode, ancestorComponent);
+		if (oldVNode!=null) {
+			const ancestor = oldVNode._component && oldVNode._component._ancestorComponent;
+			if (ancestor && ancestor[suspenseSymbol]) {
+				ancestor._vnode.suspendedContent = oldVNode;
+				// TODO: check whether parent was already removed from DOM...
+				if (oldVNode._dom) {
+					removeNode(oldVNode._dom);
+				}
+			}
+			else {
+				unmount(oldVNode, ancestorComponent);
+			}
+		}
 		if (newVNode==null) return null;
 		oldVNode = EMPTY_OBJ;
 	}
@@ -141,6 +153,11 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 
 			let prev = c._prevVNode || null;
 			c._dirty = false;
+
+			if (c[suspenseSymbol] === suspenseSymbol && c._vnode.suspendedContent) {
+				if (prev) { unmount(prev); }
+				prev = c._vnode.suspendedContent;
+			}
 			let vnode = c._prevVNode = coerceToVNode(c.render(c.props, c.state, c.context));
 
 			if (c.getChildContext!=null) {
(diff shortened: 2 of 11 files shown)


sventschui:
We could add a prop (e.g. `promise.Suspense = Symbol.for('Suspense')`) to the promises thrown by `lazy` and then skip user-land `componentDidCatch` for these errors

We should also investigate how React behaves as Suspense is catching promises in React too

sventschui:
While writing the tests one thing came to my attention. When using Suspense in React the suspended component tree doesn't loose state. It isn't unmounted, just "parked". I'm not sure whether this is possible with the current solution. Do you have an idea how to handle this?

JoviDeCroock:
That would be something very specific to a two phase render cycle and is not possible at this time as far as I know.

sventschui:
I guess most of the bytes come from a huge error message I copied from React. We should replace this with a shorter one or even with just a short-link to a wiki/error description page. Do you have a strategy in place for things like this?

The state keeping thing is quite important as it breaks all the tests and compat to React currently.

When not having suspense in core we at least should try to make it possible in a user-land extension. That means having catchErrorInComponent extendable/replaceable and something to make the state keeping possible. Honestly the dirties hack that comes to my mind here is keeping it as display hidden in user-land code but that will definitely introduce other issues. 

For the state keeping we might want to try to flag the direct children of Suspend as suspended, keep them as a prop to Suspense and not call unmount on these. During the next render simply re-run render on these children. Thinking about this makes me think of implementing Suspense not as a component but as code in the core that gets triggered if a component is of type Symbol.for("Suspense")

Sorry for my thought flush on my phone. Will look into this a bit deeper today once I get my fingers to my macbook

src/diff/index.js
```diff
@@ -378,8 +393,21 @@ function catchErrorInComponent(error, component) {
...
 				error = e;
+				isSuspend = typeof error.then === 'function';
+				suspendingComponent = component;
 			}
 		}
 	}
+
+	if (isSuspend && suspendingComponent) {
+		// TODO: what is the preact way to determine the component name?
+		const componentName = suspendingComponent.displayName || (suspendingComponent._vnode && suspendingComponent._vnode.type
```
marvinhagemeister: Maybe we can move this section into `preact/debug`. A `Suspense` component without a fallback doesn't seem to make much sense, so we could just always warn on that. This would also keep core small and tiny :tada:

test/browser/suspense.test.js
```diff
@@ -0,0 +1,235 @@
...
+		return this.state.error ? (
+			<div>
+				Catcher did catch: {this.state.error.message}
+			</div>
+		) : this.props.children;
+	}
+}
+
+const Lazy = lazy(() => new Promise((res) => {
+	setImmediate(() => {
```
marvinhagemeister: `setImmediate` is a [non-standard API](https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate) and only supported in IE

test/browser/suspense.test.js
```diff
@@ -0,0 +1,235 @@
...
+		res({ default: LazyComp });
+	});
+}));
+
+const ThrowingLazy = lazy(() => new Promise((res, rej) => {
+	setImmediate(() => {
```
marvinhagemeister: Same here :slightly_smiling_face:

test/browser/suspense.test.js
```diff
@@ -0,0 +1,235 @@
...
+		);
+
+		return tick(() => {
+			rerender();
+			expect(scratch.innerHTML).to.eql(
+				`<div>Hello ThrowingSuspense...</div>`
+			);
+		});
+	});
+});
```
marvinhagemeister: nit: missing newline at the end of the file.

src/suspense.js
```diff
@@ -0,0 +1,68 @@
...
+		this.state = {
+			l: false
+		};
+	}
+
+	componentDidCatch(e) {
+		if (e && typeof e.then === 'function') {
+			this.setState({ l: true });
+			e.then(
+				() => {
```
marvinhagemeister: Storing the callback in a variable saves `3 B` :tada: 

```js
let done = () => this.setState({ l: false });
e.then(done, done);
```

src/diff/index.js
JoviDeCroock: This also allows the recheck of isSuspsend and suspendingComponent, so that will save a lot

src/suspense.js
```diff
@@ -0,0 +1,68 @@
...
+		// mark this component as a handler of suspension (thrown Promises)
+		this[sym] = sym;
+
+		this.state = {
+			l: false
+		};
+	}
+
+	componentDidCatch(e) {
+		if (e && typeof e.then === 'function') {
```
marvinhagemeister: Idea: Maybe we can get away with just checking for the presence of a `then` property.

src/suspense.js
```diff
@@ -0,0 +1,68 @@
...
+
+export const sym = Symbol.for('Suspense');
+
+export class Suspense extends Component {
+	constructor(props) {
+		// TODO: should we add propTypes in DEV mode?
+		super(props);
+
+		// mark this component as a handler of suspension (thrown Promises)
+		this[sym] = sym;
```
marvinhagemeister: Removing the symbol and replacing it with a simple `boolean` flag saves `27 B` :100: 

```js
this._suspense = true;
```

src/suspense.js
```diff
@@ -0,0 +1,68 @@
...
+				}
+			);
+		}
+		else {
+			throw e;
+		}
+	}
+
+	render() {
+		return this.state.l ? this.props.fallback : this.props.children;
```
marvinhagemeister: Using the method arguments (only available in Preact) saves another `6 B`.

```js
render(props, state) {
  return state.l ? props.fallback : props.children;
}
```

src/diff/index.js
```diff
@@ -362,10 +363,24 @@ function doRender(props, state, context) {
...
+	let isSuspend = typeof error.then === 'function';
+	let suspendingComponent = component;
+
 	for (; component; component = component._ancestorComponent) {
 		if (!component._processingException) {
 			try {
-				if (component.constructor.getDerivedStateFromError!=null) {
+				if (isSuspend) {
+					// console.log('catchErrorInComponent component[suspenseSymbol]', component[suspenseSymbol]);
+					if (component[suspenseSymbol] === suspenseSymbol && component.componentDidCatch!=null) {
```
marvinhagemeister: I think it's safe to assume that our own `Suspense` component will always have `componentDidCatch` defined :+1:

sventschui:
Thanks for these hints. Will definitely incorporate this once I get closer to a working implementation.

The thing that currently bothers me is keeping the state of the suspended component tree (as React does). I was thinking of doing the following:

- Not unmount (but remove DOM of) children of `Suspense` (here https://github.com/developit/preact/blob/master/src/diff/index.js#L29)
- Re-render the previous children of suspense here https://github.com/developit/preact/blob/master/src/diff/index.js#L143

Can't think of how to achieve the second point. One of you guys got an idea how to solve this?

sventschui:
I justed hacked a working solution together to keep state of suspended components 🎉 . 

Would be cool if you could look into this. I don't really feel confident with the current solution. I'll invest some more time when I get to it.

ForsakenHarmony:
might want to have a test with multiple suspended components in one suspense

src/diff/index.js
sventschui: I still kept the `Missing Suspense` error here and throwing it on the `suspendingComponent`.

Would you want to save some bytes and just throw the error at the end of the function instead of the `suspendingComponent`?

test/browser/suspense.test.js
sventschui: Replaced with `setTimeout`. Thx for the heads up :)

test/browser/suspense.test.js
sventschui: Should I add an eslint rule for this?

src/suspense.js
sventschui: Done. Thx!

src/suspense.js
sventschui: Done

src/diff/index.js
JoviDeCroock: What he's saying is to throw it in a hook from preact/debug since it doesn't really matter in production anyway

src/suspense.js
sventschui: Done

src/diff/index.js
```diff
@@ -43,6 +43,9 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 
 	try {
 		outer: if (oldVNode.type===Fragment || newType===Fragment) {
+			// Passing the ancestorComponent here is needed for catchErrorInComponent to properly traverse upwards through fragments
+			// to find a parent Suspense
+			c = { _ancestorComponent: ancestorComponent };
```
sventschui: Can you please double check that this won't break anything?

src/suspense.js
```diff
@@ -0,0 +1,52 @@
...
+
+export class Suspense extends Component {
+	constructor(props) {
+		super(props);
+
+		this.state = {};
+	}
+
+	__s(e) {
+		if (typeof e.then == 'function') {
```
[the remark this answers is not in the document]
sventschui: If we assume that there never is an object like `{ then: "foo" }` thrown it would. Do you think this is safe?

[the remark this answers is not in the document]
JoviDeCroock: Hmm, things I am scared of with this is suspense -> errorboundary -> comp so you’re right in that

src/suspense.js
```diff
@@ -0,0 +1,52 @@
...
+	render(props, state) {
+		return state.l ? props.fallback : props.children;
+	}
+}
+
+export function lazy(loader) {
+	let prom;
+	let component;
+	let error;
+	return function L(props) {
```
andrewiggins: @marvinhagemeister does this function need a `displayName` to show up in devtools correctly? Similarly to `memo` and `forwardRef`?

src/diff/index.js
```diff
@@ -378,8 +393,21 @@ function catchErrorInComponent(error, component) {
 			}
 			catch (e) {
 				error = e;
+				isSuspend = typeof error.then === 'function';
```
andrewiggins: When will this `catch(e)` handle a suspend? I don't think any of the above code will ever invoke a synchronous render (`setState`, `enqueueRender`, etc. all trigger an async rerender), so the code this `catch` watches should never throw a promise.

If that's true then I think you can remove the `isSuspend` and `suspendingComponent` lines from this `catch`.

test/browser/suspense.test.js
```diff
@@ -0,0 +1,354 @@
...
+			return prom;
+		},
+		getPromise: () => prom,
+		isDone: () => done
+	};
+}
+
+class WrapperOne extends Component {
+	render() {
+		return this.props.children;
```
andrewiggins: Can this component render a the children inside of a `div`? I'd like to validate that when suspending any intermediate DOM is removed and the fallback replaces all of it.

test/browser/suspense.test.js
```diff
@@ -0,0 +1,354 @@
...
+	componentDidCatch(e) {
+		this.setState({ error: e });
+	}
+
+	render(props, state) {
+		return state.error ? <div>Catcher did catch: {state.error.message}</div> : props.children;
+	}
+}
+
+function createSuspension(name, timeout, t) {
```
andrewiggins: What's `t`? Can you give it a more legible name? Perhaps `error`?

src/suspense.js
marvinhagemeister: Good catch! Yes it does need the `displayName` property to be able to distinguish it from other components 👍

src/diff/index.js
sventschui: It is not only about error boundaries. The `ancestorComponent` is also used in `diffChildren`, `diff` and `unmount`. It is set to the diffed component as `_ ancestorComponent` prop. I need to look a bit deeper into this as I don't fully understand the possible side-effects of this change as of now

src/suspense.js
```diff
@@ -0,0 +1,52 @@
+import { Component } from './component';
+import { createElement } from './create-element';
+
+export class Suspense extends Component {
+	constructor(props) {
```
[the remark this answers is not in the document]
sventschui: I did a rewrite with a prototype based approach. Saves ~50 bytes gzipped 👍

src/suspense.js
```diff
@@ -0,0 +1,52 @@
...
+		super(props);
+
+		this.state = {};
+	}
+
+	__s(e) {
```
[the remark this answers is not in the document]
sventschui: Didn't knwo terser is doing this, renamed to `_childDidSuspend` 👍

src/suspense.js
```diff
@@ -0,0 +1,52 @@
...
+
+export function lazy(loader) {
+	let prom;
+	let component;
+	let error;
+	return function L(props) {
+		if (!prom) {
+			prom = loader();
+			prom.then(
+				({ default: c }) => { component = c; },
```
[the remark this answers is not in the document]
sventschui: Done. Together with the `{ error = e }` this saved 2 bytes

src/suspense.js
```diff
@@ -0,0 +1,52 @@
...
+				e => error = e,
```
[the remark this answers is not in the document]
sventschui: Done

src/suspense.js
sventschui: Added a `displayName` that is set once the lazy loaded component is loaded. Will check whether this works once I got a demo together

src/diff/index.js
sventschui: Done, saved 5 bytes 🎉 . 

I kept `isSuspend = false` in the catch tough. Dropping this would save another 3B but errors thrown in `_childDidSuspend` would result in a `Missing Suspense` error. What do you think? Remove the `isSuspend = false` too?

test/browser/suspense.test.js
sventschui: Done. Also added a function component wrapper

test/browser/suspense.test.js
sventschui: Done

src/suspense.js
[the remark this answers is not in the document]
marvinhagemeister: That's a very good point. Haven't thought of it that way :+1:

src/diff/index.js
sventschui: Passing `ancestorComponent` directly now. Saves 5 bytes

src/suspense.js
```diff
@@ -0,0 +1,56 @@
...
+
+		if (!component) {
+			throw prom;
+		}
+
+		return createElement(component, props);
+	}
+
+	return Lazy;
+}
```
marvinhagemeister: nit: missing newline

src/suspense.js
```diff
@@ -0,0 +1,56 @@
...
+	let component;
+	let error;
+
+	function Lazy(props) {
+		if (!prom) {
+			prom = loader();
+			prom.then(
+				(exports) => {
+					component = exports.default;
+					Lazy.displayName = 'Lazy(' + (component.displayName || component.name) + ')';
```
marvinhagemeister: Just checked against the experimental react-devtools to see how they detect both `Suspense` and `Lazy` components. Turns out that the `displayName` for `Lazy` doesn't need to be aware of the component it is rendering. It's printed like a standard component inside the tree.

```txt
<Suspense>
  <Lazy>
    <MyComponent>
```

With this in mind just doing `Lazy.displayName = 'Lazy'` is enough. That can happen outside the declaration itself.

In the long run we'll probably switch to just flagging any special components with a constant instead of a string. That way we could detect `Provider` and `Consumer` components too. The devtools render these slightly differently, but I haven't gotten around to implement that yet. This is out of scope for this PR though and the above solution is enough for now :+1:

sventschui:
I just introduced a demo of suspense and found that the current implementation messes up the order of the DOM when `Suspense` is placed inside a `Fragment` (See screenshot, `I'm (fake) lazy loaded` is supposed to go under the `lazy()` headline. DevTools show correct order, DOM is messed up). 

I added a breaking test for this but can't currently figure out why this happens.

image

src/diff/index.js
sventschui: Can you guys give me a hint on where I could add this to `preact/debug`. I did not find a proper place to add it

debug/src/debug.js
```diff
@@ -74,6 +74,22 @@ export function initDebug() {
 
 		// Check prop-types if available
 		if (typeof vnode.type==='function' && vnode.type.propTypes) {
+			if (vnode.type.name === 'Lazy') {
```
marvinhagemeister: The string name might be unsafe because minifiers always mangle the function names. We should check for `displayName` here :+1:

marvinhagemeister:
Great work on the `preact/debug` integration :+1:

LukasBombach:
@sventschui @marvinhagemeister can you check if this implementation works with Next?

I have setup a monorepo project that uses this branch with next and a next plugin similar to [@zeit/next-preact](https://github.com/zeit/next-plugins/tree/master/packages/next-preact). Basically I have copy and pasted their code and used it with this branch. Check it out here:

https://github.com/spring-media/preact-next-suspense

I have written a bit of a documentation to explain the repo in the README

sventschui:
@LukasBombach as mentioned in slack: The error is due to an old version of `preact-render-to-string`. You need to update to `>=5`

src/diff/index.js
```diff
@@ -365,10 +367,22 @@ function doRender(props, state, context) {
...
 	for (; component; component = component._ancestorComponent) {
 		if (!component._processingException) {
 			try {
-				if (component.constructor.getDerivedStateFromError!=null) {
+				if (isSuspend) {
+					if (component._childDidSuspend) {
+						component._childDidSuspend(error);
+					}
+					else {
+						continue;
```
developit: Maybe move to here?

```suggestion
		return catchErrorInComponent(Error('Missing Suspense'), suspendingComponent);
```

developit:
Question:  what's the reason for using `._childDidSuspend`?  It seems like the Suspense component could just implement componentDidCatch:

```js
class Suspense {
  componentDidCatch(e) {
    if (e && e.then) e.then(() => this.setState({}));
    else throw e;
  }
}
```

src/diff/index.js
sventschui: The `continue` results in traversing up the tree. This would break the tree traversing AFAIK

sventschui:
Well the naming of the method is irrelevant. It just needs to take precedence over regular `componentDidCatch` when the error thrown is a promise (even if between the throwing component and `<Suspense>` is another error boundary

LukasBombach:
> @LukasBombach as mentioned in slack: The error is due to an old version of `preact-render-to-string`. You need to update to `>=5`

👌

src/diff/index.js
```diff
@@ -43,7 +43,9 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 
 	try {
 		outer: if (oldVNode.type===Fragment || newType===Fragment) {
-			diffChildren(parentDom, newVNode, oldVNode, context, isSvg, excessDomChildren, mounts, c, oldDom);
+			// Passing the ancestorComponent instead of c here is needed for catchErrorInComponent
+			// to properly traverse upwards through fragments to find a parent Suspense
+			diffChildren(parentDom, newVNode, oldVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, oldDom);
```
marvinhagemeister: Good catch! This may also fix an issue I was having when working on the devtools where `Fragments` would break an upwards traversal :+1:

LukasBombach:
@sventschui @marvinhagemeister @JoviDeCroock I can confirm this PR works with Next 8.x using `preact-render-to-string@next` as @sventschui proposed:

https://github.com/spring-media/preact-next-suspense/commit/5a7de52dd39f6f8d231b1bf55e0e2ba24cb978b3

src/diff/index.js
```diff
@@ -381,8 +395,20 @@ function catchErrorInComponent(error, component) {
 			}
 			catch (e) {
 				error = e;
+				isSuspend = false;
 			}
 		}
 	}
+
+	// TODO: Add a react-like error message to preact/debug
```
38elements: Does this Todo need to be resolved ?
If it is unnecessary, it is better to remove it.

LukasBombach:
Oh wait, there is still an issue with this and `Next.js` when you run my repo and open http://localhost:3000, you'll get an error in node saying

```
[ error ] TypeError: component.props.handleStateChange is not a function
    at emitChange (preact-next-suspense/node_modules/next-server/dist/lib/side-effect.js:11:29)
    at new SideEffect (preact-next-suspense/node_modules/next-server/dist/lib/side-effect.js:26:17)
    at s (preact-next-suspense/node_modules/preact-render-to-string/dist/index.js:1:1371)
    at s (preact-next-suspense/node_modules/preact-render-to-string/dist/index.js:1:1683)
    at s (preact-next-suspense/node_modules/preact-render-to-string/dist/index.js:1:1683)
    at s (preact-next-suspense/node_modules/preact-render-to-string/dist/index.js:1:1683)
    at s (preact-next-suspense/node_modules/preact-render-to-string/dist/index.js:1:2849)
    at s (preact-next-suspense/node_modules/preact-render-to-string/dist/index.js:1:1683)
    at s (preact-next-suspense/node_modules/preact-render-to-string/dist/index.js:1:1683)
    at s (preact-next-suspense/node_modules/preact-render-to-string/dist/index.js:1:1683)
    at s (preact-next-suspense/node_modules/preact-render-to-string/dist/index.js:1:1683)
    at s (preact-next-suspense/node_modules/preact-render-to-string/dist/index.js:1:1683)
    at s (preact-next-suspense/node_modules/preact-render-to-string/dist/index.js:1:1683)
    at s (preact-next-suspense/node_modules/preact-render-to-string/dist/index.js:1:1683)
    at render (preact-next-suspense/node_modules/next-server/dist/server/render.js:86:16)
    at renderPage (preact-next-suspense/node_modules/next-server/dist/server/render.js:211:20)
```

sventschui:
@LukasBombach I once encountered this issue too. As far as I remember it was caused by context not being set correctly but this should be fixed. I can run your example without any issues

marvinhagemeister:
@LukasBombach a new version for `preact-render-to-string` with support for `creeateContext` hasn't been published yet. The code is already in `master` but I was too busy the past days. The next release should be up by the end of the week.

LukasBombach:
@sventschui have you checked your `node` console after loading the index page in your browser?

src/diff/index.js
sventschui: I'd like to tackle this (see TODO in PR description) but currently don't know how to to this without adding too many bytes to core

LukasBombach:
> **EDIT:** published it 🎉

@marvinhagemeister @sventschui that update seems to have fixed the error

debug/src/debug.js
```diff
@@ -74,6 +74,22 @@ export function initDebug() {
...
 		if (typeof vnode.type==='function' && vnode.type.propTypes) {
+			if (vnode.type.displayName === 'Lazy') {
+				const m = 'PropTypes are not supported on lazy(). Use propTypes on the wrapped component itself. ';
+				try {
+					const lazyVNode = vnode.type();
+					console.warn(m + 'Component wrapped in lazy() is ' + (lazyVNode.type.displayName || lazyVNode.type.name));
+				}
+				catch (promise) {
+					console.warn(m + 'We will log the wrapped component\'s name once it is loaded.');
+					if (promise.then) {
```
marvinhagemeister: To get coveralls green again we just need to add a case when `Lazy` throws something random as long as it's not a promise :+1:

src/suspense.js
```diff
@@ -0,0 +1,51 @@
...
+export function lazy(loader) {
+	let prom;
+	let component;
+	let error;
+
+	function Lazy(props) {
+		if (!prom) {
+			prom = loader();
+			prom.then(
+				(exports) => { component = exports.default; },
```
cristianbote: Is this the right path here? Assuming that the `loader` would always return a _default_ export? That means we won't support named exports with lazy?

JoviDeCroock: React doesn't support this either, we could probably support it by introducing a second string param that defaults to `default` but is it really needed?

pmkroeker: This seems to be the preferred method (at least for [preact-async-route](https://github.com/prateekbh/preact-async-route#usage-example). Possibly change `(exports) => { component = exports.default; },` to `(exports) => { component = exports; },`. Then the user would be able to specify what the export is?

JoviDeCroock: No, this does not return to the user so the user can't alter the used export only if it's specified at initialization time.

pmkroeker: Right, that would lock in a pre-determined export.

cristianbote: Ha, wasn't aware that React does the `default` only thing. I guess keeping it on par with React is what we should be doing, when it makes sense. I was thinking that maybe this is allowed:

```js
const Foo = lazy(() => import('./Foo').then(mod => mod.Foo));
```

Then we could basically support both of them?
```js
exports => { component = exports.default || exports },
```

This would allow the usage of default with named:
```js
const Foo = lazy(() => import('./Foo').then(mod => mod.Foo));
const Baz = lazy(() => import('./FooBaz));
```
But, feels like I might not be that familiar with `React.lazy`

JoviDeCroock: https://reactjs.org/docs/code-splitting.html#named-exports

Is the docs entry for it

marvinhagemeister: This is a great point! I think we should explore ways to use named exports after this PR has landed :+1:

debug/src/debug.js
sventschui: IIRC there are also some things to be covered in `diff/index.js`. Will check asap

LukasBombach:
🎉 😍

PR #1595, opened 2019-05-02
Ignore dangerouslySetInnerHTML during hydration

developit:
This matches React's behavior.
I can't find the issue right now but I will.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -272,7 +272,7 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
 			}
 			let oldHtml = oldProps.dangerouslySetInnerHTML;
 			let newHtml = newProps.dangerouslySetInnerHTML;
-			if (newHtml || oldHtml) {
+			if ((newHtml || oldHtml) && excessDomChildren==null) {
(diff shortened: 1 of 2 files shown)


marvinhagemeister:
Looks like we'd need to remove the `should hydrate with dangerouslySetInnerHTML` test.

developit:
Might be worth adding a warning via preact/debug for this too.

cristianbote:
Hey @developit, I updated the test that was failing and I was trying to see if I could add the warning part. But no idea how to detect this case, inside `debug/`. Any pointers?

johannesodland:
Hey @developit

Is the intention to avoid reaplying innerHTML, or is it to strip away the content of VNodes with dangerouslySetInnerHTML?

When using SSR the current solution strips away all content rendered with dangerouslySetInnerHTML during hydration. **This breaks 10.0.0-beta.2 for us**. 

We use dangerouslySetInnerHTML to render content from our CMS (and have been doing so for years). It renders perfectly on the server, but after 10.0.0-beta.2 the content is stripped away during hydration, essentially leaving the application blank.

marvinhagemeister:
@johannesodland Our future approach will to only attach event listeners in hydration and leave the DOM as is. That work is tracked in #1697

PR #1596, opened 2019-05-02
RFC: don't read attributes during hydration

developit:
This would represent a real size + performance improvement.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -264,17 +264,9 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
 			excessDomChildren = EMPTY_ARR.slice.call(dom.childNodes);
 		}
 		if (newVNode!==oldVNode) {
-			// if we're hydrating, use the element's attributes as its current props:
-			if (oldProps==null) {
-				oldProps = {};
-				if (excessDomChildren!=null) {
-					let name;
-					for (i=0; i<dom.attributes.length; i++) {
-						name = dom.attributes[i].name;
-						oldProps[name=='class' && newProps.className ? 'className' : name] = dom.attributes[i].value;
-					}
-				}
-			}
+			let oldProps = oldVNode.props || null;
+			let newProps = newVNode.props;
+
 			let oldHtml = oldProps.dangerouslySetInnerHTML;
 			let newHtml = newProps.dangerouslySetInnerHTML;
 			if ((newHtml || oldHtml) && excessDomChildren==null) {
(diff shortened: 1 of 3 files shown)


marvinhagemeister:
Looks like we just need to add a `null` check around the `dangerouslySetInnerHTML` bits in `diffElementNodes` and then we are good to go :+1:

JoviDeCroock:
The more I think about this the more I'm in favor of doing this, iterating over attributes doesn't give that much extra afaik.

andrewiggins:
Also a fan of this. Let's get some bytes back!

cristianbote:
Hey @JoviDeCroock 👋 !

So, I was looking into what's left to do here and what tests are failing. As @marvinhagemeister and you stated there are a few places that needs to be touched and the remaining tests I believe were the ones that are testing the reading of attributes when hydrating 😄.
My approach was to just send `null` to `setProperty` if `oldProps` is falsy.
```
-      setProperty(dom, keys[i], newProps[keys[i]], oldProps[keys[i]], isSvg);
+      setProperty(dom, keys[i], newProps[keys[i]], oldProps ? oldProps[keys[i]] : null, isSvg);
```
Now, maybe there could be some more gains, in terms of size, by using the `EMPTY_OBJ`? So, the lookup would not throw.

JoviDeCroock:
Yes @cristianbote want to focus in this context to be really size-aware. But yes you're right we could start off with nullish checks.

src/diff/index.js
```diff
@@ -256,20 +256,9 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
 			excessDomChildren = EMPTY_ARR.slice.call(dom.childNodes);
 		}
 		if (newVNode!==oldVNode) {
-			let oldProps = oldVNode.props;
+			let oldProps = oldVNode.props || null;
```
cristianbote: Hey @developit, as @JoviDeCroock noted, defaulting to `EMPTY_OBJ` would really get the size further down. Do you see any issues by reusing the `EMPTY_OBJ` ref instead of `null`?

developit: I'd prefer to split up the codepaths here so that attributes are diffed using a separate flow anyway. I'll check though, if the perf difference isn't huge then using an empty object here is fine.

cristianbote: Awesome!

My concern would be that using `EMPTY_OBJ` would basically do the lookup on that object where with `null` that would not happen. So, I guess `null` could entail a bit of boost? But that's just by looking at it tbh. Gonna spin-up an bench to, hopefully, clear this.

cristianbote: Alrighty! 😄 
So, using `EMPTY_OBJ` in my tests has no effect in terms of perf. The only thing would be the codepath split that you'd like to happen. Would this have an impact on splitting it up? I guess we would either do `oldProps === null` or `oldProps === EMPTY_OBJ`?

cristianbote:
@andrewiggins well, seem like we're gonna get `-32B` 🎉

developit:
Sweet, just woke up this morning wanting to check in on this PR and it's been merged :D

PR #1599, opened 2019-05-04
Code golf preact.js.gz size down by 53 B ⛳️

jviide:
This pull request applies a bunch of small code modifications that seem to reduce the overall bundle size. Here's a rough outline of the changes:

 * 276850e, 5e78b34, 37ce351, 07b42de mostly move around loop invariants to compress the code.
 * 3ec3829, d32aa17, 82f4cca reuse local variables etc. to avoid repetition.
 * 3f71109 replaces an `Array#map` call with [`Array#some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) that:
    1. should be just as supported as `Array#map` but doesn't collect all callback results to an array like `Array#map`
    1. seems to compress better than `Array#forEach`
 * 76529b7 reorganizes `createContext` extensively, getting rid of local variable and function declarations.
 * 7e53dd2 removes a superfluous check from diffProps - `newProps`'s falsyness is checked although that shouldn't be possible at that point.
 * 04646b6 doesn't explictly init `isNew` to `false`, as its default `undefined` value is sufficient.
 * The rest of the commits mostly reorder code or modify `if (x!=null)` checks to `if (x)` when appropriate.

The performance _seems_ to stay on the same level with master. I used Chrome 74, MacOS 10.14.4 and https://github.com/mathieuancelin/js-repaint-perfs for testing, so YMMV.

The size impact compared to the master is:

|bundle file|master|random-opts|delta|
|-|-|-|-|
|preact.js.gz|3548 B|3495 B|-53 B|
|preact.js.br|3242 B|3201 B|-41 B|
|preact.module.js.gz|3566 B|3517 B|-49 B|
|preact.module.js.br|3265 B|3224 B|-41 B|
|preact.umd.js.gz|3604 B|3561 B|-43 B|
|preact.umd.js.br|3293 B|3255 B|-38 B|

### DIFF
--- a/src/create-context.js
+++ b/src/create-context.js
@@ -27,7 +27,7 @@ export function createContext(defaultValue) {
 			return ctx;
 		};
 		comp.shouldComponentUpdate = props => {
-			subs.map(c => {
+			subs.some(c => {
 				// Check if still mounted
 				if (c._parentDom) {
 					c.context = props.value;
--- a/src/create-context.js
+++ b/src/create-context.js
@@ -7,49 +7,44 @@ export let i = 0;
  * @param {any} defaultValue
  */
 export function createContext(defaultValue) {
-	let context = {
+	const ctx = {};
+
+	const context = {
 		_id: '__cC' + i++,
-		_defaultValue: defaultValue
+		_defaultValue: defaultValue,
+		Consumer(props, context) {
+			return props.children(context);
+		},
+		Provider(props) {
+			if (!this.getChildContext) {
+				const subs = [];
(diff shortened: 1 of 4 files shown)


jviide:
Rebased the code on master, removed a couple of commits that used `Array#some` to replace for-loops (they seemed to bring down the performance a bit). Updated the pull request description accordingly.

kristoferbaxter:
Accidentally closed from mobile GitHub. Sorry!

PR #1602, opened 2019-05-05
(refactor) - gold Hooks (-11B) and compat (-18B)

JoviDeCroock:
Thought these packages deserved some attention aswell :D

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -171,7 +171,7 @@ function cloneElement(element) {
  * @returns {boolean}
  */
 function isValidElement(element) {
-	return element && element.$$typeof===REACT_ELEMENT_TYPE;
+	return !!element && element.$$typeof===REACT_ELEMENT_TYPE;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


compat/src/index.js
```diff
@@ -15,8 +15,7 @@ options.event = e => {
 	/* istanbul ignore next */
 	if (oldEventHook) e = oldEventHook(e);
 	e.persist = () => {};
-	e.nativeEvent = e;
-	return e;
+	return e.nativeEvent = e;
```
marvinhagemeister: neat :+1:

hooks/src/index.js
```diff
@@ -44,9 +43,9 @@ options.unmount = vnode => {
 	if (!c) return;
 
 	const hooks = c.__hooks;
-	if (!hooks) return;
-
-	hooks._list.forEach(hook => hook._cleanup && hook._cleanup());
+	if (hooks) {
+		hooks._list.some(hook => hook._cleanup && hook._cleanup());
```
ForsakenHarmony: what if this returns a truthy value?

JoviDeCroock: You are 100% right, I'll fix this tonight

marvinhagemeister: @ForsakenHarmony good catch, missed that in my review :+1:

PR #1604, opened 2019-05-05
Order issue when setState changes vnode type in Fragment

sventschui:
This PR adds a test to showcase the ordering issue after setState when I was implementing #1593 

As I don't have time to find a fix now I'll just open the PR with the test showcasing the bug. If nobody is faster than me I'll look into this tomorrow.

### DIFF
--- a/test/browser/fragments.test.js
+++ b/test/browser/fragments.test.js
@@ -1870,4 +1870,50 @@ describe('Fragment', () => {
 		rerender();
 		expect(scratch.textContent).to.equal('ABC');
 	});
+
+	it('setState should work inside a fragment when changing vnode.type returned by render', () => {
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
This is an awesome test :+1: The error is caused by `forceUpdate` always appending new elements. It's not aware of any siblings yet :S

https://github.com/developit/preact/blob/d7abd4e0a2b7a6739882245e483d2f0e69348a34/src/component.js#L76

Will look into it later.

sventschui:
Closing as the fix will land with #1605

JoviDeCroock:
Let's keep this open until we land it, helps keep a clear overview. Great you're following up on things though :D

PR #1605, opened 2019-05-05
Fix incorrect DOM when Fragment has siblings

marvinhagemeister:
Fixes the issue mentioned in #1593 and #1604 .

Currently a draft because it adds ~~`+135`~~ `+53 B`

**EDIT:** Didn't rebase against the savings in `master` when measuring size. Was way off :sweat_smile: 

Also fixes: https://github.com/developit/preact/issues/1635

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -71,22 +71,17 @@ Component.prototype.forceUpdate = function(callback) {
 		const force = callback!==false;
 
 		let mounts = [];
-		dom = diff(parentDom, vnode, vnode, this._context, parentDom.ownerSVGElement!==undefined, null, mounts, this._ancestorComponent, force, dom);
+		dom = diff(parentDom, vnode, vnode, this._context, parentDom.ownerSVGElement!==undefined, null, mounts, this._ancestorComponent, force, dom, this._parentVNode);
 		if (dom!=null && dom.parentNode!==parentDom) {
 			// The component may be rendered somewhere in the middle of the parent's
 			// children. We need to find the nearest DOM sibling to insert our
 			// newly rendered node into.
-			let nextDom;
-			let sibling = this._siblingVNode;
-			while (sibling && !(nextDom = sibling._dom)) {
-				sibling = sibling._component && sibling._component._siblingVNode;
-			}
-
-			if (nextDom) {
-				parentDom.insertBefore(dom, nextDom);
+			let nextDom = getDomSibling(vnode);
+			if (nextDom==null || nextDom.parentNode!==parentDom) {
+				parentDom.appendChild(dom);
 			}
 			else {
-				parentDom.appendChild(dom);
+				parentDom.insertBefore(dom, nextDom);
 			}
 		}
 		commitRoot(mounts, vnode);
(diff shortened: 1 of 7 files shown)


src/component.js
```diff
@@ -73,7 +73,21 @@ Component.prototype.forceUpdate = function(callback) {
...
 		dom = diff(parentDom, vnode, vnode, this._context, parentDom.ownerSVGElement!==undefined, null, mounts, this._ancestorComponent, force, dom);
 		if (dom!=null && dom.parentNode!==parentDom) {
-			parentDom.appendChild(dom);
+			// The component may be rendered somewhere in the middle of the parent's
+			// children. We need to find the nearest DOM sibling to insert our
+			// newly rendered node into.
+			let nextDom;
+			let sibling = this._siblingVNode;
+			while (sibling && !(nextDom = sibling._dom)) {
+				sibling = sibling._component && sibling._component._siblingVNode;
```
marvinhagemeister: This code assumes that the next sibling can be found in a sibling `vnode`. Need to check how it behaves for nested trees with a higher depth-level.

src/diff/children.js
```diff
@@ -24,7 +24,7 @@ import { removeNode } from '../util';
  * Fragments that have siblings. In most cases, it starts out as `oldChildren[0]._dom`.
  */
 export function diffChildren(parentDom, newParentVNode, oldParentVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, oldDom) {
-	let childVNode, i, j, oldVNode, newDom, sibDom;
+	let childVNode, i, j, oldVNode, newDom, sibDom, lastRendered;
```
JoviDeCroock: can't we default this to childVNode, saves some bytes since we can remove line 88 like that

src/component.js
JoviDeCroock: Don't really think that's a bad assumption (can't see the flaw in it)

src/component.js
```diff
@@ -73,7 +73,21 @@ Component.prototype.forceUpdate = function(callback) {
...
+			}
+
+			if (nextDom) {
```
developit: if it's smaller we could do:

```js
parentDom.insertBefore(dom, nextDom || null);
```

andrewiggins:
(Thinking out loud) I wonder if it would be smaller to just call `diffChildren` here... I think this is the only place where we enter our reconcilation through `diff` instead of `diffchildren` which leads us to copying what `diffchildren` does (insert elements into the DOM) here.

Might require creating an intermediate Fragment like top-level `render` does though

src/diff/children.js
```diff
@@ -82,6 +82,11 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 				}
 			}
 
+			if (lastRendered && lastRendered._component) {
+				lastRendered._component._siblingVNode = childVNode;
```
andrewiggins: Could we add the `_sibling` pointer to the VNode? Would remove a check and a property access.
```js
// diffChildren
if (lastRendered) {
	lastRendered._sibling = childVNode;
}
lastRendered = childVNode;
```

```js
// component.js
let sibling = vnode._sibling;
```

marvinhagemeister:
Closing because #1689 is an improved version of this PR :tada:

PR #1606, opened 2019-05-05
(feat) Remove support for style as string and other golfing (-32B)

cristianbote:
Alrighty! 🏌

This PR covers the support removal for styles as strings. It seems that no other vdom library supports it. ~~Also, I did a bench to figure out, if just accessing the style setters, for properties, would be better in term of perf. Here's the bench(https://esbench.com/bench/5ccb5a284cd7e6009ef6232e) also I added a comment about it.
Overall the perf doesn't get that much of a hit. I've seen small variation, but nothing like Safari. Safari would perform way better with setter than Chrome for example.~~

Also, this PR fixes a bug, where the css vars would get set as _dash_ separated. Like `--fooBaz` would end up like `--foo-Baz` due to camel case normalization.

Thanks to @jviide for the nice suggestion for improvement and keeping the `setProperty` approach which is faster; thanks @developit for properly bench-ing it!

All in all, this removes around:
~~`-50B`~~ 🎉
`-32B`


src/diff/props.js
```diff
@@ -39,39 +38,28 @@ const XLINK_NS = 'http://www.w3.org/1999/xlink';
...
 
 	if (name==='style') {
+		// Always clear the previous styles
+		dom.style.cssText = EMPTY_OBJ;
 
-		/* Possible golfing activities for setting styles:
-		 *   - we could just drop String style values. They're not supported in other VDOM libs.
-		 *   - assigning to .style sets .style.cssText - TODO: benchmark this, might not be worth the bytes.
-		 *   - assigning also casts to String, and ignores invalid values. This means assigning an Object clears all styles.
-		 */
```
cristianbote: These 🔝 were my guidance at hinting where to look into 🙏

src/diff/props.js
```diff
@@ -39,39 +38,28 @@ const XLINK_NS = 'http://www.w3.org/1999/xlink';
...
+			v = typeof v==='number' && IS_NON_DIMENSIONAL.test(i)===false ? (v + 'px') : v;
 
-		if (typeof value==='string') {
-			s.cssText = value;
-		}
-		else {
-			if (typeof oldValue==='string') s.cssText = '';
-			else {
-				// remove values not in the new list
-				for (let i in oldValue) {
```
cristianbote: Since the styles are cleared, like it was before, but now the _entire_ `cssText` gets removed, we don't need to go over the _old_ styles.

src/diff/props.js
```diff
@@ -39,39 +38,28 @@ const XLINK_NS = 'http://www.w3.org/1999/xlink';
...
+				dom.style.setProperty(i, v);
 			}
-			for (let i in value) {
-				const v = value[i];
-				if (oldValue==null || v!==oldValue[i]) {
-					s.setProperty(i.replace(CAMEL_REG, '-'), typeof v==='number' && IS_NON_DIMENSIONAL.test(i)===false ? (v + 'px') : v);
-				}
+			else {
+				// Bench for setter vs (setProperty + regex) https://esbench.com/bench/5ccb5a284cd7e6009ef6232e
+				dom.style[i] = v;
```
cristianbote: These are almost the same in terms of perf. The upside is in code size.

cristianbote:
Let me know what you guys think about it and what implications I might've overlooked

src/diff/props.js
```diff
@@ -39,39 +38,28 @@ const XLINK_NS = 'http://www.w3.org/1999/xlink';
  * @param {boolean} isSvg Whether or not this DOM node is an SVG node or not
  */
 function setProperty(dom, name, value, oldValue, isSvg) {
+	let v;
+
 	name = isSvg ? (name==='className' ? 'class' : name) : (name==='class' ? 'className' : name);
 
 	if (name==='style') {
+		// Always clear the previous styles
+		dom.style.cssText = EMPTY_OBJ;
```
JoviDeCroock: Elaborate, I like it :D

src/diff/props.js
JoviDeCroock: I think like this we support IE9 again.

cristianbote:
@JoviDeCroock thank you! That coverage issue just baffled me, as it was reporting that those line were not covered, but the tests are running on those lines 😄 lol. Gonna do your suggestion.

wardpeet:
I can add a codemods to https://github.com/preactjs/migrate-preact-x so we heave an easy conversion

JoviDeCroock:
> I can add a codemods to https://github.com/preactjs/migrate-preact-x so we heave an easy conversion

That would be awesome :D we could add it in the `/preact` team.

src/diff/props.js
developit: It seems odd to use an Object here. It's basically doing:

```js
dom.style.cssText = '[object Object]'
```

... because the cssText setter coerces to a String.

developit:
@cristianbote My fear is that this PR actually does impact performance in the wild quite a bit, we just don't have a performance test that shows it.

src/diff/props.js
developit: this is really expensive though. it means changing one style property changes all style properties.

src/diff/props.js
cristianbote: Gonna change it. Thanks!

src/diff/props.js
cristianbote: Yeah. That's a great point. I'll redo this PR.

src/diff/props.js
cristianbote: Not anymore 😦

src/diff/props.js
```diff
@@ -114,4 +100,4 @@ function setProperty(dom, name, value, oldValue, isSvg) {
  */
 function eventProxy(e) {
 	return this._listeners[e.type](options.event ? options.event(e) : e);
-}
+}
```
JoviDeCroock: Missing a new line here

src/diff/props.js
```diff
@@ -42,36 +43,21 @@ function setProperty(dom, name, value, oldValue, isSvg) {
...
-				}
+		const set = assign(assign({}, oldValue), value);
+		for (let i in set) {
+			if ((value || EMPTY_OBJ)[i] === (oldValue || EMPTY_OBJ)[i]) {
+				continue;
 			}
+			dom.style.setProperty(
+				(i[0] === '-' && i[1] === '-') ? i : i.replace(CAMEL_REG, '-$&'),
+				(value && (i in value))
+					? (typeof set[i]==='number' && IS_NON_DIMENSIONAL.test(i)===false)
```
JoviDeCroock: I don't know if this saves bytes but typeof should be safe to use `==` I think.

cristianbote: It's reporting the same size, sadly 😞

src/diff/props.js
cristianbote: Oops! 😄 Thank you

src/diff/props.js
jviide: The `dom.style[i]` thing is a clever way of getting rid of the camel case transformation, and something like this seems to get rid of 6 extra bytes:

```js
const v = (value && (i in value))
	? (typeof set[i]==='number' && IS_NON_DIMENSIONAL.test(i)===false)
		? set[i] + 'px'
		: set[i]
	: '';
if (i[0] === '-' && i[1] === '-') {
	dom.style.setProperty(i, v);
}
else {
	dom.style[i] = v;
}
```

cristianbote: The problem with assignment is that is a setter that would trigger parsing for that property. `setProperty` is way less overhead, hence the need to do the camelCase to dash-case parsing. Check-out this bench from @developit 
image
https://esbench.com/bench/5730613ddb965b9a00965a5b

jviide: Ah, thats true. Running the benchmark on Chrome 74 + MacOS 10.14 gives the following results:
Screenshot 2019-05-08 at 11 47 02
It must be noted that this benchmark doesn't take into account for the `setProperty` approach to run the property names through the regex. With small modifications to the `setProperty` approach that more closely resemble this pull request (run property names through the regex + use `setProperty(a, '')` instead of `removeProperty(a)`) the difference becomes somewhat smaller:
Screenshot 2019-05-08 at 11 58 10
Interestingly on Safari 12.1 and Firefox 66 the `properties with check` approach is faster (after the mentioned modifications to the benchmark).
Safari:
Screenshot 2019-05-08 at 11 56 08
Firefox:
Screenshot 2019-05-08 at 11 52 24

src/diff/props.js
ForsakenHarmony: probably doesn't matter for gzip

marvinhagemeister:
@cristianbote Can we add a test for the bug with the dash in CSS vars? That way we'll make sure we don't regress there whenever we touch this section 🙂

cristianbote:
I'd be happy to 😄 I thought I added one though 🤔 but doesn't seem like it

src/diff/props.js
cristianbote: Seems on par with my initial bench, tbh https://esbench.com/bench/5ccb5a284cd7e6009ef6232e.

@developit so, does it make sense? doing `setProperty` + regex is in the same category as calling the `setter`? By doing just the `setter` we're shaving some nice `6B` 😄 . I'm gonna try to create some more randomly, uncache-able code path and see if that still holds.

src/diff/props.js
```diff
@@ -42,36 +43,21 @@ function setProperty(dom, name, value, oldValue, isSvg) {
...
-				if (oldValue==null || v!==oldValue[i]) {
-					s.setProperty(i.replace(CAMEL_REG, '-'), typeof v==='number' && IS_NON_DIMENSIONAL.test(i)===false ? (v + 'px') : v);
-				}
+		const set = assign(assign({}, oldValue), value);
+		for (let i in set) {
+			if ((value || EMPTY_OBJ)[i] === (oldValue || EMPTY_OBJ)[i]) {
+				continue;
 			}
+			dom.style.setProperty(
+				(i[0] === '-' && i[1] === '-') ? i : i.replace(CAMEL_REG, '-$&'),
```
apfelbox: Not sure about the performance impact, but this is short: `/^--/.test(i)`

I made a ESBench, but it's all over the place though: https://esbench.com/bench/5cd320f74cd7e6009ef623bf

marvinhagemeister:
@cristianbote shoot it looks like our own css sorting function doesn't account for dashes inside variable names. From the looks of it Chrome seems to sort it differently :S

src/diff/props.js
jviide: Cool, that does cut down preact.js.gz size down by 3 bytes indeed! Changing the benchmark from `text += "match"` to something like `count += 1` (with `let count = 0;` in the setup) stabilized the results to something like this (Chrome 74, MacOS 10.14.4):
Screenshot 2019-05-08 at 22 23 32

cristianbote:
@marvinhagemeister thank you! 😄

src/diff/props.js
cristianbote: Hey @afeiship yeah, it compresses better I guess, but "direct match" is waaay faster 😄 and we might get to save 3B from somewhere else.

apfelbox: @jviide nice thanks for the analysis.

I had made tests across the browser but I had so highly varying results (sometimes the precompiled was even faster than direct.. 🙈 ) so I wasn't sure about how much of a performance hit it is (apparently a brutal one 😄 ), just wanted to put it up for discussion. 😅

jviide: Much appreciated - keep 'em coming! 👍 I was pretty surprised how big the performance difference was (and there's always the possibility that I botched the benchmark).

@marvinhagemeister pointed out that `setProperty` in `props.js` uses the same approach to identify event handlers, based on a similar benchmark: https://github.com/developit/preact/blob/7e53dd29003bbbddf5bdbc934c8e00c69b57936c/src/diff/props.js#L75-L76

Issue #1608, opened 2019-05-06
compat missing typings


pmkroeker:
Currently there are no external typings for `preact/compat`. It would be nice to add them in for those of us who would like them! I may be able to help out depending on how busy I am.

pmkroeker:
Fixed in #1609

PR #1609, opened 2019-05-06
Add types to `compat`

pmkroeker:
This is still a WIP of adding export types with `preact/compat` for those that wish to use import things directly through compat.

Some things are still being hashed out, but wanted some feedback.

Related #1608

### DIFF
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -31,8 +31,6 @@ declare namespace compat {
 
   interface PureComponent<P = {}, S = {}> extends Component {
     isPureReactComponenet: boolean;
-
-    shouldComponentUpdate?(props: Readonly<P>, state: Readonly<S>): boolean;
   }
 
   function memo<P = {}>(component: FunctionalComponent<P>, comparer?: (prev: P, next: P) => boolean): FunctionComponent<P>;
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -1,0 +1,71 @@
+import * as hooks from '../../hooks';
+import { VNode, PreactElement } from '../../src/internal';
+import { Component, createContext, createRef, Fragment, toChildArray, ComponentChildren, ComponentChild } from '../../src';
+import { FunctionalComponent } from 'preact';
+import { ForwardFn } from './internal';
+
+export { Component, createContext, createRef, Fragment } from '../../src';
+export * from '../../hooks';
+
+export const version: string;
+
+export function createPortal (vnode: VNode, container: PreactElement): VNode<any>;
+
+export function createElement (...args: VNode<any>): VNode<any>;
+
+export function cloneElement (element: VNode<any>): VNode<any>;
+
+export function render (vnode: VNode<any>, parent: PreactElement, callback?: () => void): Component | null;
+
+export function unmountComponentAtNode (container: Element | Document | ShadowRoot | DocumentFragment): boolean;
+
+export function createFactory (type: VNode["type"]): VNode<any>;
+
+export function isValidElement (element: any): boolean;
+
+export function findDOMNode (component: Component): PreactElement | null;
+
+export class PureComponent extends Component<P, S> {
+  public isPureReactComponenet: boolean;
+
+  public shouldComponentUpdate (props: P, state: S): boolean;
+}
+
+export function memo (c: FunctionalComponent, comparer: (prev: any, next: any) => boolean): FunctionalComponent;
+
+export function forwardRef (fn: ForwardFn): FunctionalComponent;
+
+export function unstable_batchedUpdates (callback: () => void, arg?: any): void;
+
+function mapFn (children: ComponentChildren, fn: (d: any, i: number) => any): Array<VNode | null> | null;
+
+export interface Children {
+  map: mapFn;
+  forEach: mapFn;
+  count: (children: ComponentChildren) => boolean;
+  only: (children: ComponentChildren) => ComponentChild;
+  toArray: toChildArray;
+}
+
+export default {
+  Children,
+  hooks,
+  version,
+  createPortal,
+  createElement,
+  cloneElement,
+  render,
+  hydrate: render,
+  unmountComponentAtNode,
+  createContext,
+  createFactory,
+  createRef,
+  Fragment,
+  isValidElement,
+  findDOMNode,
+  Component,
+  PureComponent,
+  memo,
+  forwardRef,
+  unstable_batchedUpdates,
+}
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -1,5 +1,5 @@
 import * as _hooks from '../../hooks';
-import { VNode, PreactElement } from '../../src/internal';
+import { PreactElement } from '../../src/internal';
 import * as preact from '../../src';
 import { ForwardFn } from './internal';
 
@@ -16,15 +16,15 @@ declare namespace compat {
 
   export const version: string;
 
-  export function createPortal(vnode: VNode, container: PreactElement): VNode<any>;
+  export function createPortal(vnode: preact.VNode, container: PreactElement): preact.VNode<any>;
 
-  function render(vnode: VNode<any>, parent: PreactElement, callback?: () => void): Component | null;
+  function render(vnode: preact.VNode<any>, parent: PreactElement, callback?: () => void): Component | null;
 
-  function hydrate(vnode: VNode<any>, parent: PreactElement, callback?: () => void): Component | null;
+  function hydrate(vnode: preact.VNode<any>, parent: PreactElement, callback?: () => void): Component | null;
 
   function unmountComponentAtNode(container: Element | Document | ShadowRoot | DocumentFragment): boolean;
 
-  function createFactory(type: VNode["type"]): VNode<{}>;
+  function createFactory(type: preact.VNode["type"]): preact.VNode<{}>;
 
   function isValidElement(element: any): boolean;
 
@@ -46,6 +46,6 @@ declare namespace compat {
     forEach<T extends preact.ComponentChild>(children: T | T[], fn: (child: T, i: number, array: T[]) => void): void;
     count: (children: preact.ComponentChildren) => number;
     only: (children: preact.ComponentChildren) => preact.ComponentChild;
-    toArray: (children: preact.ComponentChildren) => VNode<{}>[];
+    toArray: (children: preact.ComponentChildren) => preact.VNode<{}>[];
   }
 }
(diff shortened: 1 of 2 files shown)


compat/src/index.d.ts
```diff
@@ -0,0 +1,71 @@
...
+export function forwardRef (fn: ForwardFn): FunctionalComponent;
+
+export function unstable_batchedUpdates (callback: () => void, arg?: any): void;
+
+function mapFn (children: ComponentChildren, fn: (d: any, i: number) => any): Array<VNode | null> | null;
+
+export interface Children {
+  map: mapFn;
+  forEach: mapFn;
+  count: (children: ComponentChildren) => boolean;
```
marvinhagemeister: `count()` returns a `number`

compat/src/index.d.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export function forwardRef (fn: ForwardFn): FunctionalComponent;
+
+export function unstable_batchedUpdates (callback: () => void, arg?: any): void;
+
+function mapFn (children: ComponentChildren, fn: (d: any, i: number) => any): Array<VNode | null> | null;
+
+export interface Children {
+  map: mapFn;
+  forEach: mapFn;
```
marvinhagemeister: Although both `map` and `forEach` share the same implementation internally for byte saving reasons, we should not make that public.

```ts
export interface Children {
  map<T extends ComponentChild, R>(children: T | T[], fn: (child: T, i: number, array: T[]) => R): R[];
  forEach<T extends ComponentChild, R>(children: T | T[], fn: (child: T, i: number, array: T[]) => void): void;
  // ...
}
```

compat/src/index.d.ts
```diff
@@ -0,0 +1,71 @@
...
+  public isPureReactComponenet: boolean;
+
+  public shouldComponentUpdate (props: P, state: S): boolean;
+}
+
+export function memo (c: FunctionalComponent, comparer: (prev: any, next: any) => boolean): FunctionalComponent;
+
+export function forwardRef (fn: ForwardFn): FunctionalComponent;
+
+export function unstable_batchedUpdates (callback: () => void, arg?: any): void;
```
marvinhagemeister: The `callback` will be called with `arg`, so the type for `callback` should be `(arg?: any) => void`

compat/src/index.d.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export function findDOMNode (component: Component): PreactElement | null;
+
+export class PureComponent extends Component<P, S> {
+  public isPureReactComponenet: boolean;
+
+  public shouldComponentUpdate (props: P, state: S): boolean;
+}
+
+export function memo (c: FunctionalComponent, comparer: (prev: any, next: any) => boolean): FunctionalComponent;
```
marvinhagemeister: `comparer` should be optional and`memo` is used to compare `props`. We should add that into our types:

```ts
export function memo<P = {}>(component: FunctionalComponent<P>, comparer?: (prev: P, next: P) => boolean): FunctionComponent<P>;
```

compat/src/index.d.ts
```diff
@@ -0,0 +1,71 @@
...
+export function unmountComponentAtNode (container: Element | Document | ShadowRoot | DocumentFragment): boolean;
+
+export function createFactory (type: VNode["type"]): VNode<any>;
+
+export function isValidElement (element: any): boolean;
+
+export function findDOMNode (component: Component): PreactElement | null;
+
+export class PureComponent extends Component<P, S> {
+  public isPureReactComponenet: boolean;
```
marvinhagemeister: No need for `public`, all properties or methods are `public` by default in TS :+1:

compat/src/index.d.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export function createElement (type: VNode["type"], props?: any, Children?: ComponentChildren): VNode<any>;
+
+export function cloneElement (element: VNode<any>): VNode<any>;
+
+export function render (vnode: VNode<any>, parent: PreactElement, callback?: () => void): Component | null;
+
+export function unmountComponentAtNode (container: Element | Document | ShadowRoot | DocumentFragment): boolean;
+
+export function createFactory (type: VNode["type"]): VNode<any>;
```
marvinhagemeister: Return type should be `VNode<{}>`, because `props` will be an empty `object`.

compat/src/index.d.ts
```diff
@@ -0,0 +1,71 @@
...
+export { Component, createContext, createRef, Fragment } from '../../src';
+export * from '../../hooks';
+
+export const version: string;
+
+export function createPortal (vnode: VNode, container: PreactElement): VNode<any>;
+
+export function createElement (type: VNode["type"], props?: any, Children?: ComponentChildren): VNode<any>;
+
+export function cloneElement (element: VNode<any>): VNode<any>;
```
marvinhagemeister: Maybe we can reuse the type from `preact`?

compat/src/index.d.ts
```diff
@@ -0,0 +1,71 @@
...
+import { ForwardFn } from './internal';
+
+export { Component, createContext, createRef, Fragment } from '../../src';
+export * from '../../hooks';
+
+export const version: string;
+
+export function createPortal (vnode: VNode, container: PreactElement): VNode<any>;
+
+export function createElement (type: VNode["type"], props?: any, Children?: ComponentChildren): VNode<any>;
```
marvinhagemeister: Let's reuse the typings from `preact` :+1:

compat/src/index.d.ts
pmkroeker: Oops yep, fixed!

compat/src/index.d.ts
```diff
@@ -0,0 +1,71 @@
...
+export function createFactory (type: VNode["type"]): VNode<any>;
+
+export function isValidElement (element: any): boolean;
+
+export function findDOMNode (component: Component): PreactElement | null;
+
+export class PureComponent extends Component<P, S> {
+  public isPureReactComponenet: boolean;
+
+  public shouldComponentUpdate (props: P, state: S): boolean;
```
marvinhagemeister: same here :)

compat/src/index.d.ts
pmkroeker: Right, I'm used to my eslint config which enforces this is explicit. I have updated this.

marvinhagemeister: Or even better: Let's just reuse the type from `internal` :100:

compat/src/index.d.ts
pmkroeker: Updated.

compat/src/index.d.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+export function memo (c: FunctionalComponent, comparer: (prev: any, next: any) => boolean): FunctionalComponent;
+
+export function forwardRef (fn: ForwardFn): FunctionalComponent;
```
marvinhagemeister: We can pass the types for `props` through to the return type:

```ts
export function forwardRef<P = {}>(fn: ForwardFn<P, any>): FunctionalComponent<P>;
```

compat/src/index.d.ts
pmkroeker: Right you are. Fixed.

compat/src/index.d.ts
```diff
@@ -0,0 +1,52 @@
...
+  function memo (c: FunctionalComponent, comparer: (prev: any, next: any) => boolean): FunctionalComponent;
+
+  function forwardRef (fn: ForwardFn): FunctionalComponent;
+
+  function unstable_batchedUpdates (callback: (arg?: any) => void, arg?: any): void;
+
+
+  interface Children {
+    map<T extends ComponentChild, R>(children: T | T[], fn: (child: T, i: number, array: T[]) => R): R[];
+    forEach<T extends ComponentChild, R>(children: T | T[], fn: (child: T, i: number, array: T[]) => void): void;
```
marvinhagemeister: Made a mistake in my comment: The `R` parameter is not necessary :sweat_smile:

pmkroeker:
@marvinhagemeister thanks. Ya this is my first foray into adding types to a larger package. And I'll fix the spacing thing too!

compat/src/index.d.ts
pmkroeker: I moved things all into a `compat` namespace so they can be used via default import as well. How do I add imported types to an exported namespace? `import export` worked for hooks as it was already a namespace.

compat/src/index.d.ts
pmkroeker: Removed the duplicated entry. Fixed!

compat/src/index.d.ts
pmkroeker: Got it figured out! This has been completed

pmkroeker:
I just changed over from using `PreactElement` to `Element`. I pulled some of the initial typings from the JSDoc comments, but it seems like some of the places that use `PreactElement` in fact require just `Element`  (mostly due do passing something like `document.querySelector('#root')`.

@marvinhagemeister does this seem ok?

compat/src/index.d.ts
```diff
@@ -0,0 +1,50 @@
+import * as _hooks from '../../hooks';
+import * as preact from '../../src';
+import { ForwardFn } from './internal';
+
+export = compat;
+export as namespace compat;
+declare namespace compat {
```
ForsakenHarmony: do we need a namespace here?

pmkroeker: Likely not, I was hoping to try and facilitate both exporting everything as default, and individually using this, but it doesn't seem to work as I had anticipated.

pmkroeker: I removed the namespace. Still thinking it would be nice to support a default export as well as named, but nothing I have seem really shows a nice way of doing this.

marvinhagemeister: There is always the option to use an `object` for the default export:

```ts
export const foo: number;
export function bar() {}

export default {
  foo,
  bar,
}
```

marvinhagemeister:
It's coming along very well! I'd just add the default export like mentioned in the comments and than it's ready to be merged in my opinion :+1:

pmkroeker:
I did some testing and it seems as though the way the `index.js` file is created, TS is able to map the definitions to a default export as well.

```typescript
import React, { createPortal } from 'preact/compat';

React.createPortal(); // gives missing args, vnode not provided
createPortal(); // same message as above
```

Both of those behave the same, plus intellisense over `React` shows all the appropriate methods.

compat/src/index.d.ts
pmkroeker: Resolved in the most recent commit. Thanks!

PR #1613, opened 2019-05-07
(fix) - .some should not be used for effect invoke/cleanup

JoviDeCroock:
.some could potentially short circuit when true is returned from …effect invocation or cleanup

Thanks @ForsakenHarmony for pointing this out.


ForsakenHarmony:
map would've worked as well

PR #1617, opened 2019-05-08
(fix) - don't compare when oldVNode has no dom

JoviDeCroock:
Fixes: https://github.com/developit/preact/issues/1616

This issue is caused when we insert a value into the array, it will try to calculate an offset this way we can maintain focus when we add/remove elements.

I've found a  test where checking for oldDom effectively fails focus so this is not the way to go.

Seems like removing the focusElement is having some side-effects :/ ~~a more effective approach would be to see if the node we are comparing to has an old version or not.~~ <-- this is also faulty....

Removed my fix since it introduced a faulty scenario in the new focus test I added.

We'll need a more reliable way to check for offsets (which is hard with checking on .type) or reintroduce .focusElement.

The only thing that could fix it in both scenario's according to me is when we could say hey this element is a new one (inserted). That way we could say oldDom for this new element is null. This is because the algo notices we have an offset but doesn't know where so it starts looking for the first type match and sets the oldDom to that (which didn't used to work like that) but now it does....

The full erroneous scenario is as follows:
- we render our UI correctly
- we add an element on position 1 of the array
- while diffChildren when we arrive at our last button we notice that there is no corresponding child
- we loop over the children
- button1 is now on position 2 (where button 2 was) they are equal on type button
- button2 is now on position 3 (where button 3 was) they are equal on type button
- button3 is now on position 4 (empty), this loops over the array to find the first button occurrence which in this case is button1. This will trigger the oldDom setting since oldChild.type won't be a button and thus make button3 use the oldDom of button1 switching their position

Wondering if it's better to revisit the focus approach where when inserting it constructed an erroneous parentVDOM.childNodes.

In the end it all boils down to offsets being hard to manage without keys.

### DIFF
--- a/compat/package.json
+++ b/compat/package.json
@@ -8,6 +8,7 @@
   "module": "dist/compat.module.js",
   "umd:main": "dist/compat.umd.js",
   "source": "src/index.js",
+  "types": "src/index.d.ts",
   "license": "MIT",
   "mangle": {
     "regex": "^_"
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -1,0 +1,71 @@
+import * as _hooks from '../../hooks';
+import * as preact from '../../src';
+import { ForwardFn } from './internal';
+
+export * from '../../hooks';
+export import Component = preact.Component;
+export import createContext = preact.createContext;
+export import createRef = preact.createRef;
+export import Fragment = preact.Fragment;
+export import createElement = preact.createElement
+export import cloneElement = preact.cloneElement
+
+export declare const version: string;
+
+export declare function createPortal(vnode: preact.VNode, container: Element | Element): preact.VNode<any>;
+
+export declare function render(vnode: preact.VNode<any>, parent: Element, callback?: () => void): Component | null;
+
+export declare function hydrate(vnode: preact.VNode<any>, parent: Element, callback?: () => void): Component | null;
+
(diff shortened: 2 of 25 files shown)


JoviDeCroock:
Another option would be to see that if we're comparing a new item to skip this step:
https://github.com/developit/preact/blob/master/src/diff/children.js#L115
That would effectively make this work:

```
LOG: 'diffed', function ListItem(_ref2){ ... }, Object{children: '2'}, function ListItem(_ref2){ ... }, <input type="text">
LOG: 'diffed', function Input(){ ... }, Object{}, function Input(){ ... }, <span>3</span>
```

Focus loses because we diff like this:

```
<span>0</span> <-- gets diffed to span 1
<span>1</span> <-- gets diffed to span 3
<Input /> <-- gets diffed to an oldDom of <span>1</span> <-- unmount remound
<span>3</span>
```

andrewiggins:
I think we should take this change as it stands right now, and investigate the possibility of maintaining focus without keys in a separate PR.

@JoviDeCroock, do you mind if I push some updates to the focus tests to your PR here? Basically I just added HTML validation to all the focus tests to catch similar issues like this one in the future.

JoviDeCroock:
@andrewiggins If you have them ready go for it, else I'll push them tonight. Since I'm using them on another test branch aswell.

PR #1618, opened 2019-05-08
(fix) Self is not available in non-browser env

### LINKED ISSUE #1612: Missing `self` global creates confusing error message when running tests under Node
https://github.com/developit/preact/pull/1590 replaced a test for event handlers in an element with a reference to `self`. Since the `self` global does not exist in Node, this causes confusing error messages when trying to run tests under eg. JSDOM. See https://github.com/preactjs/enzyme-adapter-preact-pure/issues/41

I'm fairly sure the enzyme-adapter-preact-pure tests are going to fail once Preact is updated to the latest version.

cristianbote:
Fixes #1612
~~Adds `+11B`~~ ☹️
Adds `+7B`😐

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -76,7 +76,7 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 	else if (name[0]==='o' && name[1]==='n') {
 		let useCapture = name !== (name=name.replace(/Capture$/, ''));
 		let nameLower = name.toLowerCase();
-		name = (nameLower in self ? nameLower : name).substring(2);
+		name = ((typeof self!=="undefined" && nameLower in self) ? nameLower : name).slice(2);
 
 		if (value) {
 			if (!oldValue) dom.addEventListener(name, eventProxy, useCapture);
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -76,7 +76,7 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 	else if (name[0]==='o' && name[1]==='n') {
 		let useCapture = name !== (name=name.replace(/Capture$/, ''));
 		let nameLower = name.toLowerCase();
-		name = ((typeof self!=="undefined" && nameLower in self) ? nameLower : name).slice(2);
+		name = ((typeof self!=='undefined' && nameLower in self) ? nameLower : name).slice(2);
 
 		if (value) {
 			if (!oldValue) dom.addEventListener(name, eventProxy, useCapture);
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -76,7 +76,7 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 	else if (name[0]==='o' && name[1]==='n') {
 		let useCapture = name !== (name=name.replace(/Capture$/, ''));
 		let nameLower = name.toLowerCase();
-		name = ((typeof self!=='undefined' && nameLower in self) ? nameLower : name).slice(2);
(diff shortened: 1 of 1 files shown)


src/diff/props.js
```diff
@@ -76,7 +76,7 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 	else if (name[0]==='o' && name[1]==='n') {
 		let useCapture = name !== (name=name.replace(/Capture$/, ''));
 		let nameLower = name.toLowerCase();
-		name = (nameLower in self ? nameLower : name).substring(2);
+		name = ((typeof self!=='undefined' && nameLower in self) ? nameLower : name).slice(2);
```
andrewiggins: Hmm is there a reason we do `nameLower in self` instead of `nameLower in dom`?

JoviDeCroock: dom does not contain certain events like for example the animation API these are defined on window but not on dom nodes.

cristianbote: Hey @andrewiggins 👋 Yeah, so we should support the events from Global Event Interface(https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers) and some of these events, like `ontransitionend`, are not on the `dom` instance.

This was added as part of https://github.com/developit/preact/pull/1590

Do you see any issues with this?

robertknight:
I had another look into the original motivation for the change to avoid doing `nameLower in dom`. It looks like there is an inconsistency between Safari, Firefox and Chrome here. This evaluates to `true`. in Safari and Firefox but `false` in Chrome:

```js
'onanimationend' in document.body
```

I wonder whether this is simply not specced or if it is, which browser is incorrect.

robertknight:
Just a wild thought: Would we break any common use cases if the code just always lower-cased all event names and required consumers to manually attach event listeners if they wanted to listen for a non-lower-cased event?

src/diff/props.js
developit: what about `nameLower in dom.ownerDocument`?

developit:
@robertknight that's what Preact used to do. It breaks a few native DOM events, and also makes using Custom Elements really cumbersome. It was responsible for all of our failing tests on custom-elements-everywhere.com

src/diff/props.js
andrewiggins: Right! I had forgotten about that. Thanks

cristianbote: Using @robertknight suggestion `dom.ownerDocument.defaultView` is actually what were looking for? Since `dom.ownerDocument === document` and these Global Events are not defined on `document` but rather on `window`. 

This will actually turn-up smaller in size by just adding `+7B` to core

robertknight:
The [spec for defaultView](https://developer.mozilla.org/en-US/docs/Web/API/Document/defaultView) notes that it can be `null`.  For example:

```js
var parser = new DOMParser();
var doc = parser.parseFromString('<b>foo</b>', 'text/html');
doc.defaultView // null
```

Are there any other cases where this might happen that folks can think of?

It won't be `null` in the normal case of rendering into a browser, and it should be fine in JSDOM which was the problem with `self`. You could have something like the following as a fallback:

```js
eventName in (dom.ownerDocument.defaultView || dom)
```

robertknight:
I filed an upstream issue in Chrome to get some input on the inconsistency between browsers: https://bugs.chromium.org/p/chromium/issues/detail?id=961193&can=1&q=ontransitionend

developit:
This is likely to regress event handler performance by 10-20%:

image

https://esbench.com/bench/5cd8af874cd7e6009ef623fd

PR #1626, opened 2019-05-12
Make JSON injection prevention hot loader compatible

### LINKED ISSUE #1620: 10.0.0-beta.0 -> 10.0.0-beta.1, react-hot-loader no longer renders the exported component
reproduction repository: https://github.com/reaysawa/starter-preact-scss

1. clone https://github.com/reaysawa/starter-preact-scss
2. `yarn && yarn dev` should work with hot-reloading :heavy_check_mark: 
3. `yarn add preact@next` to upgrade from _10.0.0-beta._**0** to _10.0.0-beta._**1**
4. `yarn && yarn dev` again, but this time it doesn't render :red_circle:

sventschui:
Fixes #1620 / Adds 2B

### DIFF
--- a/mangle.json
+++ b/mangle.json
@@ -34,8 +34,7 @@
       "$_defaultValue": "__p",
       "$_id": "__c",
       "$_parentDom": "__P",
-      "$_self": "_",
-      "$__self": "__"
+      "$_self": "_"
     }
   }
 }
--- a/mangle.json
+++ b/mangle.json
@@ -34,7 +34,8 @@
       "$_defaultValue": "__p",
       "$_id": "__c",
       "$_parentDom": "__P",
-      "$_self": "_"
+      "$_self": "_",
+      "$__self": "__"
     }
   }
 }
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -62,7 +62,7 @@ export function createVNode(type, props, key, ref) {
 		_lastDomChild: null,
 		_component: null
 	};
-	vnode._self = vnode;
+	vnode._self = vnode.__self = {};
 
 	if (options.vnode) options.vnode(vnode);
 
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -1,5 +1,6 @@
 import options from './options';
 import { assign } from './util';
+import { EMPTY_OBJ } from './constants';
 
 /**
   * Create an virtual node (used for JSX)
@@ -62,7 +63,7 @@ export function createVNode(type, props, key, ref) {
 		_lastDomChild: null,
 		_component: null
 	};
-	vnode._self = vnode.__self = {};
+	vnode._self = EMPTY_OBJ;
 
 	if (options.vnode) options.vnode(vnode);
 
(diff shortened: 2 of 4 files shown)


JoviDeCroock:
I just had an idea to make this safer, we could maybe use `EMPTY_OBJECT` which is defined in our constants and check if _self is equal to EMPTY_OBJ

sventschui:
Change it to your proposed solution with `EMPTY_OBJ` from constants. This not only drops 3B but also makes the tests work again :)

JoviDeCroock:
Oh, thought the tests had something to do with `toChildArray`. Hope it's the best solution, will think about this a bit.

developit:
The problem is that this means it's now impossible to create static VNodes, since `._self` points to a Preact internal object.

sventschui:
There are two problems I see with using the `__proto__` approach:

- One might accidentally safe-cast a value from an API by spreading into a new object. This will happen when using redux/useReducer as spreading is quite common in these cases. I also don't know what libraries like `apollo-cache(-inmemory)` do: ``({...JSON.parse('{"foo":"bar"}')}).__proto__` results in an empty object proto and thus to be truthy.

> Another story: Did you guys ever try to use `Object.create(null)` when assigning things to a new object or in other situations? I know that `pino` and I guess also `fastify` use this approach to speed up property access on objects as the runtime does not have to traverse the proto chain.

- When using something different than `JSON.parse` it isn't guaranteed that the objects returned have a null proto. Imagine someone is using XML (yeah I know it's 2019) or protobuf.

What came to my ming is using `createElement` / `h` instead of `EMPTY_OBJ` as it is exposed from preact. What this doesn't solve though is rendering an element created by a different preact lib than the one rendering it.

We might want to assign an empty function to `_self` and check for `typeof _self === 'function` but that definitely costs more bytes. Don't know what this means for performance

JoviDeCroock:
Assigning it to h or createElement won’t be unique across different preact distributions either.

sventschui:
> or insight, that failing test was because your function child got executed because _self and other were both undefined, these are strictly equal. My tired head was not processing that yesterday.

That should be fixable by checking `newVNode._self!==undefined&&newVNode._self!==newVNode.__self` instead of `newVNode._self!==newVNode.__self` here: https://github.com/developit/preact/pull/1626/commits/5fe8fcf514d2752b2dd43fd21d7fb97e8a5e023f#diff-44070032789f10581d1b70740a2aaacfR40

> I do think that creating a function on every vnode (and every vnode render) will be quite expensive.
No need to re-create the function, we can keep it in a global variable. What is more expensive tough is the typeof check if I'm not mistaken

JoviDeCroock:
Storing in a global would imply the same issue mentioned by @developit. Window storage would also malfunction in node,...

sventschui:
by global I mean having a `const noop = () => {}` somewhere in `createElement`. When then just checking for `typeof vnode._self === 'function'` one can still create static vnodes (e.g. `vnode = { _self: () => {} }`)

JoviDeCroock:
Yes, as mentioned by @jviide we could reuse a preact function if we're using typeof checking since it doesn't really matter in terms of ref.

developit:
I should correct my __proto__ point: the check is for `undefined`, not `null`:

```js
JSON.parse('{"__proto__":null}').__proto__ // null (tainted)
// vs:
({"__proto__":null}).__proto__ // undefined (literal)
```

PR #1628, opened 2019-05-13
(chore) - support __source for jsx-sourcemaps

JoviDeCroock:
Support https://babeljs.io/docs/en/next/babel-plugin-transform-react-jsx-source.html

Fixes: https://github.com/developit/preact/issues/1571

How do we want to incorporate this in our debug messages?

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -140,10 +140,12 @@ export function initDebug() {
 	};
 
 	options.vnode = (vnode) => {
+		let source;
 		if (vnode.props && vnode.props.__source) {
-			vnode.__source = vnode.props.__source;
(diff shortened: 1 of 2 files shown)


debug/src/debug.js
```diff
@@ -140,6 +140,10 @@ export function initDebug() {
 	};
 
 	options.vnode = (vnode) => {
+		if (vnode.props && vnode.props.__source) {
```
developit: It'd be better for performance to always define this property when in dev mode:

```js
let source;
if (vnode.props && vnode.props.__source) {
  source = vnode.props.__source;
  delete vnode.props.__source;
}
vnode._source = source;
```

PR #1629, opened 2019-05-13
(fix) - Insert portal children instead of overwriting the container

JoviDeCroock:
Fixes: https://github.com/developit/preact/issues/1615

I tried this first with `preactRender(wrap, container, temp)` but it did not overwrite the replaceNode param. Thought this would function that way but apparently I was wrong. If I'm misunderstanding this concept please feel free to say so


compat/src/index.js
```diff
@@ -71,9 +71,14 @@ class ContextProvider {
  */
 function Portal(props) {
 	let wrap = h(ContextProvider, { context: this.context }, props.vnode);
-	render(wrap, props.container);
+	let container = props.container;
+	container._prevVNode = null;
```
cristianbote: Could `hydrate` be used here instead? `hydrate` does this inside it. So, we could end-up with:
```suggestion
	let temp = props.container.appendChild(document.createTextNode(''));
	hydrate(wrap, container);
```

I guess this would compress better, maybe? I couldn't verify it atm.

compat/test/browser/portals.test.js
```diff
@@ -73,7 +86,7 @@ describe('Portal', () => {
 		}
 
 		render(<App />, root);
-		expect(dialog.childNodes.length).to.equal(1);
+		expect(dialog.childNodes.length).to.equal(2);
```
JoviDeCroock: I don't know if this is acceptable or not....

marvinhagemeister: Not sure if I understand it correctly: Does this mean that we leave a DOM node behind where the `Portal` was?

JoviDeCroock: We push an empty textNode (https://github.com/developit/preact/pull/1629/files#diff-efd159ddcc099e90029db3b306b8646aR75) to ensure we are inserting the portal contents and not overriding all the already rendered children.

compat/src/index.js
```diff
@@ -71,9 +71,12 @@ class ContextProvider {
...
 function Portal(props) {
 	let wrap = h(ContextProvider, { context: this.context }, props.vnode);
-	render(wrap, props.container);
+	let container = props.container;
+	let temp = props.container.appendChild(document.createTextNode(''));
+	hydrate(wrap, container);
 	this.componentWillUnmount = () => {
-		render(null, props.container);
+		container.removeChild(temp);
+		render(null, container);
```
ForsakenHarmony: and this won't wipe the dom when it gets unmounted?

JoviDeCroock: That's a good remark, it was like this before aswell. I'll double check but it hasn't done that in tests yet.

PR #1633, opened 2019-05-15
chore: add travis-size-report

wardpeet:
Adding travis-size-report to keep our bundlesizes in check for each PR so we don't have to do it manually.

https://github.com/developit/microbundle/pull/392

### DIFF
--- a/sizereport.config.js
+++ b/sizereport.config.js
@@ -1,5 +1,5 @@
 module.exports = {
 	repo: 'developit/preact',
-	path: ['./{compat,debug,hooks,}/dist/**/!(*.js.map)'],
+	path: ['./{compat,debug,hooks,}/dist/**/!(*.map)'],
 	branch: 'master'
 };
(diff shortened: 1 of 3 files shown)


sizereport.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+	repo: 'developit/preact',
+	path: ['dist/*', 'compat/dist/*', 'debug/dist/*', 'hooks/dist/*'],
```
cristianbote: Maybe we should exclude `.map` files? Or be specific by adding `.js` as extension?
Looking into travis at what it generates, seems to report the diff for the `.map` as well. Not sure if we should _track_ those as well.

wardpeet: fixed in https://github.com/developit/preact/pull/1633/commits/51dbc5fcdf04f3943fae59bac28ba8f54af5544b

wardpeet:
Should be fixed in https://github.com/developit/preact/pull/1633/commits/51dbc5fcdf04f3943fae59bac28ba8f54af5544b

marvinhagemeister:
Let's do this! Thank you so much for your PR :+1: :100:

PR #1634, opened 2019-05-15
(chore) Pin the babel-plugin-istanbul version to fix the tests running locally

cristianbote:
As discussed previously, I'm having issues running the `test:karma` tests and also, seems like the `eslint` had something to say about the _empty render_. Got the issue with the tests, pointing to the `babel-plugin-istanbul@^5.0.1`.

The follow-up here should be enabling greenkeeper. Right @developit?

### DIFF
--- a/test/browser/lifecycle.test.js
+++ b/test/browser/lifecycle.test.js
@@ -2001,8 +2001,8 @@ describe('Lifecycle methods', () => {
 		});
 
 		it('should be called when child fails in render', () => {
-			// eslint-disable-next-line react/require-render-return
 			class ThrowErr extends Component {
+				// eslint-disable-next-line react/require-render-return
 				render() {
 					throw new Error('Error!');
 				}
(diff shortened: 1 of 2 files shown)


test/browser/lifecycle.test.js
```diff
@@ -2001,8 +2001,8 @@ describe('Lifecycle methods', () => {
 		});
 
 		it('should be called when child fails in render', () => {
-			// eslint-disable-next-line react/require-render-return
 			class ThrowErr extends Component {
+				// eslint-disable-next-line react/require-render-return
```
marvinhagemeister: These changes seem to break the linter:

```txt
/home/travis/build/developit/preact/test/browser/lifecycle.test.js
  2004:4  error  Your render method should have return statement  react/require-render-return
  2500:4  error  Your render method should have return statement  react/require-render-return
```

cristianbote: Huh, but that was failing locally though. Hmmm 🤔 back to the drawing board then. Is there a chance we could view the source files from travis?

cristianbote:
Closing since I need more time to investigate the tests issue 🙁

PR #1636, opened 2019-05-16
Better suspense

sventschui:
This PR aims to provide an implementation of Suspense that matches the behaviour of React more closely:

- Keep suspended DOM but hide it using `display: none` (Text nodes are removed)
  - This results in keeping state

*Open things*
- [x] Allow for multiple pending promises thrown in the render methods of `Suspense`s children
- [ ] More tests
- [ ] Clean up the Implementation (maybe we can make some things easier / more optimized)
- [ ] Byte-golfing
- [ ] Handle text nodes nested below children

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -101,7 +101,7 @@ module.exports = function(config) {
 
 		files: [
 			{ pattern: 'test/polyfills.js', watched: false },
-			{ pattern: config.grep || '{debug,hooks,compat,test-utils,}/test/{browser,shared}/**.test.js', watched: false }
+			{ pattern: config.grep || '{debug,hooks,compat,test-utils,}/test/{browser,shared}/suspense.test.js', watched: false }
 		],
 
 		preprocessors: {
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -101,7 +101,7 @@ module.exports = function(config) {
 
 		files: [
 			{ pattern: 'test/polyfills.js', watched: false },
-			{ pattern: config.grep || '{debug,hooks,compat,test-utils,}/test/{browser,shared}/suspense.test.js', watched: false }
+			{ pattern: config.grep || '{debug,hooks,compat,test-utils,}/test/{browser,shared}/**.js', watched: false }
 		],
 
 		preprocessors: {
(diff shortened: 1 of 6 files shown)


karma.conf.js
```diff
@@ -101,7 +101,7 @@ module.exports = function(config) {
 
 		files: [
 			{ pattern: 'test/polyfills.js', watched: false },
-			{ pattern: config.grep || '{debug,hooks,compat,test-utils,}/test/{browser,shared}/**.test.js', watched: false }
+			{ pattern: config.grep || '{debug,hooks,compat,test-utils,}/test/{browser,shared}/suspense.test.js', watched: false }
```
marvinhagemeister: If you only want to run a specific group of tests you can always use `.only` like `it.only()` or `describe.only()`. It may not speed up the globbing process as much as this change here :)

sventschui: Good point. Totally forgot about only :)

developit: I believe there's also `npm t -- --grep suspense`.

src/suspense.js
```diff
@@ -7,33 +7,37 @@ import { toChildArray } from './diff/children';
...
 	});
+	this.pendingSuspensions = [];
 }
 Suspense.prototype = new Component();
 
 /**
  * @param {Promise} promise The thrown promise
  */
 Suspense.prototype._childDidSuspend = function(promise, suspendingComponent) {
+	this.pendingSuspensions.push(promise);
```
developit: Instead of using an Array here, we can stack promises:

```js
Suspense.prototype._childDidSuspend = function(promise, suspendingComponent) {
  this._pending = this._pending.then(() => promise);
};
```

This means awaiting the pending value is as simple as `this._pending.then`.

sventschui: That's a good point. One problem is that I'd need to unregister the `then` handler of `this._pending` before re-assigning it. Otherwise the suspension would be completed when the first promise throws:

- catch first promise: `this._pending = firstPromise`
  - `this._pending.then(suspensionIsDone)` (de facto `firstPromise.then(suspensionIsDone)`)
- catch. first promise: `this._pending = this.pending.then(() => secondPromise))`
  - need to unregister `firstPromise.then(suspensionIsDone)` and replace with:
  - `this._pending.then(suspensionIsDone)`

sventschui:
Closing in favour of #1638

PR #1637, opened 2019-05-16
Simplify key matching algo

developit:


developit:
@JoviDeCroock looks like your comment got cut off.

FWIW, if Preact currently allows same-keyed vnodes with a different type to be reused, that's probably a bug.

JoviDeCroock:
Don’t understand how that happened o.O

Well yes it shouldn’t be reused so this seems like a good option to me personally.

developit:
Re-upping, I think this is mergable.

PR #1638, opened 2019-05-16
Parking approach to suspense

sventschui:
This is the "parking" approach to suspense I mentioned in slack.

It "parks" vnodes/dom that is suspended and attaches it again when the suspension is completed. This is completely different from what React does (keeping dom attached but hiding it).

- [x] I'll add some more tests in the next days to see how this approach compares to the one used in React in terms of lifecycle, state and re-renders.

Current stats:

| | #master | no Suspense at all | this PR without exporting unmount | this PR | no Suspense -> PR | master -> PR
--- | --- | --- | --- | --- | --- | ---
**core** | 3576 | 3441 | 3478 | **3487** | +46 | -89
**compat** | 2037 | 2019 | 2719 | **2397** | +378 | +360


sventschui:
I guess I made it now from a functional point of view.

After my initial byte golfing this still adds 202B. Would be awesome if someone could golf even more bytes :)

prateekbh:
Can you elaborate:

>This is the "parking" approach to suspense I mentioned in slack.

src/diff/index.js
```diff
@@ -367,22 +384,14 @@ function doRender(props, state, context) {
  * component check for error boundary behaviors
  */
 function catchErrorInComponent(error, component) {
-	// thrown Promises are meant to suspend...
-	let isSuspend = typeof error.then === 'function';
-	let suspendingComponent = component;
+	if (typeof error.then === 'function') {
```
JoviDeCroock: Is it safe to just use error.then here?

src/suspense.js
```diff
@@ -1,23 +1,102 @@
...
+
+			// reset the timeout
+			this._timeout = null;
+
+			// remove fallback
+			unmount(this.props.fallback);
+
+			// make preact think we had mounted the parkedVNode previously...
+			this._prevVNode = this._parkedVnode;
+			this._parkedVnode = null;
```
JoviDeCroock: Can we delay the this._timeout setting to here so you can do `this._timeout = this._parkedVNode = null`

src/suspense.js
```diff
@@ -1,23 +1,102 @@
...
+			this._suspensions = [];
+
+			// reset the timeout
+			this._timeout = null;
+
+			// remove fallback
+			unmount(this.props.fallback);
+
+			// make preact think we had mounted the parkedVNode previously...
+			this._prevVNode = this._parkedVnode;
```
JoviDeCroock: can we use parkedVNode for consistency?

src/diff/index.js
sventschui: 99% safe. As far as I checked it didn‘t impact gzipped size. Let me double check

src/suspense.js
sventschui: Excellent find. Will check when I‘m on my computer!

src/diff/index.js
JoviDeCroock: To be honest we could just make `options.catchError(error)` and move this to debug all together. Would shave off significant bytes

sventschui: -13B on you. Thx! Will add a test later. Improving coverage is on the TODO list anyhow

src/suspense.js
sventschui: Done. Thx!

debug/src/debug.js
```diff
@@ -75,20 +81,15 @@ export function initDebug() {
 
 		// Check prop-types if available
 		if (typeof vnode.type==='function' && vnode.type.propTypes) {
-			if (vnode.type.displayName === 'Lazy') {
+			if (vnode.type.displayName === 'Lazy' && !vnode.__lazyPropTypesWarned) {
+				vnode.__lazyPropTypesWarned = true;
```
developit: can we use a WeakMap/Map for this? it'll avoid a shape change on the vnode.

sventschui: Honestly didn't know that `WeakMap` has such a browser support. As I just tried to do this I saw that we can re-use `warnedComponents` for this :)

compat/src/index.d.ts
```diff
@@ -47,6 +45,18 @@ export declare interface Children {
   toArray: (children: preact.ComponentChildren) => preact.VNode<{}>[];
 }
 
+//
+// Suspense/lazy
+// -----------------------------------
+function lazy<T>(loader: () => Promise<{default: T}>): T;
```
pmkroeker: This should probably be `export declare function lazy...` to fall in line with the rest.

compat/src/index.d.ts
```diff
@@ -47,6 +45,18 @@ export declare interface Children {
...
+
+interface SuspenseProps {
+  children?: ComponentChildren;
+  fallback: ComponentChildren;
+}
+
+abstract class Suspense extends Component<SuspenseProps> {}
```
pmkroeker: This should also have an `export declare`  statement in front of it.

compat/src/index.js
```diff
@@ -18,6 +20,15 @@ options.event = e => {
 	return e.nativeEvent = e;
 };
 
+let oldCatchRender = options.catchRender;
+options.catchRender = (error, component) => {
+	if (oldCatchRender && oldCatchRender(error, component)) {
+		return true;
+	}
+
+	return catchRender(error, component);
```
ForsakenHarmony: this could probably just be `return oldCatchRender && oldCatchRender(error, component) || catchRender(error, component)`

src/index.d.ts
```diff
@@ -198,6 +198,17 @@ declare namespace preact {
 		hook?(component: Component): void;
 		/** Attach a hook that is invoked after a vnode has rendered. */
 		diffed?(vnode: VNode): void;
+		/** Attach a hook that is invoked after an error is catched in a component but before calling lifecycle hooks */
```
ForsakenHarmony:
```suggestion
		/** Attach a hook that is invoked after an error is caught in a component but before calling lifecycle hooks */
```

src/index.d.ts
```diff
@@ -198,6 +198,17 @@ declare namespace preact {
...
+		catchError?(error: any, component: Component): void;
+		/** 
+		 * Attach a hook that is invoked after an error is catched while executing render.
```
ForsakenHarmony:
```suggestion
		 * Attach a hook that is invoked after an error is caught while executing render.
```

compat/src/index.js
sventschui: Thx. Saved 3B!

compat/src/suspense.js
```diff
@@ -0,0 +1,143 @@
+import { Component, createElement, unmount } from 'preact';
+import { removeNode } from '../../src/util';
+
+export function catchRender(error, component) {
+	// thrown Promises are meant to suspend...
+	if (typeof error.then === 'function') {
```
JoviDeCroock: Just reasking, you can resolve if checked but e.then gave no size benefit?

sventschui: Will re-check but IIRC no

sventschui: It cut 4 bytes. I guess my statement was based on the state when this was part of the core. Thx!

sventschui:
There seems to be issues with sCU/createCoontext between Suspense and the suspending component. Let me add some tests for this situation before merging this

compat/src/suspense.js
```diff
@@ -0,0 +1,143 @@
+import { Component, createElement, unmount } from 'preact';
+import { removeNode } from '../../src/util';
+
+export function catchRender(error, component) {
+	// thrown Promises are meant to suspend...
+	if (error.then) {
+		for (; component; component = component._ancestorComponent) {
```
cristianbote: Can a `while(component = component._ancestorComponent)` do it in here? Maybe we can shave some bytes

compat/src/suspense.js
```diff
@@ -0,0 +1,143 @@
...
+ * @param {Promise} promise The thrown promise
+ */
+Suspense.prototype._childDidSuspend = function(promise) {
+	// saves 5B
+	const _this = this;
+
+	_this._suspensions.push(promise);
+	let len = _this._suspensions.length;
+
+	const suspensionsCompleted = () => {
```
cristianbote: Should these be part of the `prototype`? Thinking that these could introduce to many closures? Most of the refs are based on `scope` so, I believe these could do it?

sventschui: Well the issue is that the func needs to access `len` that is only available here and not on the proto. Any idea how to solve this? Function binding?

PR #1644, opened 2019-05-21
(fix) - support react-hot-loader

JoviDeCroock:
__proto__ gave odd issues with devtools, because the objects weren't convertible anymore. Something really odd going on there.

Adds 8B

Fixes: https://github.com/developit/preact/issues/1620

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -35,8 +35,7 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 		newType = newVNode.type, clearProcessingException;
 
 	// When passing through createElement it assigns the object
-	// ref on _self, to prevent JSON Injection we check if this attribute
-	// is equal.
+	// constructor as undefined. This to prevent JSON-injection.
 	if (newVNode.constructor !== undefined) return null;
 
 	if (tmp = options.diff) tmp(newVNode);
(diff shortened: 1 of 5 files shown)


src/diff/index.js
```diff
@@ -382,7 +381,7 @@ function catchErrorInComponent(error, component) {
 						continue;
 					}
 				}
-				else if (component.constructor.getDerivedStateFromError!=null) {
```
sventschui: Wondering why getDerivedStateFromError is called on the constructor instead of directly on the component here?

test/shared/createElement.test.js
```diff
@@ -32,9 +32,10 @@ describe('createElement(jsx)', () => {
 		expect(<Test />).to.have.property('type', Test);
 	});
 
-	it('should set VNode._self property to prevent json injection', () => {
+	it('should set VNode.constructor property to prevent json injection', () => {
 		const vnode = <span />;
-		expect(vnode._self).to.equal(vnode);
+		// eslint-disable-next-line no-proto
```
sventschui: Guess this one is a left over from the proto approach?

JoviDeCroock: Good spot, thanks 👍

src/diff/index.js
JoviDeCroock: I think it's the same story as for the other static method `getDerivedStateFromProps` also called on newType which is the constructor.

sventschui: I now get why this is needed. `getDerivedStateFromError` is static and thus not available on the instance but on the constructor only 🤦‍♂

PR #1646, opened 2019-05-22
Golf bytes

sventschui:
Golf some bytes

-26B

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -333,7 +333,7 @@ export function unmount(vnode, ancestorComponent, skipRemove) {
 
 	vnode._dom = vnode._lastDomChild = null;
 
-	if ((r = vnode._component)) {
+	if (r = vnode._component) {
 		if (r.componentWillUnmount) {
 			try {
 				r.componentWillUnmount();
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -81,7 +81,7 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 			}
 			else {
 				// Instantiate the new component
-				if (newType.prototype && newType.prototype.render) {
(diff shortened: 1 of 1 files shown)


src/diff/index.js
```diff
@@ -352,7 +352,7 @@ export function unmount(vnode, ancestorComponent, skipRemove) {
 		}
 	}
 
-	if (dom!=null) removeNode(dom);
+	if (dom) removeNode(dom);
```
JoviDeCroock: We should test this with an empty textNode/textNode with 0. Not entirely sure if this would still work

sventschui: I'll try to dig into this later

src/diff/index.js
```diff
@@ -328,12 +328,12 @@ export function unmount(vnode, ancestorComponent, skipRemove) {
...
 	let dom;
 	if (!skipRemove && vnode._lastDomChild==null) {
-		skipRemove = (dom = vnode._dom)!=null;
+		skipRemove = (dom = vnode._dom);
 	}
 
 	vnode._dom = vnode._lastDomChild = null;
 
-	if ((r = vnode._component)!=null) {
+	if ((r = vnode._component)) {
```
sventschui:
```suggestion
	if (r = vnode._component) {
```

src/diff/index.js
developit: if `dom instanceof Text` it should work fine. IIRC the reason for using `!=null` and `==null` all over the codebase was performance + byte size. We'll want to check for perf regressions if we swap them, and make sure we use them uniformly.

developit:
@andrewiggins exactly. The "no change" and "repeat diff" benchmarks can sometimes show these, but it'd be better to run them through the main public benchmarks.

sventschui:
Closing this as per the perf remarks 👍

PR #1647, opened 2019-05-22
(fix) - should call unmount when replacing components

JoviDeCroock:
Fixes: https://github.com/developit/preact/issues/1645

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -21,7 +21,7 @@ export function render(vnode, parentDom, replaceNode) {
 	diffChildren(
 		parentDom,
 		replaceNode ? vnode : (parentDom._prevVNode = vnode),
-		replaceNode ? undefined : oldVNode,
+		oldVNode,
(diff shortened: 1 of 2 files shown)


src/render.js
```diff
@@ -21,7 +21,7 @@ export function render(vnode, parentDom, replaceNode) {
 	diffChildren(
 		parentDom,
 		replaceNode ? vnode : (parentDom._prevVNode = vnode),
-		replaceNode ? undefined : oldVNode,
```
ForsakenHarmony: what was the point of having undefined here?

JoviDeCroock: Hmm I don't really know @LukasBombach was there a reason to put this on undefined back then? Maybe I'm breaking something that isn't tested?

LukasBombach: I think I got it wrong there. I thought there was no oldVNode if your hydrate from nothing, but it was probably just a brainfart

PR #1649, opened 2019-05-23
(chore) - Update links

JoviDeCroock:
Now that we moved to the preactjs org it's time to update the links 

coveralls etc still reporting bad so I wouldn't merge just yet


JoviDeCroock:
Seems like there's a missing setting in coveralls to switch from branch coverage to line coverage. Can't exactly locate it yet.

PR #1650, opened 2019-05-23
(fix) - CI build

JoviDeCroock:
The CI seemed to have lost its cache with the org move, this pins some dependencies as a temp fix so our PR's would run again.

We just miss a repo to drop our coveralls reports


cristianbote:
Thank you @JoviDeCroock! 😄

PR #1651, opened 2019-05-23
Consistently use key check logic (-12 B)

andrewiggins:
Apply the key check simplification from #1637 to the other place where we looking for matching keys. Reusing the same logic saves us a decent amount of bytes.


JoviDeCroock:
Coveralls seems to act up more now with the new version...

Issue #1656, opened 2019-05-29
react-spring useImperativeHandle problem - Preact X


yusufgenc:
React-spring uses useImperativeHandle but doesn't provide any ref to it. Current preact/hook implementation doesn't handle this scenario and makes it unusable. 

Current useImperativeHandle code.
```javascript
export function useImperativeHandle(ref, createHandle, args) {
	const state = getHookState(currentIndex++);
	if (argsChanged(state._args, args)) {
		state._args = args;
		ref.current = createHandle();
	}
}
```
Error
```
Uncaught TypeError: Cannot set property 'current' of undefined
    at l (src.a2b27638.js:567)
    at useSprings (src.a2b27638.js:2796)
    at useSpring (src.a2b27638.js:2845)
    at p.Card [as constructor] (src.a2b27638.js:3951)
    at p.j [as render] (src.a2b27638.js:379)
    at C (src.a2b27638.js:307)
    at C (src.a2b27638.js:307)
    at g (src.a2b27638.js:244)
    at L (src.a2b27638.js:403)
    at Object.b (src.a2b27638.js:754)
```

My fix for this problem (preact/hooks/src/index.js)
```javascript
export function useImperativeHandle(ref, createHandle, args) {
	if (ref) {
		const state = getHookState(currentIndex++);
		if (argsChanged(state._args, args)) {
			state._args = args;
			ref.current = createHandle();
		}
	}
}
```
Enviroment:
"preact": "^10.0.0-beta.1"
"react-spring": "^9.0.0-beta.8"

I hope this would help someone.

PR #1657, opened 2019-05-29
(fix) - prevent crash with nullish ref

JoviDeCroock:
Fixes: https://github.com/preactjs/preact/issues/1656

Apparently some libraries pass in null as a ref until one is available. Our approach would throw in this scenario

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -140,9 +140,9 @@ export function useImperativeHandle(ref, createHandle, args) {
 	const state = getHookState(currentIndex++);
 	if (argsChanged(state._args, args)) {
 		state._args = args;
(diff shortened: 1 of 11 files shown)


hooks/src/index.js
```diff
@@ -140,7 +140,9 @@ export function useImperativeHandle(ref, createHandle, args) {
 	const state = getHookState(currentIndex++);
 	if (argsChanged(state._args, args)) {
 		state._args = args;
-		ref.current = createHandle();
+			if (ref) {
```
cristianbote: Really, really small nit 😄 I believe there's 3x tabs in here? And should it be just 2x?

PR #1658, opened 2019-05-29
Consistently use `_children` & combine Fragment and Component diffing (-73 B)

andrewiggins:
## Summary

~~NOTE: this might be a good candidate for merging after we release 10.0. Given the size of the change, it may not be worth de-stabilizing master with more changes. Thoughts?~~ I might need this change to improve #1605, so perhaps it would be worthwhile to get in soon. Still investigating.

* Replace `component._prevVNode` with `vnode._children`
* Combine Fragment and Component diffing

### Replace `component._prevVNode` with `vnode._children`

Previously, `_children` was a cache for a normalized `props.children`. With a DOM VNode and Fragment VNode, you could access the children of the VNodes on the this `_children` property, but on VNodes for components, you need to go through `vnode._component._preVNode`.

This PR formalizes the concept behind `_children`. It is now the container for the rendered children of all VNodes. If you think of `props.children` as input to render functions, `_children` is the output of the `render`. And DOM VNodes and Fragments are just render functions that return `props.children`. The implementation is a little different (DOM VNodes don't have a render function, per se), but this concept applies.

### Combine Fragment and Component diffing

Previously, if `diff` was called with a Component, it would call `c.render`, coerce the result to a single VNode (e.g. convert arrays to a Fragment), and then call `diff`.  However, building on the `_children` change (and the concept mentioned there), we now instead treat Components as always returning arrays, and coerce result of render into a children of VNodes to continue rendering. So, `diff` always calls `diffChildren` after diff'ing a VNode, regardless of its type. Another way to think about this is that everything is Fragment now :smile:

Being able to depend on `diffChildren` -> `diff` -> `diffChildren` for all VNodes allows us to do operations that require knowing the `parentVNode` in `diffChildren` without having to special case the `diff` -> `diff` situation. For example, there is now only one place where oldVNode matching happens: `diffChildren` (we can remove some code from the top of `diff`). Further, mounting and unmounting DOM only happens in `diffChildren` as well (if unmounting wasn't recursive, we could inline it).

### Future Considerations

With this PR, their are only 2 place `diff` is called: `diffChildren` and `forceUpdate`. If we can change `forceUpdate` to also enter the diff through `diffChildren`, then it is conceivable that we could inline `diff` into `diffChildren` and make our diff a single function! I think this could give us some good byte savings by sharing more local variables and reducing duplicate structures (e.g. have fewer than 5 `try {} catch {}` like we currently do). It may reduce the approachability of our code, but perhaps we can mitigate that with a well-defined sectional comment format to make up for the lost function names. Either way, I think it might be worth an exploration.

### Breaking change

There are some breaking changes in this PR.

* Component refs aren't re-invoked after each render. Preact v8 and v10 both re-invoke component refs on every render. This PR changes that behavior so they are only invoked on initial render, when the ref changes, or on unmount. This new behavior [appears to match React](https://codesandbox.io/s/reinvoking-preact-refs-90oj9).

## Per change summary

91f043b Only copy _dom from oldVNode to newVNode if necessary (-1 B)
3159b71 Replace c._prevVNode with vnode._children (-2 B)
b73ab00 Call `diffChildren` after rendering components (-8 B)
b09318f Copy the oldVNode in `forceUpdate` to prevent side effect modification (+7 B)
dc86517 Set `vnode._dom` pointer in diffChildren (+2 B)
fe7f20d Only remove excessDomChildren after diff'ing DOM VNodes (+5 B)
6f06653 Remove redundant vnode type checks in `diff` (-17 B)
40212eb Apply refs after unmounting in diffChildren (+25 B)
910f98c Combine Fragment and Component diffing (-60 B)
27116bc Remove Array coercion from coerceToVNode (-8 B)
6b18f6f Remove more null checks from diff (-13 B)
c0c070e Normalize oldVNode to EMPTY_OBJ once (-2 B)
9d89d10 Fix Fragment _dom pointers (-10 B)
3470318 Flatten top-level Fragments returned from render (+21 B)
31036a3 Remove unused append in forceUpdate (-14 B)

~~Total change: -70 B 😁~~
~~Total change: -78 B 😁~~
~~Total change: -57 B 😃~~
Total change: -73 B 😁

NOTE: Individual commit size diffs may not be exact due to drifting caused by merges from master. They are intended to give an estimate of relative cost.


test/browser/refs.test.js
```diff
@@ -83,7 +83,7 @@ describe('refs', () => {
 
 		ref.resetHistory();
 		render(<Foo ref={ref} />, scratch);
-		expect(ref).to.have.been.calledOnce;
+		expect(ref).not.to.have.been.called;
```
andrewiggins: These test modifications demonstrate the breaking change

test/browser/fragments.test.js
```diff
@@ -495,6 +495,7 @@ describe('Fragment', () => {
 		expect(scratch.innerHTML).to.equal('<div>Hello</div>');
 	});
 
+	// TODO: Investigate
```
andrewiggins: This is the only test in core that is currently failing. Previously we converted array returns into unkeyed Fragments, so an unkeyed Fragment render return and an array return were viewed as the same:
```js
render() { return [<div />]; }
// becomes
c._prevVNode = <Fragment><div /></Fragment>

render() { return <Fragment><div /></Fragment>; }
// becomes
c._prevVNode = <Fragment><div /></Fragment>
```

However, this PR changes that behavior such that if render returns an array we flatten it, and if render returns a Fragment we just put it into the `_children` array :/
```js
render() { return [<div />]; }
// becomes
vnode._children = [<div />];

render() { return <><div /></>; }
// becomes
vnode._children = [<Fragment><div /></Fragment>];
```

Any thoughts on the importance of this feature?

andrewiggins: In other words, `render()` used to always return a single VNode, and we coerced array returns into Fragments. Now, `render()` always returns an array and no coercion takes place.

JoviDeCroock:
This is more a question but does this solve the issues added in: https://github.com/preactjs/preact/pull/1639

src/diff/index.js
```diff
@@ -24,13 +24,6 @@ import options from '../options';
  * Fragments that have siblings. In most cases, it starts out as `oldChildren[0]._dom`.
  */
 export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, force, oldDom) {
-	// If the previous type doesn't match the new type we drop the whole subtree
```
andrewiggins: These checks are not necessary when `diff` is always called by `diffChildren` (there is one exception, `forceUpdate`, discussed below).  `diffChildren` already does proper null checking on newVNode and passes in EMPTY_OBJ if oldVNode is `null`. `diffChildren` will never pass two vnodes of non-equal types. It will instead always pass `null` if it can't find a matching oldVNode for newVNode.

The only place `diff` is called not inside `diffChildren` is `forceUpdate`. However, `forceUpdate` always calls `diff` with non-null vnodes of the same type. It will never pass in a null newVNode or oldVNode cuz it always passes the VNode that represents the component. If the component is in the tree, it can't have a null VNode or else it doesn't exist in the tree. It can render null, but that is handled by `diffChildren`.

In other words, all of this logic is duplicated in `diffChildren`. Since we now call always `diffChildren` after `diff`, we no longer need to duplicate that logic here.

src/diff/children.js
```diff
@@ -75,12 +75,16 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
...
-					oldVNode = null;
+					oldVNode = EMPTY_OBJ;
 				}
 			}
 
 			// Morph the old element into the new one, but don't append it to the dom yet
-			newDom = diff(parentDom, childVNode, oldVNode, context, isSvg, excessDomChildren, mounts, ancestorComponent, null, oldDom);
+			newDom = diff(parentDom, childVNode, oldVNode || EMPTY_OBJ, context, isSvg, excessDomChildren, mounts, ancestorComponent, null, oldDom);
+
+			if ((j = childVNode.ref) && (oldVNode || EMPTY_OBJ).ref != j) {
```
cristianbote: Do you think we could shave more bytes by always defaulting the `oldVNode` to `EMPTY_OBJ`? Therefore we would not need to have the `OR` in here?

andrewiggins:
@JoviDeCroock I'm not sure. I don't think so, but I haven't verified that hypothesis yet.

src/diff/children.js
```diff
@@ -110,15 +114,32 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 				}
 
 				oldDom = newDom.nextSibling;
+
+				if (firstChildDom == null) {
+					firstChildDom = newDom;
+				}
+
+				if (typeof newParentVNode.type == 'function') {
+					newParentVNode._lastDomChild = newDom;
```
andrewiggins: Hmm I think `_lastDomChild` has become a proxy for `typeof vnode.type == 'function' && (vnode's children have been diffed)`. Perhaps there is a better way to capture this information that is more explicit and smaller in size...

src/internal.d.ts
```diff
@@ -10,7 +10,7 @@ export interface FunctionalComponent<P = {}> extends preact.FunctionComponent<P>
 export type ComponentFactory<P> = preact.ComponentClass<P> | FunctionalComponent<P>;
 
 export interface PreactElement extends HTMLElement {
-	_prevVNode?: VNode<any> | null;
+	_children?: VNode<any> | null;
```
robertknight: It looks to me from a skim of the rest of the code that `_children` can be a VNode or an array or `null`, is that the case?

andrewiggins: My intention was to make it always an array (or null before it is defined) so if that's not true it is likely a bug or an oversight. I'll double check

test/browser/fragments.test.js
andrewiggins: Here is the relevant check and comment in React for this behavior: https://git.io/fj0P9

Perhaps it wouldn't be too big to add a similar check...

andrewiggins: Its 21 B to coerce top level non-keyed fragments to arrays.
```js
let isTopLevelFragment = tmp != null && tmp.type == Fragment && tmp.key == null;
toChildArray(isTopLevelFragment ? tmp.props.children : tmp, newVNode._children=[], coerceToVNode, true);
```

compat/src/suspense.js
```diff
@@ -15,20 +15,21 @@ export function catchRender(error, component) {
...
-		if (tmp = tmp._prevVNode) removeDom(tmp);
-	}
-	else if ((tmp = vnode._children) && vnode._lastDomChild) {
-		for (let i = 0; i < tmp.length; i++) {
-			if (tmp[i]) removeDom(tmp[i]);
+function removeDom(children) {
+	// TODO: We need to recurse down the tree cuz _dom pointers aren't bubbled up
+	// when a component suspends. See comment about "early exit" in diff/index.js
+	// for a possible resolution
+	for (let i = 0; i < children.length; i++) {
```
andrewiggins: @sventschui Can you take a look particularly at these suspense changes, and double check that they makes sense to you?

debug/test/browser/devtools.test.js
```diff
@@ -278,19 +285,29 @@ describe('devtools', () => {
 
 	describe('getChildren', () => {
 		it('should get component children', () => {
-			let a = createElement('div', { foo: 1 });
+			const Foo = () => <div>foo{null}bar</div>;
```
andrewiggins: @marvinhagemeister Can you double check that my updates here make sense and stay true to the original intent? I quickly tapped around the devtools with these changes and it seems everything worked

src/diff/children.js
andrewiggins: Saved 2 bytes! In order for null placeholders to work, if `oldVNode` is exactly `null`, it needs to stay that way while trying to find a match. So I coerce it to `EMPTY_OBJ` after the matching is finished.

src/diff/children.js
andrewiggins: I'm wrong. The real purpose for `_lastDomChild` is to properly resume the diff from the end of Fragment (line 92 in children.js)

test/browser/fragments.test.js
```diff
@@ -228,12 +228,9 @@ describe('Fragment', () => {
 			span(2)
 		].join('')));
 		expectDomLogToBe([
-			'<span>1.remove()',
-			// TODO: Why is it re-appending all the children 😢
```
andrewiggins: Woohoo! Fixed this de-op in the diffing!

karma.conf.js
```diff
@@ -127,8 +127,8 @@ module.exports = function(config) {
 						exclude: /node_modules/,
 						loader: 'babel-loader',
 						options: {
-							comments: false,
-							compact: true,
+							// comments: false,
```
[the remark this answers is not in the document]
marvinhagemeister: I'm wondering if we should remove these options from our karma config. Last time I checked they were breaking sourcemaps and thus any breakpoints.

debug/test/browser/devtools.test.js
marvinhagemeister: Clicked around for a while but I can't find any new issues with the devtools :+1:

marvinhagemeister:
Finally had a chance to read through the changes in depth. Merging `_prevVNode` and `_children` is such a relief. It was causing me a lot of trouble with the devtools and always using `_children` makes it so much simpler :tada:

Removing the special `diff -> diff` case in particular is very exciting! Making every component a `Fragment` is a more sensible path forward and should help us normalize more things :+1: I'll need to have a closer look again after JSConf is over, but I'm hoping that it will help us find a better solution to the sibling issue #1605 .

The `forceUpdate` stuff can be postponed to another PR in my opinion. It may take a bit of thought to get it right too. Although I'm confident that this PR will make it a lot easier. I think you have a really good handle on that :)

Regarding refs: Calling them on every render seems unnecessary. I'm with you that just calling them when something changes is sufficient :100: 

Overall this PR is (like your others) top notch and the creme de la creme. It's just that good :+1: :+1:

test/browser/fragments.test.js
andrewiggins: I went ahead and added this code. Figured it was useful for components like a loader HOC:

```js
render() {
  if (this.state.loading) {
    return <Fragment>Loading...</Fragment> // or something more complex
  }
  else {
    return props.children; // or some other array of children
  }
}
```

robertknight:
Nice work! I'm not as deeply familiar with the diffing logic as other devs but conceptually it looks like a good simplification. Assuming this lands before the next beta, I will need to update the Enzyme adapter accordingly, as it accesses the rendered children of vnodes.

compat/src/suspense.js
sventschui: Looks good 👍

src/component.js
```diff
@@ -71,10 +70,7 @@ Component.prototype.forceUpdate = function(callback) {
 		const force = callback!==false;
 
 		let mounts = [];
-		dom = diff(parentDom, vnode, vnode, this._context, parentDom.ownerSVGElement!==undefined, null, mounts, this._ancestorComponent, force, dom);
-		if (dom!=null && dom.parentNode!==parentDom) {
-			parentDom.appendChild(dom);
```
andrewiggins: Removed this because code coverage was complaining (showing this line was never hit), which I take to mean that either it is no longer needed, or that we are missing a test case to justify it's purpose. However, since #1605 is going to completely rewrite this code anyway, I'll just remove this now to keep coveralls happy.

JoviDeCroock:
The maxDuration line in compat/suspense seems to have lost its hit

marvinhagemeister:
@JoviDeCroock I'm running into this even on `master`. Not sure how it got in :S

/cc @sventschui

JoviDeCroock:
@marvinhagemeister maybe the coverall settings reset again and it's showing branch coverage or because of an update in istanbul maybe :/

sventschui:
I guess maxDuration was never tested. I don‘t know how to test this in mocha. I know jest supports mocking timers :/

I‘m in a university seminar this week. Can look into this next week ✌️

andrewiggins:
I think coveralls is failing cuz I changed the number of lines in suspense.js, so the one line that wasn't covered caused a tiny percentage change cuz it is now a greater percentage of lines (since the total number of lines is smaller).

I think we can proceed without coveralls approval since it doesn't actually represent a change in functional coverage.

Are people okay if I merge this?

karma.conf.js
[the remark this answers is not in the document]
andrewiggins: I accidentally left this in (whoops). If this causes problems, we can undo it.

package.json
```diff
@@ -17,7 +17,7 @@
     "build:hooks": "microbundle build --raw --cwd hooks",
     "build:test-utils": "microbundle build --raw --cwd test-utils",
     "build:compat": "microbundle build --raw --cwd compat --globals 'preact/hooks=preactHooks'",
-    "dev": "microbundle watch --raw --format cjs,umd",
+    "dev": "microbundle watch --raw --format cjs",
```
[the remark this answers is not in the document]
andrewiggins: Whoops, forgot this one too 🤦‍♂. When comparing sizes I usually just look at `cjs` (for no real reason, it's just first in the list), so I remove the `umd` cuz it just slows down the build. If there is a reason to keep both, happy to undo and change my habits lol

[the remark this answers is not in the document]
ForsakenHarmony: oh no 👀 

I think it was for direct usage with unpkg

PR #1660, opened 2019-05-30
Fix initially suspended component under sCU

sventschui:
So the parking approach to suspense has an issue when the suspending component suspends on the first render (as lazy() does) and is under a component with an sCU that returns false. This is represented by the newly added `should work throughout sCU when initially being suspended` test.

My first approach was to patch the `_parentDom` attribute and then call `forceUpdate()` on the component that suspended. The problem with this is, that when this component is directly under `Suspense` that it will show up (modify the DOM) too soon.

The second approach (this PR) is to not patch the `_parentDom` but call `forceUpdate()` on the `_ancestorComponent` of the suspending component. This seems to work but feels wrong. It also breaks other tests since it will result in some render methods being called too often.

Other thoughts I will explore:
- ~~Patch `_parentDom` and call `forceUpdate` on all suspending components once all of them finished.~~ This seems wrong as a suspending component might suspend again.

### DIFF
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -3,10 +3,11 @@ import { removeNode } from '../../src/util';
 
 export function catchRender(error, component) {
 	// thrown Promises are meant to suspend...
+	let suspendingComponent = component;
 	if (error.then) {
 		for (; component; component = component._ancestorComponent) {
 			if (component._childDidSuspend) {
-				component._childDidSuspend(error);
+				component._childDidSuspend(error, suspendingComponent);
 				return true;
 			}
 		}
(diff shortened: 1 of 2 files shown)


ForsakenHarmony:
guess it's similar to what we do for context

https://github.com/preactjs/preact/blob/master/src/create-context.js#L25

sventschui:
Yes it's similar but with the issue that we want to "postpone" the update of the dom but not the re-render because:
- the update of the dom will result in the fallback being overriden if there is no DOM between `<Suspense>` and the suspending component
- calling render as soon as possible is crucial as it might suspend multiple times

Thompson1985 (approved):

JoviDeCroock:
Closing: draft experiment against the 2019 reconciler; compat Suspense was rewritten in v11 and this approach (which failed tests) no longer applies.

PR #1667, opened 2019-05-31
Add preact-charts to Component Libraries

pmkroeker:
With preact-X now in beta.2, I have moved preact-charts up to version 1! With support for both preact 8.4.2 and X! It would be great to be included in the README.


jt3k:
now the link is not working :(

pmkroeker:
I unfortunately had to remove the lib from github, I'll post a PR. If I have some free time I would love to re-start on it. It was starting to get pretty specific to what we needed so a public version made less sense.

PR #1675, opened 2019-06-03
Retype style to error on strings

pmkroeker:
In #1606, support for string styles was removed. This change to the typings enforces that you must pass an object to `style`. This change will help improved typescript users transition to non-string stylings, with an error being thrown.

Could also possible add in something like [csstype](https://www.npmjs.com/package/csstype) to fill out the possibilities more fully, but I think that it would add extra stuff people would need to download with the package.

**EDIT**: Not sure why coverage decreased.

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -601,7 +601,7 @@ export namespace JSXInternal {
 		srcSet?: string;
 		start?: number;
 		step?: number | string;
-		style?: any;
+		style?: {[key: string]: string | number};
 		summary?: string;
 		tabIndex?: number;
 		target?: string;


src/jsx.d.ts
```diff
@@ -601,7 +601,7 @@ export namespace JSXInternal {
 		srcSet?: string;
 		start?: number;
 		step?: number | string;
-		style?: any;
+		style?: {[key: string]: string | number};
```
ForsakenHarmony: kinda wondering if there would be any lib where this would not be enough

JoviDeCroock:
Build is probably hiccup, restarted it to get correct coverage

src/jsx.d.ts
pmkroeker: Would this be a better fit maybe? I think all style values should fit into either `string | number`, but there may be some edge cases (like `grid-template-area`, where you can pass multiple strings)
```suggestion
		style?: {[key: string]: any};
```

ForsakenHarmony: but even there it would be a string in js? idk

was thinking like one of the css-in-js libs that passes stuff to style might break, but I'm fine with merging it like this and potentially making it broader if anyone runs into issues

pmkroeker: Ok, why don't I just keep it as is for now, and we can always approach it again if someone finds a lib where it breaks.

Issue #1666, opened 2019-05-31
[TypeScript] Issue with h alias of createElement


SolarLiner:
When importing `h` from preact, TypeScript complains that `typeof h` has "no compatible call signatures". I haven't had the time to investigate, but this is breaking every one of my projects with Preact X (none of them being critical of course) as they all import `h` instead of `createElement as h`.

Example error report:
```
 ./src/elements/title.tsx
[tsl] ERROR in /home/solarliner/Documents/Projets/VSCode/preact-bulma/src/elements/title.tsx(16,5)
      TS2349: Cannot invoke an expression whose type lacks a call signature. Type 'typeof h' has no compatible call signatures.
```
Example source code:
```typescript
import classnames from "classnames";
import { h, RenderableProps } from "preact";

interface ITitleProps {
  size?: 1 | 2 | 3 | 4 | 5 | 6;
  el?: string;
  spaced?: boolean;
}
// tslint:disable-next-line:variable-name
export const Title = createHeading("title");
// tslint:disable-next-line:variable-name
export const Subtitle = createHeading("subtitle");

function createHeading(mode: string) {
  return (props: RenderableProps<ITitleProps>) =>
    h(
      props.el || "p",
      { class: classnames(mode, { [`is-${props.size}`]: !!props.size, "is-spaced": !!props.spaced }) },
      props.children
    );
}
```
from https://github.com/solarliner/preact-bulma

SolarLiner:
Here's a repro project: https://codesandbox.io/s/youthful-shape-uir85

In `src/index.ts`, the `h` call is getting flagged by TypeScript - which crashes my build system, but seems to run past it on CodeSandbox - also, it seems like TypeScript doesn't type check the the h function when used from JSX, which hides the error in most usages.

The "natural" fix would be to replace the h definition in the types to `const h: typeof createElement` but this doesn't carry the namespace with it. Since I don't know why exactly we need the `createElement`function to be augmented like that, I don't want to mess it up this way.

Knowing that JSX is not affected by this bug means that I can just use the workaround import in a lot fewer places - but this is still not ideal.

pmkroeker:
Hmm, using `h` to define JSX seems to work ok, but when [using it as a function](https://codesandbox.io/embed/preactxhtypeerror-zu7lo), it will flag and error (TS2349).

PR #1676, opened 2019-06-03
fix: don't use typeof check for h

JoviDeCroock:
Fixes: https://github.com/preactjs/preact/issues/1666


ForsakenHarmony:
this seems to be the only working option :/

JoviDeCroock:
> this seems to be the only working option :/

Sadly enough noticed the same thing, tried a bit but no others

PR #1679, opened 2019-06-04
(refactor) - implement useRef with useState

JoviDeCroock:
-3Bytes for the hooks package


developit:
Q: would this be even smaller via useMemo?

JoviDeCroock:
> Q: would this be even smaller via useMemo?

Sorry, I'm not sure if I understand. you mean doing `useMemo(() => { current: initialValue }, []);`?

PR #1682, opened 2019-06-05
Add LimeApp in real-world apps

gmarcos87:
LimeApp is a PWA written on preact and running on the LibreMesh nodes. LibreMesh is a custom firmeware to create mesh networks in a simple and self-administrable way.


robertknight:
@gmarcos87 - The app failed to load when I looked just now. There was a Mixed Content error reported in the terminal. The URLs in this PR though are correct.

gmarcos87:
Yes, the app is designed to work from http and embedded in a router with a particular firmeware. It is not visible from the web, maybe i should remove the link on github pages as it does not support http anymore.

developit:
@gmarcos87 - would you be able to open a PR for the website repo instead of this one? If so that'd be great! Here's an example:
https://github.com/developit/preact-www/pull/289

cristianbote:
Hey @gmarcos87 have you had a chance to open a PR against https://github.com/preactjs/preact-www?

Issue #1681, opened 2019-06-05
Test utils `act()` disables normal deferred rendering


robertknight:
When the act function is called for the first time, it [installs a debounceRendering hook](https://github.com/preactjs/preact/blob/896e77d8c11a36ea642fe3999ac7087ac5110961/test-utils/src/index.js#L19) in order to be able to flush effects and state updates synchronously.

It does not restore the hook before it returns. As a result, any subsequent effects or `setState` calls which get executed outside the `act()` call will no longer trigger deferred rendering, which is a pretty confusing experience to debug.

Is there any reason not to restore any `options` hooks before `act` returns?

JoviDeCroock:
Good point, I'll fix this when I get home :D thanks for noticing it I'll probably try and dive into the other issue as well.

robertknight:
Thanks @JoviDeCroock. If there is any way I can help, please let me know!

PR #1683, opened 2019-06-05
(fix) - restore debounceRendering after act

JoviDeCroock:
Fixes: https://github.com/preactjs/preact/issues/1681


test-utils/src/index.js
```diff
@@ -36,6 +37,7 @@ export function act(cb) {
 		rerender();
 	}
 
+	options.debounceRendering = previousDebounceRendering;
```
andrewiggins: Should `act` call `teardown` since it calls `setupRerender()`? `teardown` not only restores `debounceRendering` but also flushes any updates called so that the next test begins with an empty update queue.

robertknight: My thought is the same reading this. I think the setup and teardown processes should be symmetrical in `act`. Otherwise someone might add code to `setupRerender` in future and that wouldn't get undone at the end of `act`.

JoviDeCroock:
@robertknight it should be fixed now

PR #1687, opened 2019-06-06
(fix) - ensure correct unmount

JoviDeCroock:
Fixes: https://github.com/preactjs/preact/issues/1632
Adds a test for the wrong dom scenario (isn't fixed yet) but I aim at doing this in the future


JoviDeCroock:
Hmm, with the changes from @andrewiggins this change does not seem to be needed... Anyone sees a scenario where we should at it for safety?

JoviDeCroock:
@marvinhagemeister would you like me to revert the change and just PR with the tests?

andrewiggins:
I went ahead and merged this cuz I'm building on top of it for my next PR

PR #1688, opened 2019-06-06
Replace ancestorComponent with vnode parent pointer (+5 B)

andrewiggins:
## Summary

This is pre-req for #1689 (which fixes #1593, #1604, #1605).

As a side benefit, it continues to normalize our VNode structure to make our virtual tree easier to reason and work with.

2b087d5 Replace ancestorComponent with parent vnode pointer (+8 B)
52f2161 Move _depth property to VNode (-3 B)

## Breaking changes

`catchError` and `catchRender` now take VNodes as their arguments instead of Components


src/diff/index.js
```diff
@@ -171,7 +167,7 @@ export function commitRoot(mounts, root) {
 			c.componentDidMount();
 		}
 		catch (e) {
-			catchErrorInComponent(e, c._ancestorComponent);
+			catchErrorInComponent(e, c._vnode);
```
ForsakenHarmony: isn't this missing `._parent`?

Seems like this could make the component be able to catch it's own error which I'm not sure is possible in react

src/index.d.ts
```diff
@@ -208,16 +208,18 @@ declare namespace preact {
 		/** Attach a hook that is invoked after a vnode has rendered. */
 		diffed?(vnode: VNode): void;
 		/** Attach a hook that is invoked after an error is caught in a component but before calling lifecycle hooks */
-		catchError?(error: any, component: Component): void;
+		catchError?(error: any, vnode: VNode): void;
```
ForsakenHarmony: wondering if there would be a use for vnodes here

src/diff/index.js
andrewiggins: That's true! I'll add a test and fix this

src/index.d.ts
andrewiggins: Honestly, I don't think there really is... I'm not super happy with this change. I have more thoughts about this. Will post more soon

src/diff/index.js
andrewiggins: Interesting... Does appear that in React, a component can catch it's own ref though: https://codesandbox.io/s/react-throwing-refs-lk958 (that actually matches the behavior in this branch already lol)

JoviDeCroock: Yes a component can catch errors in its own render/lifecycles, I think we're aligning more with React like this.

src/diff/index.js
```diff
@@ -258,32 +252,32 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
...
-		catchErrorInComponent(e, ancestorComponent);
+		catchErrorInComponent(e, parentVNode);
 	}
 }
 
 /**
  * Unmount a virtual node from the tree and apply DOM changes
  * @param {import('../internal').VNode} vnode The virtual node to unmount
- * @param {import('../internal').Component} ancestorComponent The parent
+ * @param {import('../internal').VNode} parentVNode The parent
```
robertknight: The comment here still refers to a "parent component" rather than a vnode. Can the parent VNode be obtained via the `vnode._parent` link here instead of passing an argument?

src/diff/index.js
andrewiggins: Hmmm this codesandbox seems to show that a component can't catch errors in it's own lifecyles. Does this look right? https://codesandbox.io/s/react-throwing-lifecycles-kjwky

src/diff/index.js
andrewiggins: Thanks! I'll update the comment. Since this function is recursive, I wanted to keep `parentVNode` as a pointer to first parent that isn't being unmounted, thinking it should handle errors of children that are being unmounted (should an unmounting error boundary still handle errors for children as they are unmounted?). However, let me check what React does here... perhaps we can save some bytes?

test/browser/lifecycle.test.js
```diff
@@ -1999,6 +2000,14 @@ describe('Lifecycle methods', () => {
 	});
 
 	describe('#componentDidCatch', () => {
+
+		/** @type {Error} */
+		let expectedError;
```
andrewiggins: @rpetrich FYI, I've updated the error boundary tests a bit. Lemme know if you have any thoughts!

andrewiggins: I think the biggest change is specifying that throwing refs can be caught in the component that defines them

src/diff/index.js
andrewiggins: Hmm seems React does swallow errors if the nearest error boundary to a thrown error is being unmounted (https://codesandbox.io/s/react-throwing-unmount-1vdvn) (<- could someone double check me on this?)

I think I'll leave our behavior for now just cuz I think I'd prefer to make noise on errors (e.g. failing to unsubscribe some resource which could lead to a memory leak).

src/index.d.ts
marvinhagemeister: Personally, I don't mind the change for now. It could prove useful in the future for `preact/debug`. We could theoretically catch top-level errors and display a nice message in the console.

src/diff/index.js
marvinhagemeister: TIL about react catching `ref` errors in the same component. Apart from that I'm not aware of components being able to catch errors inside their own lifecycle functions like the codesandbox demonstrates :+1:

src/diff/index.js
marvinhagemeister: Agree, swallowed errors is always painful to debug :+1:

src/index.d.ts
andrewiggins: #1692 contains where my brain went when making this change. Let me know what you think!

marvinhagemeister: Awesome, marking it as private is an even better solution :+1:

test/browser/lifecycle.test.js
rpetrich: Updated error boundary tests look good and check more behaviour 👍

robertknight:
To check my understanding, does every vnode now have a parent pointer, regardless of type (dom, component or fragment)? This means to get the ancestor component, it might be necessary to follow several parent pointers up the tree until we get to a component node?

marvinhagemeister:
@robertknight yup, your comment describes the changes perfectly :+1:

PR #1689, opened 2019-06-06
Begin diff with next DOM sibling in forceUpdate

andrewiggins:
When a Component isn't currently associated with a DOM node, we need to begin the diff from it's next DOM sibling so that if this render of the Component adds DOM, it will be correctly inserted into it's parent (instead of always appended like before).

I stole this idea from @marvinhagemeister's #1605, but re-implemented it so it no longer needs a `_sibling` pointer which should make this a little smaller. 

I tried exploring other ideas (like two loops in diffChildren) but they didn't pan out like I had hoped (would be just as big a change and couldn't get Fragments working after a while 😞). I figured it would be better to just get something working in master first.

With #1658, I think this solution works out pretty elegantly anyway (`diffChildren` is still the only place where DOM is added or removed).

Total change: +75 B 😬

NOTE: depends on #1688 (+5 B) so don't merge until after #1688 is in master. I'll rebase this PR on master once that happens.

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -70,30 +70,43 @@ Component.prototype.forceUpdate = function(callback) {
 		// shouldComponentUpdate
 		const force = callback!==false;
 
-		// let oldRoot = createElement(Fragment, {}, vnode);
-		// oldRoot._children = [vnode];
-		// let newRoot = createElement(Fragment, {}, cloneElement(vnode));
+		// // let oldRoot = createElement(Fragment, {}, vnode);
+		// // oldRoot._children = [vnode];
+		// // let newRoot = createElement(Fragment, {}, cloneElement(vnode));
 
-		// let oldVNodeCopy = assign({}, oldVNode);
-		// let oldRoot = createElement(Fragment, {}, oldVNodeCopy);
-		// oldRoot._children = [oldVNodeCopy];
-		// let newRoot = createElement(Fragment, {}, oldVNode);
+		// // let oldVNodeCopy = assign({}, oldVNode);
+		// // let oldRoot = createElement(Fragment, {}, oldVNodeCopy);
+		// // oldRoot._children = [oldVNodeCopy];
+		// // let newRoot = createElement(Fragment, {}, oldVNode);
 
-		let oldRoot = { _children: [assign({}, oldVNode)] };
-		let newRoot = { _children: [oldVNode] };
+		// let oldRoot = { _children: [assign({}, oldVNode)] };
+		// let newRoot = { _children: [oldVNode] };
 
-		// TODO: Consider passing in vnode._parent and the index to start and stop at:
-		// start (i): vnode._parent._children.indexOf(vnode)
-		// end (newChildrenLength): vnode._parent._children.indexOf(vnode) + 1
+		// // TODO: Consider passing in vnode._parent and the index to start and stop at:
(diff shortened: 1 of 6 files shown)


test/browser/fragments.test.js
```diff
@@ -1902,4 +1902,206 @@ describe('Fragment', () => {
 		rerender();
 		expect(scratch.textContent).to.equal('ABC');
 	});
+
+	it('should replace node in-between children', () => {
```
andrewiggins: TODO: Add some tests similarly to these where the component returns a Fragment, to make sure that this code properly handles components that insert Fragments *panics - hopefully it just works!*

test/browser/getDomSibling.test.js
```diff
@@ -0,0 +1,353 @@
+import { h, render, Fragment } from '../../src/index';
+import { getDomSibling } from '../../src/component';
+import { setupScratch, teardown } from '../_util/helpers';
+
+/** @jsx h */
+
+describe('getDomSibling', () => {
```
andrewiggins: @marvinhagemeister FYI, I copied your tests from #1605 and added a bunch. I couldn't get the implementation of `getDomSibling` in #1605 to pass all the new tests which is why I re-built it from scratch.

test/browser/getDomSibling.test.js
```diff
@@ -0,0 +1,353 @@
...
+			</div>
+		), scratch);
+		let vnode = getRoot(scratch)._children[0]._children[0];
+		expect(getDomSibling(vnode)).to.equal(scratch.firstChild.childNodes[1]);
+	});
+
+	it('should find sibling with nested placeholder', () => {
+		render((
+			<div key="0">
+				<Fragment key="0.0">
```
andrewiggins: I added keys to some of the tests here just cuz they made it easier to debug `getDomSibling` with a couple of well-placed `console.log(vnode.key)`

src/component.js
```diff
@@ -87,6 +87,37 @@ Component.prototype.forceUpdate = function(callback) {
  */
 Component.prototype.render = Fragment;
 
+/**
+ * @param {import('./internal').VNode} vnode
+ * @param {number | null} [childIndex]
+ */
+export function getDomSibling(vnode, childIndex) {
```
andrewiggins: I forget - does JS support optimized tail recursion? I think this function could be re-written to be tail-recursive if it isn't already (I forget what exactly qualifies as tail recursive).

JoviDeCroock: V8 had it for a while but not anymore, tracking bug: https://bugs.chromium.org/p/v8/issues/detail?id=4698

test/browser/getDomSibling.test.js
marvinhagemeister: Ohh that's a very neat way to debug ordering. I'll use that on my next PR :+1:

test/browser/getDomSibling.test.js
marvinhagemeister: I'm glad you did that as my version was a bit hacked together to get a good sense of the problem space, but I didn't have time to clean it up. The variant in this PR is much cleaner and the `childIndex` handling is ace :100:

test/browser/fragments.test.js
andrewiggins: Done

PR #1690, opened 2019-06-07
Delete React debugging props __self and __source

### LINKED ISSUE #1058: Compiled HTML looks faulty: __source / __self [object Object]
## Hey guys!
I'm sitting with a freshly installed **create-react-app** which I've converted to **Preact** with the simple steps. I get this weird issue when compiling my jsx files with Preact, and can't really find the root of cause. 

#### My Index.jsx
```jsx
import registerServiceWorker from './utilities/registerServiceWorker';

render(
  <div className="app">
    <p className="app-intro">
      To get started, edit <code>src/App.js</code> and save to reload.
    </p>
  </div>,
  document.getElementById('root')
);

registerServiceWorker();
```

#### The compiled HTML file
```jsx
<div class="app" __source="[object Object]" __self="[object Object]">
  <button class="app-intro" __source="[object Object]" __self="[object Object]">
To get started, edit <code __source="[object Object]" __self="[object Object]">src/App.js</code> and save to reload.
  </button>
>
```

### Does anybody have an idea of what might could be the problem?

mxstbr:
The @babel/preset-react preset includes two plugins which add `__source` and `__self` attributes to every element included development, which React uses for debugging. This muddles up the DOM and is pretty annoying.

For users of preact/compat, it is hard two remove these plugins as Babel does not allow disabling plugins included in a preset (see https://github.com/babel/babel/issues/3016#issuecomment-156150431). We could fork it, but unfortunately these plugins are included two levels deep so we would have to fork `babel-preset-react-app` _and_ `@babel/preset-react`, and then keep updating them with upstream changes, which is a bit too much work for such a small change.

Instead, this is a tiny patch that (in development only) deletes these two attributes from every VNode if they exist. This should not impact production bundle size or performance, but would be a big win for the DX for our project.

Questions:

1. Is this the right place to put the filtering? Should this live in `preact/compat` instead?
2. Is the development check done correctly? I could not find any other instances of a NODE_ENV check outside of `preact/debug`, so I am not sure if that does what we need it to do.

Closes #1058

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -119,11 +119,16 @@ export function initDebug() {
 	};
 
 	options.vnode = (vnode) => {
-		let source;
+		let source, self;
 		if (vnode.props && vnode.props.__source) {
 			source = vnode.props.__source;
 			delete vnode.props.__source;
 		}
+		if (vnode.props && vnode.props.__self) {
+			self = vnode.props.__self;
+			delete vnode.props.__self;
+		}
+		vnode.__self = self;
 		vnode.__source = source;
 		Object.defineProperties(vnode, deprecatedAttributes);
 		if (oldVnode) oldVnode(vnode);
--- a/debug/test/browser/debug.test.js
+++ b/debug/test/browser/debug.test.js
@@ -97,6 +97,14 @@ describe('debug', () => {
 		expect(vnode.props.__source).to.be.undefined;
 	});
 
+	it('should add __self to the vnode in debug mode.', () => {
+		const vnode = h('div', {
+			__self: {}
+		});
+		expect(vnode.__self).to.deep.equal({});
+		expect(vnode.props.__self).to.be.undefined;
+	});
+
 	it('should throw an error when using a hook outside a render', () => {
 		class App extends Component {
 			componentWillMount() {
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -33,12 +33,6 @@ export function createElement(type, props, children) {
 	let key = props.key;
 	if (ref!=null) delete props.ref;
 	if (key!=null) delete props.key;
-	if (process.env.NODE_ENV==='development') {
-		// The @babel/preset-react preset includes two plugins which add __source and
-		// __self attributes to every element in dev, which React uses for debugging
-		if (props.__source!=null) delete props.__source;
-		if (props.__self!=null) delete props.__self;
-	}
 
 	return createVNode(type, props, key, ref);
 }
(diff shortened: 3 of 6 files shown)


cristianbote:
Hey @mxstbr 👋!

Thanks for taking the time to look and come up with a solution for this! 💯

Have you seen the `options` object from `preact` package, `import { options } from "preact"`?

The `options` object basically lets you jump into different phases of vnode creation and/or rendering. As you'll notice in here https://github.com/preactjs/preact/blob/master/src/create-element.js#L67 that hook is called, whenever a new `vnode` is created. So, one could easily define a `vnode` hook and handle whatever particular case one could find himself into. For example, if you already use the `/compat` that would require you to basically define a new `.vnode` hook - since there's one already defined for compat layer.

```js
import { options } from "preact";

const old = options.vnode;
options.vnode = vnode => {
  // Call the previous 'hook'
  old(vnode);

  // Do the changes here
  delete vnode.props.__self;
  delete vnode.props.__source;
}
```

What do you think?

We do have a wip for integration with the devtools from React, @marvinhagemeister is working on that.

Thanks again for opening this! 👍

mxstbr:
Ohh that works amazingly well, thank you, that solves my problem 100%! :tada:

> We do have a wip for integration with the devtools from React, @marvinhagemeister is working on that.

Does that mean the `__source` and `__self` attributes will be used in the future and will be filtered anyway?

cristianbote:
Really happy that works for you 😄 

Umm, I just checked with @marvinhagemeister, and seems like it's not part of the devtools integration. So, no, these won't be filtered by that, but the `vnode` hook will still work whatsoever.

JoviDeCroock:
__source is used for jsx sourcemaps, it was a feature request for /debug: https://www.npmjs.com/package/@babel/plugin-transform-react-jsx-source

marvinhagemeister:
Yup, in the long-term we'd like to use both `__source` and `__self` for nicer error messages. We just haven't got around to it yet :slightly_smiling_face: 

Both properties should be only added during development, because from my understanding those babel plugins are not used in production. For that we stuff everything development related into `preact/debug`. We already remove `__source` from `props` there, but not `__self`.

cc @developit

mxstbr:
Oh I see, I had not been using `preact/debug`! I thought that was only for the DevTools integration but that seems to work even without `preact/debug`... somehow 🤔 

So the right fix would be to add a filter for `__self` to `preact/debug` in the exact same manner that `__source` is filtered, am I understanding that correctly?

mxstbr:
Implemented in the latest commit @marvinhagemeister—let me know if that looks better!

developit:
Nice work all, thanks for the PR @mxstbr!

btw I think you make a good point - we need to clarify that the "elevator pitch" for `preact/debug` is similar to React's development version.

PR #1697, opened 2019-06-09
Bypass props during hydration

developit:
This fixes a couple things that end up being best done together:

- instead of special-casing the `multiple` prop by applying it before diffing other props, it leverages the fact that preact already special-cases the `value` and `checked` props. Pulling these out into `diffElementNodes()` ensures they are applied last.

- during hydration (`excessDomChildren==null`), never call `diffProps()`. This makes it so that Preact no longer diffs attributes and props at all during hydration - previously only unknown attributes were ignored, but for performance reasons and to match other VDOM libraries it's worth also not diffing known attributes

- don't diff children when a node has dangerouslySetInnerHTML applied. This avoids the case where you have a node with both dangerouslySetInnerHTML and children specified, which previously would continually remove children. The solution in this PR probably doesn't correctly unmount children when transitioning from JSX children to dangerouslySetInnerHTML. This could perhaps be fixed by having the existence of dangerouslySetInnerHTML for a VNode assign `vnode._children=[]`.

- **styles:** this should be a bit of a performance optimization for styles. It's hard to verify, but I think I'm seeing some relatively consistent numbers in the perf tests showing this to be the case. I'd originally made the change in order to fix the style issue someone reported this week (can't find an open issue about it for some reason)

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -232,9 +232,9 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
 					dom.innerHTML = newHtml && newHtml.__html || '';
 				}
 			}
-
-			diffProps(dom, newProps, oldProps, isSvg);
 		}
+
+		diffProps(dom, newProps, oldProps, isSvg, excessDomChildren!=null);
 
 		// If the new vnode didn't have dangerouslySetInnerHTML, diff its children
 		if (!newHtml) {
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -222,18 +222,30 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
 
 		let oldHtml = oldProps.dangerouslySetInnerHTML;
 		let newHtml = newProps.dangerouslySetInnerHTML;
-		if ((newHtml || oldHtml) && excessDomChildren==null) {
-			// Avoid re-applying the same '__html' if it did not changed between re-render
-			if (!newHtml || !oldHtml || newHtml.__html!=oldHtml.__html) {
-				dom.innerHTML = newHtml && newHtml.__html || '';
+
+		// During hydration, props are not diffed at all (including dangerouslySetInnerHTML)
+		// @TODO we should warn in debug mode when props don't match here.
+		if (excessDomChildren == null) {
+			if (newHtml || oldHtml) {
+				// Avoid re-applying the same '__html' if it did not changed between re-render
+				if (!newHtml || !oldHtml || newHtml.__html!=oldHtml.__html) {
+					dom.innerHTML = newHtml && newHtml.__html || '';
+				}
 			}
-		}
-		if (newProps.multiple) {
-			dom.multiple = newProps.multiple;
+
+			diffProps(dom, newProps, oldProps, isSvg);
 		}
 
-		diffChildren(dom, newVNode, oldVNode, context, newVNode.type==='foreignObject' ? false : isSvg, excessDomChildren, mounts, EMPTY_OBJ);
-		diffProps(dom, newProps, oldProps, isSvg);
+		// If the new vnode didn't have dangerouslySetInnerHTML, diff its children
+		if (!newHtml) {
+			diffChildren(dom, newVNode, oldVNode, context, newVNode.type==='foreignObject' ? false : isSvg, excessDomChildren, mounts, EMPTY_OBJ);
+		}
+
+		// (as above, don't diff props during hydration)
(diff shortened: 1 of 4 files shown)


src/diff/index.js
```diff
@@ -236,18 +236,30 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
...
+			// @TODO we should warn in debug mode when props don't match here.
+			if (excessDomChildren == null) {
+				if (newHtml || oldHtml) {
+					// Avoid re-applying the same '__html' if it did not changed between re-render
+					if (!newHtml || !oldHtml || newHtml.__html!=oldHtml.__html) {
+						dom.innerHTML = newHtml && newHtml.__html || '';
+					}
 				}
+
+				diffProps(dom, newProps, oldProps, isSvg);
```
marvinhagemeister: Does this mean we still diff or is `oldProps` just an empty object? My understanding is that all new props are applied regardless of any existing attributes that may have been present in the dom.

src/diff/index.js
```diff
@@ -236,18 +236,30 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
...
+			if (!newHtml) {
+				diffChildren(dom, newVNode, oldVNode, context, newVNode.type==='foreignObject' ? false : isSvg, excessDomChildren, mounts, ancestorComponent, EMPTY_OBJ);
 			}
 
-			diffChildren(dom, newVNode, oldVNode, context, newVNode.type==='foreignObject' ? false : isSvg, excessDomChildren, mounts, ancestorComponent, EMPTY_OBJ);
-			diffProps(dom, newProps, oldProps, isSvg);
+			// (as above, don't diff props during hydration)
+			if (excessDomChildren == null) {
+				if (('value' in newProps) && newProps.value !== dom.value) dom.value = newProps.value==null ? '' : newProps.value;
+				if (('checked' in newProps) && newProps.checked !== dom.checked) dom.checked = newProps.checked;
```
marvinhagemeister: So good! :raised_hands:

marvinhagemeister:
Love the direction this is going! This way of special casing some attributes looks a lot cleaner :+1:

The first test failure seems to be a test case that needs to be updated for the changes in this PR. The other one may likely be only related to a different ordering in the assertions. I vaguely remember having swapped the attributes assertion order last time I touched `diffProps` :four_leaf_clover:

marvinhagemeister:
Played a bit with the changes in this PR and it looks like event handlers won't be attached anymore on hydrate. Made a new test case for this and it works on `master`, but fails in this branch.

```js
it('should attach event handlers on hydrate', () => {
  let spy = sinon.spy();
  scratch.innerHTML = '<span>Test</span>';
  let vnode = <span onClick={spy}>Test</span>;

  hydrate(vnode, scratch);

  scratch.firstChild.click();
  expect(spy).to.be.calledOnce;
});
```

src/diff/index.js
developit: ah that's something I didn't think to check for (and one of the failing tests). It seems like my modification here also disables event handler attachment.

marvinhagemeister:
FYI: Rebased it against `master` and added the failing test from the comments.

JoviDeCroock:
Added this to project board since it fixes https://github.com/preactjs/preact/issues/1719

marvinhagemeister:
@developit I'd love if you can take another look at the recent changes. To be able to attach event listeners during hydration we need to call `diffProps`. Because `oldProps` is always an empty object during hydration the first loop will do nothing. For the second loop where the properties are created, I added a condition to check if we're in hydration mode or not.

src/diff/props.js
```diff
@@ -9,27 +8,32 @@ import { assign } from '../util';
...
+		if (!(i in newProps)) {
+			setProperty(dom, i, null, oldProps[i], isSvg);
 		}
 	}
 
-	for (i in oldProps) {
-		if (i!=='children' && i!=='key' && !(i in newProps)) {
-			setProperty(dom, i, null, oldProps[i], isSvg);
+	for (i in newProps) {
+		if ((!hydrate || typeof newProps[i]=='function') && i!=='value' && i!=='checked' && oldProps[i]!==newProps[i]) {
```
developit: @marvinhagemeister idea: check for `i[0]=='o' && i[1]=='n'` instead? it should be ~free.

marvinhagemeister: Tried that but it's `+8 B` bigger :S

marvinhagemeister:
Rebased and the byte size actually is the same as in `master` :tada:

PR #1700, opened 2019-06-10
Bubble up Component dom changes up the virtual tree (+51 B)

andrewiggins:
If a component changed it's DOM node and was wrapped in other Components, the parent Components `_dom` and `.base` pointer could become out of date. Since `this.base` is a public API feature on Components, I figured this need to be fixed.

This might slightly decrease our memory footprint as we won't have pointers to useless DOM nodes anymore. Also fixes some cases where we would needlessly append DOM nodes to parents that are no longer mounted (see comment in fragments.test.js)

~~Total change: +36 B 😐~~
~~Total change: +48 B 😕~~
Total change: +51 B 😕


test/browser/fragments.test.js
```diff
@@ -2575,7 +2575,6 @@ describe('Fragment', () => {
 			div('B')
 		].join(''), 'updateB');
 		expectDomLogToBe([
-			// TODO: Extra appends happen here in actual. Why?
```
andrewiggins: These spurious appends no longer happen!

src/diff/children.js
```diff
@@ -100,6 +100,11 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 					// have a non-null _lastDomChild. Continue the diff from the end of
 					// this Fragment's DOM tree.
 					newDom = childVNode._lastDomChild;
+
+					// Eagerly cleanup _lastDomChild. We don't need to persist the value because
+					// it is only used by `diffChildren` to determine where to resume the diff after
+					// diffing Components and Fragments.
+					childVNode._lastDomChild = null;
```
andrewiggins: Instead of trying to keep `_lastDomChild` up to date, I figured it was better to just clean up the pointer after we no longer used it to reduce the likelihood of having out of date pointers to DOM nodes no longer in the DOM.

andrewiggins:
I believe this issue was already occurring. I think it has always been broken in Preact X actually. However, I don't think `c.base` is a well known or often used property (though we use it at work as a `forwardRef` cheat).

I checked out 699111d (the PR before my `_children` refactor) and 1566a31f3 (alpha.1) and the tests fail in both those commits so it's been with us for sometime.

marvinhagemeister:
Can confirm that we always had this issue in X (maybe even in 8.x?). This PR is the missing piece for our new `Fragment` algorithm :+1: We may revisit it when we'll work on 2-phase render, but for now this is the best we can do for the current architecture :100:

src/component.js
```diff
@@ -61,16 +61,30 @@ Component.prototype.setState = function(update, callback) {
...
 
 		let mounts = [];
-		diff(parentDom, vnode, assign({}, vnode), this._context, parentDom.ownerSVGElement!==undefined, null, mounts, force, dom == null ? getDomSibling(vnode) : dom);
+		let newDom = diff(parentDom, vnode, assign({}, vnode), this._context, parentDom.ownerSVGElement!==undefined, null, mounts, force, oldDom == null ? getDomSibling(vnode) : oldDom);
 		commitRoot(mounts, vnode);
+
+		if (newDom != oldDom) {
+			// Update parent component's _dom and c.base pointers
+			while (
+				vnode._parent
```
marvinhagemeister: Are there cases where the `_parent` pointer can be `null`? If that's the case for root nodes we could try using our `EMPTY_OBJ` ref for that. Maybe that could save a byte or two :tada:

test/browser/components.test.js
```diff
@@ -1573,4 +1573,319 @@ describe('Components', () => {
...
+		child.setState({ tagName: 'span' });
+		rerender();
+
+		expect(scratch.innerHTML).to.equal('<div><p></p><span>Hello</span></div>');
+		expect(child.base).to.equalNode(scratch.firstChild.lastChild);
+		expect(parent2.base).to.equalNode(child.base);
+		expect(sibling.base).to.equalNode(scratch.firstChild.firstChild);
+		expect(parent1.base).to.equalNode(sibling.base);
+		expect(parentDom1.base).to.equalNode(scratch.firstChild);
+	});
```
marvinhagemeister: I think we're missing one test scenario where the first child is non-null and unmounted on the second render. That seems to occur frequently on the apps I've worked on:

```jsx
// First render
<div>
  <Component>
    <Child />
    <Sibling />
  </Component>
</div>

// Second render
<div>
  <Component>
    {null}
    <Sibling />
  </Component>
</div>
```

test/browser/fragments.test.js
```diff
@@ -1295,7 +1295,7 @@ describe('Fragment', () => {
 		expectDomLogToBe([
 			'<ol>012345.insertBefore(<li>4, <li>0)',
 			'<ol>401235.insertBefore(<li>5, <li>0)',
-			// TODO: see issue #193 - Hmmm why does this extra append happen?
+			// TODO: Hmmm why does this extra append happen?
```
marvinhagemeister: Unrelated to this PR: Maybe we're wrongly recreating the child here in `diffChildren`. Something to check for another PR :+1:

test/browser/components.test.js
andrewiggins: Good call! Will add

src/component.js
andrewiggins: Oh true! I'll investigate

test/browser/components.test.js
```diff
@@ -1573,4 +1574,834 @@ describe('Components', () => {
...
+
+	it('should update parent c.base if first child becomes null', () => {
```
marvinhagemeister: Had a look into this and it seems like sometimes the `_dom` pointer of components is not set properly. When I log the `vnode` in `unmount()` `_dom` is null whereas `c.base` points to the correct DOM node.

andrewiggins: Yea, when looking at your comment about a missing test scenario, I realized a while bunch of edge cases I didn't handle lol. Still working on it

compat/src/suspense.js
```diff
@@ -27,18 +34,18 @@ function removeDom(children) {
 	for (let i = 0; i < children.length; i++) {
 		let child = children[i];
 		if (child != null) {
-			if (child._children) {
-				removeDom(child._children);
-			}
-			if (child._dom) {
+			if (typeof child.type !== 'function' && child._dom) {
```
andrewiggins: @sventschui FYI, this PR enables the `removeDom` function to now only remove the first DOM children parents under Suspense, instead of recursing all the way down the tree and removing all the nodes bottom up. Should be a bit faster now for deep trees!

src/component.js
```diff
@@ -61,16 +61,32 @@ Component.prototype.setState = function(update, callback) {
...
 
 		let mounts = [];
-		diff(parentDom, vnode, assign({}, vnode), this._context, parentDom.ownerSVGElement!==undefined, null, mounts, force, dom == null ? getDomSibling(vnode) : dom);
+		let newDom = diff(parentDom, vnode, assign({}, vnode), this._context, parentDom.ownerSVGElement!==undefined, null, mounts, force, oldDom == null ? getDomSibling(vnode) : oldDom);
 		commitRoot(mounts, vnode);
+
+		if (newDom != oldDom) {
+			updateParentDomPointers(vnode);
+
+			// // 3496 B
```
andrewiggins: TODO: Remove all the commented code before checking in. Wanted to first give people an idea of some of the golfing techniques I tried.

P.S. Surprised this iterative approach isn't smaller than the recursive function:

Minified:
```js
// Iterative: 3496 B
for (; null != (o = o.__p) && null != o.__c; )
    for (o.__e = o.__c.base = null, r = 0; r < o.__k.length; r++)
        if (null != (i = o.__k[r]) && null != i.__e) {
            o.__e = o.__c.base = i.__e;
            break;
        }
```

```js
// Recursive: 3484 B
function m(n) {
	var l, u;
	if (null != (n = n.__p) && null != n.__c) {
		for (n.__e = n.__c.base = null, l = 0; l < n.__k.length; l++)
			if (null != (u = n.__k[l]) && null != u.__e) {
				n.__e = n.__c.base = u.__e;
				break;
			}
		return m(n);
	}
}
```

compat/src/suspense.js
sventschui: Awesome 🎉

src/component.js
sventschui: Do we have some guidelines about bytes saving vs. function call overhead? Asking for a friend 😂

andrewiggins: I've seen a bench (assuming it can be trusted 🤷‍♂) where in one situation recursion was faster than the iterative solution, so I typically go for the byte savings first on something like this, especially since this code is only called once per update.

I think it would be hard to measure the perf difference between an iterative solution and recursive in this case since it isn't called very often in a diff (as opposed to the children loop in `diffChildren` which is called for every VNode, for example).

But that's just me. I don't spend as much time doing perf tests as @developit or others, so I defer to them.

andrewiggins:
I'll clean it up and merge it today

PR #1701, opened 2019-06-10
Clean up scratch DOM after tests

andrewiggins:


compat/test/browser/portals.test.js
```diff
@@ -29,6 +29,8 @@ describe('Portal', () => {
 		render(<Foo>foobar</Foo>, scratch);
 
 		expect(root.innerHTML).to.equal('foobar');
+
+		root.parentNode.removeChild(root);
```
marvinhagemeister: Wondering if we should just delete every child of `document.body` in `teardown()` instead. If the test fails we'd be left with a dangling child. As long as we do that change in `helpers` instead of the user facing `test-utils` package we should be fine. What do you all think?

andrewiggins: Karma puts some stuff under body that I'm uncertain we can remove

marvinhagemeister: Ohh that's true! Totally forgot about karma :+1:

PR #1703, opened 2019-06-11
Update compat typings

### LINKED ISSUE #1693: [10.0.0-beta.2] TS7026 error when using module alias between react and preact/compat on tsconfig.json
Hi, committers!

When I set this module alias between `react` and `preact/compat` on `tsconfig.json`,
```json
    ...
    "baseUrl": "./",
    "paths": {
      "react": ["node_modules/preact/compat"],
      "react-dom": ["node_modules/preact/compat"]
    },
    ...
```

I see `tsc` has TS7026 errors ("JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists.").
```
$ yarn tsc
yarn run v1.16.0
$ /Users/babie/src/github.com/babie/preact-cli-template/template/node_modules/.bin/tsc
src/App.tsx:5:5 - error TS7026: JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists.

5     <h1 className="title">Hello, Preact!</h1>
      ~~~~~~~~~~~~~~~~~~~~~~
...
```

And when I write this patch,
```patch
diff --git a/node_modules/preact/compat/src/index.d.ts b/node_modules/preact/compat/src/index.d.ts
index bdc044f..6314d96 100644
--- a/node_modules/preact/compat/src/index.d.ts
+++ b/node_modules/preact/compat/src/index.d.ts
@@ -11,6 +11,7 @@ export import createRef = preact.createRef;
 export import Fragment = preact.Fragment;
 export import createElement = preact.createElement
 export import cloneElement = preact.cloneElement
+export import JSX = preact.JSX
 
 export declare const version: string;
 
```

Then `tsc` has no error.
```
$ yarn tsc
yarn run v1.16.0
$ /Users/babie/src/github.com/babie/preact-cli-template/template/node_modules/.bin/tsc
✨  Done in 1.72s.
```

Probably, TypeScript needs a global(top level) `JSX` type. 

Here is a sample repository:
- found problem: https://github.com/babie/preact-cli-template/tree/b87af42e002e00e3e77febd9531319eabd628660
- fix problem: https://github.com/babie/preact-cli-template/tree/a57e2ea1276cac7c18bcefecf26e326064b428e6
- diff: https://github.com/babie/preact-cli-template/commit/a57e2ea1276cac7c18bcefecf26e326064b428e6

pmkroeker:
Initially this was just going to try and fix #1693, but got a bit away from me!

- Since Suspense was moved to `compat`, I figured the TS tests for it should move as well
- Component typing was being exported in the default object incorrectly
- Added JSX to the exports (fixes #1693)
- Namespaced the compat typings, such that they will work with React. This also allows importing both as `import * as React from 'preact/compat'` and `import React from preact/compat`. The later does require the `"allowSyntheticDefaultImports": true,` flag to be set in `tsconfig.json`. I figured that this was an ok compromise as the [TypeScript documentation](https://www.typescriptlang.org/docs/handbook/react-&-webpack.html) uses the `*` import method.

Unfortunately, this does hit this issue (https://github.com/microsoft/TypeScript/issues/30153). That while has been resolved, apparently can still crop up. Removing `export import JSX = preact.JSX;` does allow the compiler to run properly, but then JSX types are removed.

Task List:
- [x] Export JSX
- [x] Namespace for easier import method
- [x] Test when aliasing React + TS ✔️
- [x] Test using React + preact/compat in same project. ✔️


pmkroeker:
As a not I namespaced them under `React` as its the most likely usage if people are using `compat` as React (which is when these typings come into play)

compat/src/index.d.ts
```diff
@@ -1,75 +1,75 @@
...
-
-export declare interface PureComponent<P = {}, S = {}> extends preact.Component {
-  isPureReactComponenet: boolean;
+import * as _Suspense from './suspense';
+
+// export default React;
+export = React;
+declare namespace React {
+	// Export JSX
+	export import JSX = preact.JSX;
```
pmkroeker: This line currently causes `tsc` to crash. Removing it causes jsx to not be typed throwing errors.

JoviDeCroock: tsc crashes on this but we aren't seeing this in our tests or?

pmkroeker: Currently, JSX is not being exported in the compat typings (an oversight on my part in the initial PR). I added some specific tests for the typings, and running tsc with this line in will crash it, but removing the line causes `JSX.IntrinsicElements does not exist on JSX: any`.

pmkroeker: Should also note, that running `tsc` in `./tests/ts` also bails out with tsc crashing.

pmkroeker: Also, I modified an app from work to test this out, and it runs past this line with no problem (I use `awesome-typescript-loader` with webpack). On build it gives some different warnings:
```
WARNING in C:/Users/Peter/Documents/GitHub/preact/compat/dist/compat.module.js 1:5343-5344
"export 'Fragment' (imported as 'v') was not found in 'preact'
```

This may be build related as it doesn't seem TS related. Not sure.

pmkroeker: Fixed! After looking through the travis-ci logs I noticed that the typescript tests were passing, so it couldn't have been the TS version. Updated my global TS install to 3.5.1 and the issue is resolved!

compat/src/index.d.ts
```diff
@@ -1,75 +1,77 @@
...
-export declare function findDOMNode(component: preact.Component): Element | null;
-
-export declare interface PureComponent<P = {}, S = {}> extends preact.Component {
-  isPureReactComponenet: boolean;
+import * as _Suspense from './suspense';
+
+// export default React;
+export = React;
+export as namespace React;
+declare namespace React {
```
ForsakenHarmony: could this conflict with `@types/react`?

pmkroeker: It could as they also use the `React` namespace. I'll have to try them with react. I think that it you're using React + TS and using a bundle it will preferentially use the React typings.

I'll test this first thing tomorrow.

pmkroeker: Tested using my [Preact/TS boilerplate](https://github.com/pmkroeker/pmkroeker.github.io), with installed React + @types/react and its components and it works as intended. The typescript compiler runs before the webpack alias is resolved.

ForsakenHarmony: right, but if you don't import compat directly, do these typings even do anything?

pmkroeker: That's correct. Unless you either import from `preact/compat` or use an alias in tsconfig for `react: preact/compat`, these typings will not be used.

ForsakenHarmony: but what if we do import it 

and have `@types/react` somewhere else, i.e. a dependency that gets aliased

pmkroeker: I believe it will pick the typings that the dependency uses. With typescript its more difficult to mix and match React / Preact, as you have to specify the jsxFactory in your config. If they are both somehow referenced, the worst case scenario is the namespaces get merged together, which should be fine.

pmkroeker: I just tested out importing a React Component into a preact/compat component.
The basic pieces worked, but there is a difference in the typings between the 2 in how `componentDidCatch` is typed, which causes a build error.

pmkroeker: https://github.com/preactjs/preact/blob/a19aee98c64fca277f4ea06527c105f4c5363170/src/diff/index.js#L336 is the difference between the 2 typings. React passes both the error and errorInfo into `componentDidCatch(error, errorInfo)`, where as we just pass error. Adding the second argument to the typings resolved the issue, but possibly should be addressed.

ForsakenHarmony: well, `jsxFactory` doesn't matter for dependencies, because they're generally compiled

pmkroeker: True, I don't see any potential issues. If they call React internally they will use the React typings.

src/index.d.ts
```diff
@@ -94,7 +94,7 @@ declare namespace preact {
 		componentWillUpdate?(nextProps: Readonly<P>, nextState: Readonly<S>, nextContext: any): void;
 		getSnapshotBeforeUpdate?(oldProps: Readonly<P>, oldState: Readonly<S>): any;
 		componentDidUpdate?(previousProps: Readonly<P>, previousState: Readonly<S>, snapshot: any): void;
-		componentDidCatch?(error: any): void;
+		componentDidCatch?(error: any, errorInfo: any): void;
```
pmkroeker: This is needed as 2 args are passed into `componentDidCatch` in React, which causes an incompatible type equality. Currently, we only pass a single argument into `componentDidCatch`

pmkroeker: I think we could probably pass a part of the returned error object back as the second part. Could pass through `error.message`.

PR #1708, opened 2019-06-12
Fix cross browser assertions

marvinhagemeister:
This PR attempts to fix all cross-browser issues and attempts to make the CI green again. Size increase is mainly due to the IE11 workaround for `select.value`.

~~Adds `+41 B`~~ 😐
Adds `+35 B` 🙂

### DIFF
--- a/compat/test/browser/portals.test.js
+++ b/compat/test/browser/portals.test.js
@@ -83,7 +83,7 @@ describe('Portal', () => {
 		let set;
 
 		function Foo(props) {
-			const [additionalProps, setProps] = useState({ style: { background: 'red' } });
+			const [additionalProps, setProps] = useState({ style: { backgroundColor: 'red' } });
 			set = (c) => setProps(c);
 			return (
 				<div>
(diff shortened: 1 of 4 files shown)


marvinhagemeister:
Only 1 remaining test to go :muscle:

marvinhagemeister:
Dug a little deeper on the IE11 select issue with a real IE11 instance. Turns out that using `select.value = "A"` doesn't work at all in IE11 :/

These are the other alternatives, but they all require iteration:

- Use `select.selectIndex`
- Set `options.selected`

andrewiggins:
@marvinhagemeister Re: the `select.value` fix, what was failing? Was it the saucelabs tests? How did you test the IE11 select.value fix? I'm curious cuz I'd love to get the 35 bytes back to invest in other features (like an internal vnode object that wraps the public one).

I downloaded an IE11 Win7 VM from [modern.ie](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/) and ran this [jsbin](https://jsbin.com/fawezag) in IE11 on that VM and `select.value` seems work.

marvinhagemeister:
@andrewiggins I'm happy to revert it if a smaller fix can be found 👍 Yup, tested it with the same VM and a real Win7 IE11 machine we have at work in our QA department.

Just checked again and the linked jsbin works for both examples. What doesn't work is when the `select` is pasted into the demo app. For that I used the following component and reverted this PR locally:

```jsx
function App() {
	const [v, update] = useState(0);
	return (
		<div>
			<select value={v}>
				<option value="0">0</option>
				<option value="1">1</option>
				<option value="2">2</option>
				<option value="2">2</option>
				<option value="3">3</option>
			</select>
			<button onClick={() => update(Math.floor(Math.random() * 4))}>Update</button>
		</div>
	);
}

render(<App />, document.body);
```

Screencast of the bug:

ie11_select

PR #1713, opened 2019-06-13
(fix) - portals 2.0

JoviDeCroock:
# Current problems

## Initial render

I moved this to a solution where we prepend children to the root's _children array and after that call `forceUpdate` since roots are components. This in theory could be a good approach but when we initially render, forceUpdate won't do anything, since _vnode etc isn't defined yet. (I have tried setting an empty state didn't work either).

## Inserting multiple roots

Our former solution where we used `hydrate('', container)` makes it so that the second insertion hydrates the first portal. This in turn makes it so that we can't unmount the first one with `render(null, container)` anymore. This was a signifier that we should avoid using hydrate for portals.

## Switching container

At the moment I rely on `render(null, container)` to work, in theory this should be good enough for the time being but when we have a better solution we should test more thorough.

## Singular insertions

Note that the singular insertions at the moment are working since these fallback to `hydrate` and `render`, these tests are always on initial render etc so don't take these passes for granted.

## How did we get here?

First we encountered the issue that we didn't insert portals, we made the assumption that portals were rendering into empty roots.
We solved this by prepending the portalChild, this in turn unmounted all children when using `render(null, container)`. This was solved by using `hydrate('', container)`.

I could really use a few pointers on this since I feel like I've been going in the same circle for a while now.

Test cases added for the new issue report, thanks to @MichaelRando for the reproductions

Found a missing case to allow changing container, these two failing tests leave me jumbled though.
Been researching the hint provided about Fragments but this feels like odd behavior

- [x] support switching container
- [ ] support multiple portals in one root
- [ ] avoid irrelevant children from rerendering

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -2,7 +2,7 @@ import { hydrate, render as preactRender, cloneElement as preactCloneElement, cr
 import * as hooks from 'preact/hooks';
 export * from 'preact/hooks';
 import { Suspense as _Suspense, lazy as _lazy, catchRender } from './suspense';
-import { assign } from '../../src/util';
+import { assign, removeNode } from '../../src/util';
 
 const version = '16.8.0'; // trick libraries to think we are react
 
@@ -56,6 +56,11 @@ function handleElementVNode(vnode, props) {
  * @returns {import('./internal').Component | null} The root component reference or null
  */
 function render(vnode, parent, callback) {
+	// React destroys any existing DOM nodes, see #1727
+	while (parent.firstChild) {
+		removeNode(parent.firstChild);
+	}
+
 	preactRender(vnode, parent);
 	if (typeof callback==='function') callback();
 
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -20,8 +20,8 @@ options.event = e => {
 };
 
 let oldCatchRender = options._catchRender;
-options._catchRender = (error, component) => (
-	oldCatchRender && oldCatchRender(error, component) || catchRender(error, component)
+options._catchRender = (error, newVNode, oldVNode) => (
+	oldCatchRender && oldCatchRender(error, newVNode, oldVNode) || catchRender(error, newVNode, oldVNode)
 );
 
 /**
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -85,6 +85,10 @@ function Portal(props) {
 		this.container = container;
 	}
 
+	this.componentDidUpdate = () => {
+		this.container.__preact._component.forceUpdate();
+	};
+
 	render(wrap, container);
 	this.componentWillUnmount = () => {
 		render(null, container);
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -76,19 +76,45 @@ class ContextProvider {
 	}
 }
 
-function addRenderRoot(vnode, parent) {
-	if (parent._children && parent._children._component && parent._children._component._vnode) {
-		parent._children.props.children.unshift(vnode);
-		parent._children._component.forceUpdate();
-	}
-	else {
-		// We should insert here aswell but since the vnode doesn't
-		// exist yet we can't forceUpdate.... Most commonly during
-		// initial render.
-		hydrate('', parent);
-		render(vnode, parent);
-	}
-}
+// Solution 1 I tried:
+// This fails because it doesn't hydrate and just removes the already present DOM
+//
+// function Portal(props) {
+// 	let wrap = h(ContextProvider, { context: this.context }, props.vnode);
+// 	let container = props.container;
+// 	let temp = document.createTextNode('');
+// 	if (container.hasChildNodes()) {
+// 		container.insertBefore(temp, container.firstChild);
+// 	}
+// 	else {
+// 		container.appendChild(temp);
+// 	}
+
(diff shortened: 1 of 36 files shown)


compat/test/browser/portals.test.js
```diff
@@ -79,6 +79,80 @@ describe('Portal', () => {
...
+		rerender();
+		expect(scratch.innerHTML).to.equal('<div>foobar</div><div>foobar</div><div><p>Hello</p></div>');
+
+		toggle2();
+		rerender();
+		expect(scratch.innerHTML).to.equal('<div>foobar</div><div><p>Hello</p></div>');
+
+		toggle();
+		rerender();
+		expect(scratch.innerHTML).to.equal('<div><p>Hello</p></div>');
```
JoviDeCroock: I think due to hydrating it won't unmount the supplementary node here. Option: use the unmount method exposed by preact?

compat/test/browser/portals.test.js
```diff
@@ -79,6 +79,80 @@ describe('Portal', () => {
...
+				</div>
+			);
+		}
+
+		render(<Foo><div>foobar</div></Foo>, scratch);
+		expect(scratch.innerHTML).to.equal('<div><p>Hello</p></div>');
+
+		toggle();
+		rerender();
+		expect(scratch.innerHTML).to.equal('<div>foobar</div><div><p>Hello</p></div>');
```
JoviDeCroock: Due to the children starting out as false something goes wrong here, still investigating this one.

marvinhagemeister: This could be related to Fragment diffing because of the root Fragment that gets created for each `render()`

JoviDeCroock: So calling render twice with other insides won't change due to that diffing?

compat/test/browser/portals.test.js
```diff
@@ -79,6 +79,143 @@ describe('Portal', () => {
...
+		toggle2();
+		rerender();
+		expect(scratch.innerHTML).to.equal('<div>foobar</div><div><p>Hello</p></div>');
+
+		toggle();
+		rerender();
+		expect(scratch.innerHTML).to.equal('<div><p>Hello</p></div>');
+	});
+
+	it('should work with changing the container', () => {
```
JoviDeCroock: A case we seemed to be missing

compat/src/index.js
```diff
@@ -76,11 +76,12 @@ class ContextProvider {
  * @param {object | null | undefined} props
  */
 function Portal(props) {
-	let wrap = h(ContextProvider, { context: this.context }, props.vnode);
+	let wrap = h(ContextProvider, { context: this.context }, props.vnode || null);
```
JoviDeCroock: TODO: remove this null defaulting it didn't help

src/render.js
```diff
@@ -1,9 +1,23 @@
 import { EMPTY_OBJ, EMPTY_ARR } from './constants';
-import { commitRoot } from './diff/index';
-import { diffChildren } from './diff/children';
+import { commitRoot, diff } from './diff/index';
 import { createElement, Fragment } from './create-element';
 import options from './options';
 
+function addRenderRoot(vnode, parent) {
```
andrewiggins: Should this live with Portal in compat? I think only Portals should add to the render root. Calling `render` multiple times on the same container is suppose to replace the elements in the container. Portal is special cuz it doesn't. Then perhaps you can do `root.children.push()`

andrewiggins: So instead of `render(wrap, container)` in Portal, perhaps it should be `addRenderRoot(wrap, container)` and `addRenderRoot` just calls `render` if there is no existing root, else it appends the child to the existing root's children and runs `forceUpdate`.

JoviDeCroock: In that case we need to be able to check if prev and nextNode are different no? Else we call addRenderRoot every render

JoviDeCroock: Well, something needs to change anyway too much tests goofing up, trying to find out why first :D

JoviDeCroock:
The problem with the hydrate approach seems to clearly be when we insert two portals in to the same root. This goofes up the unmounting of the second insertion... There does not seem to be a clearcut solution for this on first render. First iteration we can't use forceUpdate since the _vode isn't present on the dom node yet.

JoviDeCroock:
# Current problems

## Initial render

I moved this to a solution where we prepend children to the root's _children array and after that call `forceUpdate` since roots are components. This in theory could be a good approach but when we initially render, forceUpdate won't do anything, since _vnode etc isn't defined yet. (I have tried setting an empty state didn't work either).

## Inserting multiple roots

Our former solution where we used `hydrate('', container)` makes it so that the second insertion hydrates the first portal. This in turn makes it so that we can't unmount the first one with `render(null, container)` anymore. This was a signifier that we should avoid using hydrate for portals.

## Switching container

At the moment I rely on `render(null, container)` to work, in theory this should be good enough for the time being but when we have a better solution we should test more thorough.

## How did we get here?

First we encountered the issue that we didn't insert portals, we made the assumption that portals were rendering into empty roots.
We solved this by prepending the portalChild, this in turn unmounted all children when using `render(null, container)`. This was solved by using `hydrate('', container)`.

I could really use a few pointers on this since I feel like I've been going in the same circle for a while now.

JoviDeCroock:
Closing this due to remake

PR #1715, opened 2019-06-14
Begin rendering with diff instead of diffChildren and other golf ⛳

andrewiggins:
Beginning the render with diff means that the initial Fragment will be rendered as a real Component making our `parent._children` pointer more useful (will have meaningful `_dom` pointers, will have `forceUpdate` function after initial render).

Total change: -12 B


andrewiggins:
Coveralls is failing cuz I skipped two tests in debug cuz the implementation doesn't work as expected. Let's fix that debug feature in a separate PR.

PR #1716, opened 2019-06-18
Flatten children array as it is diffed

andrewiggins:
Instead of passing over the newVNode's children array twice (once in `toChildArray` to flatten it, and once again in `diffChildren` to diff the children), we now pass over the children once, diffing the children and flattening the list as we go.

Total change:
~~+6 B (22e0fba)~~
-4 B (ffc1713)

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -24,9 +24,8 @@ import { getDomSibling } from '../component';
  * @param {boolean} isHydrating Whether or not we are in hydration
  */
 export function diffChildren(parentDom, newParentVNode, oldParentVNode, context, isSvg, excessDomChildren, mounts, oldDom, isHydrating) {
-	let childVNode, i, j, oldVNode, newDom, sibDom, firstChildDom, refs;
+	let i, j, oldVNode, newDom, sibDom, firstChildDom, refs;
 
-	let newChildren = newParentVNode._children || toChildArray(newParentVNode.props.children, newParentVNode._children=[], coerceToVNode, true);
 	// This is a compression of oldParentVNode!=null && oldParentVNode != EMPTY_OBJ && oldParentVNode._children || EMPTY_ARR
 	// as EMPTY_OBJ._children should be `undefined`.
 	let oldChildren = (oldParentVNode && oldParentVNode._children) || EMPTY_ARR;
@@ -49,8 +48,8 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 		}
 	}
 
-	for (i=0; i<newChildren.length; i++) {
-		childVNode = newChildren[i] = coerceToVNode(newChildren[i]);
+	i=0;
+	newParentVNode._children = toChildArray(newParentVNode._children, childVNode => {
 
 		if (childVNode!=null) {
 			childVNode._parent = newParentVNode;
@@ -150,7 +149,10 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 				}
 			}
 		}
-	}
+
+		i++;
+		return childVNode;
+	});
 
 	newParentVNode._dom = firstChildDom;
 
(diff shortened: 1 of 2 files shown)


andrewiggins:
If anyone has the time and setup, would love to know if this PR moves the perf at all

src/diff/children.js
```diff
@@ -48,8 +47,8 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 		}
 	}
 
-	for (i=0; i<newChildren.length; i++) {
-		childVNode = newChildren[i] = coerceToVNode(newChildren[i]);
+	i=0;
+	newParentVNode._children = toChildArray(newParentVNode._children, childVNode => {
```
developit: @andrewiggins any chance you tried checking the size with this inner function pulled out? I feel like the argument passing to account for locals would be a nightmare though.

src/diff/index.js
```diff
@@ -238,6 +238,7 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
 			dom.multiple = newProps.multiple;
 		}
 
+		newVNode._children = newVNode.props.children;
```
developit: I'm probably missing something - isn't this duplicative of line 51?

cristianbote:
Hey @andrewiggins took a stab at it and tried to resolve the conflicts 🕺 If something it's not right, please let me know.

src/diff/index.js
andrewiggins: Yea, looking again, this is one of the weirder parts of this change. `diffChildren` expects the children of the VNode to already be on the `_children` property. It then loops through the `_children` property and flattens the result which is re-assigned to `_children`.

I did it this way because the children of a VNode are different for Components vs DOM VNodes. Components' children are the result of render while DOM VNodes are the JSX children (`props.children`). To keep the logic of determining what the children are in this function I needed a way to pass the resolved children into `diffChildren`. I did this by sticking the resolved children onto `_children`.

So this line puts the JSX children onto `_children`. Line 118 is the corresponding line for Components.

src/diff/children.js
andrewiggins: I did not. I think I talked myself out of that because, assuming `toChildArray` works as a callback, it'd significantly complicate this code. We'd have to capture all the args to call the callback with in an array, add the `childVNode` to that list, and then call `.apply` on the callback. So something like:

```js
toChildArray(vnode._children, diffChild, [...captured locals...]);

function diffChild(...capturedArgs..., childVNode) {
    // actual diff code
}

function toChildArray(children, callback, args) {
    if (...) {
        args.push(childVNode);
        flattened.push(callback.apply(null, args));
    }
    else {
         // similar code to above
    }
}
```

This felt like a lot of ceremony and I didn't think it'd save us bytes. If you think it might, I'd be happy to give it a shot.

For this PR I also didn't consider a non-recursive solution cuz in my small experience they've been bigger but perhaps that's worth an exploration in the future?

src/diff/children.js
```diff
@@ -169,27 +171,27 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
...
- * will be applied on each child if the `vnode` is not `null`
- * @param {boolean} [keepHoles] wether to coerce `undefined` to `null` or not.
- * This is needed for Components without children like `<Foo />`.
+ * @param {(vnode: import('../internal').VNode) => import('../internal').VNode} [callback]
+ * A function to invoke for each child before it is added to the flattened list.
+ * @param {import('../internal').VNode[]} [flattened] An flat array of children to modify
+ * @returns {import('../internal').VNode[]}
  */
-export function toChildArray(children, flattened, map, keepHoles) {
+export function toChildArray(children, callback, flattened) {
```
developit: IIRC the first two args here are used externally. Maybe worth changing to `toChildArray(children, flattened, callback)`?

marvinhagemeister: I think that change was made for golfing purposes in `children.js`. There it's used without the `flattened` argument. Either way we've only advertised the first arg so far in our types and elsewhere. So I think it's fine to swap them like it's done here.

PR #1717, opened 2019-06-18
Add functions to `debug` for inspecting the rendered VNode tree

robertknight:
This adds a set of functions to `preact/debug` that can be used by development/testing tools to inspect the rendered component tree:

```ts
// Get the root vnode that was rendered into a DOM element.
getVNodeFromContainer(dom): VNode|null

// Get the output that were produced when a VNode or Component instance was last rendered.
getLastRenderOutput(containerOrVNode): VNode[]|null

// For rendered vnodes with a string `type` or text nodes, get the corresponding DOM node.
getDOMNode(vnode): Node|null

// For rendered vnodes with a function `type`, get the `Component` instance.
getComponent(vnode): Component|null
```

The initial use case is for the Enzyme adapter, which currently relies on [its own version of this](https://github.com/preactjs/enzyme-adapter-preact-pure/blob/master/src/preact10-internals.ts) and is vulnerable to breaking if Preact's internals change (eg. the `_prevNode` => `_children` change).

The API only exposes the minimal amount of information that the Enzyme
adpater needs, with the hope of not getting in the way of future changes
to Preact's internals.

At this stage what I'd like feedback on is whether other devs are happy with
exposing this information via APIs in the debug package, as well as any feedback
on the function names etc. If/when folks are happy with the APIs being added, I'll finish up the tests.

----

- [ ] Get approval on the new APIs being added
- [ ] Finish up the tests

### DIFF
--- a/debug/src/index.js
+++ b/debug/src/index.js
@@ -1,5 +1,6 @@
 import { initDebug } from './debug';
 import { initDevTools } from './devtools';
+export { getVNodeFromContainer, getLastRenderOutput, getDOMNode, getComponent } from './inspect';
 
 if (process.env.NODE_ENV==='development') {
 	initDebug();
--- a/debug/src/inspect.js
+++ b/debug/src/inspect.js
@@ -1,0 +1,58 @@
+/**
+ * Functions for inspecting a rendered VNode tree.
+ */
+
+/**
+ * Return the VNode that was most recently rendered into a DOM container using
+ * `render`.
+ *
+ * The rendered children of the returned VNode can be inspected using
+ * `getLastRenderOutput`.
+ *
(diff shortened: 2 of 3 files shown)


marvinhagemeister:
The more I work on devtools, the more I'm leaning in favor of this PR. I just created similar accessor functions for them which look exactly like the ones here. One function I'd like to add is `getVNodeParent` for `vnode._parent` and something for `_lastDomChild`

Thompson1985 (approved):

JoviDeCroock:
Closing: unfinished draft adding debug VNode-tree inspection for the enzyme adapter (now unmaintained); preact-devtools serves this today.

PR #1720, opened 2019-06-19
use object.is instead of === to check the same state in hooks

hackerxian:
use object.is can check NaN for hook state change, and can follow react : [https://reactjs.org/docs/hooks-reference.html#bailing-out-of-a-state-update](https://reactjs.org/docs/hooks-reference.html#bailing-out-of-a-state-update)

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -9,6 +9,15 @@ let currentComponent;
 /** @type {Array<import('./internal').Component>} */
 let afterPaintEffects = [];
 
+
+/**
+ * Object.is polyfill
+ * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is
+ */
+function is(x, y) {
+	return (x === y && (x !== 0 || 1 / x === 1 / y)) || (x !== x && y !== y);
+}
+
 let oldBeforeRender = options._render;
 options._render = vnode => {
 	if (oldBeforeRender) oldBeforeRender(vnode);
@@ -84,7 +93,7 @@ export function useReducer(reducer, initialState, init) {
 
 			action => {
 				const nextValue = reducer(hookState._value[0], action);
-				if (hookState._value[0]!==nextValue) {
+				if (!is(hookState._value[0], nextValue)) {
 					hookState._value[0] = nextValue;
 					hookState._component.setState({});
 				}
--- a/hooks/test/browser/useState.test.js
+++ b/hooks/test/browser/useState.test.js
@@ -80,6 +80,32 @@ describe('useState', () => {
 		expect(Comp).to.be.calledOnce;
 	});
 
+	it('does not rerender on same value', () => {
+		let lastState;
+		let doSetState;
+
+		const Comp = sinon.spy(() => {
+			const [state, setState] = useState(NaN);
+			lastState = state;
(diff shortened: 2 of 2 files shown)


marvinhagemeister:
`Object.is` is [not available in IE11](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is#Browser_compatibility), so we'd need a polyfill for that.

hackerxian:
add Object.is polyfill

hackerxian:
Has shared method or dir to have commonly used?

robertknight:
I note that React defines that new-style Context update bailouts [are also defined to use `Object.is`](https://reactjs.org/docs/context.html#contextprovider).

andrewiggins:
Hi @hackerxian! Welcome :smile: Thanks for taking the time to open a PR into Preact.

It looks like supporting `Object.is` adds 34 bytes (+4%) to the hooks bundle. Originally I think we went with just triple-equals check to save on bytes because we (at least I) didn't think `+0`, `-0`, and `NaN` were commonly used values to justify the cost.

Did you run into a situation where you needed to differentiate between `+0` and `-0` (by default `-0 === 0 === +0`) or compare `NaN` values? Could you describe that situation?

Again, thanks!

hackerxian:
Hi,  @andrewiggins , Thanks for replying. 
My situation maybe less, because I handle this `parseInt('asdf') vs parseInt('123')` to set correct number in the form input and set illegal number to avoid rerender. I know it’s easy that I can resolve this to add validate form. 

I just want to stay the same with react and avoid `NaN === NaN` to rerender. If you think this will add more bytes and no earnings. That's ok.

Connormiha:
I think the library should not worry about polyfills and just use modern browser's API `as is`. Real projects injects `core-js` or other polyfill in frontend bundle if they need. @marvinhagemeister

robertknight:
> I think the library should not worry about polyfills and just use modern browser's API as is. 

That would be a breaking change for users, so needs some careful consideration. Preact currently works in IE 11 and later without any polyfills at all (not even `Promise` or `Object.assign`), which is quite nice from the point of view of making it easy to use the library for some small widget on a site.

That said, we could probably make some moderate size savings + performance improvements by requiring newer standard library features.

JoviDeCroock:
> I think the library should not worry about polyfills and just use modern browser's API `as is`. Real projects injects `core-js` or other polyfill in frontend bundle if they need. @marvinhagemeister

People don't target `node_modules` with `babel-loader` so this won't be added when the application uses the `usage` paramter in `babel-preset-env`. This will be added with `entry`.

developit:
In order to switch to leveraging newer stdlib features, we'd have to provide two copies of the library. That's increasingly workable via bundlers, but it'll still be a support burden.

cristianbote:
Hey @hackerxian, thank you so much for taking the time to open this PR and go through the hurdles! We really appreciate your contribution!
With the above thread, seems like supporting this type of same value check, would add bytes that currently we can't afford. Also, not adding the polyfill would mean drop support for certain browsers as @robertknight stated.
If you don't mind we could close this and down the road try to revisit. Seems like also this could be solved in user-land, as you mentioned.

Once again, awesome job! 💯

Issue #1722, opened 2019-06-19
componentWillUnmount not called


ivaynberg:
Given the following

```
class MyComponent extends Component {
  componentDidMount() {
    console.log("Component mounted.");
  }
  componentWillUnmount() {
    console.log("Component unmounted");
  }
  render() {
    return <div>My Component</div>;
  }
}

var container=document.createElement("div");
document.body.appendChild(container);
render(<MyComponent />, document.body, container);
render(<div>Not my component</div>, document.body, container);
```

I expect componentWillUnmount() to be called, and it has been in X alphas and beta0, but in beta1 and beta2 it no longer is.

Sandbox: https://codesandbox.io/s/1288-5mfm5

pmkroeker:
Which browser are you using? I just opened your Sandbox and I'm getting `Component unmounted` in the console. On Chrome.

JoviDeCroock:
Same here, it logs the unmount for me.

ivaynberg:
using Chrome

JoviDeCroock:
Your reproduction was not inline with your issue statement, I corrected it. Is there a reason as to why you are using replaceNode? I'll look into the issue though, as far as I can see you can avoid using replaceNode

ivaynberg:
i am using preact to enhance existing select components, so i need to render a preact component after a specific dom node. i do this by inserting the placeholder div into the right place in the dom and rendering a preact component into it.

since an existing select component does not have to be a last child of a container i cannot use the two arg variant of render()

JoviDeCroock:
All right, looking into it as we speak, I have an inclination as to what is causing this

PR #1723, opened 2019-06-19
(fix) - replaceNode should unmount when diffing twice with replaceNode

JoviDeCroock:
Fixes: https://github.com/preactjs/preact/issues/1722

Adds 7Bytes

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -13,13 +13,16 @@ import options from './options';
  */
 export function render(vnode, parentDom, replaceNode) {
 	if (options._root) options._root(vnode, parentDom);
-	let oldVNode = parentDom._children;
+	let oldVNode = replaceNode ?
+		replaceNode._children || parentDom._children :
+		parentDom._children;
+
 	vnode = createElement(Fragment, null, [vnode]);
 
 	let mounts = [];
 	diff(
 		parentDom,
-		replaceNode ? vnode : (parentDom._children = vnode),
+		replaceNode ? (replaceNode._children = vnode) : (parentDom._children = vnode),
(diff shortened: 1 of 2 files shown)


src/render.js
```diff
@@ -13,13 +13,16 @@ import options from './options';
  */
 export function render(vnode, parentDom, replaceNode) {
 	if (options._root) options._root(vnode, parentDom);
-	let oldVNode = parentDom._children;
+	let oldVNode = replaceNode ?
+		replaceNode._children || parentDom._children :
```
JoviDeCroock: This `||` is needed for the case when we first did `render(x, y, undefined)` and transition to a render with replaceNode. The second is needed for when we use replaceNode twice in a row.

JoviDeCroock:
@cristianbote you are right, it seems to be after the most recent master merge, probably something that can be gzipped more efficiently since it occurs more.

src/render.js
developit: What about this?

```suggestion
replaceNode && replaceNode._children || parentDom._children
```

src/render.js
```diff
@@ -13,13 +13,16 @@ import options from './options';
...
+		parentDom._children;
+
 	vnode = createElement(Fragment, null, [vnode]);
 
 	let mounts = [];
 	diff(
 		parentDom,
-		replaceNode ? vnode : (parentDom._children = vnode),
+		replaceNode ? (replaceNode._children = vnode) : (parentDom._children = vnode),
```
developit: Possibly smaller:

```suggestion
		(replaceNode || parentDom)._children = vnode,
```

PR #1724, opened 2019-06-20
Update suspense so tests only rely on public API

andrewiggins:
The goal of this PR is to abstract the suspense tests away from the implementation to enable experimentation. Ideally, our tests should use the public api as much as possible to match how our users will use features. This makes our tests a good validation that our code is meeting our user's expectations.

With that in mind, I tried to avoid using internal properties on VNode - they may change over time, and ideally we can change them without having rewrite these tests each time. I also try to avoid using `forceUpdate` directly and instead rely on our `rerender` method helper to simulate our rerender queue flushing.

While working on this I also rewrote `_childDidSuspend` to reduce it's size a bit. I removed the `maxDuration` feature since we don't have any tests for it yet and it's not yet released in React. If we preemptively ship `maxDuration` before React, we may accidentally introduce some compat problems that aren't fixable without a major version bump. So personally I'd prefer to wait for them first.

Total change in compat:
-63 B :smile:

### DIFF
--- a/compat/src/internal.d.ts
+++ b/compat/src/internal.d.ts
@@ -1,7 +1,7 @@
 import { Ref } from '../..';
 import {
-  Component as PreactComponent,
-  VNode as PreactVNode,
+	Component as PreactComponent,
+	VNode as PreactVNode,
 	FunctionalComponent as PreactFunctionalComponent
 } from '../../src/internal';
 import { SuspenseProps } from './suspense';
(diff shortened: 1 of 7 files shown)


compat/src/internal.d.ts
```diff
@@ -1,30 +1,41 @@
...
-  shouldComponentUpdate?(nextProps: Readonly<P>): boolean;
-  _forwarded?: true;
+	shouldComponentUpdate?(nextProps: Readonly<P>): boolean;
+	_forwarded?: true;
 }
 
 export interface VNode<T = any> extends PreactVNode<T> {
-  $$typeof: symbol | string;
-  preactCompatNormalized: boolean;
+	$$typeof?: symbol | string;
```
andrewiggins: Made these optional cuz they are only on the VNode after the `vnode` option is invoked so technically they are undefined for a moment, but really because it makes VSCode happy lol

compat/src/suspense.js
```diff
@@ -30,15 +30,15 @@ export function catchRender(error, newVNode, oldVNode) {
 	return false;
 }
 
-function removeDom(children) {
+function detachDom(children) {
```
andrewiggins: Renamed this to `detachDom` since it only detaches a subtree from the screen and doesn't remove every element

compat/src/suspense.js
```diff
@@ -59,70 +59,36 @@ Suspense.prototype = new Component();
  * @param {Promise} promise The thrown promise
  */
 Suspense.prototype._childDidSuspend = function(promise) {
-	// saves 5B
-	const _this = this;
 
-	_this._suspensions.push(promise);
-	let len = _this._suspensions.length;
+	/** @type {import('./internal').SuspenseComponent} */
+	const c = this;
```
andrewiggins: Went with `c` here instead of `_this` cuz it matches `diff` a little more. We use `c` to refer to the currently rendering component in `diff` so I though aligning the name here would make it more consistent.

PR #1725, opened 2019-06-20
typings: add context as 3rd type parameter for the Component interface

mk0x9:
So TypeScript compiler will type check it :)

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -84,24 +84,25 @@ declare namespace preact {
 	// Type alias for a component instance considered generally, whether stateless or stateful.
 	type AnyComponent<P = {}, S = {}> = FunctionComponent<P> | Component<P, S>;
 
-	interface Component<P = {}, S = {}> {
+	interface Component<P = {}, S = {}, C = any> {
 		componentWillMount?(): void;
 		componentDidMount?(): void;
 		componentWillUnmount?(): void;
 		getChildContext?(): object;
-		componentWillReceiveProps?(nextProps: Readonly<P>, nextContext: any): void;
-		shouldComponentUpdate?(nextProps: Readonly<P>, nextState: Readonly<S>, nextContext: any): boolean;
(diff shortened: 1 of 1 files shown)


src/index.d.ts
```diff
@@ -84,24 +84,25 @@ declare namespace preact {
...
-	interface Component<P = {}, S = {}> {
+	interface Component<P = {}, S = {}, C = any> {
 		componentWillMount?(): void;
 		componentDidMount?(): void;
 		componentWillUnmount?(): void;
 		getChildContext?(): object;
-		componentWillReceiveProps?(nextProps: Readonly<P>, nextContext: any): void;
-		shouldComponentUpdate?(nextProps: Readonly<P>, nextState: Readonly<S>, nextContext: any): boolean;
-		componentWillUpdate?(nextProps: Readonly<P>, nextState: Readonly<S>, nextContext: any): void;
+		componentWillReceiveProps?(nextProps: Readonly<P>, nextContext: C): void;
```
developit: Thinking we should mark context as Readonly here like props and state are treated.

src/index.d.ts
```diff
@@ -115,7 +116,7 @@ declare namespace preact {
 
 		state: Readonly<S>;
 		props: RenderableProps<P>;
-		context: any;
+		context: C;
```
developit: This might be odd when using `contextType`? Not sure. Using `contextType` assigns `this.context` to the current value of the given context provider.

mk0x9: You mean that it's possible to assign any `Context` to the static `contextType`? Or that `contextType` could be dynamically switched?

src/index.d.ts
mk0x9: Makes sense, thanks!

pmkroeker:
It would be great if this could be looked at again. Currently using contextType does not set the type of context in a class. It does set the proper type of class.contextType.

Thompson1985 (approved):

Thompson1985 (approved):

PR #1734, opened 2019-06-25
Golf bytes

wizzardo:
moved typeof to separate function (-101B)

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -1,4 +1,4 @@
-import {assign, isFunction} from './util';
+import { assign, isFunction } from './util';
 import { diff, commitRoot } from './diff/index';
 import options from './options';
 import { Fragment } from './create-element';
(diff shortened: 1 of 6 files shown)


wizzardo:
Interesting.. I compared js before compression

JoviDeCroock:
gzip compression is based on recurring mentions. We always base our lost bytes minzipped. This can be seen by executing the `build` script.

wizzardo:
Sorry, didn't know this detail, my target was to minimise raw js because browser needs to parse it every time, but js.gz can be cached

wizzardo:
Ok, so please decline this pr as it's not relevant

JoviDeCroock:
No problem, @wizzardo it was a fun exploration to see and thank you for spending your time trying this.

PR #1735, opened 2019-06-25
Add return type to _childDidSuspend

pmkroeker:
Having strict ts settings throws if function does not specify a return type. Adding `: void` fixes the issue.


pmkroeker:
My pleasure, slowly switching our apps over to X, so may come across a few more in the next little while (but hopefully not).

PR #1740, opened 2019-06-26
Fix up compat types and tests

andrewiggins:


compat/src/index.js
```diff
@@ -179,7 +178,6 @@ function normalizeVNode(vnode) {
 function cloneElement(element) {
 	if (!isValidElement(element)) return element;
 	let vnode = normalizeVNode(preactCloneElement.apply(null, arguments));
-	vnode.$$typeof = REACT_ELEMENT_TYPE;
```
andrewiggins: This is already set in `options.vnode` hook which `preactCloneElement` calls so no need to reset here I think

compat/src/index.d.ts
```diff
@@ -53,11 +53,11 @@ declare namespace React {
 
 	export function unmountComponentAtNode(container: Element | Document | ShadowRoot | DocumentFragment): boolean;
 
-	export function createFactory(type: preact.VNode["type"]): preact.VNode<{}>;
+	export function createFactory(type: preact.VNode["type"]): (props?: any, ...children: preact.ComponentChildren[]) => preact.VNode<any>;
```
andrewiggins: https://reactjs.org/blog/2014/10/14/introducing-react-elements.html#creating-a-reactelement

compat/src/index.d.ts
```diff
@@ -53,11 +53,11 @@ declare namespace React {
...
 	export function isValidElement(element: any): boolean;
 	export function findDOMNode(component: preact.Component): Element | null;
 
-	export interface PureComponent<P = {}, S = {}> extends preact.Component {
+	export class PureComponent<P = {}, S = {}> extends preact.Component {
```
andrewiggins: `PureComponent` is an actual class that should be extended by other classes, not an interface that any object implement

compat/src/index.d.ts
```diff
@@ -67,7 +67,7 @@ declare namespace React {
 
 	export function unstable_batchedUpdates(callback: (arg?: any) => void, arg?: any): void;
 
-	export interface Children {
+	export const Children: {
```
andrewiggins: Updating the type to be a real object, not an interface for some other object to implement

compat/src/index.js
```diff
@@ -366,14 +364,14 @@ options.vnode = vnode => {
...
+ * @param {(arg: Arg) => void} callback function that triggers the updatd
+ * @param {Arg} [arg] Optional argument that can be passed to the callback
  */
 // eslint-disable-next-line camelcase
-function unstable_batchedUpdates(callback, arg) {
-	callback(arg);
-}
+const unstable_batchedUpdates = (callback, arg) => callback(arg);
 
+export * from 'preact/hooks';
```
andrewiggins: Moved all exports to the bottom of the file so they are grouped together

compat/src/internal.d.ts
```diff
@@ -19,7 +19,7 @@ export interface Component<P = {}, S = {}> extends PreactComponent<P, S> {
 
 export interface FunctionalComponent<P = {}> extends PreactFunctionalComponent<P> {
 	shouldComponentUpdate?(nextProps: Readonly<P>): boolean;
-	_forwarded?: true;
+	_forwarded?: boolean;
```
andrewiggins: Makes VSCode happier when used in `compat/index.js`

compat/src/suspense.d.ts
```diff
@@ -10,4 +10,4 @@ export interface SuspenseProps {
   fallback: preact.ComponentChildren;
 }
 
-export abstract class Suspense extends Component<SuspenseProps> {}
+export class Suspense extends Component<SuspenseProps> {}
```
andrewiggins: Suspense isn't supposed to be extended so removed the abstract keyword

compat/src/index.js
```diff
@@ -366,14 +364,14 @@ options.vnode = vnode => {
 /**
  * Deprecated way to control batched rendering inside the reconciler, but we
  * already schedule in batches inside our rendering code
- * @param {(a) => void} callback function that triggers the updatd
- * @param {*} [arg] Optional argument that can be passed to the callback
+ * @template Arg
+ * @param {(arg: Arg) => void} callback function that triggers the updatd
```
marvinhagemeister: Minor typo in the end. It was present before but maybe can be fixed Addy with this PR 🙂

andrewiggins: Oh yup! Good catch. Fixed :)

ForsakenHarmony:
What about the coverage?

Also types and tests would maybe be a bit better in separate prs 🤔

andrewiggins:
I think the coverage failing was a fluke. It passed on a second run.

And yea, perhaps they are better as separate PRs. Whoops. Next time.

PR #1742, opened 2019-06-27
Replace catchRender with catchError

andrewiggins:
This PR changes the behavior of the `catchError` option to enable composing of catchError option hooks (similar to middleware). The initial value of `options.catchError` is set to `catchErrorInComponent`. Any code that overrides `options.catchError` should store the previous value of `options.catchError` and invoke it if they want to rely on the default implementation of `catchError`. This pattern allows code to skip the default implementation of if an option hook chooses not to invoke the previous catchError behavior. It also gives options the ability to change what the default implementation is invoked with.

Two new features this pattern enables:
1. Using catchError to replace catchRender (to implement Suspense)
2. Adding debug error for "Missing Suspense"

In order to enable using `catchError` for Suspense, I had to change the behavior the of `catchError` to take in the VNode that threw error, instead of the first parent. An initial implementation caused some problems in how errors were caught that our tests didn't catch so I modified some tests to cover this scenario. Namely, if a direct child of an error boundary throws, the error boundary should catch that error. The initial implementation was skipping the first parent if a child threw in unmount and refs.

Change:
preact.js: -18 B
compat.js: -10 B

### DIFF
--- a/compat/mangle.json
+++ b/compat/mangle.json
@@ -23,7 +23,6 @@
       "$_render": "__r",
       "$_hook": "__h",
       "$_catchError": "__e",
-      "$_catchRender": "__E",
       "$_unmount": "_e"
     }
   }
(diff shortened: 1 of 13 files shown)


compat/mangle.json
```diff
@@ -23,7 +23,6 @@
       "$_render": "__r",
       "$_hook": "__h",
       "$_catchError": "__e",
-      "$_catchRender": "__E",
```
marvinhagemeister: I'm not sure if we can remove the mangle setting here as `Suspense` in compat still relies on a consistent name with core.

PR #1743, opened 2019-06-28
Optimize creating array of childrens

Connormiha:
I just optimized createing array of children from arguments. I used slice instead of `for loop` + `push`.
It works 2x faster in Chrome and Firefox.
https://jsperf.com/preact-create-element-children


src/create-element.js
```diff
@@ -13,10 +13,7 @@ export function createElement(type, props, children) {
 	props = assign({}, props);
 
 	if (arguments.length>3) {
-		children = [children];
-		for (let i=3; i<arguments.length; i++) {
-			children.push(arguments[i]);
-		}
+		children = Array.prototype.slice.call(arguments, 2);
```
cristianbote: Really like this one 👍 😄I feel like the initial version was stuff that got generated from a spread operator 🤔? Also, this comes with a nice `-29B` so I _really_ dig it!

Question: what if we drop the `if(arguments.length > 3)` completely? And just rely on:
```
children = Array.prototype.slice.call(arguments, 2);
```

Connormiha: If we get rid of `if(arguments.length > 3)`, than we can change logic, because `children` will be an array always.

Connormiha: If we have only 3 arguments, than we will be create additional array. We need to check if it hits performance.

cristianbote: Ah, actually `children` could be something else than an `array`. I think it can't be dropped the `if` I mean.
Also, this actually shaves `-12B` which is still not bad 🎉!

developit:
It seems worth checking a few other approaches here before settling on one.
Here's a comparison of 4:
https://jsperf.com/preact-create-element-children/6

Based on those results, it seems like `[].slice.call(arguments, 2)` is best.

Connormiha:
I suggest to use slice literal instead of prototype(as bonus reduce size). Safari has a much smaller users count than Chromium + Firefox.

lukeed:
Avoiding reassignment to the `children` argument seems to be the biggest gain across the board. IIRC, doing so still de-opts and would explain the gain.

I've been hunting down similar optimizations for a while in my experimentation. I ported a few of the candidates over to the jsperf here since it's a good environment 👍 

https://jsperf.com/preact-create-element-children/10

developit:
@lukeed that seems correct, since we have no guarantee that Preact is loaded in strict mode.

lukeed:
Ah, right, sorry. My `_if` tests preserve that behavior (pulled from experiment) but forgot to include that for the `_noreuse` tests.

Updated: https://jsperf.com/preact-create-element-children/12

Connormiha:
@developit So which option to choose?

cristianbote:
IMO `slice` version, has the highest score on Chrome only. With that being said, we should go for `for_noreuse` version, but, we need to reuse it to keep the functionality in place. Maybe it's best to keep what we have now?

developit:
I think the only change we could justify here would be dropping the base-case. Instead of initializing `children=[arguments[2]]` and iterating from `arguments[3..k]`, just initialize `children=[]` and iterate from `2..k`. This would likely save a couple bytes, and offers a tiny performance improvement. It's unlikely to amount to anything visible in a benchmark though.

lukeed:
@developit That looks to be `for_noreuse`, right? Pasting here to save a click – it does seem to be within the Top 3 performers across browsers.

```js
  function for_noreuse(type, props, child) {
    let i=2, len=arguments.length, children=child;
  	if (len > 3) {
  		for (children=[]; i < len; i++) {
  			children.push(arguments[i]);
  		}
  	}
    return { type, props, children } 
  }
```

GZip may be better if `arguments.length` is not extracted to a variable

src/create-element.js
```diff
@@ -13,10 +13,7 @@ export function createElement(type, props, children) {
 	props = assign({}, props);
 
 	if (arguments.length>3) {
-		children = [children];
-		for (let i=3; i<arguments.length; i++) {
-			children.push(arguments[i]);
-		}
+		children = [].slice.call(arguments, 2);
```
cristianbote: As @developit said, the best change we could have right now, would be these:

Start from 2, with and empty `children`
```suggestion
		children = [];
		for (let i=2; i<arguments.length; i++) {
			children.push(arguments[i]);
		}
```

cristianbote: And also change the condition inside the if to `arguments.length>2`

cristianbote: This shaves `-4B` 😃

Connormiha: `arguments.length>2` => `arguments.length>2` will effect the next `if`.
```js
if (children!=null) {
    props.children = children;
}
```
Children never be null.

developit: right, I'd leave the other conditional as-is.

PR #1747, opened 2019-07-01
(fix) - calling setState callback in will mount

JoviDeCroock:
Adds 10 Bytes to the main package
Doesn't fix this yet: https://github.com/preactjs/preact/issues/556
Does fix willMount getting called

At this point the call order is still: cbWillMount --> didMount while in React other way around


test/browser/lifecycles/componentWillMount.test.js
```diff
@@ -40,5 +43,66 @@ describe('Lifecycle methods', () => {
...
+				}
+			}
+
+			render(<Foo />, scratch);
+
+			expect(spy).to.be.calledOnce;
+			expect(componentState).to.deep.equal({ value: 1 });
+		});
+
+		it.skip('should call setState callback in correct order', () => {
```
JoviDeCroock: I don't think this is possible yet, since we call `diff` in `forceUpdate` and then `commitRoot` --> this function calls the `didMount`. in `diff` however we resolve the setState callbacks.

src/component.js
```diff
@@ -49,8 +49,11 @@ Component.prototype.setState = function(update, callback) {
 	// Skip update if updater function returned null
 	if (update==null) return;
 
+	// The check for _renderCallbacks is because we can't call
+	// this when we're in the constructor but we should call it in
+	// will mount.
+	if (callback && Array.isArray(this._renderCallbacks)) this._renderCallbacks.push(callback);
 	if (this._vnode) {
-		if (callback) this._renderCallbacks.push(callback);
```
developit: What about creating it here if it doesn't exist?

```js
		if (callback) (this._renderCallbacks || (this._renderCallbacks = [])).push(callback);
```

JoviDeCroock: That would make constructors work in constructor as well, is this intended?

developit: I don't know, haha. It seems intuitive to me that the callback would always be called. It seems like this would call the setState callback after initial render, which seems reasonable. Not sure what React does.

JoviDeCroock:
@developit the newly proposed solution adds 4 more bytes but enables us to call the CB in constructor

Issue #1741, opened 2019-06-27
Callback on nested setState doesn't called with update state


edimitchel:
Hi !

On this issue (#556), I discuss with @JoviDeCroock about an issue when I use `setState` on the component~~Will~~**Did**Mount.

My issue concerns the callback of a nested `setState` (into another `setState` callback). The state inside the nested callback is not updated (was it called at the same time as the parent callback ?). This weird way to handle state is not mine, it's a legacy code, but it's usefull for some SSR needs (we use same components for React, React with SSR and Preact).

You can see the wrong behavior on [this codesandbox](https://codesandbox.io/s/556-glj62).

Issue #1736, opened 2019-06-25
`options._commit` is called on child although parent commit is in progress


marvinhagemeister:
While working on the new devtools adapter in #1549 , I came across an issue where `options._commit` is called on a child although there is a pending commit already in progress on a parent. This can be reproduced with the following snippet ([codesandbox](https://codesandbox.io/s/optionscommit-6bgox) link for the lazy):

```jsx
import { h, render, Component, options } from "preact";

let i = 0;
// options.commit, should only be called once
options.__c = vnode => console.log(i++, vnode);

class App extends Component {
  componentDidMount() {
    this.forceUpdate();
  }
  render() {
    return <div>foo</div>;
  }
}

render(<App />, document.getElementById("app"));
```

This pattern may seem strange, but it's actually something that is used in `mobx-react`. It can be observed with our people demo app when you navigate to `/people` (make sure the detail view is blank via a reload) and then click on a person name in the left to mount the detail view.

JoviDeCroock:
All right just thinking about this so the forceUpdate should be added to the pending commit right?

andrewiggins:
I think this may point to a difference in how React's `forceUpdate` works and ours. I believe React's `forceUpdate` schedules an update to happen after the current render/commit phase (sans any optimizations they may apply). If that's true, their `forceUpdate` behaves more like our `setState({})`.

Our `forceUpdate` synchronously does a diff right in the middle of whatever call stack it is in.

Perhaps we should change our `forceUpdate`  to call `setState({})` and inline the current forceUpdate implementation inside of the `processQueue` function? Might save us some difficult compat issues down the road (and also makes thinking about `forceUpdate` easier since it behaves just like `setState` instead of doing something wildly different)

I used this codesandbox to test my theory: https://codesandbox.io/s/gallant-hellman-4cfkk. Notice how in React `=== Parent cDM ===` is logged before the second `ForceUpdate.render` (implying React schedules the forceUpdate to occur after the current diff stack) whereas in Preact `=== Parent.cDM ===` is logged after the second `ForceUpdate.render`.

JoviDeCroock:
That change would imply that we would have to change our way of testing I assume, since if we move to setState we would have to call `rerender()` everytime we use `forceUpdate`

jeremy-coleman:
would just wrapping with promise.resolve().then(()=>forceUpdate()) fix the timing? It would still be sync, but im pretty sure it would cause forceUpdate to run after cdm

[...]

PR #1748, opened 2019-07-01
(fix) - nested setState calls with accurate this.state

JoviDeCroock:
Adds 7bytes to the main package
Fixes: https://github.com/preactjs/preact/issues/1741


developit:
This is clever. @JoviDeCroock I think I assumed the fix here would exist in setState() - what's the reason for moving it into the update loop?

src/diff/index.js
```diff
@@ -128,7 +128,10 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 
 			c.base = newVNode._dom;
 
-			while (tmp=c._renderCallbacks.pop()) tmp.call(c);
+			while (tmp=c._renderCallbacks.pop()) {
+				if (c._nextState) { c.state = c._nextState; }
```
developit: This is clever. @JoviDeCroock I think I assumed the fix here would exist in setState() - what's the reason for moving it into the update loop?

test/browser/components.test.js
```diff
@@ -132,6 +132,42 @@ describe('Components', () => {
...
+			let finalState;
+			class Foo extends Component {
+				constructor(props) {
+					super(props);
+					this.state = { a: 'b' };
+				}
+
+				componentDidMount() {
+					states.push(this.state);
+					// eslint-disable-next-line
```
developit: (we could definitely remove this lint rule globally IMO)

src/diff/index.js
JoviDeCroock: Well we could do it in setState but I thought it would be easier to make this happen in the callbacks personally.

It looked like the correct place for me, also felt like the other was increasing size more drastically. Would you prefer me making the second as an example as well?

developit: I guess I would be curious to see the difference. Did you have the other version lying around?

I wonder... we have this at the beginning of setState:

```js
let s = (this._nextState!==this.state && this._nextState) || (this._nextState = assign({}, this.state));
```

Maybe that just needs to be moved into forceUpdate?

JoviDeCroock: I can take some time to reproduce it this afternoon, hmm why move that to forceUpdate? The callbacks are fired subsequently no?

developit: yeah, just I'm wondering if there's a way to have setState() and the render loop both use the same shared logic, rather than different cases for each. (I might be way off here!)

JoviDeCroock: I'll try to see if this is possible, this could maybe be a first step to solving https://github.com/preactjs/preact/issues/1736

JoviDeCroock: I think doing so would make `componentDidUpdate` reflect invalid prevState

edimitchel:
Is there any hope of getting this correction before the release of the stable version of Preact X?
In any case, thanks a lot for your hard work !

marvinhagemeister:
@edimitchel I've actually come across a project which had a minor but valid use case for nesting `setState`. So we're merging this one here :tada:

PR #1749, opened 2019-07-01
(fix) - portals

JoviDeCroock:
Adds ~~101~~ ~~81~~  ~~123~~ ~~114~~ ~~103~~ 88B to the compat package 👎 

- [x] support switching container
- [x] avoid irrelevant children from rerendering
- [x] support placeholder vnodes
- [x] support multiple portals in one root
- [x] support nested portals

This is a remake of: https://github.com/preactjs/preact/pull/1713
I needed to clear my head a bit and start over I felt like I was stuck in a rut.

I have almost gotten to the fix the only remaining issue is that when we insert another portal in the same container and unmount one of them we lose both. I don't know if this is an issue or not.... Let's assume it is, working on a fix for it but it can possibly get quite complicated.

One of the options would be to check the _children of the array for an existing portal and if there is one remove from the array.... I’m not sure about this

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -80,8 +80,8 @@ class ContextProvider {
  * @param {object | null | undefined} props
  */
 function Portal(props) {
-	let wrap = h(ContextProvider, { context: this.context }, props.vnode);
 	let container = props.container;
+	let wrap = h(ContextProvider, { context: this.context }, props.vnode);
 
 	// When we change container we should clear our old container and
 	// indicate a new mount.
@@ -100,12 +100,7 @@ function Portal(props) {
 			// wrap into the container.
 			hydrate('', container);
 			// If it has child nodes we need to insert before the first child.
-			if (container.hasChildNodes()) {
-				container.insertBefore(temp, container.firstChild);
-			}
-			else {
-				container.appendChild(temp);
-			}
+			container.insertBefore(temp, container.firstChild);
 			this.hasMounted = true;
 			this.container = container;
 			// Render our wrapping element into temp.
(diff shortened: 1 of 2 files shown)


compat/test/browser/portals.test.js
```diff
@@ -178,8 +178,145 @@ describe('Portal', () => {
 		}
 
 		render(<App />, root);
-		expect(dialog.childNodes.length).to.equal(1);
+		expect(dialog.childNodes.length).to.equal(2);
```
JoviDeCroock: This is because of the empty textNode, one would expect `replaceNode` to delete it but for some reason it isn't deleted.

compat/src/index.js
```diff
@@ -80,18 +80,56 @@ class ContextProvider {
...
+			_this.hasMounted = true;
+			_this.container = container;
+			// Render our wrapping element into temp.
+			preactRender(wrap, container, _this.temp);
+		}
+		else {
+			// When we have mounted and the vnode is present it means the
+			// props have changed or a parent is triggering a rerender.
+			// This implies we only need to call render.
+			render(wrap, container);
```
developit: Why render here instead of preactRender?  If there isn't a difference, this conditional could be reduced:

```js
let temp;
if (props.vnode) {
  if (!_this.hasMounted) {
    temp = document.createTextNode('');
    hydrate('', container);
    container.insertBefore(temp, container.firstChild);
    _this.hasMounted = true;
    _this.container = container;
  }
  preactRender(wrap, container, temp);
}
```

JoviDeCroock: Using `replaceNode` here isn't possible since this can only be used on initially rendering

PR #1752, opened 2019-07-04
Fix compat types

### LINKED ISSUE #1751: forwardRef in TypeScript
```typescript
import { h } from 'preact'
import { useRef, useImperativeHandle } from 'preact/hooks'
import { ForwardFn } from 'preact/compat/src/internal'
import { forwardRef } from 'preact/compat'

export interface MyInputProps {
  id: string
}
export interface MyInputHandles {
  focus (): void
}

const MyInput: ForwardFn<MyInputProps, MyInputHandles> = (props, ref) => { // ts error
  const inputRef = useRef<HTMLInputElement>()

  useImperativeHandle(ref, () => ({ // ts error 2
    focus: () => {
      if (inputRef.current) {
        inputRef.current.focus()
      }
    }
  }))

  return <input {...props} ref={inputRef} />
}

export default forwardRef(MyInput)
```

Will report error:

```
TS2322: Type '(props: MyInputProps, ref: Ref<MyInputHandles>) => Element' is not assignable to type 'ForwardFn<MyInputProps, MyInputHandles>'.
  Type 'Element' is missing the following properties from type 'VNode<any>': $$typeof, preactCompatNormalized, _children, _parent, and 4 more.
```

```
TS2345: Argument of type 'Ref<MyInputHandles>' is not assignable to parameter of type 'Ref<MyInputHandles | null | undefined>'.
  Type 'RefObject<MyInputHandles>' is not assignable to type 'Ref<MyInputHandles | null | undefined>'.
    Property 'current' is optional in type 'RefObject<MyInputHandles>' but required in type 'Ref<MyInputHandles | null | undefined>'.
```

marvinhagemeister:
Fixes #1751 .

cc to our typings god @pmkroeker

### DIFF
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -64,7 +64,7 @@ declare namespace React {
 
 	export function memo<P = {}>(component: preact.FunctionalComponent<P>, comparer?: (prev: P, next: P) => boolean): preact.FunctionComponent<P>;
 
-	export function forwardRef<P = {}>(fn: _internal.ForwardFn<P, any>): preact.FunctionalComponent<P>;
+	export function forwardRef<R, P = {}>(fn: _internal.ForwardFn<P, R>): preact.FunctionalComponent<P>;
 
 	export function unstable_batchedUpdates(callback: (arg?: any) => void, arg?: any): void;
 
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -1,7 +1,7 @@
 import * as _hooks from '../../hooks';
 import * as preact from '../../src';
 import { JSXInternal } from '../../src/jsx'
-import { ForwardFn } from './internal';
+import * as _internal from './internal';
 import * as _Suspense from './suspense';
 
 // export default React;
@@ -43,6 +43,7 @@ declare namespace React {
 	export import lazy = _Suspense.lazy;
 
 	// Compat
+	export import ForwardFn = _internal.ForwardFn;
 	export const version: string;
 
 	export function createPortal(vnode: preact.VNode, container: Element): preact.VNode<any>;
@@ -63,7 +64,7 @@ declare namespace React {
 
 	export function memo<P = {}>(component: preact.FunctionalComponent<P>, comparer?: (prev: P, next: P) => boolean): preact.FunctionComponent<P>;
 
-	export function forwardRef<P = {}>(fn: ForwardFn<P, any>): preact.FunctionalComponent<P>;
+	export function forwardRef<P = {}>(fn: _internal.ForwardFn<P, any>): preact.FunctionalComponent<P>;
 
 	export function unstable_batchedUpdates(callback: (arg?: any) => void, arg?: any): void;
 
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -58,7 +58,7 @@ declare namespace React {
 	export function isValidElement(element: any): boolean;
 	export function findDOMNode(component: preact.Component): Element | null;
 
-	export class PureComponent<P = {}, S = {}> extends preact.Component {
+	export abstract class PureComponent<P = {}, S = {}> extends preact.Component {
 		isPureReactComponent: boolean;
 	}
 
--- a/compat/src/suspense.d.ts
+++ b/compat/src/suspense.d.ts
@@ -10,4 +10,4 @@ export interface SuspenseProps {
   fallback: preact.ComponentChildren;
 }
 
-export class Suspense extends Component<SuspenseProps> {}
+export abstract class Suspense extends Component<SuspenseProps> {}
--- a/compat/src/suspense.d.ts
+++ b/compat/src/suspense.d.ts
@@ -1,4 +1,4 @@
-import { Component } from '../../src';
+import { Component, ComponentChild } from '../../src';
 
 //
 // Suspense/lazy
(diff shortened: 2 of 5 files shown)


weijarz:
Does `CreateHandle` also need to be parameterized?

```typescript
type CreateHandle<T> = () => T;

export function useImperativeHandle<T>(ref: PreactRef<T>, create: CreateHandle<T>, inputs?: Inputs): void;
```

compat/src/index.d.ts
```diff
@@ -57,13 +58,13 @@ declare namespace React {
...
 
-	export class PureComponent<P = {}, S = {}> extends preact.Component {
+	export abstract class PureComponent<P = {}, S = {}> extends preact.Component {
 		isPureReactComponenet: boolean;
 	}
 
 	export function memo<P = {}>(component: preact.FunctionalComponent<P>, comparer?: (prev: P, next: P) => boolean): preact.FunctionComponent<P>;
 
-	export function forwardRef<P = {}>(fn: ForwardFn<P, any>): preact.FunctionalComponent<P>;
+	export function forwardRef<P = {}>(fn: _internal.ForwardFn<P, any>): preact.FunctionalComponent<P>;
```
JoviDeCroock: We offer forwarding a ref but we can't type the ref being forwarded and this seems to be possible in React under the following:

```
function forwardRef<T, P = {}>(
  Component: RefForwardingComponent<T, P>
): ComponentType<P & ClassAttributes<T>>
```

JoviDeCroock: We offer forwarding a ref but we can't type the ref being forwarded and this seems to be possible in React under the following:

function forwardRef<T, P = {}>(
  Component: RefForwardingComponent<T, P>
): ComponentType<P & ClassAttributes<T>>

compat/src/suspense.d.ts
```diff
@@ -10,4 +10,4 @@ export interface SuspenseProps {
   fallback: preact.ComponentChildren;
 }
 
-export class Suspense extends Component<SuspenseProps> {}
+export abstract class Suspense extends Component<SuspenseProps> {}
```
developit: What's the reason for this being abstract? Just to avoid `new Suspense`?

marvinhagemeister: Close, it's because we declared `Component` as abstract and it expects classes that extend it to declare a render method. Because the definition didn't have one this failed. Maybe the better way is just to declare the render method here.

ForsakenHarmony: alternatively `export const Suspense: Component<SuspenseProps>;`, but not sure if that would be idiomatic typescript

pmkroeker: Declaring render is probably the best solution. Prevents people from thinking its extendable.

hooks/src/index.d.ts
```diff
@@ -74,7 +74,7 @@ type CreateHandle = () => object;
  * ref.current
  * @param inputs If present, effect will only activate if the values in the list change (using ===).
  */
-export function useImperativeHandle<T>(ref: Ref<T>, create: CreateHandle, inputs?: Inputs): void;
+export function useImperativeHandle<T>(ref: PreactRef<T>, create: CreateHandle, inputs?: Inputs): void;
```
ForsakenHarmony: as mentioned above, `CreateHandle` should probably also have the type param, could copy from the react typings

```suggestion
export function useImperativeHandle<T, R extends T>(ref: PreactRef<T>, create: () => R, inputs?: Inputs): void;
```

ForsakenHarmony: also have to remove the type above though

sometimes think maybe we should just extend the react types instead of maintaining them independently?

compat/src/index.d.ts
```diff
@@ -57,13 +58,13 @@ declare namespace React {
 	export function isValidElement(element: any): boolean;
 	export function findDOMNode(component: preact.Component): Element | null;
 
-	export class PureComponent<P = {}, S = {}> extends preact.Component {
+	export abstract class PureComponent<P = {}, S = {}> extends preact.Component {
```
pmkroeker: I agree with `PureComponent` being abstract. Classes are derived from it so abstract makes sense.

compat/src/suspense.d.ts
pmkroeker: `render` should pull from the same typings as the abstract:
```
render(props?: RenderableProps<P>, state?: Readonly<S>, context?: any): ComponentChild;
```

compat/src/index.d.ts
ondratra: It shouldn't be abstract. Otherwise it couldn't be used as `<Suspense>` tag. Imho this is proper declaration:
```
import { Component, RenderableProps, ComponentChild } from '../../src';
...
export class Suspense extends Component<SuspenseProps> {
    render(props?: RenderableProps<SuspenseProps>, state?: Readonly<{}>, context?: any): ComponentChild;
}
```

More reasoning in https://github.com/preactjs/preact/pull/1785

pmkroeker: My comment here was for `PureComponent`, which should be abstract.

ondratra: Sorry, entered comment into wrong box.

compat/src/suspense.d.ts
ondratra: Imho the typing should be like this:
```
import { Component, RenderableProps, ComponentChild } from '../../src';
...
export class Suspense extends Component<SuspenseProps> {
    render(props: RenderableProps<SuspenseProps>, state: Readonly<{}>): ComponentChild;
}
```
Because in file `suspense.js` there is implementation accepting props and state
```
Suspense.prototype.render = function(props, state) {
	return state._parkedChildren ? props.fallback : props.children;
};
```

hooks/src/index.d.ts
marvinhagemeister: That's something we can investigate once X is out :+1:

PR #1755, opened 2019-07-06
Minor debug enhancements.

marvinhagemeister:
This PR improves code coverage in `preact/debug` and fixes some debug errors not being thrown.

- `diff()` will never receive a `null` vnode. `diffChildren` will skip `diff` if encounters a null-vnode
- `props` can never be `null` either
- Fix "Missing Suspense"-error never being thrown
- Improve overall code-coverage of `preact/debug`

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -32,6 +32,12 @@ export function initDebug() {
 					break;
 				}
 			}
+
+			// We haven't recovered and we know at this point that there is no
+			// Suspense component higher up in the tree
+			if (error instanceof Error) {
+				throw error;
+			}
 		}
 
 		oldCatchError(error, vnode, oldVNode);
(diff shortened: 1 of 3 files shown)


debug/src/debug.js
```diff
@@ -77,7 +81,7 @@ export function initDebug() {
 			throw new Error('Invalid type passed to createElement(): '+(Array.isArray(type) ? 'array' : type));
 		}
 
-		if ((type==='thead' || type==='tfoot' || type==='thead') && parentVNode.type!=='table') {
```
JoviDeCroock: Damn, that's my bad how did I overlook that

marvinhagemeister: Well I missed it in my review, so it's on me :sweat_smile:

PR #1761, opened 2019-07-07
Inline compat vnode hook

ForsakenHarmony:
Removes one of the side effects for the rest of the app

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -129,7 +129,15 @@ let Children = {
 function createElement(...args) {
 	let vnode = h(...args);
 
+	vnode.$$typeof = REACT_ELEMENT_TYPE;
+
+	applyEventNormalization(vnode);
 	let type = vnode.type, props = vnode.props;
+	if (type && type._forwarded && vnode.ref) {
+		vnode.props.ref = vnode.ref;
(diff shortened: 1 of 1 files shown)


ForsakenHarmony:
Hmm ok, cloneElement uses it as well

compat/src/index.js
```diff
@@ -129,7 +129,15 @@ let Children = {
 function createElement(...args) {
 	let vnode = h(...args);
 
+	vnode.$$typeof = REACT_ELEMENT_TYPE;
```
developit: IIRC this means it's not possible to do this anymore:


```js
import { createElement } from 'preact';
import { isValidElement } from 'react-is';

isValidElement(<div />)
```

This seems important since CLI creates non-compat vnodes for all JSX.

developit:
I wonder if there is a way we could move the side-effecting version into a getter? So that it would get installed globally, but only when something like `Component` is referenced.

ForsakenHarmony:
realised the big implicit change I hate is the event thing where it changes `onChange` to `onInput` where I explicitly wanted `onChange`

ForsakenHarmony:
not entirely sure how to solve it, but this is probably not it

PR #1762, opened 2019-07-08
Add Web Maker to real world app list

chinchang:


developit:
@chinchang make sure to add it to the website too!

chinchang:
@developit already added there :)

developit:
FWIW we are likely phasing out the readme list in favor of the website list(s). We're going to do a better job of highlighting users in the future!

PR #1763, opened 2019-07-10
Add fallback for requestanimationframe for when tab is not focused

### LINKED ISSUE #1758: useEffect is not called when tab is not focused
I found a difference from react :grimacing:. It looks like Preact isn't calling components' `useEffect` callback when that tab isn't focused. I've been looking into why this is, but I haven't been able to figure it out yet.

I created codesandboxes to compare the React and Preact behaviors:

https://codesandbox.io/s/react-useeffect-p3f7f
https://p3f7f.codesandbox.io/
https://codesandbox.io/s/preact-useeffect-60zl9
https://60zl9.codesandbox.io/

How to reproduce the issue with the Preact sandbox:
1. [Open the sandbox link](https://60zl9.codesandbox.io/)
2. Open devtools
3. Refresh the page and as quickly as possible switch away from that tab, before it loads.
4. Wait for ~10 seconds
5. Go back to the tab
6.
   **Expected behavior**
   - Preact should have fired the root component callback while you were on another tab
   - The subcomponent should have started rendering, and called its own effect callback
   - By the time you returned, the subcomponent's fake "network data" promise should have already resolved, so the page should display: `Subcomponent data: "much data"`

   **Actual behavior**
   - The root component callback does not fire until you go back to the tab (see the `console.log`s)

React is behaving as expected

calebeby:
Closes #1758 

Is there a way to write tests for this? I manually tested it, and it worked fine, but it would be nice to have automated tests for it.

~+18b gz~ :slightly_frowning_face: Any ideas for improvement?
+25b gz


hooks/src/index.js
```diff
@@ -227,7 +223,11 @@ if (typeof window !== 'undefined') {
 				options.requestAnimationFrame(flushAfterPaintEffects);
```
developit: As @sventschui mentioned in chat, I think we can compress this:

```js
const RAF_TIMEOUT = 100;

function safeRaf(callback) {
	const done = () => {
		clearTimeout(timeout);
		cancelAnimationFrame(raf);
		setTimeout(callback);
	};
	const timeout = setTimeout(done, RAF_TIMEOUT);
	const raf = requestAnimationFrame(done);
}

	afterPaint = (component) => {

		(options.requestAnimationFrame || safeRaf)(flushAfterPaintEffects);
```

marvinhagemeister:
Thank you so much for the PR :100:

calebeby:
Thanks, glad to help out 😄

PR #1766, opened 2019-07-10
Move babel testing config into Karma config

developit:


marvinhagemeister:
It looks we need to have a global `babelrc` file for the shared tests. I'm wondering if we should drop them entirely as they only duplicate some of the existing browser tests.

karma.conf.js
```diff
@@ -131,10 +131,21 @@ module.exports = function(config) {
...
+									loose: true,
+									exclude: ['transform-es2015-typeof-symbol'],
+									targets: {
+										browsers: ['last 2 versions', 'IE >= 9']
+									}
+								}]
+							],
+							plugins: [
+								'transform-object-rest-spread',
+								'transform-react-jsx',
```
JoviDeCroock: For some reason this isn't being picked up, the tests aren't handling jsx

PR #1769, opened 2019-07-12
(feat) - compat mode for preact 8

JoviDeCroock:
This is in no way a definite solution yet, this will point to what would need to change in order to support our previous way of working correctly.

Small POC at 8-compat mode, this is in core for now this should probably move but we don't want to force an import of preact/compat. This would entail a 2KB increase for what could be several bytes instead.

I still have one failing test but this is more or less to show what the implications would be in supporting 8 mode, my knowledge of 8 is quite limited so I'd rather have some more eyes on this.


src/diff/index.js
```diff
@@ -100,7 +101,7 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 			}
 
 			oldProps = c.props;
-			oldState = c.state;
+			oldState = c._prevState || {};
```
JoviDeCroock: to handle the last failing test case this should be altered to `|| c.state`

src/LooseMode.js
```diff
@@ -0,0 +1,19 @@
+import { Component, options } from 'preact';
+
+export default function LooseMode() {}
+
+LooseMode.prototype = new Component();
+
+LooseMode.prototype.componentDidMount = function() {
+	let oldVnodeHook = options.vnode;
```
developit: Right now this re-installs the hook on every mount. Just needs a flag to flip once installed so it doesn't happen again:

```js
let oldVnodeHook = false;

LooseMode.prototype.componentDidMount = function() {
  if (oldVnodeHook===false) {
    oldVnodeHook = options.vnode;
    options.vnode = vnode => {
      // etc
    };
  }
};
```

src/LegacyComponent.js
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Update component state and schedule a re-render.
+ * @param {object | ((s: object, p: object) => object)} update A hash of state
+ * properties to update with new values or a function that given the current
+ * state and props returns a new partial state
+ * @param {() => void} [callback] A function to be called once component state is
+ * updated
+ */
+Component.prototype.setState = function(update, callback) {
```
developit: Hmm - should this be LegacyComponent?

```suggestion
LegacyComponent.prototype.setState = function(update, callback) {
```

src/LegacyComponent.js
```diff
@@ -0,0 +1,25 @@
+import { assign } from './util';
+import { Component, enqueueRender } from './Component';
+
+export function LegacyComponent(){}
+
+LegacyComponent.prototype = Component;
```
developit:
```suggestion
LegacyComponent.prototype = new Component;
```

src/component.js
```diff
@@ -38,9 +38,9 @@ export function Component(props, context) {
  * updated
  */
 Component.prototype.setState = function(update, callback) {
+	if (this.state) this._prevState = this.state;
```
developit: A thought:  could we (ab)use a getter/setter pair on `LegacyComponent#state` to restore the old behaviour?

```js
// emulates forward-looking .state by storing values in ._state and prioritizing return of `._nextState`.
const stateDescriptor = {
  get() {
    return this._nextState || this._state;
  },
  set(newState) {
    this._state = newState;
  }
};

function LegacyComponent(p, c) {
  Object.defineProperty(this, 'state', stateDescriptor);
  Component.call(this, p, c);
}
```

Issue #1770, opened 2019-07-12
Portal bug after update from 10.0.0-beta.3 to 10.0.0-rc.0


pronevich:
I am using antd modal in react project:

```
Uncaught (in promise) TypeError: Cannot read property 'forEach' of null
    at preact__WEBPACK_IMPORTED_MODULE_0__.options.diffed (debug.js?c91e:203)
    at preact__WEBPACK_IMPORTED_MODULE_0__.options.diffed (index.js?d798:137)
    at preact__WEBPACK_IMPORTED_MODULE_0__.options.diffed (index.js?a178:26)
    at P (index.js?c988:144)
    at g (children.js?b0ec:78)
    at P (index.js?c988:98)
    at Portal.p.forceUpdate (component.js?dcb2:65)
    at Portal.createContainer (Portal.js?2ca0:41)
    at Portal.componentDidMount (Portal.js?2ca0:21)
    at N (index.js?c988:160)
```

JoviDeCroock:
Could you check if it works when you don't use `preact/debug`. I'm just wondering, this seems to be because `_list` on the component at hand is undefined.

pronevich:
Possibly `preact/debug` added by preact-cli, don't see any way to turn it off

MichaelRando:
removing `preact/debug` fixes it for my app. But adding `preact/debug` to my tests doesn't trigger it; interestingly.

JoviDeCroock:
This error could be related to the order of the options.diffed being overridden.... so defaulting in /debug for effect checks should fix the issue

PR #1773, opened 2019-07-13
(fix) - debug import

JoviDeCroock:
When debug would be imported before compat/hooks this would result in a crash since _list and so on wasn't initialized yet.

Fixes: https://github.com/preactjs/preact/issues/1770


debug/src/debug.js
```diff
@@ -198,9 +198,11 @@ export function initDebug() {
 	};
 
 	options.diffed = (vnode) => {
+		if (oldDiffed) oldDiffed(vnode);
```
JoviDeCroock: Moving this up top ensures the call for the older one first.

debug/src/debug.js
```diff
@@ -198,9 +198,11 @@ export function initDebug() {
...
+
 		if (vnode._component && vnode._component.__hooks) {
 			let hooks = vnode._component.__hooks;
-			hooks._list.forEach(hook => {
+			(hooks._list || []).forEach(hook => {
```
JoviDeCroock: Default anyway if there are for example two or more positions where optionss.diffed is used.

PR #1779, opened 2019-07-16
Add Partial to setState

### LINKED ISSUE #1778: (TypeScript) Why can't I setState with an object containing the keys of my state type
Why doesn't this work:

```typescript
export interface IState {
    username: string;
    password: string;
}

// ...

private onInputChange(e: Event, field: keyof IState) {
    const target = e.target as HTMLInputElement;
    this.setState({ [field]: target.value });
}
```

image

Am I incorrect in thinking this should work? `field` is by definition one of the keys of `IState`, so defining an object containing only `field` as keys should be a valid `IState` object. Is there an issue with the preact type declarations, or have I misunderstood something?

pmkroeker:
Fixes some typing issues when using `keyof State` as a key in setState. Fixes #1778.

Allows for this:
```typescript
interface IState {
  username: string;
  password: string;
}

function onInputChange(e: Event, field: keyof IState) {
  const target = e.target as HTMLInputElement;
  setState({ [field]: target.value });
}
```

Without throwing errors of missing fields.


marvinhagemeister:
Sweet :+1: :100: Thank you so much for the PR. Let's add a little test case to make sure were not running into this in the future again. Just declaring the component in the linked issue should do it :tada:

PR #1781, opened 2019-07-17
Fix Portal children always being mounted

### LINKED ISSUE #1780: [10.0.0-rc.0] Portals DOM always destroyed on parent update
Still can't switch to preactX, even on rc.0. I tried it again and I'm presently seeing the focus lost per text character input from the keyboard, on the form in the modal. It's not #1694 nor fixed by #1691; but my test case from before does show the prob. I've updated it at:
https://github.com/MichaelRando/preact-text-problem

marvinhagemeister:
This PR fixes an issue where the children of a `Portal` component would always be treated as new children, leading to incorrect `componentDidMount` calls. By switching to the raw preact `render` function we skip the `dom` from being destroyed in an update scenario.

~~The downside of this is that this definitely breaks the stacking scenario where multiple trees are rendered into the same `dom` node. I'm wondering if we should reject that case in general and push rethinking root handling after the X release.~~

Nevermind the stacking bit. Found a way to fix that too :tada: 

Fixes #1780

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -103,6 +103,7 @@ function Portal(props) {
 			_this._container = container;
 			// Render our wrapping element into temp.
 			preactRender(wrap, container, _this._temp);
+			_this._children = this._temp._children;
 		}
 		else {
 			// When we have mounted and the vnode is present it means the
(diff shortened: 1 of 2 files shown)


steffenmllr:
Awesome fix @marvinhagemeister - this is the bug I just ran into :) Do you plan on making a `10.0.0-rc.1` release before the X release ?

marvinhagemeister:
@steffenmllr Good call, we should release another RC version soon. We need a bit more time on the docs site anyway. Th site is mostly ready and we're currently working on updating the content :tada:

steffenmllr:
@marvinhagemeister sounds great! Is there an easy way to install from git master? 
I've tried `npm install preactjs/preact#master` but this fails because of the monorepo (I think)

JoviDeCroock:
if you npm install from a branch, then you will have to build. So you would have to include ```cd node_modules/preact && npm i && npm run build```

Issue #1783, opened 2019-07-18
Render bug with replaceNode parameter after update from 10.0.0-beta.3 to 10.0.0-rc.0


jasson2788:
Hello guys. After upgrading preact to its newest version the styles are no longer applied when using a replaceNode parameter in the render function.

Here's an example: 

```
import { h, render, Component, JSX } from 'preact';

const parent = document.getElementById('parent') as HTMLElement;
const child = document.createElement('div') as HTMLDivElement;
parent.appendChild(child);

class Comp extends Component {
    public render = (): JSX.Element => {
        return <div style={{ height: '100px' }}>Preact</div>;
    }
}
	
render(<Comp />, parent, child);
```

When you call the render function without the replaceNode parameter the styles are applied correctly:

`render(<Comp />, parent);`

Have a great day everyone.

cristianbote:
Hi @jasson2788, thanks for opening this issue.

So what's happening here, is that the `third` param for the `render` method is used as a starting point for kicking in the hydration process. Since `-rc.0` we're ignoring the props during hydration.

PR #1786, opened 2019-07-20
(feat) - diff props when replacing a node and don't diff when hydrating

JoviDeCroock:
One concern I'm having is what we export in /compat since we just export  our render function as hydrate, this does not seem right with the thigns we are changing at this moment in time.

Seems to add 15B

Fixes: https://github.com/preactjs/preact/issues/1783

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -13,8 +13,13 @@ import options from './options';
  */
 export function render(vnode, parentDom, replaceNode) {
 	if (options._root) options._root(vnode, parentDom);
-	let oldVNode = replaceNode && replaceNode._children || parentDom._children;
 
+	let oldVNode = replaceNode && replaceNode._children || parentDom._children;
+	let isHydrating = false;
+	if (replaceNode === true) {
+		isHydrating = true;
+		replaceNode = undefined;
+	}
 	vnode = createElement(Fragment, null, [vnode]);
 
 	let mounts = [];
@@ -32,6 +37,7 @@ export function render(vnode, parentDom, replaceNode) {
 		mounts,
 		false,
 		replaceNode || EMPTY_OBJ,
+		isHydrating,
 	);
 	commitRoot(mounts, vnode);
 }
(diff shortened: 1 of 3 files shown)


src/render.js
```diff
@@ -13,8 +13,13 @@ import options from './options';
  */
 export function render(vnode, parentDom, replaceNode) {
 	if (options._root) options._root(vnode, parentDom);
-	let oldVNode = replaceNode && replaceNode._children || parentDom._children;
 
+	let oldVNode = replaceNode && replaceNode._children || parentDom._children;
+	let isHydrating = false;
+	if (replaceNode === true) {
```
marvinhagemeister: Random though (not sure if it will work): We could try using `null` for `replaceVNode` so that it stays in the same type for v8. That way we wouldn't need to reset it to `undefined` two lines later.

JoviDeCroock: Good idea, implemented!

cristianbote:
Agh, this adds so much size... 😞
`+30B` https://travis-ci.org/preactjs/preact/builds/562138925#L1733.
Wondering if there's something going on with the diff or just this is the reality? 🤔

JoviDeCroock:
@cristianbote Last time I checked it was 22B

src/render.js
```diff
@@ -13,8 +13,12 @@ import options from './options';
  */
 export function render(vnode, parentDom, replaceNode) {
 	if (options._root) options._root(vnode, parentDom);
-	let oldVNode = replaceNode && replaceNode._children || parentDom._children;
 
+	let oldVNode = replaceNode && replaceNode._children || parentDom._children;
+	let isHydrating = false;
```
JoviDeCroock: Maybe leaving this out will save some bytes as in let `let isHydrating = replaceNode === null`

JoviDeCroock: This cut 7B

src/render.js
```diff
@@ -44,5 +46,5 @@ export function render(vnode, parentDom, replaceNode) {
  */
 export function hydrate(vnode, parentDom) {
 	parentDom._children = null;
-	render(vnode, parentDom);
+	render(vnode, parentDom, null);
```
developit: would it be possible to use Boolean `true` here? Semantically it might make a bit more sense, but also it avoids someone passing `null` values for replaceNode without realizing it - this is actually pretty likely, since DOM getters always return `null` for empty values (eg `foo.firstElementChild`).

PR #1787, opened 2019-07-20
Add p5.js example to demos

Gherciu:
#### Description
A small example of integration p5.js library with preact, to create different stuff with canvas and the p5.js library.


cristianbote:
Hi @Gherciu, thanks for opening this PR! We really appreciate it. The `demo` folder is intended as a sandbox more than anything, useful to debug implementation details. With that said, we're using that just for that purpose and not exemplifying integrations of other libraries.

Thanks again for opening this PR, but we're gonna close it, since we're expecting to keep that as a sandbox for debugging internals.

PR #1797, opened 2019-07-24
TS: improve typing of "style" attribute using CSSType

mnkhouri:
This commit adds the [CSSType library](https://www.npmjs.com/package/csstype), which provides TypeScript and Flow definitions for CSS, generated by data from MDN. It provides autocompletion and type checking for CSS properties and values.

The library is used by [most other popular css-in-js frameworks](https://www.npmjs.com/browse/depended/csstype) for TS typechecking, including:
- [@types/react](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/2ccc2ff364f965e7407387f9045c8b7496d734a3/types/react/index.d.ts#L1656)
- [@material-ui/styles](https://github.com/mui-org/material-ui/blob/4d8206d3d44580345141405e5f78cd02fbb7ef99/packages/material-ui-styles/src/withStyles/withStyles.d.ts#L14)
- [@types/styled-components](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/4e23957ab6ef42ced4ac67048b33df9567474d9c/types/styled-components/index.d.ts#L25)
- glamorous
- csx
- @types/jss

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ coverage
 *.log
 package/
 preact-*.tgz
+jsx-csstype.d.ts
--- a/config/copy-csstype.js
+++ b/config/copy-csstype.js
@@ -1,0 +1,6 @@
+const fs = require('fs');
+const path = require('path');
+
+const src = path.join(__dirname, '..', 'node_modules', 'csstype', 'index.d.ts');
+const dest = path.join(__dirname, '..', 'src', 'jsx-csstype.d.ts');
+fs.copyFileSync(src, dest);
--- a/package-lock.json
+++ b/package-lock.json
@@ -3547,6 +3547,12 @@
 				"css-tree": "1.0.0-alpha.37"
 			}
 		},
+		"csstype": {
+			"version": "2.6.13",
+			"resolved": "https://registry.npmjs.org/csstype/-/csstype-2.6.13.tgz",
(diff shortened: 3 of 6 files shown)


nikolay-borzov:
Why not use `CSSStyleDeclaration` available in `node_modules\typescript\lib\lib.dom.d.ts`?

```
style?: string | Partial<CSSStyleDeclaration>;
```

nitsky:
@nikolay-borzov `CSSStyleDeclaration` is not the same as what react/preact expect for the `style` prop. `CSSStyleDeclaration` is what is returned by `someElement.style`, so it has a number of properties on it such as `getPropertyPriority(property: string): string` that are not suitable for the `style` prop.

The types for react also use the `csstype` package: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/7caf9e949ed3779010d39927ee3f97f71708db7c/types/react/index.d.ts#L1754

creepinson:
Is there an update on this? I can't type my stylesheets if they are declared with variables and its really annoying.

marvinhagemeister:
Put it up as a discussion point in the team. Personally I'm in favour of merging this. Main reason we're a bit hesitant is that we got a lot of backlash from users for adding a type only dependency in the past.

sventschui:
As discussed on Slack I think I'd be in favour of copy-ing the types during postinstall and then ship them as part of the preact package. Is there any downside to this?

marvinhagemeister (approved):
Added a commit that inlines csstype at build time. Didn't use `postinstall` because that hook runs for consumers too, but we just want to do the inlining upon publishing.

marvinhagemeister:
@mnkhouri Thank you so much for your PR! Your changes will be included in the next Preact release :raised_hands:

PR #1802, opened 2019-07-25
Use internal reference for hydration flag

developit:
This avoids users accidentally passing `null` as the third argument to render:

```js
const root = document.createElement('div');
document.body.appendChild(root);

render(<App />, root, root.firstElementChild);
                      //   ^^^^^^^^^^^^^^^^^ this will be `null`
```

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -3,6 +3,8 @@ import { commitRoot, diff } from './diff/index';
 import { createElement, Fragment } from './create-element';
 import options from './options';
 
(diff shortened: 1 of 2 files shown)


src/render.js
```diff
@@ -45,6 +47,5 @@ export function render(vnode, parentDom, replaceNode) {
  * update
  */
 export function hydrate(vnode, parentDom) {
-	parentDom._children = null;
```
JoviDeCroock: I think to make the tests work you'll have to reset this.

Issue #1798, opened 2019-07-25
[10.0.0-rc.0] preact/compat and onTransitionEnd


fjorgemota:
Hey!

I'm using Preact X (version 10.0.0-rc.0) on my project, with `preact/compat` and [rmwc](https://rmwc.io)...And I found a problem: Apparently, `onTransitionEnd` event doesn't work, because Preact only checks if the lower-case attribute exists on the DOM element to define if it should use the lower-case event name (spoiler: it should!) or not.

The problem is...I found #1589 issue already (which is closed, but IMHO it shouldn't, but I will explain my point soon) AND #1590 which was a pretty smart way to detect which event to use. Okay, I found #1631 too which reverts all of that, and the final suggestion is to just use the lowercase format to be able to work around this limitation.

That's okay, in my opinion...for pure preact projects. But, for `preact/compat`, which should work fine with other React libraries (like RMWC, mentioned above), there's no way to fix that issue, because...well, unless I use `patch-package` to replace `onTransitionEnd` with `ontransitionend` on the RMWC project (which is a pretty bad "solution" IMHO), Preact will happily add the `TransitionEnd` event to the DOM element, which....well, will never be triggered, we know that, because that's not the correct name of the event.

With that said...I think there's two ways to fix that problem on Preact itself:

1) Try to cache the specific attributes of the global variable on a Preact's specific variable so it's faster to check, fixing the performance penalty of checking of the lookup (would add a few more bytes and probably would mean a slower first lookup);
2) Fix that bug at least on `preact/compat` itself, for compatibility with other libraries;

IMHO, the first way would be far better...because if the problem is the lookup for the attribute then caching that data somewhere would at least fix that problem. But the two possibilities are valid, in my opinion.

What do you think? 

Edit: Fix reference to the correct PR with the fix.

PR #1804, opened 2019-07-25
(fix) - transitions

JoviDeCroock:
Fixes: https://github.com/preactjs/preact/issues/1798

This PR should serve as a deciding point whether or not we want to support these cases to improve React compat.


fjorgemota:
Hey.

Thanks for the PR! :)

Even throught I mentioned specifically `onTransitionEnd` on the issue, I think there's other events that probably would be worked around too on `preact/compat`. #1589 with it's `onAnimationEnd` is one of these examples. 

So...maybe we should use a generic code like the one found on #1590 to detect which events to patch for, and maybe verifying if the attribute doesn't exist on DOM element first?

I don't know..these are just ideas. But anyway thanks for the PR. I hope I can help on the discussion. =)

JoviDeCroock:
@fjorgemota this PR is intended as a starting point for discussing if we want to support these events. The solution of #1589 can't be used since this breaks testing envs. If we would agree on this I would suggest looking for the postfix start/end and using that to determine a toLowerCase()

fjorgemota:
@JoviDeCroock 	oh. Ok then. Fair enough. I'm waiting for the reviews and further discussion, then. Thanks! :)

developit:
A thought: for DOM nodes (`typeof vnode.type == 'string'`), loop over props and lowercase any keys matching `/^on(ani|tra)/`.

JoviDeCroock:
@developit yes, so we want to support these?

compat/src/index.js
```diff
@@ -174,6 +174,13 @@ function createElement(...args) {
 
 	let type = vnode.type, props = vnode.props;
 	if (typeof type!='function') {
+		Object.keys(props).forEach(key => {
+			if (/^on(Ani|Tra)/.test(key)) {
```
marvinhagemeister: Quick question: Is this true for all `onAnimation*` and all `onTransition*` events?

JoviDeCroock: Hmm only for start and end I presume we could be taking capture here too

cristianbote: Hey @JoviDeCroock I was looking over our `compat` code and there's a `applyEventNormalization` function. Should this be part of that function? I'm thinking it'll be easier to tackle updates on that code in the future since all the compat event handlers would be in one place, maybe?

PR #1808, opened 2019-07-26
Exclude devtools tests from coverage

marvinhagemeister:


debug/test/browser/devtools.test.js
```diff
@@ -137,790 +137,797 @@ function getRoot(element) {
 	return element._children;
 }
 
-describe('devtools', () => {
+let ua = window.navigator.userAgent;
+if (
+	ua.indexOf('Edge') === -1 &&
+	ua.indexOf('Chrome') > -1 &&
+	ua.indexOf('Firefox') > -1
+) {
```
marvinhagemeister: This `if`-statement is the only change here in this file. Everything else is just indentation. I tried using the `this.skip`-API like the mocha docs recommend, but it will still run some tests contrary to what the docs say.

marvinhagemeister:
Hm I'm not sure if there is a way to tell coveralls to pull the coverage from one browser only.

debug/test/browser/devtools.test.js
ForsakenHarmony: could you like exit or return or something for early exit instead of indenting everything?

developit: I believe this works:

```js
const desc = /Chrome|Firefox/.test(navigator.userAgent) && !/Edge/.test(navigator.userAgent) ? describe : describe.skip;

desc('..', () => {
  ..
```

developit:
@marvinhagemeister we could probably have the last browser always be Chrome?

marvinhagemeister:
@developit I'm not sure how can ensure that it's the last one. Once Saucelabs takes over the order seems to be in their hand.

debug/test/browser/devtools.test.js
marvinhagemeister: That's clever! I like it :+1:

marvinhagemeister: @ForsakenHarmony tried the early return, and it works too. I think I'll go with @developit 's approach though, it's more clear and prints that the `devtools` test have been skipped :slightly_smiling_face:

marvinhagemeister: @ForsakenHarmony I was wrong! It doesn't make any difference :joy:

debug/test/browser/devtools.test.js
```diff
@@ -137,7 +137,11 @@ function getRoot(element) {
 	return element._children;
 }
 
-describe('devtools', () => {
+// Devtools are only supported in Chrome and Firefox anyway
+const supported = /Chrome|Firefox/.test(navigator.userAgent) && !/Edge/.test(navigator.userAgent);
```
cristianbote: Not sure how picky should we be here 😄 but this returned `true` for a mobile userAgent. We could iterate over it and maybe combine them rules? 
```js
/(Chrome)|(Firefox)[^(Edge)]/i.test(navigator.userAgent)
```

marvinhagemeister:
Tried a few other things, but I don't see any other way than to disable coverage reporting for the devtools code...

PR #1810, opened 2019-07-27
Add VNode.constructor to src/internal.d.ts

38elements:
This adds `VNode.constructor` to type definition.


src/internal.d.ts
```diff
@@ -52,6 +52,7 @@ export interface VNode<P = {}> extends preact.VNode<P> {
 	 */
 	_lastDomChild: PreactElement | Text | null;
 	_component: Component | null;
+	constructor: undefined;
```
38elements: https://github.com/preactjs/preact/blob/b4e059ad00ff2d4dc71b0d69d4a148a204da6d5f/src/create-element.js#L66

Issue #1811, opened 2019-07-27
JSX type definition in Preact 8 is moved into preact namespace or JSX type definition is added to global for compatibility 


38elements:
The namespace of JSX was changed.
This is breaking change.
I think that it is better to specify in the release notes.

https://github.com/preactjs/preact/blob/b4e059ad00ff2d4dc71b0d69d4a148a204da6d5f/src/index.d.ts#L6-L8

https://github.com/preactjs/preact/blob/3aaa578270c0c6cd1fb7f5cf992737eec2934433/src/preact.d.ts#L168-L170

To solve the following issue, I think that it is better to add JSX type definition to global.

https://github.com/developit/unistore/pull/158

I think it is better that JSX type definition in global  be deprecated.

ForsakenHarmony:
> To solve the following issue, I think that it is better to add JSX type definition to global.

> I think it is better that JSX type definition in global be deprecated.

So uhh which one are you asking for?

global was removed because it conflicted with react

38elements:
> So uhh which one are you asking for?

JSX is added to global.
And JSX exists global until Preact 8 support is terminated.

38elements:
I think it is better that JSX type definition in Preact 8 is moved into preact namespace.

ForsakenHarmony:
Release in v8.5.0

aMarCruz:
Great, now prting my 8.x preact app to 10 breaks my toolchain because "JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists".

![image](https://user-images.githubusercontent.com/6636980/63655883-2c3e3c80-c753-11e9-9b99-13b848cd8e97.png)

Using @types/react fixes this issue but breaks other things, I'm not using preact/compat.

38elements:
In 8.4.x, JSX exists in global namespace.
In 8.5+, JSX exists in global namespace and preact namespace.
In 10, JSX exists in preact namespace.

PR #1818, opened 2019-07-30
JSX type definition in Preact 8 is moved into preact namespace

38elements:
This is related to #1811.

### DIFF
--- a/src/preact.d.ts
+++ b/src/preact.d.ts
@@ -936,3 +936,777 @@ type Defaultize<Props, Defaults> =
 			// Include the remaining properties from Props
 			& Pick<Props, Exclude<keyof Props, keyof Defaults>>
 		: never;
+
+declare global {
+	namespace JSX {
+		interface Element extends preact.VNode<any> {
+		}
+
+		interface ElementClass extends preact.Component<any, any> {
+		}
+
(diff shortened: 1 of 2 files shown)


src/preact.d.ts
```diff
@@ -154,18 +154,7 @@ declare namespace preact {
...
-type Defaultize<Props, Defaults> =
-	// Distribute over unions
-	Props extends any
-		? 	// Make any properties included in Default optional
-			& Partial<Pick<Props, Extract<keyof Props, keyof Defaults>>>
-			// Include the remaining properties from Props
-			& Pick<Props, Exclude<keyof Props, keyof Defaults>>
-		: never;
 
-declare global {
```
ForsakenHarmony: removing it would probably be a breaking change, I guess we'll need both and the only way to do it might be to copy paste

38elements:
@ForsakenHarmony 
Thank you for reviewing.
I changed this.

ForsakenHarmony:
cc @pmkroeker

pmkroeker:
Looks good to me!

ForsakenHarmony:
Released in v8.5.0

38elements:
@ForsakenHarmony @pmkroeker 
Thanks.

PR #1821, opened 2019-07-31
Remove a duplicate test in render tests

OrKoN:
It looks like `should render nothing node given null` and `should not render null` are testing the same thing. Perhaps it's worth removing one of them. It might be that `scratch.childNodes` and `scratch.innerHTML` checks are not equivalent though, but they look the same to me.

### DIFF
--- a/test/browser/render.test.js
+++ b/test/browser/render.test.js
@@ -44,12 +44,6 @@ describe('render()', () => {
 		logCall(Element.prototype, 'remove');
 	});
 
-	it('should render nothing node given null', () => {
-		render(null, scratch);
-		let c = scratch.childNodes;
-		expect(c).to.have.length(0);
-	});
-
 	it('should render an empty text node given an empty string', () => {
 		render('', scratch);
 		let c = scratch.childNodes;


cristianbote:
Hey thanks for opening this PR. Well, we could argue that `render(null, scratch)` should be checked that it doesn't render anything and also that doesn't create empty children. Take this example:

```js
const div = document.createElement("div");
div.appendChild(document.createTextNode("")); // empty text node

div.innerHTML === ""; // true ✅ 
div.childNodes.length; // 1 <-- oops 🔴  
```
So, we still need that test case in order to properly handle the case were children could be created with empty text.

ForsakenHarmony:
should we merge the tests then?

cristianbote:
@ForsakenHarmony indeed! 😄 we could totally do that. @OrKoN are you ok with that? Would you like to (re)open the PR with both of the assertion in one case?

OrKoN:
@cristianbote yes, it might be that also the 4 tests after `should not render null` might need to check for child nodes. What do you think?

P.S. I am on vacation now, will take a look in a week, greets from Italy!

ForsakenHarmony:
Yeah enjoy your vacation first 🎉

PR #1822, opened 2019-07-31
fixes xvfb

prateekbh:
- uses new travis service


prateekbh:
FWIW this is the 3rd project in last one hour to which i've sent the same fix 😆

PR #1831, opened 2019-08-03
Fix last element always losing focus

### LINKED ISSUE #1829: [10.0.0-rc.1] Component with conditional render result forces remount of siblings
Codepen: <https://codepen.io/alebastr/pen/ZNgWPZ?editors=0010#0>

Certain wrapping component causes preact to remount all subsequent siblings without calling any lifecycle methods. However, when I inline the same logic (`{this.props.open && <Modal {...this.props} />}`), it works correctly. 
```javascript
class Wrapper extends React.Component {
    render() {
        if (this.props.open) {
            return <Modal {...this.props} />;
        }
        return null;
    }
}
```
From the codepen you can see that the button before the wrapper component preserves focus, but the button after does not. This happens because after unmounting Wrapper and returning focus, preact temporarily removes second button from DOM.

The issue was introduced between 10.0.0-beta.0 and 10.0.0-beta.1.

marvinhagemeister:
Fixes #1829 .
Adds `+10 B`.

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -112,7 +112,12 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 					// The values only have the same type when `null`.
 
 					outer: if (oldDom==null || oldDom.parentNode!==parentDom) {
-						parentDom.appendChild(newDom);
+						// Skip appendind if the last element is the same as the new dom
+						// node. This is necessary to prevent the last element losing focus
+						// if it had focus before the diff.
+						if (parentDom.lastChild!==newDom) {
+							parentDom.appendChild(newDom);
+						}
 					}
 					else {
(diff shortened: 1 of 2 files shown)


alebastr:
This actually works only for the specific element order from my initial reproduction script.
The issue remains if there's more than one node after Wrapper component, i.e. <https://codepen.io/alebastr/pen/OKxjgZ>

src/diff/children.js
```diff
@@ -112,7 +112,12 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 					// The values only have the same type when `null`.
 
 					outer: if (oldDom==null || oldDom.parentNode!==parentDom) {
```
JoviDeCroock: Can't we add the check here?

developit:
@alebastr has a good point. I'm thinking about this one.

marvinhagemeister:
yeah this is probably not the right fix for that.

PR #1835, opened 2019-08-04
Avoid accessing fn.prototype where possible

developit:
image

Hoping this has minimal byte size impact. Perhaps there are other places where we do this?


jonathantneal:
Is there a place I can see the before and after perf change? The image only shows the before, correct?

PR #1848, opened 2019-08-10
(fix) - offsets

JoviDeCroock:
This is an initial draft, I just need some more eyes on this. I feel like we can get this sorted but I'm stuck on the insert on position 0 case.


src/diff/children.js
```diff
@@ -61,8 +84,12 @@ export function diffChildren(parentDom, newParentVNode, oldParentVNode, context,
 			// We use `undefined`, as `null` is reserved for empty placeholders
 			// (holes).
 			oldVNode = oldChildren[i];
-
-			if (oldVNode===null || (oldVNode && childVNode.key == oldVNode.key && childVNode.type === oldVNode.type)) {
+			// When we are dealing with an inserted child we should ensure that this is diffed as a hole
+			// and not versus a previously active vnode.
+			if (insertions.includes(childVNode)) {
```
developit:
```suggestion
			if (~insertions.indexOf(childVNode)) {
```

PR #1853, opened 2019-08-11
Fix undefined styles not being cleared

### LINKED ISSUE #1850: Material UI example not working as in react
Hi, I just tried using the Material UI demo "Hide app bar" with preact 10 rc1 (using preact/compat in an alias configuration).

Using React, it can be observed, that the App Bar moves out when scrolling down (passing 100px) and in again when scrolling in again (above 100px or when last scroll direction was up).

Using preact, after hiding the App Bar, it never comes back. Assuming the goal of compatibility with React, I think it should work exactly the same.

Documentation: https://material-ui.com/components/app-bar/#hide-app-bar
Code Sandbox using React: https://codesandbox.io/s/h8ls7

Reproduction Code Sandbox using preact/compat: https://codesandbox.io/s/material-demo-7v6qq ... well, that's not working yet and I'm running out of time for today.
I don't know how to configure the alias (tried in package config as guessed from what I saw in https://github.com/codesandbox/codesandbox-client/pull/853). Changing the imports in the two files isn't enough because material-ui imports react as well. Can somebody help?

marvinhagemeister:
Fixes #1850 .
Adds `+2 B` 🎉

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -31,7 +31,7 @@ function setStyle(style, key, value) {
 		style.setProperty(key, value);
 	}
 	else {
-		style[key] = typeof value==='number' && IS_NON_DIMENSIONAL.test(key)===false ? value+'px' : value;
+		style[key] = typeof value==='number' && IS_NON_DIMENSIONAL.test(key)===false ? value+'px' : value || null;
 	}
 }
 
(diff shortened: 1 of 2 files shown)


cloudkite:
I believe this change breaks setting unitless styles values to 0 for example `style={{opacity: 0}}` since 0 is falsy therefore will be set to null

PR #1854, opened 2019-08-11
Implement support for async `act` callbacks and nested calls to `act`

robertknight:
This PR implements support for passing async callbacks to `act` and nesting calls to `act`, matching [React 16.9.0](https://reactjs.org/blog/2019/08/08/react-v16.9.0.html#async-act-for-testing). This is useful in cases where triggering an effect or state change involves async steps, such as waiting for a `fetch` call to resolve.

Following [React's behavior](https://github.com/facebook/react/issues/15472), `act` returns a Promise which resolves immediately if the callback is synchronous or after the callback resolves if it is asynchronous. Awaiting the result of `act` before making assertions is optional if the callback is sync, but required if the callback is async, as effects/updates are not flushed until the callback's result has resolved in that case.

When calls to `act` are nested, effects and updates are only flushed when the outer `act` call returns. This also follows React's behavior.

One feature of React's `act` that is not implemented yet is warning if you pass an async callback to `act` but forget to await the result.

- The first commit adds a Babel transform for await/async using the same low-footprint transform that is used in microbundle, for convenience writing the tests for the later commits ([Slack discussion](https://preact.slack.com/archives/G9T60GUQ0/p1565514000140700))
- The second commit adds support for async callbacks to `act`
- The third commit adds support for nested calls to `act`


test-utils/src/index.js
```diff
@@ -10,11 +10,37 @@ export function setupRerender() {
...
+ *
+ * Returns a Promise which resolves "immediately" if the callback is
+ * synchronous or when the callback's result resolves if it is asynchronous.
+ *
+ * @param {() => void|Promise<void>} cb The function under test. This may be sync or async.
+ * @return {Promise<void>}
  */
 export function act(cb) {
+	++actDepth;
+	if (actDepth > 1) {
```
marvinhagemeister: Super super tiny nit: We can combine this line and the above one:

```js
if (++act.Depth > 1) { ... }
```

test-utils/test/shared/act.test.js
```diff
@@ -191,4 +191,144 @@ describe('act', () => {
...
+				return <button onClick={increment}>{count}</button>;
+			}
+
+			render(<Button />, scratch);
+			const button = scratch.querySelector('button');
+			expect(button.textContent).to.equal('0');
+
+			act(() => {
+				act(() => {
+					button.dispatchEvent(new Event('click'));
```
marvinhagemeister: This will fail in IE11 because it doesn't support `new Event()`. I have a PR ready which introduces a helper method for that: https://github.com/preactjs/preact/pull/1856/files#diff-d429c5e5b8692f8b67a9d38592332a1eR8

test-utils/test/shared/act.test.js
```diff
@@ -191,4 +191,144 @@ describe('act', () => {
...
+				return <button onClick={forceUpdate}>test</button>;
+			}
+
+			act(() => {
+				render(<Widget />, scratch);
+				const button = scratch.querySelector('button');
+				expect(counter).to.equal(0);
+
+				act(() => {
+					button.dispatchEvent(new Event('click'));
```
marvinhagemeister: Same IE11 issue as below

test-utils/test/shared/act.test.js
robertknight: Thanks for catching this. I've updated the code to use the helper.

robertknight:
Feedback / IE 11 issues addressed. Thanks for the super-swift review :)

PR #1857, opened 2019-08-12
Warn when calling setState in a constructor

### LINKED ISSUE #1840: setState() callback not invoked (since 8.3.0)
I'm upgrading a project from `preact^8.2` to `preact^10.0.0-rc1` and ran into a problem.

From my component's constructor, I'm invoking a private method that calls `setState()` after initially loading some search results. This `setState()` call needs to update a private field after the update has finished, so I'm using the `setState()` callback.

Up to and including release `8.2.9`, this worked fine.

Starting with release `8.3.0`, the callback is no longer being called.

Current latest release `10.0.0-rc.1` is also affected.

Here's a repro:

https://codesandbox.io/s/silly-bhabha-x2roy

You should see `wut?` in the console, and the text should change to `loaded`, but never does.

Granted, calls to `setState()` don't work at all in React, so it's not an incompatibility with React - but it is a breaking change from previous versions of Preact, and it was tagged as a feature release, so in semver terms, this is either versioned incorrectly, or it's simply a regression/bug.

If this is difficult to fix for some reason, I'd suggest a proper error message, e.g. "setState() cannot be called from the constructor". (ReactDOM gives an error message along those lines. I know you guys are counting bytes - but isolating a silent error like this can take a while.)

Known work-around: defer the `setState()` call using e.g.:

    setTimeout(() => this.setState({...}), 0);

Could be related to #556 or #1741. (If those are related, and if this issue isn't going to be addressed because of size concerns, it might make sense to close all of these issues together - perhaps adding an error message or documenting this pitfall.)

marvinhagemeister:
This PR adds a warning message in `debug` when `this.setState` is called inside the class `constructor`.

Fixes #1840 .

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -1,6 +1,6 @@
 import { checkPropTypes } from './check-props';
 import { getDisplayName } from './devtools/custom';
-import { options } from 'preact';
+import { options, Component } from 'preact';
 import { ELEMENT_NODE, DOCUMENT_NODE, DOCUMENT_FRAGMENT_NODE } from './constants';
 
 function getClosestDomNodeParent(parent) {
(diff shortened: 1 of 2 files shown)


marvinhagemeister:
Looks like coveralls is down... Will try again later.

PR #1858, opened 2019-08-12
Throw error when passing plain objects as children

### LINKED ISSUE #1677: Warn or throw an error in debug when an object is passed as a child
The following code triggers an error in React but not in Preact:

```js
function App() {
  return <div>{{}}</div>
}
```

Since it likely indicates a bug in the app's code, and can trigger confusing error messages later if other code tries to treat the object as an element, it would be useful if `preact/debug` were to warn or throw an error if this happens.

marvinhagemeister:
Fixes #1677 .

### DIFF
--- a/debug/mangle.json
+++ b/debug/mangle.json
@@ -17,6 +17,7 @@
   "props": {
     "cname": 6,
     "props": {
+      "$_depth": "__b",
       "$_dirty": "__d",
       "$_prevState": "__u",
       "$_nextState": "__s",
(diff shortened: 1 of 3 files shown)


developit:
LGTM.

I was going to comment that a similar vnode hook would be nice for getting the right error callsite, but we can't actually know that a vnode's children will be forwarded into the renderer unmodified (eg, this would be valid: `const F = ({ children }) => createElement('div', children);render(<F>{{id:'foo'}}</F>`).

marvinhagemeister:
@developit yup, that's the reason behind moving it to `options.diffed`. We only need to validate the actual rendered children.

Issue #1813, opened 2019-07-28
[10.0.0-rc.0] Suspense's fallback not being unmounted when it should


fjorgemota:
Hey.

I found a pretty weird issue on Suspense's fallback implementation of Preact X. Here's a [small codesandbox](https://codesandbox.io/s/preact-suspense-lazy-issue-euz15) with the issue (please navigate using the links on the demo to see the issue ocurring).

It's pretty simple, actually: Basically, when using a Component as a fallback, the content rendered by that Component isn't unmounted **after** Switch rendered the new page of the router.

Now, curiously, the issue is basically resolved when I use the content rendered by the Component directly as the fallback, OR when I use `() => <Loading />` with the fallback. I assume there's some flag on the component object that is making Preact to ignore it instead of just unmounting it, but as I don't know many of the Preact internals I'll post this issue here and hope for a solution.

Hope this helps to discover the cause of the issue. And keep the great work! :+1:

fjorgemota:
Okay, apparently the suggested workaround (using `() => <Loading />` as fallback)  bugs with `preact/debug`, causing an exception "TypeError: Invalid value used as weak map key" to be throw. At least that's a very specific `preact/debug` issue.

:thinking:

marvinhagemeister:
@fjorgemota `preact/debug` errors because `() => <Loading />` is not a valid component instance. You're passing the raw function instead.

```jsx
// This is invalid
<div>{() => <Loading />}</div>
```

To pass Components around they need to be wrapped in `h/createElement`:

```jsx
<div>{h(() => <Loading />)}</div>

// ... or via native jsx
const MyComponent = () => <Loading />;
<div><MyComponent /></div>
```

fjorgemota:
Oh, okay, this makes sense.

Unfortunately, with that behavior the issue originally reported just...happens again. But okay, I'll let the issue open anyway, and I'll ignore the `preact/debug` issue for now.

Thanks for the reply anyway =)

sventschui:
Sorry for the late reply. Was quite busy the past days. I guess I found the cause and a a fix (see #1859). Unfortunately I did not yet manage to reproduce this issue outside the codesandbox

aduh95:
[...]

PR #1859, opened 2019-08-12
Allow fallback of Suspense to be mounted multiple times

sventschui:
I failed to reproduce #1813 in our demo app but with some debugging I came across the fact that the `_dom` property was missing on the fallback vnode. The reason for that seems to be the `coerceToVNode` in render since it clones the vnode. What I do not get is why I can not reproduce this in the demo app.

This PR adds 14B to preact and compat each.

Anybody has any ideas how to repro in demo?

TODO:
- [ ] Tests
- [ ] Revert demo


sventschui:
Just as a short update. I found that the duplicate dom of some components can be prevented by adding a guard to _childDidSuspend that prevents tracking promises multiple times. It does not solve the stale fallback dom to still be present. 

Durinh these investigations I found further issues with dom nodes that are mounted as siblings to a suspending component. As a result of these issues I started investigating two-phase rendering.

TL;DR I looked into the issue and think Suspense is not possible with the current implementation of preact. The two phase render will take a lot of time to be built and will land post 10.0.0 release

sventschui:
Closing this one as we need a two-phase diff algo to properly solve these issues

PR #1860, opened 2019-08-12
Support adding refs to memo'ed components

### LINKED ISSUE #1845: 10.0.0-rc1 difficult to track down crash
Hi there, any advice on how I can track down a crash on code that works fine with react but breaks in preact - is it possible to force webpack/babel to use the source dist of the module rather than the dist? The error is `Uncaught (in promise) TypeError: Cannot read property 'call' of undefined  at y.r [as shouldComponentUpdate...` but I cannot figure out the call process properly from the minified code to give you a decent bug report?

sventschui:
Support adding refs to memo'ed components. If I am right this should fix #1845

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -346,7 +346,7 @@ function memo(c, comparer) {
 	function shouldUpdate(nextProps) {
 		let ref = this.props.ref;
 		let updateRef = ref==nextProps.ref;
-		if (!updateRef) {
+		if (!updateRef && ref) {
 			ref.call ? ref(null) : (ref.current = null);
 		}
 		return (!comparer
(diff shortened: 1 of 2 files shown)


sventschui:
The stack of #1845 (see below) says that we try to access the `call` property on `undefined` in a function in compat that is set to a property called `shouldComponentUpdate`.

Not only yields a find of `.call` only one result but it is also in a function that is stored on a prop called `shouldComponentUpdate`. :)

```
Uncaught (in promise) TypeError: Cannot read property 'call' of undefined
    at d.t [as shouldComponentUpdate] (compat.module.js:49)
```

marvinhagemeister:
So good 👍👍

PR #1861, opened 2019-08-12
Add isValidElement to core

marvinhagemeister:
This came up on slack while we were looking at ways to remove the `isValidElement` import in [wouter](https://github.com/molefrog/wouter/blob/master/preact/react-deps.js).

Adds `+14 B`.


marvinhagemeister:
cc @kristoferbaxter

developit:
build failure is coveralls outage

marvinhagemeister:
@andrewiggins @developit changed it back to `isValidElement` as suggested in slack :+1:

PR #1866, opened 2019-08-13
Disable coveralls until their outage is resolved

marvinhagemeister:
This PR disables reporting the collected code coverage stats to `coveralls`. They've been experiencing various outages since more than a day ago and it will probably at least another day until their back up again: https://status.coveralls.io . This unblocks merging of PRs as the CI currently always fails as soon as the report is sent to coveralls.


cristianbote:
Awesome! 💯

PR #1869, opened 2019-08-15
Restructure devtools code in preparation for V4

marvinhagemeister:
This PR prepares the current devtools code for V4. I decided to split #1549 to make reviews easier. The changes here are only cosmetic and introduce no new features. They mainly shuffle the code around so that we can easily switch between multiple devtool versions.

The goal is that we can smooth out the transition period. Our code will (with #1549 ) be able to pick the right adapter depending on the devtools extension the user has installed.

### DIFF
--- a/debug/src/devtools/index.js
+++ b/debug/src/devtools/index.js
@@ -1,45 +1,25 @@
 /* istanbul ignore file */
-import { options, Component, Fragment } from 'preact';
-import { Renderer } from './renderer';
+import { options, Component } from 'preact';
+import { now, catchErrors } from './util';
+import { createLegacyAdapter } from './legacy/connect';
 
-/**
- * Wrap function with generic error logging
- *
- * @param {*} fn
- * @returns
- */
-function catchErrors(fn) {
-	return function(arg) {
-		try {
-			return fn(arg);
-		}
-		catch (e) {
-			/* istanbul ignore next */
-			console.error('The react devtools encountered an error');
-			/* istanbul ignore next */
-			console.error(e); // eslint-disable-line no-console
-		}
-	};
-}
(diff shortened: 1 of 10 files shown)


debug/src/devtools/index.js
```diff
@@ -51,62 +31,24 @@ export function initDevTools() {
...
-
-			let roots = hook.getFiberRoots(rid);
-			root = helpers.handleCommitFiberRoot(root);
-			if (!roots.has(root)) roots.add(root);
-		});
-
-		onCommitUnmount = catchErrors(vnode => {
-			hook.onCommitFiberUnmount(rid, vnode);
-		});
+		let adapter = createLegacyAdapter(config, hook);
```
marvinhagemeister: With #1549 we'll add a switch statement here.

debug/test/browser/devtools.test.js
```diff
@@ -70,23 +74,23 @@ function createMockHook() {
...
+ * @param {VNode} vnode
+ * @param {Set<VNode>} seen
  * @returns {boolean}
  */
 function checkPreceding(vnode, seen) {
 	if (vnode==null) return true;
 
 	// If a leaf node is a Fragment and has no children it will be skipped
-	if (vnode.type===Fragment && vnode._children==0) return true;
+	if (vnode.type===Fragment && (vnode._children==null || vnode._children.length===0)) return true;
```
marvinhagemeister: Found an actual bug in the test assertions here: `vnode._children` is never of type `number` :tada:

marvinhagemeister:
Ok, I don't think this is necessary anymore seeing that the react time just published V4 as stable :tada:

PR #1871, opened 2019-08-15
Fix incorrect dom pointer with sCU

### LINKED ISSUE #1864: [10.0.0-rc1] Really weird issue with react-redux connect()
So I'm running slightly newer than -rc1 based on the git from ~18h ago. Running with a particular route under react-router-dom with a route like `/foo/:id` I have a relatively simple component which is connected to the store with react-redux and toggles between `null` and a sub-component view based on the value from the store - in test case below the string `foo`. When the id value changes a different prop (`song`) is passed to the subcomponent. The bug is that when the component is showing on a page and the store changes to make `render()` return null, the component hides correctly. But when I navigate to a new id (ie doesnt unmount, but props change) the component is rendered in its old form even though `render()` is still called and returns null.

If I switch away from using the `connect()` function to directly subscribing and changing component state based on the store value then the component behaves correctly. Switching to react 16.9.0 also does not display this issue.

Smallest case I can reproduce with in my app (ie inside the routing) is below. This is the one that doesn't work:

```
class _PresenterView extends Component {
    render() {
        const { song, is_active } = this.props;

        if( !is_active )
            return null;
    
        return 'foo';
    }
}
export const PresenterView = connect(state => ({ is_active: state.cast.active }))(_PresenterView);
```

And if we connect manually to the store then this works correctly:

```
export class PresenterView extends Component {
    componentDidMount() {
        store.subscribe(() => this.setState({ is_active: store.getState().cast.active }));
    }

    render() {
        const { is_active } = this.state;
        const { song } = this.props;

        if( !is_active )
            return null;
    
        return 'foo';
    }
}
```

marvinhagemeister:
Tasks:

- [x] Fix incorrect DOM pointers
- [x] Add unit tests
- [x] golf

Adds ~~`+13 B`~~ `+6 B`

cc @JoviDeCroock @jviide .

Fixes #1864

### DIFF
--- a/demo/index.js
+++ b/demo/index.js
@@ -20,6 +20,7 @@ import DevtoolsDemo from './devtools';
 import SuspenseDemo from './suspense';
 import Redux from './redux';
 import TextFields from './textFields';
+import ReduxBug from './reduxUpdate';
 
 let isBenchmark = /(\/spiral|\/pythagoras|[#&]bench)/g.test(window.location.href);
 if (!isBenchmark) {
@@ -77,6 +78,7 @@ class App extends Component {
 						<Link href="/redux" activeClassName="active">Redux</Link>
 						<Link href="/suspense" activeClassName="active">Suspense / lazy</Link>
 						<Link href="/textfields" activeClassName="active">Textfields</Link>
+						<Link href="/reduxBug/1" activeClassName="active">Redux Bug</Link>
 					</nav>
 				</header>
 				<main>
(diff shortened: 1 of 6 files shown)


mzealey:
This may have fixed an example case or two, but it is not fixing my real-world case. If I return 'foo' instead of `null` from the render function I get displayed `foofoo<old component>` so I'm not sure it is related to the render() exactly returning null but rather some difference in what it was returning.

mzealey:
So when the app first starts, it prints a 'foo' in place of the subcomponent correctly. Then I switch state to load the subcomponent and it works fine. When I turn off the subcomponent I get 'foofoo' and then when I switch to another route with a different id (or, I suspect do anything to change the props/state) it becomes 'foofoo<old subcomponent>'

marvinhagemeister:
Let's continue the discussion here: https://github.com/preactjs/preact/pull/1879

PR #1879, opened 2019-08-18
Fix more incorrect dom pointers with sCU

### LINKED ISSUE #1864: [10.0.0-rc1] Really weird issue with react-redux connect()

marvinhagemeister:
We missed a few cases where it would still keep old DOM around. One test is still failing.

Fixes #1864 :crossed_fingers:

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -90,7 +90,7 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 					c.state = c._nextState;
 					c._dirty = false;
 					c._vnode = newVNode;
-					newVNode._dom = oldDom!=null ? oldVNode._dom : null;
+					newVNode._dom = oldDom!=null ? oldDom!==oldVNode._dom ? oldDom : oldVNode._dom : null;
 					newVNode._children = oldVNode._children;
 					break outer;
 				}
--- a/test/browser/lifecycles/shouldComponentUpdate.test.js
+++ b/test/browser/lifecycles/shouldComponentUpdate.test.js
@@ -294,9 +294,9 @@ describe('Lifecycle methods', () => {
 		});
 
 		// issue #1864
-		it('should update dom pointers correctly', () => {
+		it('should update dom pointers correctly when returning null', () => {
 			function Child({ showMe, counter }) {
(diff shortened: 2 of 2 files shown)


mzealey:
The issue I reported at bottom of #1871 still manifests with these changes.

mzealey:
I've found another similar issue with a node (material UI button) suddenly reappearing in -rc2 and -rc3 but it is a bit complex to trigger even within my application. I think there are still some edge cases missing... In this case I have a `<div>` with some buttons under which are conditionally rendered. Sometimes the non-conditional ones switch to being first and the conditional ones go to the end of the div. And then sometimes they duplicate after this.

mzealey:
@marvinhagemeister  was the above issue with the patch addressed prior to closing - i don't see any additional work on this?

marvinhagemeister:
@mzealey yup, the same fix was included in #1888, which has landed in `master` already :tada:

PR #1883, opened 2019-08-19
Add test case for commit queues

### LINKED ISSUE #1736: `options._commit` is called on child although parent commit is in progress

marvinhagemeister:
This is necessary for the new devtools to work reliably in X. It's a hard requirement for #1549 .

Will some day fix #1736 .

### DIFF
--- a/test/browser/spec.test.js
+++ b/test/browser/spec.test.js
@@ -1,5 +1,5 @@
 import { setupRerender } from 'preact/test-utils';
-import { createElement as h, render, Component } from '../../src/index';
+import { createElement as h, render, Component, options } from '../../src/index';
 import { setupScratch, teardown } from '../_util/helpers';
 
 /** @jsx h */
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
Closing, commits are nicely queued in `master` already

PR #1887, opened 2019-08-19
TS: Add key to all jsx elements

marvinhagemeister:
This allows every `JSX` element to have an optional `key` prop (see the test snippet).

cc @pmkroeker

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -14,6 +14,10 @@ export namespace JSXInternal {
 		? Defaultize<Props, Defaults>
 		: Props;
 
+	interface IntrinsicAttributes {
+		key?: any;
+	}
+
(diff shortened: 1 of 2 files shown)


src/jsx.d.ts
```diff
@@ -14,6 +14,10 @@ export namespace JSXInternal {
 		? Defaultize<Props, Defaults>
 		: Props;
 
+	interface IntrinsicAttributes {
+		key?: any;
```
pmkroeker: Can a key be anything? I would have expected it to be `string|number`. Not sure the internal methods that use key.

marvinhagemeister: We don't have any restrictions in X because we just do a referentially equality check. So it can be indeed anything in our case.

PR #1888, opened 2019-08-19
(fix) - scu without children

JoviDeCroock:


JoviDeCroock:
Hey @mzealey if you feel up to it, feel free to test with this branch. I think all cases SHOULD be resolved. Thank you so much for the error reports and following this up

mzealey:
Yes this fixes my live use-case thank you!

src/component.js
```diff
@@ -111,7 +111,7 @@ export function getDomSibling(vnode, childIndex) {
 			// Since updateParentDomPointers keeps _dom pointer correct,
 			// we can rely on _dom to tell us if this subtree contains a
 			// rendered DOM node, and what the first rendered DOM node is
-			return sibling._dom;
+			return sibling._component ? sibling._component.base : sibling._dom;
```
andrewiggins: Hi! (Popping in from vacation 🌴 for a moment 😁. I have limited internet so I'll be away, so don't wait for me)

My thinking when playing with this method last was that `_dom` was the source of truth and that `c.base` was just a reflection of that value (perhaps one day it is a getter property). Is there a situation where `v._component.base != v._dom`? Perhaps we should fix the code that updates `c.base` and not `v._dom` (assuming my understanding is anywhere near close to meaningful 😬)?

src/diff/index.js
```diff
@@ -90,7 +90,7 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 					c.state = c._nextState;
 					c._dirty = false;
 					c._vnode = newVNode;
-					newVNode._dom = oldDom!=null ? oldVNode._dom : null;
+					c.base = newVNode._dom = oldDom!=null ? oldDom!==oldVNode._dom ? oldDom : oldVNode._dom : null;
```
andrewiggins: To double check my understanding, `c == newVNode._component` here?

test/browser/lifecycles/shouldComponentUpdate.test.js
```diff
@@ -294,9 +294,9 @@ describe('Lifecycle methods', () => {
 		});
 
 		// issue #1864
-		it('should update dom pointers correctly', () => {
+		it('should update dom pointers correctly when returning null', () => {
```
andrewiggins: Should child return null instead of '' per the test title?

src/diff/index.js
JoviDeCroock: Yes

src/component.js
JoviDeCroock: I'll try and debug it again, last time when debugging this I put `debugger;` everywhere the _dom got reset but this resulted in no triggers. Still, the `_dom` was set correctly when SCU returned false but then when it did have to update the `_dom` magically turned into `null`. This might be something with a ref somewhere but yes, did not find anything that made sense, I'll try and take a look at it as mentioned but can't promise anything.

JoviDeCroock: I've spent the last 2 hours debugging this again and it's really odd... Some mutation is goofing up this behavior since c.base is correct and even here https://github.com/preactjs/preact/pull/1888/files#diff-44070032789f10581d1b70740a2aaacfR93

The value is correct but somewhere it gets mutated, I even think it's mutated as a part newVNode and reflected on oldVNode due to us treating `.dom` as a mutable thing...

sventschui:
When a component bails out using `sCU===false` it's children are not diffed. This lead to outdated `_parent` pointers on the children what ultimately lead to `updateParentDomPointers` update the `_dom` pointer of a wrong/old parent vnode. Thanks to @JoviDeCroock for the support in finding this.

My fix in #229c135 fixes this. It adds some bytes to the bundle. Since I couldn't get my local build to run (terser issue) I could not yet do advanced bytes golfing, so feel free to try your luck reducing the footprint of this fix 🍀

sventschui:
```
for (tmp = 0; tmp < newVNode._children.length; tmp++) {
	if (newVNode._children[tmp]) newVNode._children[tmp]._parent = newVNode;
}
```
vs.
```
newVNode._children.some(child => {
	if (child) child._parent = newVNode;
});
```

where the later saves 11B but is slower during runtime. What to prefer?

developit:
@sventschui my suggestion would be that we use the `for` loop for now, but we should do a PR to convert not only this but other loops to other syntax at the same time - that will let us measure the effect without it being confounded by mixed syntax.

Also I wondered if this was smaller?

```js
const c = newVNode._children;
for (tmp = c.length; tmp--; ) {
    if (c[tmp]) c[tmp]._parent = newVNode;
}
```

Issue #1896, opened 2019-08-26
preact.render third argument causes weird behavior


laino:
https://codepen.io/laino/pen/XWrMyda?editors=0010 Edit: fixed link.

When using the third "replace node" argument, you won't get anything at all.

Comment either of those ``render`` and it will work.

hectorromo:
@laino The render function has changed and works now out of the box like you'd expect it to. Repeated renders will correctly render into the container instead of always appending to it.
https://github.com/marvinhagemeister/preact-faq/blob/master/migrate-to-x.md#remove-the-3rd-argument-to-render

laino:
@hectorromo The TypeScript types don't reflect that though and clearly the third argument does *something*.

JoviDeCroock:
It's the replaceNode argument that got added back in a later release. https://preactjs.com/guide/v10/api-reference/#render

laino:
@JoviDeCroock It doesn't work though. As shown in that codepen. It just removes everything.

laino:
Wait.... I linked the wrong codepen.

laino:
https://codepen.io/laino/pen/XWrMyda?editors=0010

Here's the right one.

JoviDeCroock:
This thirth argument does not work for the `/compat` only for the pure Preact

laino:
@JoviDeCroock It's not using compat? At least not unless there's some magic happening I'm not aware of.

JoviDeCroock:
I've made a small example to show it works since I can't edit on codepen: https://codesandbox.io/s/pensive-mahavira-jgh4h

laino:
@JoviDeCroock Your example doesn't use ReactX

laino:
@JoviDeCroock I fixed it to demonstrate the broken behavior:

laino:
You'd expect to see "3" there, but instead you get an empty element.

JoviDeCroock:
Well again, you are double rendering. I don't see the point in that let's maybe look at what use case that is used for?

laino:
It even happens when you render something different. It doesn't have to be the same element.

laino:
@JoviDeCroock There, now I'm trying to render something different and have it replace the result of my first render: https://codesandbox.io/s/cool-fog-zl4jj

laino:
Except.... you get nothing.

JoviDeCroock:
[...]

PR #1900, opened 2019-08-27
(fix) - replacing a node

JoviDeCroock:


marvinhagemeister:
Weird, GitHub seems to be bugging out for me. The `master` branch says that this PR is merged, but that's not the case in the UI here.

src/diff/index.js
```diff
@@ -212,6 +212,10 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
 
 	if (newVNode.type===null) {
 		if (oldProps !== newProps) {
+			if (excessDomChildren!=null) {
+				const index = excessDomChildren.indexOf(dom);
+				if (index > -1) excessDomChildren[index] = null;
```
cristianbote: We could, maybe, just directly do `excessDomChildren[excessDomChildren.indexOf(dom)] = null;`. Would this improve the size in any way? This will end-up with a property on the `excessDomChildren["-1"]` but maybe we could live with that?

cristianbote:
This fixes #1896

PR #1901, opened 2019-08-28
(fix) Add scu check for context.Consumer

cristianbote:
This makes sure that `Consumer` doesn't render twice for same context. This was discovered when trying to add support for `preact x` on `preact-redux`.

~~Adds `14B`~~ 😞 
Adds `13B` 😞

### DIFF
--- a/src/create-context.js
+++ b/src/create-context.js
@@ -15,7 +15,7 @@ export function createContext(defaultValue) {
 		Consumer(props, context) {
 			this.shouldComponentUpdate = function (_props, _state, _context) {
 				return _context !== context;
-			}
+			};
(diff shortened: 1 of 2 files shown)


src/create-context.js
```diff
@@ -13,6 +13,9 @@ export function createContext(defaultValue) {
 		_id: '__cC' + i++,
 		_defaultValue: defaultValue,
 		Consumer(props, context) {
+			this.shouldComponentUpdate = function (_props, _state, _context) {
+				return _context !== context;
```
marvinhagemeister: Closing over the closure arguments is a neat trick :+1: Love it :100:

PR #1906, opened 2019-09-01
Two phase render

sventschui:
This is a first attempt to two phase render that then hopefully lets us resolve the issues with Suspense.

There are a lot of issues and even more refactoring open but I wanted to share my first approach.

If you'd like to help out I suggest to take a look at the lifecycle and render tests first.

Things that will be "breaking" with this PR:
- `getSnapshotBeforeUpdate` is called later than before. This matches the behaviour of react more closely

My todo list:
- [ ] Try to get rid of the `_isNew` flag on the component instance (or at least replace the `isNew` in `renderTree` with references to that flag to golf some bytes)
- [ ] Revisit the handling of the `shouldComponentUpdate`
- [ ] Check whether `options._diff` is still placed correctly in renderTree
- [ ] Check whether `options.diffed` should be moved to the commit phase
- [ ] Check the use of the `base` property on the component instance as we do no longer set it during `renderTree`
- [ ] Check that the `setState` callbacks are called at the right time (currently in `renderTree`)
- [ ] Make sure `getSnapshotBeforeUpdate` is called with correct `oldProps`/`oldState`
- [ ] Make sure that `_parentDom` on the component instance is updated correctly in `commit` phase
- [ ] Investigate/double check added `coerceToVNode` in `toChildArray` call on line ~143 in `renderTree`
- [ ] Remove commented code
- [ ] Get rid of the `_used` flag on vnodes
- [ ] Reduce bundle size
- [ ] Do we need to bring back `c.base` (and it's tests)?
- [ ] Check call order of `componentWillUnmount`
- [ ] Check call order of `componentDidUpdate`
- [ ] Check call order expectation of `componentDidMount` in `should handle hoisted component vnodes without DOM`

```
723 tests completed
22 tests skipped
30 tests failed
```

Failing tests:

- [ ] Some of the perf tests (but these were failing on my machine previously too)
- [ ] Lifecycle method order (`getSnapshotBeforeUpdate` is called in wrong order)
- [x] Error handling of `getSnapshotBeforeUpdate` as this is now in commit phase
- [x] Some of the `keys` tests
- [x] Some of the `Fragments` tests (good old children insertion)
- [x] `should handle hoisted component vnodes without DOM`
- [x] `can access refs from within a layout effect callback`
- [ ] Suspense
- [ ] Portal
- [x] `should support useImperativeHandle`


JoviDeCroock:
Closing: exploratory two-phase-render draft (30 failing tests at the time). The v11 reconciler rewrite is the successor to this exploration.

Issue #1903, opened 2019-08-29
Missed ref


Crysp:
Preact `forwardRef` missing ref object after re-render

React ✅
https://codesandbox.io/s/react-forwardref-2wxkq

Preact ❌
https://codesandbox.io/s/preact-forwardref-pr1n1

It happens because of
https://github.com/preactjs/preact/blob/a16af189c7194916f540f666d524eb3fcadf9164/compat/src/index.js#L390

Why?

JoviDeCroock:
Hey @Crysp 

No it shouldn't be related to the linked code fragment. The code fragment dictates that we are removing `ref` from the props passed into the vnode. Because we keep our ref on vnode.ref instead, else we would be endlessly passing on ref if we wouldn't remove that from the props. Imagine a scenario where you do this.

```
const App = (props) => {
  return <X {...props} />
}
```

This would effectively make you pass on your ref further down, this feels like a bug with `useImperativeHandle`. I'll look into that tonight. Thanks for the issue report

Crysp:
@JoviDeCroock thx

PR #1909, opened 2019-09-02
(fix) - register the handles after rendering

JoviDeCroock:
Fixes: https://github.com/preactjs/preact/issues/1903

When we put a ref on a html-node this will effectively remove the handle, that's why we have to register these after our render occurs

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -30,8 +30,8 @@ options.diffed = vnode => {
 
 	const hooks = c.__hooks;
 	if (hooks) {
+		hooks._handles = bindHandles(hooks._handles);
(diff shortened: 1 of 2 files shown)


hooks/src/index.js
```diff
@@ -31,6 +31,7 @@ options.diffed = vnode => {
 	const hooks = c.__hooks;
 	if (hooks) {
 		hooks._pendingLayoutEffects = handleEffects(hooks._pendingLayoutEffects);
+		hooks._handles = handleHandles(hooks._handles);
```
JoviDeCroock: In retrospect I think I should flip these two

JoviDeCroock:
Test case like the updated issue report where state gets updated without ref bound on input.

PR #1904, opened 2019-08-30
Remove postinstall "ad"

ForsakenHarmony:
https://www.zdnet.com/article/npm-bans-terminal-ads/

PR #1911, opened 2019-09-04
allow user to silence opencollective message with environment variables

rettgerst:
whether or not you decide to remove the opencollective postinstall message, I think it's reasonable to request users' wishes not to see it.

this patch changes the script so that nothing is shown if the environment variables `ADBLOCK`, `OPENCOLLECTIVE_HIDE` or `DISABLE_OPENCOLLECTIVE` are set.


cristianbote:
Hey rett, we already considered removing it all together. Here https://github.com/preactjs/preact/pull/1904.
Thanks for opening it. I'll close yours and you can keep an eye on the first PR.

Thanks again for taking the time to open it

Issue #1916, opened 2019-09-09
`createElement` generates circular data structure in 10.0.0-rc.2


robertknight:
The `h`/`createElement` function in Preact 10.0.0-rc.2 can generate a circular data structure, causing an infinite loop when the result is later passed to `render`.

**Steps to reproduce**

```js
const { h, render } = require('preact');

const el = h('div', null, h('span', null), h('br', null));

console.log('VDOM node is', el);
```

**Expected:** Non-circular data structure
**Actual:** Circular data structure. This causes an infinite loop in `toChildArray` when rendering.

Slack discussion: https://preact.slack.com/archives/G9T60GUQ0/p1568062137383300

jviide:
The problem is here: https://github.com/preactjs/preact/blob/9fb01cbb6a2e654bd0288fcec24ca70ddf94fa96/src/create-element.js#L12-L20

This bug is triggered when `h` gets more than three arguments (`arguments.length > 3`).

Because children is an argument then the statement `children =  []` is functionally equivalent to `arguments[2] = []`. Due to this the first iteration of the for-loop (i.e. `children.push(arguments[2])`) is functionally equivalent to `children.push(children)`. This in turn causes the cyclical structure.

developit:
@JoviDeCroock I wonder if it'd be better to revert the optimization we merged that caused this?

antidotech:
Temporal workarounds for people facing this issue:

Because it's about how `arguments` is handled by some modes, using esmodules should work. E.g. if you're using webpack, 
```ts
alias: { preact: 'preact/dist/preact.module.js' }
```
(make sure that file is not excluded in bundler/compiler)

cf: https://github.com/preactjs/preact/pull/1917#issuecomment-529694815

PR #1917, opened 2019-09-09
(fix) - avoid making a circular ref for children

JoviDeCroock:
Fixes: https://github.com/preactjs/preact/issues/1916

Can't seem to reproduce this in tests for some reason


jviide:
(Sorry for the wall of text 🙂)

The trouble reproducing the bug (in tests) may be due to a difference between strict and non-strict modules/functions. According to https://www.ecma-international.org/ecma-262/6.0/#sec-arguments-exotic-objects NOTE 1:
> For non-strict functions the integer indexed data properties of an arguments object whose numeric name values are less than the number of formal parameters of the corresponding function object initially share their values with the corresponding argument bindings in the function’s execution context. This means that changing the property changes the corresponding value of the argument binding and vice-versa. This correspondence is broken if such a property is deleted and then redefined or if the property is changed into an accessor property. For strict mode functions, the values of the arguments object’s properties are simply a copy of the arguments passed to the function and there is no dynamic linkage between the property values and the formal parameter values.

I.e. the following piece of code works differently depending on whether it's a part of a strict or a non-strict module:

```js
function f(a) {
  a += 1;
  return arguments[0];
}
```

A strict version of `f(1)` returns `1` and a non-strict version returns `2`.

The minified `dist/preact.js` module is a non-strict module (it doesn't have the `"use strict";` line in the beginning), so using it directly triggers the bug. Adding the `"use strict";` line in to the beginning of `dist/preact.js` module ✨magically✨ suppresses the bug.

However, the Babel transform that Mocha uses compiles modules into strict mode, thus hiding the bug. It can be reproduced (rather cumbersomely) by adding the line `"ignore": ["./dist"]` to `.babelrc` and then importing  `../../dist/preact.js` instead of `../../src/index` in `shared/createElement.test.js` and then adding a triggering test.

developit:
@jviide right - so the clever bit of the `3+n` loop is that it assigns to arguments[2], which it never accesses again. I remember feeling hesitant when we were looking at merging this optimization but couldn't put my finger on why, my guess would be that I'd been bitten by this.

I think @JoviDeCroock's PR here to go back to the previous solution seems like the best course of action. Otherwise we'd have to create an intermediary variable, which is likely to be a fair number more bytes.

robertknight:
Is there a reason we don't make the entire Preact bundle operate in strict mode, which would also have avoided the issue here by changing the way `arguments` behaves? The source code is written as ES modules which are supposed to always be strict. So for Preact devs it looks like the code is strict, and _behaves_ strict in unit tests but isn't strict at runtime production builds depending on how the code is loaded. That seems to me like a footgun that could end up causing other problems in future.

test/browser/createElementProduction.test.js
```diff
@@ -0,0 +1,10 @@
+import { h } from '../../dist/preact';
+
+describe('createElement production build', () => {
+	it('should produce a correct structure', () => {
+		const el = h('div', null, h('span', null), h('br', null));
+		expect(JSON.stringify(el)).to.equal(
+			'{"type":"div","props":{"children":[{"type":"span","props":{},"__k":null,"__p":null,"__b":0,"__e":null,"l":null,"__c":null},{"type":"br","props":{},"__k":null,"__p":null,"__b":0,"__e":null,"l":null,"__c":null}]},"__k":null,"__p":null,"__b":0,"__e":null,"l":null,"__c":null}'
```
marvinhagemeister: Instead of checking against the shape directly, we could just check if `JSON.stringify` doesn't throw an error. According to [the spec](https://tc39.es/ecma262/#sec-json.stringify) it's must throw a `TypeError` when it encounters a cyclic structure.

cristianbote: That's really smart 😁

lishoulong (commented):
good work

PR #1923, opened 2019-09-10
Composition api

porfirioribeiro:
This pull request is a implementation of the [Vue Composition API](https://vue-composition-api-rfc.netlify.com/) in Preact.

This started as a proof of concept of implementing the API in Preact, but ended up being a full implementation. It adds  bytes to Preact itself and can be used standalone just like hooks, so both Composition and Hooks can be used also together. 

There are some differences from the Vue Api as Preact works differently internally.
I wriote a [blog post](https://dev.to/porfirioribeiro/preact-composition-api-ef1) to show the differences between Vue composition Api and React Hooks

There are still missing tests, docs, better typings, more examples and probably benchmark it to see the differences between this, hooks and classes.

If someone else find this concept interesting, i can work more on it and bring those missing parts.

Examples:
* [7guis](https://codesandbox.io/s/preact-composition-7guis-2i1l3)
* [async](https://codesandbox.io/s/preact-composition-github-demo-hfdsq)
* [provider](https://codesandbox.io/s/preact-composition-provider-demo-94lzq)
* [samples](https://codesandbox.io/s/github/porfirioribeiro/preact-composition-samples)


marvinhagemeister:
Wow, that's really cool! It's good to know that the Preact code base remains approachable even with the new addon structure :)

This is an area where I've done a lot of prototypes and experiments myself. Back when the react team announced hooks (and even earlier) we've been toying with a similar idea in spirit like the one vue is pursuing. The gist of it was to have constructor functions for components, like you have in this PR, and to follow a more observable like approach.

I'd love for this to become a solid alternative to hooks. I wouldn't focus too much on remaining 1:1 API compatible with vue and would rather tackle it from the perspective of what's best for Preact :+1: 

Regarding `computed`: I've written several variations of that with different implementation ideas and it turns out that it's fairly easy to add automatic dependency tracking. The same can be reused for `watch()`. The tricky bit is making sure that no glitches happen in the system (if that's a desireable requirement).

Anyhow, gotta run. I really love where this is going and I'm very interested in this. Feel free to hit me up on [slack](https://preact.slack.com/).

porfirioribeiro:
Thanks @marvinhagemeister 
I think the API could be fine tuned to be a solid alternative to hooks.

Would be nice to collaborate in the API design and add the missing pieces to make the API production ready.
I'm on Preact slack too, maybe we can chat there about this

developit:
This is ridiculously cool and I'm still digesting the syntax. It looks like observables!

porfirioribeiro:
Wow, thanks for your feedback. I guess better docs and samples would help...
In fact i don't know much about observables :cold_sweat: so probably it looks like it

porfirioribeiro:
Started adding some tests, trying to mimic the ones in hooks.
Some may not make sense as the API's have a different flow from hooks
But i will try to do them all, and then some more specific for the composition API

maiah:
Looks like Mobx React-lite https://mobx-react.js.org

porfirioribeiro:
Added more tests, achieved 100% coverage
While adding tests i found and fixed some bugs.

Also introduced a new feature provide/inject
It works like context but you don't need to wrap jsx with Ctx.Provider
Contexts are accessed using a simple string, if you pass a `reactive` or `ref` to it, everytime it changes it will update the components injecting it.
Folows the name in Vue: https://vue-composition-api-rfc.netlify.com/api.html#provide-inject and works mostly like that or Svelte get/setContext https://svelte.dev/docs#setContext

I'll work on a more elaborated demo (https://codesandbox.io/s/preact-composition-provider-demo-94lzq) inspired by https://github.com/hperrin/svelte-material-ui/blob/master/packages/list/List.svelte#L68

After creating more demos i would probably work on docs.

composition/src/index.js
```diff
@@ -0,0 +1,258 @@
...
+	if (isReactive(src))
+		ctx._component.__compositions.w.push({ src, cb: () => c.setState({}) });
+
+	return src;
+}
+
+export function reactive(value) {
+	const c = currentComponent;
+
+	return new Proxy(
```
JoviDeCroock: I don't think proxies support our current  browser support

ForsakenHarmony: well, could add it with a note that `preact/composition` only works with Proxies

porfirioribeiro:
Last night i've added some simple benchmark tests, mostly composition is faster than hooks.
Apart from `reactive`, i was expecting Proxy it to be slower, but not 90% slower:
```
compositionRef: 78127 kHz
compositionReactive: 6855 kHz (-91%)
hooks: 67721 kHz (-13%)
```
I've already think about a way to ditch Proxy, with minimal API changes, using Object.defineProperty
The biggest difference would be that when you call reactive you need to always pass an object with all properties.

With Proxy this works:
```js
const state= reactive({});
//later
state.someProp= 125;
```
With `defineProperty` it won't, as reactive don't now about `someProp`, but that can be solved with:
```js
const state= reactive({someProp: undefined});
//later
state.someProp= 125;
```

So i think i can ditch Proxy without breaking any tests, once i have time i'll work on it.

The other place where Proxy are used is passing props to the `createComponent` function, that can be changed to a getProps function, will probably even save some bytes.
```js
//now
const Comp = createComponent(props => {
	const logName = () => console.log(props.name);
	return () => <button onClick={logName}>name</button>;
});
//after
const Comp = createComponent(getProps => {
	const logName = () => console.log(getProps().name);
	return () => <button onClick={logName}>name</button>;
});
```

composition/src/index.js
porfirioribeiro: I will try to remove the need for Proxies, so no need for polyfill, and will not change browser support, hopefully

See the comment below for more information

porfirioribeiro: Removed Proxy from the PR

porfirioribeiro:
Changed the implementation to not use Proxys, it even saved +- 80 bytes.

It achieved a better performance with Object.defineProperties, but the `reactive` is still 50% slower than `ref`, blame for immutability Object.assign to shallow copy the inner object every time we change a value.

@developit  would it be possible to export `enqueueRender` and call that directly instead of `c.setState({})`, in a quick test, it seems to be like 6% faster. It could also be used in hooks.
There might be some unexpected quirk i didn't found

composition/src/index.js
```diff
@@ -0,0 +1,249 @@
...
+	let x = value;
+	const c = currentComponent;
+
+	const reactiveProperty = {
+		get() {
+			return x;
+		},
+		set(v) {
+			x = v;
+			c.setState({});
```
ForsakenHarmony: You can pull the object out as a constant, might make perf a bit better

porfirioribeiro: I did that change, i cant tell if perf is better

porfirioribeiro:
Working on https://codesandbox.io/s/github/porfirioribeiro/preact-composition-samples/tree/master/

Porting some samples from https://codesandbox.io/s/github/kentcdodds/learn-react-hooks/tree/master/
To compare with preact composition

https://kentcdodds.com/blog/dont-sync-state-derive-it

mindplay-dk:
@porfirioribeiro asked me to bring this up here for discussion.

I like this approach - a lot!

This proposal appears to address one of the two things I passionately dislike about hooks.

First: the apparent repeated initializations - e.g. `useState(createMyInitialState())` inside a functional component *appears* as though it's repeatedly initializing the state - and of course `createMyInitialState()` (or whatever expression you have in there) will evaluate and execute for no real reason at every update, with the initial state value getting discarded after the first update.

My gripe there isn't with the potentially adverse performance implications of that design, but with the simple fact that this code doesn't do what it appears to do. So kudos for properly separating initialization from rendering! This is a much better design - from a performance point of view as well, but mainly because this code does what it appears to do! 👍

Second: hiding things in global state. When you see a stand-alone statement such as `watch(thing)`, you have to wonder, how does the `watch` function know which component is watching the thing? What looks like just a flat function of course is reaching out for a "current component" in global state somewhere.

Why do we need to hide the real dependencies of `watch` or other functions in global state? In my opinion, hiding dependencies in favor of typing slightly less is a somewhat short-sighted payoff. For one, having global state always complicates testing - a plain function with no hidden dependencies is usually easier to understand, but, moreover, it's usually immediately understood by someone seeing a piece of code that calls that function.

Can we expose that hidden dependency and make it visible?

For example:

```js
const Comp = createComponent(component => {
    const filterRef = ref(component, 'ALL');
    const filteredItems = watch(
        component,
        [props => props.items, filterRef],
        ([items, filter]) => filterItems(items, filter)
    );
    return () => <ItemList items={filteredItems.value} />;
});
```

In my opinion, this should raise fewer eyebrows on first sight: this code just does what it appears to do. (It even tends to reads out as meaningful english - e.g. "watch component", which is exactly what you're asking that function to do: watch the component for changes to these properties...)

React hooks deviates from what I consider obvious, idiomatic JS in two ways - you've fixed the first.

Fix the latter, and I think you'll have me *all the way* sold on this idea. 🤩

Whatever you decide, I think this is definitely a step up from hooks. 👍

porfirioribeiro:
Passing component as a prop to `createComponent` function has good things and drawbacks.

the good thing is that it would help me with typescript support , as i could know what props we should expect
It would also make it simpler to access props as `component.props` would always point to the latest props

On other side, that would be exposing implementation details. If somehow the underline implementation change, it could break. Also passing the component to all composition functions kinda looks like boilerplate, but i can agree that it is more explicit.

I would love to ear other people opinions on this!

mindplay-dk:
> On other side, that would be exposing implementation details.

I see this not as exposing implementation details, but merely exposing the dependencies your code is really dependent on.

> If somehow the underline implementation change, it could break.

But that's true whether these functions obtain the component object as an argument or fetch it from global state.

> Also passing the component to all composition functions kinda looks like boilerplate, but i can agree that it is more explicit.

It is boilerplate.

But it's not difficult to implement - it's *easy* boilerplate.

In my opinion, don't abstract or hide things unless you're *helping* - if it's already *easy*, complicating things by introducing global state, hiding or over-abstracting things, if well-intentioned, is misguided.

I like simple things. If something is difficult, I might hide or abstract something - but I'll always weigh the added complexity against the improved ergonomics: it has to *really* pay off to be worth while.

If something is already easy, I don't try to make it easier.

porfirioribeiro:
Finished Typescript definitions, some complications on `watch/effect` second parameter, it's hard to extract a tuple.
Now to add more samples, [implemented 5](https://codesandbox.io/s/preact-composition-7guis-2i1l3) of the [7guis](https://eugenkiss.github.io/7guis/) already.

porfirioribeiro:
As things are getting to a more concrete point, i would like to have a review of what has been done and to know if there is any chance of this to be merged into Preact.

So if @developit, @JoviDeCroock , @marvinhagemeister , @ForsakenHarmony or any other Preact core team could have a look at this and review, either the implementation but also the API

@mindplay-dk pointed some options that might be considered or not, would be good to have some help on this! Thanks.

developit:
Hi @porfirioribeiro! this has come a long way since I last checked in!

Regarding `enqueueRender()`, I don't know that we can export that API. However, it is possible to obtain it with 1:1 performance using the following:

``sjs
import { Component } from 'preact';

const forceUpdate = Component.prototype.forceUpdate;
const enqueueRender = component => forceUpdate.call(component);
```

**Comment by porfirioribeiro** [comment]
Hi @developit, forceUpdate sees to be a little faster in some small tests, 
i just don't know if `_force` will not fore all children to re-render?

It's not a big deal and might be considered a premature optimization, i'll leave that for other time.

I'm focusing today on finish some final issues, so i can remove the WIP state

**Comment by porfirioribeiro** [comment]
The API for Preact composition is getting more far from Vue composition API.
And that's mostly because of the difference of the nature of both libraries.
I've added some more commits, and i think the API is on a good stage now.

* Use _renderCallbacks array for onMounted instead of creating a watcher with no args, was more messy to understand
* Removed support to passing ref's from `createdRef` to watch and effect, its not necessary and i have no way to assure that the passed argument is a ref or is just any object with a current property.
* as `watch` is mostly used to compute and memoize, i decided to simplify it by spreading the resolved args instead of passing an array of new args, other with old args and the onCleanup function. That will continue to be the behavior of `effect` as it is supposed to be used to subscribe or do some imperative stuff

TypeScript is the only bottleneck still bothers me, especially on `watch` and `effect`

File: composition/src/index.js [hunk 3]
```diff
@@ -0,0 +1,260 @@
...
+options.unmount = vnode => {
+	if (oldBeforeUnmount) oldBeforeUnmount(vnode);
+
+	/** @type {import('./internal').Component} */
+	const c = vnode._component;
+	if (c && c.__compositions) {
+		// cleanup `effect`s onCleanup
+		c.__compositions.e.forEach(cleanupEffect);
+		// call all onUnmounted lifecycle callbacks
+		c.__compositions.u.forEach(f => f());
```
JoviDeCroock: Bundle size nits, if you aren't returning anything from `cleanupEffect` and `f` it's better to do { f() } since now your transpilation will add `return` in front of it increasing bundle size.
Also `some` will save you some extra bytes compared to `forEach`

composition/src/index.js
```diff
@@ -0,0 +1,260 @@
...
+		c.__compositions.c = {};
+		c.getChildContext = () => c.__compositions.c;
+	}
+	c.__compositions.c[`__sC_${name}`] = { _component: c, _value };
+}
+
+export function inject(name, defaultValue) {
+	const c = currentComponent;
+
+	const ctx = c.context[`__sC_${name}`];
```
JoviDeCroock: you only seem to use `c` once I'd inline `currentComponent` here

composition/src/index.js
```diff
@@ -0,0 +1,260 @@
...
+		set(v) {
+			x = v;
+			c.setState(EMPTY_STATE);
+		}
+	};
+	return Object.defineProperties(
+		{},
+		Object.keys(value).reduce(
+			(acc, key) =>
+				Object.assign(acc, {
```
JoviDeCroock: We use an inhouse `assign` util which has a perf benefit over this one (also this one isn't supported too good in IE11)

composition/src/index.js
```diff
@@ -0,0 +1,260 @@
...
+					[key]: {
+						enumerable: true,
+						get() {
+							return x[key];
+						},
+						set(v) {
+							if (v !== x[key]) {
+								x = Object.assign({}, x);
+								x[key] = v;
+								c.setState(EMPTY_STATE);
```
JoviDeCroock: `EMPTY_STATE` is referentially equal won't this bailout?

composition/src/index.js
```diff
@@ -0,0 +1,260 @@
...
+	return Object.defineProperties(
+		{},
+		{
+			[$Reactive]: { get },
+			value: {
+				get,
+				set(newValue) {
+					if (v !== newValue) {
+						v = newValue;
+						c.setState(EMPTY_STATE);
```
JoviDeCroock: You only use this once, I'd inline it

composition/src/index.js
porfirioribeiro: That removed 6b from bundle, thanks

composition/src/index.js
porfirioribeiro: It's also used on line [112](https://github.com/preactjs/preact/pull/1923/files#diff-23369f01c5ea98c3654a47ca1abd2de7R112) to update the component when the injected variable is a value or reactive

composition/src/index.js
porfirioribeiro: Switched it, good catch

composition/src/index.js
porfirioribeiro: Actually in the beginning i had `{}` it does not make any difference.

composition/src/index.js
porfirioribeiro: At the point this function is called here, `currentComponent` may have already change, so i need to keep a reference to the component this belongs to.
It's kinda like what we do with hooks, but hooks are called every render, so `currentComponent` is the component the hook refers too, while compositions are only called once at setup phase

porfirioribeiro:
> This is looking really good already, made a first round of comments

Thanks @JoviDeCroock, i answered or solved your questions :+1:

cztomsik:
random idea: `createComponent` could be called `reactiveComponent`

```
export const Counter = reactiveComponent(() => ...)
```

I'm sure you were already thinking about that and maybe you have even better name :-)

porfirioribeiro:
Thanks for the excitement @cztomsik :exclamation: 
That's exactly my view and why i created this!
I hope to work a more on it soon!

fantasticsoul:
> This pull request is a implementation of the [Vue Composition API](https://vue-composition-api-rfc.netlify.com/) in Preact.
> 
> This started as a proof of concept of implementing the API in Preact, but ended up being a full implementation. It adds bytes to Preact itself and can be used standalone just like hooks, so both Composition and Hooks can be used also together.
> 
> There are some differences from the Vue Api as Preact works differently internally.
> I wriote a [blog post](https://dev.to/porfirioribeiro/preact-composition-api-ef1) to show the differences between Vue composition Api and React Hooks
> 
> There are still missing tests, docs, better typings, more examples and probably benchmark it to see the differences between this, hooks and classes.
> 
> If someone else find this concept interesting, i can work more on it and bring those missing parts.
> 
> Examples:
> 
> * [7guis](https://codesandbox.io/s/preact-composition-7guis-2i1l3)
> * [async](https://codesandbox.io/s/preact-composition-github-demo-hfdsq)
> * [provider](https://codesandbox.io/s/preact-composition-provider-demo-94lzq)
> * [samples](https://codesandbox.io/s/github/porfirioribeiro/preact-composition-samples)


I've implement one for react already in [Concent](https://github.com/concentjs/concent), see this demo:
https://codesandbox.io/s/concent-guide-xvcej

here is a article about setup feature:
https://dev.to/fantasticsoul/use-concent-release-react-hook-s-maximum-energy-2ldf

the most interesting thing about Concent's setup is that it can be reused in function component and class component both.

Welcome to understand and give me feedback

porfirioribeiro:
This was a great experiment! Allowed me to understand much about how Preact works internally.
Also gave me the chance to know and interact with very nice people, from which I'm always learning a lot!!

It was too ambitious for me 😄 but I hope at least some ideas here were good. And I'm really happy to see reactivity coming in the form of preact/signals

Closed in favor of https://github.com/preactjs/signals

PR #1924, opened 2019-09-10
(fix) vnode._children should keep their type as array when diffing

cristianbote:
When trying to pave the way for `preact-redux` with the latest rc, seems that the `_children` is loosing it's array type whenever `react-redux` is trying to `.forceUpdate()`.
This is the relevant error and log https://travis-ci.org/developit/preact-redux/builds/583356189#L390

Adds `+2B`

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -1,7 +1,7 @@
 import { EMPTY_OBJ, EMPTY_ARR } from '../constants';
 import { Component, enqueueRender } from '../component';
 import { Fragment } from '../create-element';
-import { diffChildren } from './children';
+import { diffChildren, toChildArray } from './children';
 import { diffProps } from './props';
 import { assign, removeNode } from '../util';
 import options from '../options';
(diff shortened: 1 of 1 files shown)


cristianbote:
I still need to add a test for this.

src/diff/index.js
```diff
@@ -118,7 +118,7 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 
 			tmp = c.render(c.props, c.state, c.context);
 			let isTopLevelFragment = tmp != null && tmp.type == Fragment && tmp.key == null;
-			newVNode._children = isTopLevelFragment ? tmp.props.children : tmp;
+			newVNode._children = toChildArray(isTopLevelFragment ? tmp.props.children : tmp);
```
developit: nit: should we pass the Array directly? (`toChildArray(.., [])`)

cristianbote: Well the third argument of `toChildArray` is actually the returned flattened array, and if that's missing is defined inside the function and returned. Was thinking that without passing it we could save some bytes.

marvinhagemeister:
@cristianbote did you manage to find a test case for this?👀 I had in mind that it would be added before merging, but my comment wasn't clear at all

cristianbote:
Oh 💩 totally forgot about it! Will add it asap. Thanks for reminding me! 😟

PR #1929, opened 2019-09-11
Fix controlled input values out of sync with dom

### LINKED ISSUE #1899: Uncontrolled input
Same logic with different libraries works different.

React - ✅
https://codesandbox.io/s/react-controlled-input-egxgu

Preact - ❌
https://codesandbox.io/s/preact-controlled-input-kib3p

marvinhagemeister:
Fixes #1899 .
Adds `+15 B`

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -255,6 +255,10 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
 		// (as above, don't diff props during hydration)
 		if (!isHydrating) {
 			if (('value' in newProps) && newProps.value!==undefined && newProps.value !== dom.value) dom.value = newProps.value==null ? '' : newProps.value;
+			// preact/#1899
+			// We need this value for input masking.
+			dom._lastValue = dom.value;
+
 			if (('checked' in newProps) && newProps.checked!==undefined && newProps.checked !== dom.checked) dom.checked = newProps.checked;
 		}
 	}
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -122,5 +122,6 @@ function setProperty(dom, name, value, oldValue, isSvg) {
  * @private
  */
 function eventProxy(e) {
-	return this._listeners[e.type](options.event ? options.event(e) : e);
+	this._listeners[e.type](options.event ? options.event(e) : e);
+	if (this._lastValue!=this.value) this.value = this._lastValue;
 }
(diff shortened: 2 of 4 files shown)


src/diff/props.js
```diff
@@ -122,5 +122,6 @@ function setProperty(dom, name, value, oldValue, isSvg) {
  * @private
  */
 function eventProxy(e) {
-	return this._listeners[e.type](options.event ? options.event(e) : e);
+	this._listeners[e.type](options.event ? options.event(e) : e);
```
andrewiggins: Do we need to still capture the return value from the handler and return it?

src/diff/props.js
```diff
@@ -122,5 +122,6 @@ function setProperty(dom, name, value, oldValue, isSvg) {
...
+	if (this._lastValue!=this.value) this.value = this._lastValue;
```
andrewiggins: This logic will apply for any event added to an input element, right? So if I add `onFocus`, I'll need to also make it a controlled input in order for the input to work? Just checking if I understand how this works.

src/diff/index.js
```diff
@@ -255,6 +255,10 @@ function diffElementNodes(dom, newVNode, oldVNode, context, isSvg, excessDomChil
 		// (as above, don't diff props during hydration)
 		if (!isHydrating) {
 			if (('value' in newProps) && newProps.value!==undefined && newProps.value !== dom.value) dom.value = newProps.value==null ? '' : newProps.value;
+			// preact/#1899
+			// We need this value for input masking.
+			dom._lastValue = dom.value;
```
developit: should we add a `mangle.json` mapping for this? We could reuse a property name that's only used on component or vnode objects right now, like `__s` ("next state").

src/diff/props.js
developit: ahhhh good catch. Maybe this should guard and only apply for `onChange`/`onInput`? Though right now it means this works with either, and guards against state mutations in focus/blur.

src/diff/props.js
marvinhagemeister: I don't think we do. We attach `eventProxy` like a standard event handler and those don't return anything:

```
dom.addEventListener("input"; e => {
  // No need to return anything here
});
```

marvinhagemeister:
I'll leave this open as I want to check if this PR leads to cursor jumps or something.

src/diff/index.js
marvinhagemeister: Done :ballot_box_with_check:

marvinhagemeister:
Just checked, this solution leads to cursor jumping :cry:

PR #1931, opened 2019-09-14
Start refactor to provider check for emit instead of SCU Consum…

### LINKED ISSUE #1933: 10.0.0-rc.2 break antd/popover
After updating rc2, the Popover component from the ant package stopped displaying a pop-up message.

It is likely that the problem is caused by https://github.com/preactjs/preact/pull/1901

error reproduction:
rc1: https://nuzsk.github.io/preact-x-consumer-bug/10.0.0-rc.1/dist/index.html
rc2: https://nuzsk.github.io/preact-x-consumer-bug/10.0.0-rc.2/dist/index.html
source:
https://github.com/Nuzsk/preact-x-consumer-bug

ps: rc3/master have the same rc2 behavior

JoviDeCroock:
I've started this but will probably be getting back to it next week, Bad news I've found a potential stale dom issue with the SCU I've added in the tests

Fixes #1933

EDIT: Add list of resolved issues - @marvinhagemeister

### DIFF
--- a/test/browser/createContext.test.js
+++ b/test/browser/createContext.test.js
@@ -420,7 +420,7 @@ describe('createContext', () => {
 			constructor(props) {
 				super(props);
 				this.state = {
-					x: 'y'
+					status: 'initial'
 				};
 
 				this.renderInner = this.renderInner.bind(this);
@@ -428,13 +428,13 @@ describe('createContext', () => {
 				app = this;
 			}
 
-			renderInner() {
-				return <p>{this.state.x}</p>;
+			renderInner(value) {
+				return <p>{value}: {this.state.status}</p>;
 			}
 
 			render() {
 				return (
-					<Provider value="irrelevant">
+					<Provider value="value">
 						<Consumer>
 							{this.renderInner}
 						</Consumer>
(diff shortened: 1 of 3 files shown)


mxstbr:
Note to readers, here is the most minimal reproduction of the issue: https://codesandbox.io/s/double-context-preventing-value-propagation-rj00e

cristianbote:
Hey @mxstbr thanks for the sandbox. The thing is, that with master that works. Got merged in here https://github.com/preactjs/preact/pull/1925/files#diff-e959fc84d230f763daeb8f8fa0a571e7R18. Is there another case where this _breaks_?

mxstbr:
I am running the latest master in our project locally and it is definitely still broken. Here is the slightly-less-minimal reproduction that I sent @JoviDeCroock privately: https://codesandbox.io/s/issue-1864-urv9j

test/browser/createContext.test.js
```diff
@@ -352,6 +353,10 @@ describe('createContext', () => {
 	});
 
 	it('should not re-render the consumer if the context doesn\'t change', () => {
+
+		// Disable the debounce rendering
+		options.debounceRendering = f => f();
```
marvinhagemeister: That should be done via `act()`. That we we don't have to manually restore it :)

Nuzsk:
Hi, should this pull request fix issue https://github.com/preactjs/preact/issues/1933?

marvinhagemeister:
@Nuzsk Just checked the repro in the linked issue with the changes in this PR and it fixes the issue for me. The popup appears as expected with this PR :+1:

marvinhagemeister:
Let-a-go :mushroom: :+1:

PR #1938, opened 2019-09-18
Adding volume to HTML Attributes

### LINKED ISSUE #1934: audio tag not support volume attribute in typescripts
I use 10.0.0-rc.1, and I want to use <audio {...someProps} volume={0.1}></audio>.
But in ts, I got an error :
Property 'volume' does not exist on type 'HTMLAttributes'.
Then I found in index.d.ts file where HTMLAttributes just don't include this attribute.
Is there any chance to add this?

jessicabyrne:
Fixes #1934 

@marvinhagemeister 
@cristianbote

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -750,6 +750,7 @@ export namespace JSXInternal {
 		ul: HTMLAttributes;
 		"var": HTMLAttributes;
 		video: HTMLAttributes;
+		volume: HTMLAttributes;
 		wbr: HTMLAttributes;
 
 		//SVG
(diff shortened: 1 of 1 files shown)


src/jsx.d.ts
```diff
@@ -750,6 +750,7 @@ export namespace JSXInternal {
 		ul: HTMLAttributes;
 		"var": HTMLAttributes;
 		video: HTMLAttributes;
+		volume: HTMLAttributes;
```
pmkroeker: This should be in HTMLAttributes instead.
```typescript
interface HTMLAttributes {
	...
	volume?: string | number
}
```

jessicabyrne: Right you are. New commit has this change, removed from Intrinsic Elements.

PR #1946, opened 2019-09-24
Fix UNSAFE_* lifecycles being overwritten

### LINKED ISSUE #1941: Preact@10.0.0-rc.3 preact/compat not calling UNSAFE* functions
I'm working with react-table in the latest version 6.10.3 they have renamed the deprecated methods by adding the UNKOWN prefix (https://github.com/tannerlinsley/react-table/pull/1461). However when working with Peact X and preact/compat these functions are not being called. 

Should preact/compat patch the unsafe functions on Component? 
This https://github.com/preactjs/preact/blob/master/compat/src/index.js#L368 seems to patch the functions on the Component.

In this issue https://github.com/preactjs/preact/issues/1461 you mention keeping them until they are removed from react. Does Preact X with preact/compat support the UNSAFE* functions ?

marvinhagemeister:
This was caused by transpilers overwriting our patched property when inheritance was in play. It's based on the assumption that having both the prefixed `UNSAFE_*` and the unprefixed lifecycle hook is invalid.

Fixes #1941
Adds `+29 B` to compat.

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -364,19 +364,6 @@ function memo(c, comparer) {
 	return Memoed;
 }
 
-// Patch in `UNSAFE_*` lifecycle hooks
-function setUnsafeDescriptor(obj, key) {
-	Object.defineProperty(obj.prototype, 'UNSAFE_' + key, {
-		configurable: true,
-		get() { return this[key]; },
-		set(v) { this[key] = v; }
-	});
-}
-
-setUnsafeDescriptor(Component, 'componentWillMount');
-setUnsafeDescriptor(Component, 'componentWillReceiveProps');
-setUnsafeDescriptor(Component, 'componentWillUpdate');
-
 /**
  * Pass ref down to a child. This is mainly used in libraries with HOCs that
  * wrap components. Using `forwardRef` there is an easy way to get a reference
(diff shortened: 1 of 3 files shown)


compat/src/index.js
```diff
@@ -406,6 +404,16 @@ options.vnode = vnode => {
 		vnode.props.ref = vnode.ref;
 		vnode.ref = null;
 	}
+
+	// We can't just patch the base component class, because components that use
+	// inheritance and are transpiled down to ES5 will overwrite our patched
+	// getters and setters. See #1941
+	if (typeof type === 'function' && !type._patchedLifecycles && type.prototype) {
+		setSafeDescriptor(type, 'componentWillMount');
```
developit: A thought: passing `type.prototype` here could save a couple bytes and property lookups.

JoviDeCroock:
Are we okay with merging this and doing the optimisation as a patch release?

compat/src/index.js
marvinhagemeister: Tried it out, but it looks like gzip already plays in our favor. Using a variable to store the `prototype` will lead to the same size and passing it directly is indeed `1B` smaller :tada:

Issue #1951, opened 2019-09-26
Missing disableRemotePlayback attribute in jsx.d.ts


johannesodland:
`disableRemotePlayback` should be defined in HTMLAttributes:

https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/disableRemotePlayback

Issue #1953, opened 2019-09-27
styled-components ref type


marvinhagemeister:
Got this bug report via email, so I'm pasting it here:

> (...)I've just got a quick question regarding the index.d.ts file of Preact, more exactly this line. Is there any reason why it's not ref?: Ref<any> | null; instead? I'm asking this since I'm having some troubles with setting up some styled components libraries with Typescript and the latest Preact rc, and adding the ? to this line fixed it. (...)

38elements:
> (...)I've just got a quick question regarding the index.d.ts file of Preact, more exactly this line. Is there any reason why it's not ref?: Ref | null; instead? I'm asking this since I'm having some troubles with setting up some styled components libraries with Typescript and the latest Preact rc, and adding the ? to this line fixed it. (...)

I think the sender of the email should show the detailed content of "some troubles".

xiel:
I'm also getting a similar error from the same type of ref when I want to use `react-spring` with preact.

`<animated.div style={animatedStyles} />`

**Property 'ref' is missing in type 'ReactElement'**
```
    Error:Error:line (117)TS2605: JSX element type 'ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)>) | (new (props: any) => Component<any, any, any>)>' is not a constructor function for JSX elements.
  Property 'ref' is missing in type 'ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)>) | (new (props: any) => Component<any, any, any>)>' but required in type 'Element'.
```

when I manually change (preact/src/index.d.ts)

```
interface VNode<P = {}> {
  ...
  ref: Ref<any> | null;
  ...
}
```

to be
`ref?: Ref<any> | null;`
the errors go away.

From what I see in the React.Element types, ref should be defined optional.

xiel:
Was broken with this PR it seems: https://github.com/preactjs/preact/pull/1447

victordidenko:
Still not working with many libraries: [react-spring](https://www.react-spring.io/), [astroturf](https://github.com/4Catalyzer/astroturf), [react-fontawesome](https://github.com/FortAwesome/react-fontawesome)

<img width="853" alt="Screenshot 2019-11-20 at 11 57 07" src="https://user-images.githubusercontent.com/15016227/69224364-75086480-0b8d-11ea-8388-7693e5a27649.png">

zaninime:
[...]

PR #1955, opened 2019-09-29
(chore) - fix some typing issues

JoviDeCroock:
Some minor typing related fixes

fixes: https://github.com/preactjs/preact/issues/1951

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -14,7 +14,7 @@ declare namespace preact {
 		type: ComponentType<P> | string | null;
 		props: P & { children: ComponentChildren } | string | number | null;
 		key: Key;
-		ref?: Ref<any> | null;
+		ref: Ref<any> | null;
 		/**
 		 * The time this `vnode` started rendering. Will only be set when
 		 * the devtools are attached.
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -14,7 +14,7 @@ declare namespace preact {
 		type: ComponentType<P> | string | null;
 		props: P & { children: ComponentChildren } | string | number | null;
 		key: Key;
-		ref: Ref<any> | null;
+		ref?: Ref<any> | null;
 		/**
 		 * The time this `vnode` started rendering. Will only be set when
 		 * the devtools are attached.
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -526,6 +526,7 @@ export namespace JSXInternal {
 		defer?: boolean;
 		dir?: string;
 		disabled?: boolean;
+		disableRemotePlayback?: boolean;
 		download?: any;
 		draggable?: boolean;
 		encType?: string;


src/index.d.ts
```diff
@@ -14,7 +14,7 @@ declare namespace preact {
 		type: ComponentType<P> | string | null;
 		props: P & { children: ComponentChildren } | string | number | null;
 		key: Key;
```
JoviDeCroock: Shouldn't key also be `?`

src/index.d.ts
```diff
@@ -14,7 +14,7 @@ declare namespace preact {
 		type: ComponentType<P> | string | null;
```
JoviDeCroock: Maybe replace string with KeyOf(HTMLTypes) to increase safety

src/index.d.ts
```diff
@@ -14,7 +14,7 @@ declare namespace preact {
...
 		props: P & { children: ComponentChildren } | string | number | null;
 		key: Key;
-		ref: Ref<any> | null;
+		ref?: Ref<any> | null;
```
38elements: I think that `?` is unnesseary if only `createVNode()` provides VNode.
https://github.com/preactjs/preact/blob/870cf60352c46a86e110dd598efe7be9179ba979/src/create-element.js#L53-L73

JoviDeCroock: Hmm, then I'll maybe have to dig deeper into this issue https://github.com/preactjs/preact/issues/1953

JoviDeCroock: Well I've re thought this through and if createVNode provides undefined as a ref then it should be ? since it's just undefined if the user doesn't provide one which makes it optional, no? Please correct me if I'm wrong

38elements: > PS: do you agree with the other comments I made?

No, I do not.
Since I think `ref` or `ref?` is exsiting `ref` key or not in object.

```ts
let a : {b?: undefined} = {}
let c : {b: undefined} = {} // error
```

http://www.typescriptlang.org/play/index.html

src/index.d.ts
38elements: Is it necessary to think about custom elements?

JoviDeCroock: Good point

src/index.d.ts
38elements: I think this is same issue as `ref`.

src/index.d.ts
marvinhagemeister: Yep, @38elements ist correct. The `?` sign means that the property doesn't exist which is different from allowing`undefined` as a value

xiel: you broke this https://github.com/preactjs/preact/issues/1953

JoviDeCroock: There is no change that happened here to ref, look at https://github.com/preactjs/preact/pull/1955/files please

xiel: Yes sorry, I mixed up the tabs and diffs, you just changed it back and forth. I landed here because you referenced the issue and the type error is still there.
But the ref **?** was removed here initially: #1447 (from what I see)

PR #1965, opened 2019-10-02
(fix) Fix the href not getting removed (#1943)

### LINKED ISSUE #1943: Undefined props not handled correctly.
Per Jason, this line

https://github.com/preactjs/preact/blob/master/src/diff/props.js#L17 

causes incorrect behavior for hyperlinks in some rare cases.

cristianbote:
This fixes #1943 

~Adds `+4B` 🙁~
~Shaves `-1B` 😁~
~Adds `+15B` 😢~ 
Adds `+12B` 😢


cristianbote:
Look into this and basically `href` property exists on `dom`. So, it's set to empty string when `null` value is passed along.

https://github.com/preactjs/preact/pull/1965/files#diff-642a704212e8abfc3d854f17a1aaacbeR101

marvinhagemeister:
Is `href` the only attribute where that error pops up?

cristianbote:
@marvinhagemeister Well that's kinda the issue here. Basically anything that `name in dom` will end up with empty string, whenever it's gonna be removed from props.

I did tried that before and everything passed. But, is that the proper behaviour? Would it break custom elements 💯 support? 🤔

src/diff/props.js
```diff
@@ -14,7 +14,7 @@ export function diffProps(dom, newProps, oldProps, isSvg, hydrate) {
 	let i;
 
 	for (i in oldProps) {
-		if (!(i in newProps)) {
+		if (typeof newProps[i] === 'undefined') {
```
marvinhagemeister: FIY: I checked if `newProps[i] === undefined` would be smaller, but it leads to the same size :tada:

cristianbote: Uh, I think we've used this somewhere else, and I hoped to get better size doing the `typeof`. But yeah seems to be the same 😄 

Thanks for looking into it!

cristianbote:
This is breaking the uncontrolled input behaviour, which is true. Looking to fix that.

cristianbote:
I think this is ready. The added size is bothering me but oh well. Let me know what you think

src/diff/props.js
```diff
@@ -96,7 +96,19 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 		&& !isSvg
 		&& (name in dom)
 	) {
-		dom[name] = value==null ? '' : value;
+		if (value==null) {
+			// This condition here guards against modifying a unonctrolled input's value
```
developit:
```suggestion
			// This condition here guards against modifying an uncontrolled input's value
```

src/diff/props.js
```diff
@@ -96,7 +96,19 @@ function setProperty(dom, name, value, oldValue, isSvg) {
...
+			if (name!=='value' && name!=='checked') {
```
developit: ooooh crap I hadn't thought of this implication. There are a bunch more names here that would have to be disabled, I think this might be a sign that we need to try different overall solutions...

cristianbote: Yup, you're right. I'm gonna start looking at it from other angles

cristianbote:
I'm gonna close this one since I'm pursuing another option here. Will open a different one.

PR #1968, opened 2019-10-02
(feat) - start migrating hooks to core

JoviDeCroock:
Open for golfing!

TODO: https://github.com/preactjs/preact/pull/1992/files#diff-34ef6e66e4ad48c6d14a981b05075ab9R219


src/diff/index.js
```diff
@@ -110,7 +111,12 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 			c.props = newProps;
 			c.state = c._nextState;
 
-			if (tmp = options._render) tmp(newVNode);
```
JoviDeCroock: options._render is now unused

JoviDeCroock: Double checked if this is not using the hook https://github.com/duxiaofeng-github/preact-responder-event-plugin/blob/master/src/index.js

porfirioribeiro: > options._render is now unused (not quite sure if this is breaking since it's an internal field)

Well actually it's being used here on the composition API PR
https://github.com/preactjs/preact/pull/1923/files#diff-23369f01c5ea98c3654a47ca1abd2de7R9
Removing that would break it for good

JoviDeCroock: Thanks for notifying me, I'll put it back and pick up work where I left up when the thought process behind this is done.

Issue #1806, opened 2019-07-26
10.0.0-rc.0 portal render differs from React when portal container element has existing children


rmales:
Hi,

I noticed that when using a portal in Preact the portal content is mounted in a different order compared to React when the portal container has existing child nodes. In Preact its prepended while in React its appended.
Example: https://codesandbox.io/s/patient-dream-0b83r

I also noticed that on rerender the existing child nodes are removed in Preact, but maybe that issue is fixed by https://github.com/preactjs/preact/pull/1781 (is there a way to test that commit in CodeSandbox)?

Thanks,
Richard

JoviDeCroock:
Oh, that would be my mistake then, thought React prepended as well. Will fix that later thanks for the issue report!

PR #1971, opened 2019-10-03
Append portal node to container instead of prepend

toraora:
This fixes the issue described in #1806. Expected test outcomes were
generated by taking the test code and running it with React instead of
Preact.

### DIFF
--- a/compat/test/browser/portals.test.js
+++ b/compat/test/browser/portals.test.js
@@ -76,7 +76,7 @@ describe('Portal', () => {
 
 		toggle();
 		rerender();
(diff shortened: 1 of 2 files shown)


compat/test/browser/portals.test.js
```diff
@@ -452,6 +452,6 @@ describe('Portal', () => {
 
 		toggle();
```
JoviDeCroock: For sanity's sake could you duplicate this test with `<Modal open={open}>Hello</Modal>` as firstChild, I remember trying to avoid `append` for some reason if that works out I'll test something myself and else approve the PR

toraora: Added the test

PR #1974, opened 2019-10-03
(feat) - queue effects for dom correctness

JoviDeCroock:
Could use an extra pair of eyes for that last test

### DIFF
--- a/test/browser/hooks/useEffect.test.js
+++ b/test/browser/hooks/useEffect.test.js
@@ -32,13 +32,12 @@ describe('useEffect', () => {
 			return null;
 		}
 
-		render(<Comp />, scratch);
-		render(<Comp />, scratch);
+		act(() => {
+			render(<Comp />, scratch);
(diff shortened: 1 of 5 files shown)


test/browser/hooks/useEffect.test.js
```diff
@@ -32,13 +32,12 @@ describe('useEffect', () => {
...
-		render(<Comp />, scratch);
-
-		expect(cleanupFunction).to.be.not.called;
-		expect(callback).to.be.calledOnce;
+		act(() => {
+			render(<Comp />, scratch);
+			render(<Comp />, scratch);
+		});
 
-		render(<Comp />, scratch);
```
JoviDeCroock: I'm honestly scared because act is now a hard requirement for useEffect which it wasn't before... This could break some tests :(

andrewiggins: Yea, agreed

PR #1976, opened 2019-10-04
Reduce size of adding hooks to core (-117 B)

andrewiggins:
## Summary

Individual commit messages have more detail about the change if you are curious.

897bb21 Replace `globalHookState` with `resetHookState` function (-70 B)
4954398 Fix bug introduced by descturcturing `useImperativeHandle` hook state (+2 B)
ecb99ee Reuse `vnode._component` null check in `unmount` to call hook cleanup callbacks (-5 B)
401ed47 Move `pendingLayoutEffects` flush into component diff to reuse `c` variable (-12 B)
4756817 Minify `useImperativeHandle` hook state props (-10 B)
bb9acdd Schedule imperative handle creation as a layout effect (-31 B)
bd45a91b Fix `useContext` args change check (+9 B)
8e4de8f7 Move `handleEffects` into hooks.js so most hooks code is in hooks file (+0 B)

Total size:
4264 B: preact.js.gz (-117 B) 😁 
3884 B: preact.js.br (-112 B) 😁

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -149,10 +149,6 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 			}
 
 			if (c.__hooks) {
-				c.__hooks._handles.some(handle => {
-					if (handle._ref) handle._ref.current = handle._createHandle();
-				});
-				c.__hooks._handles = [];
 				c.__hooks._pendingLayoutEffects = handleEffects(c.__hooks._pendingLayoutEffects);
 			}
 
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -150,7 +150,7 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 
 			if (c.__hooks) {
 				c.__hooks._handles.some(handle => {
-					if (handle.ref) handle.ref.current = handle.createHandle();
+					if (handle._ref) handle._ref.current = handle._createHandle();
 				});
 				c.__hooks._handles = [];
 				c.__hooks._pendingLayoutEffects = handleEffects(c.__hooks._pendingLayoutEffects);
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -148,6 +148,14 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 				c.componentDidUpdate(oldProps, oldState, snapshot);
 			}
 
+			if (c.__hooks) {
+				c.__hooks._handles.some(handle => {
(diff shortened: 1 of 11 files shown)


src/diff/index.js
```diff
@@ -158,15 +161,6 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 		}
 
 		if (tmp = options.diffed) tmp(newVNode);
-
-		const hooks = newVNode._component && newVNode._component.__hooks;
-		if (hooks) {
-			hooks._handles.some(handle => {
-				if (handle.ref) handle.ref.current = handle.createHandle();
-			});
-			hooks._handles = [];
```
andrewiggins: Note: I removed this code by scheduling imperative handle creation as a layout effect, reusing the layout flush to call `createHandle` so you won't see it anywhere.

src/diff/index.js
```diff
@@ -385,14 +379,3 @@ function doRender(props, state, context) {
 
 	throw error;
 };
-
-export function handleEffects(effects) {
```
andrewiggins: I just moved this function into the hooks file since it mostly deals with hook data

src/hooks.js
```diff
@@ -80,24 +80,28 @@ export function useEffect(callback, args) {
...
 
-export function useImperativeHandle(ref, createHandle, args) {
-	let { _args } = getHookState(globalHookState.currentIndex++);
-	if (argsChanged(_args, args)) {
-		_args = args;
-		globalHookState.currentComponent.__hooks._handles.push({ ref, createHandle });
-	}
+export function useImperativeHandle(_ref, _createHandle, args) {
+	// -37 B
+	// useMemo(() => { if (_ref) { _ref.current = _createHandle(); } }, args);
```
andrewiggins: I think #1909 prevents us from saving bytes by using `useMemo` here. I'll remove this before checking in.

src/index.js
```diff
@@ -1,9 +1,26 @@
...
+	isValidElement
+} from './create-element';
 export { Component } from './component';
 export { cloneElement } from './clone-element';
 export { createContext } from './create-context';
 export { toChildArray } from './diff/children';
 export { unmount as _unmount } from './diff';
 export { default as options } from './options';
-export * from './hooks';
+export {
```
andrewiggins: Manually listing exports to avoid exporting the `resetHookState` function

src/hooks.js
andrewiggins: Yup. `useMemo` causes the test in #1909 to fail. Going with `useLayoutEffect`

PR #1979, opened 2019-10-05
Improve VNode typings

andrewiggins:


src/index.d.ts
```diff
@@ -12,7 +12,7 @@ declare namespace preact {
 
 	interface VNode<P = {}> {
 		type: ComponentType<P> | string | null;
-		props: P & { children: ComponentChildren } | string | number | null;
```
andrewiggins: I don't think we expose Text VNodes (the only VNodes where props isn't an object) to the user, so I removed them.

src/index.d.ts
```diff
@@ -208,7 +208,7 @@ declare namespace preact {
 	// Preact helpers
 	// -----------------------------------
 	function createRef<T = any>(): RefObject<T>;
-	function toChildArray(children: ComponentChildren): Array<VNode | null>;
+	function toChildArray(children: ComponentChildren): Array<VNode | string | number>;
```
andrewiggins: `toChildArray` removes `null`s from `props.children`

src/index.d.ts
marvinhagemeister: Good point :+1:

PR #1980, opened 2019-10-05
(fix) - should reorder memoized children

JoviDeCroock:
So apparently the fix introduced in the other PR where we make children point to their parents make this line irrelevant.

Save bytes + fix code 🎉 

-15B

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -90,7 +90,7 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 					c.state = c._nextState;
 					c._dirty = false;
 					c._vnode = newVNode;
-					newVNode._dom = oldDom!=null ? oldDom!==oldVNode._dom ? oldDom : oldVNode._dom : null;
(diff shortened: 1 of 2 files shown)


marvinhagemeister:
Oh that'd be really cool 👍💯

We should double check with the rest cases from https://github.com/preactjs/preact/pull/1879 just to be sure that we don't introduce any regressions

JoviDeCroock:
@marvinhagemeister can confirm the demo still works

src/diff/index.js
```diff
@@ -90,7 +90,7 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 					c.state = c._nextState;
 					c._dirty = false;
 					c._vnode = newVNode;
-					newVNode._dom = oldDom!=null ? oldDom!==oldVNode._dom ? oldDom : oldVNode._dom : null;
+					newVNode._dom = oldVNode._dom;
```
cristianbote: This line always confused me 😄 anyway. Good catch!

PR #1981, opened 2019-10-06
Small edit in util

nagayev:
Use Object.assign if it's possible

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -6,6 +6,7 @@
  * @returns {O & P}
  */
 export function assign(obj, props) {
+	if(Object.assign) return Object.assign(obj, props);
 	for (let i in props) obj[i] = props[i];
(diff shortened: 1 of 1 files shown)


JoviDeCroock:
This would increase bundle size and reduce performance in environments with Object.assign

marvinhagemeister:
Awesome, thanks taking the time to make a PR :tada: We benchmarked the performance of native `Object.assign` vs our custom `assign` function in all current and many older browsers. Turns out that our solution has the same performance and is faster in older browsers like IE11 which is part of our browser scope.

Adding two different code-paths may choose the appropriate strategy at runtime, but is not really doable for us given our 3kB size constraints as @JoviDeCroock wrote. For this reason we unfortunately can't accept this PR.

Rest assured that we do really appreciate experimenting with different solutions :+1:

PR #1984, opened 2019-10-07
Always pull force flag from component

### LINKED ISSUE #1983: 10.0.0 break context update with mobx.toJs usage
After updating 10.0.0 child components can't observe store passed by context when mobx.toJs used in their parent.

explanation of example:
1. MobX store passed into Context.Provider value in App component
2. contenxt.store.count serialized & printed in Test component
3. contenxt.store.count printed in Counter component
4. user click button and compare clicks/raw value on your screen
5. the values are the same for 10.0.0-rc.1 and differ for 10.0.0 release

error reproduction (cache cleaning may be required):
rc1: https://nuzsk.github.io/preact-x-consumer-bug/10.0.0-rc.1/dist/index.html
rel: https://nuzsk.github.io/preact-x-consumer-bug/10.0.0/dist/index.html
source:
https://github.com/Nuzsk/preact-x-consumer-bug

note: removing toJs allows the child component to update

marvinhagemeister:
The way MobX works is that every observed component returns `false` from `shouldComponentUpdate` unless `props` have changed. They than trigger the child renders themselves via direct `forceUpdate` calls.

Our issue was that we only set the `force` flag sort-of globally during a render and would ignore any children that have been enqueued via `forceUpdate` in the same render cycle. This PR addresses that by ensuring that the `force` flag is always the one from the current component.

**Tasks:**
- [x] Add a test case

Fixes #1983
Size `-24 B` :100:

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -123,13 +123,13 @@ export function getDomSibling(vnode, childIndex) {
 function renderComponent(component) {
 	let vnode = component._vnode,
 		oldDom = vnode._dom,
-		parentDom = component._parentDom,
-		force = component._force;
-	component._force = false;
+		parentDom = component._parentDom;
+
 	if (parentDom) {
 		let mounts = [];
-		let newDom = diff(parentDom, vnode, assign({}, vnode), component._context, parentDom.ownerSVGElement!==undefined, null, mounts, force, oldDom == null ? getDomSibling(vnode) : oldDom);
+		let newDom = diff(parentDom, vnode, assign({}, vnode), component._context, parentDom.ownerSVGElement!==undefined, null, mounts, oldDom == null ? getDomSibling(vnode) : oldDom);
 		commitRoot(mounts, vnode);
+		component._force = false;
 
 		if (newDom != oldDom) {
 			updateParentDomPointers(vnode);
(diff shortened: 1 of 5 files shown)


cristianbote:
Wow! 😁

andrewiggins:
@marvinhagemeister Woo! Love it when we can save bytes and fix bugs :)

In `src/components.js:132` you clear the force flag: `component._force = false`. Should this be somewhere else that guarentees the flag is always cleared even for children of whose `_force` flag is true.

In the UT that you added, do we also need to clear the <Inner> components `_force` flag? What I'm thinking happens is that both Inner and Outer queue up a rerender, but `renderComponent` is only called with Outer, because Inner's `_dirty` flag is set to `false` when Outer's rerender is processed. Since Inner never goes into `renderComponent` (the condition on `src/component.js:198` prevents it).

I took a stab at this but wanted to drop a message here so someone can look at it in case I forget.

marvinhagemeister:
@andrewiggins Good catch! That's something I didn't think of. Maybe we just need to reset the flag at the end of `diff()`, once a component has finished rendering instead of in `renderComponent`

PR #1987, opened 2019-10-08
(fix) - resolve attributes of reused nodes that weren't present in vdom

JoviDeCroock:


test/browser/render.test.js
```diff
@@ -464,6 +464,26 @@ describe('render()', () => {
...
+			scratch.appendChild(div);
+
+			const App = () => (
+				<div>
+					<span>Bye</span>
+				</div>
+			);
+
+			render(<App />, scratch);
+			expect(scratch.innerHTML).to.equal('<div class=""><span>Bye</span></div>');
```
cristianbote: Huh... I guess `class` should be missing entirely here, based on this https://github.com/preactjs/preact/pull/1965. Does it makes sense? A previous attribute, removed entirely from node if that's missing. I guess it kinda does.

JoviDeCroock: className also resolves to an empty string it seems

cristianbote: Yup. I guess this will surface in my PR. I'll handle it then and ping you 😛

PR #1990, opened 2019-10-09
Add _force to internal.d.ts

38elements:


src/internal.d.ts
```diff
@@ -57,23 +57,27 @@ export interface VNode<P = {}> extends preact.VNode<P> {
...
 
 export interface Component<P = {}, S = {}> extends preact.Component<P, S> {
+	// When component is functional component, this is reset to functional component
 	constructor: preact.ComponentType<P>;
 	state: S; // Override Component["state"] to not be readonly for internal use, specifically Hooks
 	base?: PreactElement;
 
 	_dirty: boolean;
-	_renderCallbacks: Array<() => void>;
+	_force?: boolean | null;
```
38elements: https://github.com/preactjs/preact/blob/cd6445054718c5a24f4fc332d7726dfca0e0aa07/src/component.js#L53-L70

https://github.com/preactjs/preact/blob/cd6445054718c5a24f4fc332d7726dfca0e0aa07/src/diff/index.js#L149-L151

src/component.js
```diff
@@ -13,20 +13,6 @@ import { Fragment } from './create-element';
...
-	// this._renderCallbacks = []; // Only class components
-
-	// Other properties that Component will have set later,
-	// shown here as commented out for quick reference
-	// this.base = null;
-	// this._context = null;
-	// this._vnode = null;
-	// this._nextState = null; // Only class components
-	// this._processingException = null; // Always read, set only when handling error
-	// this._pendingError = null; // Always read, set only when handling error. This is used to indicate at diffTime to set _processingException
```
38elements: I think this comment is the same role as the type definition. 
Therefore, I think it is better that this comment moves to type definition.

PR #2003, opened 2019-10-12
Rewrite useImperativeHandle to use useLayoutEffect (-35 B)

andrewiggins:
This PR brings some improvements to hooks that @JoviDeCroock and I have been playing with in other branches for a little while into master.

@JoviDeCroock I thought I'd bring these in a separate PR cuz I have a couple different branches I'd like to build on top of them. That alright?


hooks/src/index.js
```diff
@@ -61,18 +60,27 @@ function getHookState(index) {
...
 	if (index >= hooks._list.length) {
 		hooks._list.push({});
 	}
 	return hooks._list[index];
 }
 
+/**
+ * @param {import('./index').StateUpdater<any>} initialState
+ */
 export function useState(initialState) {
```
andrewiggins: When I tested it, converting this to `useReducer.bind` seemed to increase the size so I left it as is for now.

hooks/src/index.js
```diff
@@ -132,19 +140,16 @@ export function useRef(initialValue) {
...
-		currentComponent.__hooks._handles.push({ ref, createHandle });
-	}
-}
-
-function bindHandles(handles) {
-	handles.some(handle => {
-		if (handle.ref) handle.ref.current = handle.createHandle();
-	});
-	return [];
+	args.push(ref);
```
sventschui: Not sure how our general stance on this topic is but we are modifying an array from user-land here. This might have strange side effects for users. What do you think about this?

JoviDeCroock: I’ve thought about this and this introduces non-deterministic behavior what if the user sends no array (happens on render) crash/limitation

andrewiggins: Oh good points. I'll fix up and add tests

developit: Suggestion

```suggestion
	args = args.concat(ref);
```

JoviDeCroock: Problem with that is that this can diminish the on every render to become once and never again

hooks/src/index.js
```diff
@@ -239,8 +244,10 @@ function safeRaf(callback) {
 if (typeof window !== 'undefined') {
 	let prevRaf = options.requestAnimationFrame;
 	afterPaint = (component) => {
-		if ((!component._afterPaintQueued && (component._afterPaintQueued = true) && afterPaintEffects.push(component) === 1) ||
-		    prevRaf !== options.requestAnimationFrame) {
+		if (
+			(!component._afterPaintQueued && (component._afterPaintQueued = afterPaintEffects.push(component)) === 1)
+			|| prevRaf !== options.requestAnimationFrame
```
developit: Hmm - I'm wondering how common the case of swapping `options.requestAnimationFrame` is that we'd need this. Was there a specific case where this was troublesome? Only thing I could think of was injecting a super long rAF and then wanting to cancel.

hooks/src/index.js
```diff
@@ -239,8 +244,10 @@ function safeRaf(callback) {
 if (typeof window !== 'undefined') {
 	let prevRaf = options.requestAnimationFrame;
 	afterPaint = (component) => {
-		if ((!component._afterPaintQueued && (component._afterPaintQueued = true) && afterPaintEffects.push(component) === 1) ||
-		    prevRaf !== options.requestAnimationFrame) {
+		if (
+			(!component._afterPaintQueued && (component._afterPaintQueued = afterPaintEffects.push(component)) === 1)
```
developit: Question: In core's state queue, we always push onto the state array and just dedupe the resulting array when flushing (invoking the callback sets a boolean that makes subsequent invocations inert). I wonder if that'd be cheaper here?

hooks/src/index.js
```diff
@@ -268,6 +281,10 @@ function invokeEffect(hook) {
 	if (typeof result === 'function') hook._cleanup = result;
 }
 
+/**
+ * @param {any[]} oldArgs
+ * @param {any[]} newArgs
+ */
 function argsChanged(oldArgs, newArgs) {
 	return !oldArgs || newArgs.some((arg, index) => arg !== oldArgs[index]);
```
developit: I know it's not part of this PR, but I wonder if this is smaller after gzip?

```suggestion
    if (!oldArgs || oldArgs.length!==newArgs.length) return;
    for (let i=newArgs.length; i--; ) if (newArgs[i]!==oldArgs[i]) return;
    return true;
```

hooks/src/index.js
developit: Hmm - I'm not sure what you mean. My understanding is that `useImperativeHandle()` is always going to get called on every render (even if it returns a cached value). The `args` value will always be new unless someone hoists it out of render, in which case modifying it in-place will cause an infinite loop.

JoviDeCroock: Yes but the args array is optional meaning that it can't cache at all. If we always make it an array it is dependent on a mutable ref which potentially does not change. This makes it so that useImperativeHandle(() => ) is equivalent to useImperativeHandle(() => , []) which is not deterministic

andrewiggins: Updated this code so that `useImperativeHandle` should support all the functionality of `useLayoutEffect` and added to tests to capture missing features I think

hooks/src/index.js
andrewiggins: Oh interesting... Yea that probs is.

Though in the next PR I plan on opening I rewrote this mechanism to no longer need the `_afterPaintQueued` at all. [Relevant commit is here](https://github.com/preactjs/preact/commit/650e93f5436d75c17d3ab7f13a7c7ecc107956d6) if you are curious. So I'll leave this for now and we can perhaps discuss again in the next PR. Thoughts?

hooks/src/index.js
andrewiggins: No worries. All comments welcome 😄 

This function returns `true` if the args are different so I think the `return` values need to be swapped like so:

```suggestion
	if (!oldArgs || oldArgs.length!==newArgs.length) return true;
	for (let i=newArgs.length; i--; ) if (newArgs[i]!==oldArgs[i]) return true;
	return;
```

In my tests it adds 21 B. However, given that `argsChanged` is potentially run a lot in `render` methods, perhaps it might be worth the bytes for the perf boost? (I'm assuming a `for` loop is marginally faster than `.some`).

hooks/src/index.js
andrewiggins: I don't think its very common and personally would love to save some bytes here (-24 bytes!). It was orginally added in #1851 so perhaps @robertknight could chime and let us know his thoughts? Perhaps if we could fix the original `act` issue from #1851 in a different way, could we perhaps remove this check here?

hooks/src/index.js
andrewiggins: For anyone curious, [here is the relevant React code](https://git.io/Je8mb) I believe

hooks/src/index.js
robertknight: I am sure that the issue captured by the tests in #1851 should stay fixed, i.e. `act` needs to reliably flush any renders/effects triggered during its callback regardless of whether there are renders/effects happening outside of `act` callbacks or not. Tests can break in very confusing ways otherwise.

The Enzyme adapter installs a [`debounceRendering` hook](https://github.com/preactjs/enzyme-adapter-preact-pure/blob/67a70088f1aff9ddf0ef1df9e3a63e1e9c77f9aa/src/debounce-render-hook.ts#L21) to ensure that it can flush pending state updates when Enzyme asks for the current render output. It doesn't access `requestAnimationFrame` though, so if you can find an alternative way to keep the tests added in #1851 passing, that would be fine from my point of view.

hooks/src/index.js
andrewiggins: Gonna leave it for now. PR #2012 has the change where we no longer need to track if the effect has been queued.

PR #2006, opened 2019-10-12
Import and re-export hooks explicitly to allow for Webpack tree shaking

### LINKED ISSUE #1529: Maybe provide tree-shakeable alternative to preact/compat?
I've read the release notes for 10.0.0-alpha.0 (https://github.com/developit/preact/releases/tag/10.0.0-alpha.0) where you say that

> preact/compat […] still has both named and a default export to remain compatible with react

I absolutely get that, but some of the functions in compat can actually be useful on their own! Yesterday, I wanted to use `memo`, but it more than doubled my bundle size because now I was importing all of compat. (I then copied just `memo` into my codebase, but I don't think that's necessarily the best way to go about this. :D)

Do you think there may be value in providing some of the functions on their own? Maybe something like 

```js
import { memo } from 'preact/utils'
```

On first look, these functions could be especially useful on their own:
- Portal
- PureComponent
- memo

davidbailey00:
Addresses #1529 - by eliminating the function call, bundlers can optimize more efficiently. There's still a bundle size increase of ~1.8 KB in projects using `preact/compat` when it exports a default object at all, compared to only named exports, but this PR reduces that amount from ~4.5 KB.

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -1,5 +1,5 @@
 import { hydrate, render as preactRender, cloneElement as preactCloneElement, createRef, h, Component, options, toChildArray, createContext, Fragment, _unmount } from 'preact';
-import * as hooks from 'preact/hooks';
+import { useState, useReducer, useEffect, useLayoutEffect, useRef, useImperativeHandle, useMemo, useCallback, useContext, useDebugValue } from 'preact/hooks';
 import { Suspense, lazy } from './suspense';
 import { assign, removeNode } from '../../src/util';
 
@@ -455,7 +455,17 @@ export {
 };
 
 // React copies the named exports to the default one.
-export default assign({
+export default {
+	useState,
+	useReducer,
+	useEffect,
+	useLayoutEffect,
+	useRef,
+	useImperativeHandle,
+	useMemo,
+	useCallback,
+	useContext,
+	useDebugValue,
 	version,
 	Children,
 	render,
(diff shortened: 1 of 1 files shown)


developit:
@davidbailey00 good catch! can we also replace the re-export on [Line 131](https://github.com/preactjs/preact/pull/2006/files#diff-efd159ddcc099e90029db3b306b8646aR431) with this approach? For the CJS and UMD bundles we only ship a single exports object.

davidbailey00:
> can we also replace the re-export on Line 431 with this approach?

There's no difference in changing this line as far as I can tell since it's already exporting the "native" way, i.e. bundlers are already able to optimize.

davidbailey00:
Re: the additional size reduction by removing the default export entirely, perhaps there's some mutation happening to one of the properties of the default export object, which is preventing Webpack being able to entirely remove those functions even though they're unused.

JoviDeCroock:
There are a few side effects being run which also increases size, removing these would not be ideal, we could however change it so that the first ever `createElement`/`render` calls an init function that does these side-effects but I would consider that out of scope for this PR

davidbailey00:
Update on that - seems that Webpack 5 beta produces code output of the same size either way 🎉

JoviDeCroock:
Thanks a lot for this @davidbailey00 great optimization for every user!

davidbailey00:
@JoviDeCroock you're most welcome!!

PR #2010, opened 2019-10-14
Properly flush nested setState callbacks (-6 B) 

andrewiggins:
Previously, we would synchronously flush nested renderCallbacks (i.e. setState and forceUpdate callbacks). This behavior meant that nested setState calls would flush before the state was properly applied in diff, meaning DOM, props, etc. wouldn't have been updated for that setState call when the callback was run (though the state would be correct due to a shortcut we took).

This PR fixes this behavior by only invoking the renderCallbacks for the current setState call synchronously. Nested setState callbacks are queued and executed when those setState calls are applied.

Also includes a small change to make `setState` and `forceUpdate` a little more similar which I assume helps gzip.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -132,10 +132,8 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 
 			c.base = newVNode._dom;
 
-			while (tmp=c._renderCallbacks.pop()) {
-				if (c._nextState) { c.state = c._nextState; }
-				tmp.call(c);
-			}
+			tmp = c._renderCallbacks; c._renderCallbacks=[];
+			tmp.some(cb => cb.call(c));
 
 			// Don't call componentDidUpdate on mount or when we bailed out via
 			// `shouldComponentUpdate`
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -132,8 +132,9 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 
 			c.base = newVNode._dom;
 
-			tmp = c._renderCallbacks; c._renderCallbacks=[];
-			tmp.some(cb => cb.call(c));
+			tmp = c._renderCallbacks;
+			c._renderCallbacks=[];
+			tmp.some(cb => { cb.call(c); });
 
 			// Don't call componentDidUpdate on mount or when we bailed out via
 			// `shouldComponentUpdate`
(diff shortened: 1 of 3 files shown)


src/diff/index.js
```diff
@@ -132,10 +132,8 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 
 			c.base = newVNode._dom;
 
-			while (tmp=c._renderCallbacks.pop()) {
-				if (c._nextState) { c.state = c._nextState; }
-				tmp.call(c);
-			}
+			tmp = c._renderCallbacks; c._renderCallbacks=[];
```
JoviDeCroock: Nit: let's newline this

src/diff/index.js
```diff
@@ -132,10 +132,8 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
...
+			tmp.some(cb => cb.call(c));
```
JoviDeCroock: I would prefer not returning here since this could have weird consequences if they really return a value since `.some` would bailout on first return that can be casted to `true`

andrewiggins: Right! Good call.

src/diff/index.js
```diff
@@ -132,10 +132,9 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 
 			c.base = newVNode._dom;
 
-			while (tmp=c._renderCallbacks.pop()) {
-				if (c._nextState) { c.state = c._nextState; }
-				tmp.call(c);
-			}
+			tmp = c._renderCallbacks;
+			c._renderCallbacks=[];
+			tmp.some(cb => { cb.call(c); });
```
kristoferbaxter: Should be safe to omit the curlies here, but won't change the ES5 output.

`tmp.some(cb => cb.call(c));`

andrewiggins: The ES5 output is different. `cb => cb.call(c)` becomes `function (cb) { return cb.call(c); }` whereas `cb => { cb.call(c); }` becomes `function (cb) { cb.call(c); }` ([babel repl](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBAhjAvDYAjJA-FqB0w4A2BAFMAJQBQoks6yamMA3tnoSeQNwwC-QA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2%2Ces2015-loose&prettier=true&targets=&version=7.6.4&externalPlugins=)).

I think we need to curlies so that `.some` doesn't bail out if the callback returns a truthy value [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some). Since these can be user provided callbacks we can't trust the return value.

PR #2011, opened 2019-10-15
Re-purpose renderCallbacks as a general per-component commit ph…

### LINKED ISSUE #648: Post-mount with componentDidMount

### LINKED ISSUE #1223: componentDidMount() hooks of child are called after componentDidUpdate() hook of the parent
version: preact@8.3.1 

I found the child's componentDidMount hook is called after parent's componentDidUpdate hook.

this is the demo: https://codepen.io/xwchris/pen/vVYVWx

1. click the `toggle show` button
2. see the console log

andrewiggins:
Currently the only lifecycle method that is invoked after all work for a diff is done is `componentDidMount`. This PR re-purposes the `_renderCallbacks` queue in each component to be a general "after work" or "commit phase" callback queue. I use this queue to queue up several kinds of methods:

1. componentDidMount
2. componentDidUpdate
3. setState callbacks
4. forceUpdate callbacks

I believe all of these methods are intended to be invoked after a diff is completed and all dom work is done. This PR makes that true! And the best part, only for 11B!

I'm hoping that we can further reuse this queue for the `useLayoutEffect` hook since this hook is basically just a way to directly "hook" into this mechanism (running a callback synchronously after a diff is complete), but that'll come in another PR.

I included test from #884 (which previously tried to add this behavior for cDU) and fixed the test from #506 to verify the changes.

Fixes #648
Fixes #1223


src/component.js
```diff
@@ -104,17 +104,17 @@ export function getDomSibling(vnode, childIndex) {
...
 function renderComponent(component) {
 	let vnode = component._vnode,
 		oldDom = vnode._dom,
 		parentDom = component._parentDom;
 
 	if (parentDom) {
-		let mounts = [];
-		let newDom = diff(parentDom, vnode, assign({}, vnode), component._context, parentDom.ownerSVGElement!==undefined, null, mounts, oldDom == null ? getDomSibling(vnode) : oldDom);
-		commitRoot(mounts, vnode);
+		let commitQueue = [];
```
andrewiggins: Do people like this rename of `mounts` to `commitQueue`? Since the `mounts` array isn't just used for `componentDidMount` anymore I figured it probably needed a new name that better communicated that.

src/diff/index.js
```diff
@@ -101,10 +109,13 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 				if (c.componentWillUpdate!=null) {
 					c.componentWillUpdate(newProps, c._nextState, cctx);
 				}
-			}
 
-			oldProps = c.props;
-			oldState = c.state;
+				if (c.componentDidUpdate!=null) {
+					c._renderCallbacks.push(() => {
+						c.componentDidUpdate(oldProps, oldState, snapshot);
```
andrewiggins: Note, I've moved the queuing of `cDU` up here to reuse the `isNew` check and to also no longer need to check `oldProps == null` as a way to determine if `sCU` returned `false`. Plus it lives with the rest of the lifecycle methods so I think reasoning about lifecycle methods will be a little easier.

src/diff/index.js
```diff
@@ -161,16 +166,17 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
...
-export function commitRoot(mounts, root) {
-	let c;
-	while ((c = mounts.pop())) {
+export function commitRoot(commitQueue, root) {
+	commitQueue.some(c => {
 		try {
-			c.componentDidMount();
+			commitQueue = c._renderCallbacks;
+			c._renderCallbacks = [];
+			commitQueue.some(cb => { cb.call(c); });
```
andrewiggins: @developit, it seems back in #648, you [expressed perf concerns about a general callback queue](https://github.com/preactjs/preact/issues/648#issuecomment-295326760), which I think is what I've implemented here. Any concerns with this implementation?

test/browser/components.test.js
```diff
@@ -1588,21 +1590,14 @@ describe('Components', () => {
...
+		// Statically create X element
 		const A = <X />;
 
 		class App extends Component {
 			constructor(props) {
 				super(props);
 				this.state = { i: 0 };
-			}
-
-			componentDidMount() {
```
andrewiggins: Syncrhonously calling rerender inside of `cDM` doesn't effectively replicate how Preact's diffing algorithm works (we batch updates and diff them later). Plus it causes an infinite loop so I rewrote this test.

test/browser/keys.test.js
```diff
@@ -372,21 +372,21 @@ describe('keys', () => {
 		ops = [];
 		render(<Foo moved={false} />, scratch);
 
-		expect(ops).to.deep.equal(['Mount Stateful2', 'Mount Stateful1']);
+		expect(ops).to.deep.equal(['Mount Stateful1', 'Mount Stateful2']);
```
andrewiggins: Previously Preact called `cDM` for siblings in reverse order. Now we call `cDM` for siblings from top to bottom. Since `cDM` is invoked after all DOM operations are done, I don't think this order matters as much, but for what its worth, our order matches React now ([codesandbox](https://codesandbox.io/s/react-lifecycle-render-callback-ordering-rmvwp)).

This sibling re-ordering also impacted the `should handle hoisted component vnodes without DOM` test.

src/component.js
marvinhagemeister: +1 for the name change. It better reflects the actual thing it's used for 👍

andrewiggins:
Thanks! :smile: 

@JoviDeCroock I think `cDU` always fired when props change (see [this lifecycle test](https://github.com/preactjs/preact/blob/81ac3c4aee57a8a28dfa7223bcf2aabd1cd0456e/test/browser/lifecycles/lifecycle.test.js#L115)). And unmounted components don't go through diff when they are unmounted so `cDU` won't get invoked when unmounting ([relevant lifecycle test](https://github.com/preactjs/preact/blob/81ac3c4aee57a8a28dfa7223bcf2aabd1cd0456e/test/browser/lifecycles/lifecycle.test.js#L195)). So I think this PR maintains the current behavior.

PR #2012, opened 2019-10-17
Store components with layout effects in a global queue and run during  (+8 B)

### LINKED ISSUE #1893: UseLayoutEffect inconsistent behavior
Assuming the `useLayoutEffect` should fire _after_ DOM is painted, there's a bug where the elements are not yet flushed to DOM for the initial invocation of `useLayoutEffect` when it's used in a component that's rendered as a nested child:

Case 1:
```js
return <CompoUsingLayoutEffect/>
```
Case 2:
```js
return <div>
	<CompoUsingLayoutEffect/>
</div>
```

In case 2, the `useLayoutEffect` is initially fired _before_ the DOM is painted. While in case 1, it works as expected.

Consecutive executions of `useLayoutEffect` are called after the DOM is painted.

Best if you check the repro yourself:

Preact: https://codesandbox.io/s/agitated-morse-b6zrp?expanddevtools=1
React: https://codesandbox.io/s/hardcore-hertz-p5up6?expanddevtools=1

In the repros above, all textareas should be initially resized to fit the content, but in Preact the first textarea is initially set to `height: 0` because the `scrollHeight` is read before the `textarea` is in the DOM (check the console for `0 false` which indicates the element isn't in the `document`).

Preact `10.0.0-rc.1`

andrewiggins:
I ended up with a solution pretty similar to #1974 and #1992 so thanks to @JoviDeCroock for his work!

I originally wanted to try just queuing the layout effects directly to an array, but it ended up not working out.

Fixes #1893


andrewiggins:
Yeah, I need to do that. Was looking into related issues last night. Will report back here

JoviDeCroock:
I think this PR hits the same timings as mine, that's why I asked. Maybe I'm missing a detail or something

andrewiggins:
I copied the Textarea bug repo into the demo app and ran it against this branch and all 4 text areas were properly sized! I think this PR does fire layout effects at the same time as #1974 but for the most part leaves `useEffect` untouched so those tests don't change.

image

andrewiggins:
I'll address the build failure when other PRs go in. One thing at time lol. I think I have branch that is 3 branches deep off of master hahaha

JoviDeCroock:
So when I refresh with CPU slowdown it had a delay mainly seeing of that’s still the case. If not you found a better timing.

hooks/test/browser/combinations.test.js
```diff
@@ -254,19 +254,27 @@ describe('combinations', () => {
 			'effect outer call <span>hello 1</span>'
 		]);
 
+		// NOTE: this order is (at the time of writing) intentionally different from
```
JoviDeCroock: We can support this if we want. I had it in my PR by doing it separately

robertknight:
Is there anything more that you want to do before this is merged @andrewiggins?  I see there is a skipped test, but the changes in this PR look valuable to me even before that is resolved.

hooks/test/browser/combinations.test.js
andrewiggins: Yup. I think I'd like to see if its necessary though. I think we can save some bytes (and possibly runtime perf for large apps) if we only have to loop through this array once instead of twice (once to run all cleanups, again to run all effects). If someone reports an issue where they have a valid use case for React's behavior, we can look at doing the two loops again.

andrewiggins:
Feel free to leave more comments. Just ping me on Slack if you do.

PR #2015, opened 2019-10-17
Fix internal vnode typings

pmkroeker:
I think #1979 causes some regression in the internal VNode typings.

The internal VNode type incorrectly extended preact.VNode.

The issue showed up in my projects webpack build. The below changes fix the issues.


andrewiggins:
Oh interesting. Does your TS build look at internal.d.ts?

pmkroeker:
I guess it must. I changed the typings in my node_modules and it ran through fine. VSCode also showed an error on the VNode line.

andrewiggins:
Oh okay, I think I see it now. This fails when using `preact/compat` and it appears that our TS tests for `preact/compat` don't run on build like our other tests, which is how this got through.

To be clear, my hesitation is that, internally, `VNode.type` is `null` sometimes and `VNode.props` isn't an always object so those internal types are correct in spirit (obviously still need to figure out how to express that properly in TypeScript though since what I did in #1979 doesn't compile).

Regardless, errors in internal types should never impact consumers, so why that's happening at all is concerning...

Issue #2020, opened 2019-10-17
useImperativeHandle function isn't invoked after first render


lorenzomigliorero:
Unlike React, the `console.log` function isn't invoked after first render.

```json
"preact": "^10.0.0-rc.3",
"preact-render-to-string": "^5.0.7",
```

```javascript
{
    resolve: {
      alias: {
        react: 'preact/compat',
        'react-dom': 'preact/compat',
        'react-dom/server': 'preact-render-to-string',
      },
    },
  }
```

```javascript
import React, { useImperativeHandle } from 'react';

const Bar = React.forwardRef((props, ref) => {
  useImperativeHandle(ref, () => ({ foo: 100 }));
  return null;
});

export default () => <Bar ref={console.log} />;
```

PR #2021, opened 2019-10-17
(fix) - call function when ref is one

JoviDeCroock:
fixes: https://github.com/preactjs/preact/issues/2020 (--> next year am I right?)


hooks/src/index.js
```diff
@@ -142,7 +142,8 @@ export function useImperativeHandle(ref, createHandle, args) {
 
 function bindHandles(handles) {
 	handles.some(handle => {
-		if (handle.ref) handle.ref.current = handle.createHandle();
+		if (typeof handle.ref === 'function') handle.ref(handle.createHandle());
+		else if (handle.ref) handle.ref.current = handle.createHandle();
```
marvinhagemeister: FYI: This looks pretty similar to `applyRef` in core. There is potential for golfing in the PR where hooks are merged into core :tada:

PR #1865, opened 2019-08-13
Reproduce issue #1813 in demo

aduh95:
It seems there is some kind of race condition here, because sometimes the loading elements disappear, sometimes they don't.

This also shows a problem with nested Suspense components.

#### Expected HTML output (`body` tag)

```html
<body>
  <script type="text/javascript" src="/main.js"></script>
  <div>DropZone (component #0)</div>
  <div>
    <main>
      <div>
        GenerateComponents (component #1)
        <div>Lazy loaded sub component</div>
      </div>
      <div>AddNewComponent (component #2)</div>
    </main>
    <aside>
      <section>
        <div>
          GenerateComponents (component #3)
          <div>Lazy loaded sub component</div>
        </div>
        <div>AddNewComponent (component #4)</div>
      </section>
    </aside>
  </div>
  <footer>Footer here</footer>
</body>
```

#### Actual HTML output (`body` tag)

```html
<body>
  <script type="text/javascript" src="/main.js"></script>
  <div class="Editor">
    <main>
      <div>
        GenerateComponents (component #1)
        <div>Lazy loaded sub component</div>
      </div>
    </main>
    <aside>
      <section>
        <div>
          GenerateComponents (component #3)
          <div>Lazy loaded sub component</div>
        </div>
      </section>
    </aside>
  </div>
  <footer>Footer here</footer>
  <div>DropZone (component #0)</div>
  <div class="Editor">
    <main>
      <div>Loading...</div>
      <div>
        GenerateComponents (component #1)
        <div>Lazy loaded sub component</div>
      </div>
      <div>AddNewComponent (component #2)</div>
    </main>
    <aside>
      <section>
        <div>Loading...</div>
        <div>
          GenerateComponents (component #3)
          <div>Lazy loaded sub component</div>
        </div>
        <div>AddNewComponent (component #4)</div>
      </section>
    </aside>
  </div>
  <footer>Footer here</footer>
</body>
```

The `<div>Loading...</div>` node should be removed from the DOM, and the `<div class="Editor">`, `<footer>` nodes should not be duplicated.

Refs: https://github.com/preactjs/preact/pull/1859
Refs: https://github.com/preactjs/preact/issues/1813

PR #2025, opened 2019-10-18
Support re-suspending

### LINKED ISSUE #1738: lazy/Suspense behavior different with react
```tsx
// index.tsx
import { h, render }      from 'preact';
import { lazy, Suspense } from 'preact/compat';

const Outer = lazy(() => import('./Outer'));

render(
  <Suspense fallback={<div>loading outer...</div>}>
    <App/>
  </Suspense>,
  document.getElementById('app')!,
);
```

```tsx
// Outer.tsx
import { FunctionComponent, h } from 'preact';
import { lazy, Suspense }       from 'preact/compat';

export const Outer: FunctionComponent = () => {

  const Inner = lazy(() => import('./Inner'));

  return (
    <div>
      <div>Outer</div>
      {/*<Suspense fallback={<div>loading inner...</div>}>*/}
        <Inner/>
      {/*</Suspense>*/}
    </div>
  );
};

export default Outer;
```

```tsx
// Inner.tsx
import { FunctionComponent, h } from 'preact';

export const Inner: FunctionComponent = () => <div>Inner</div>;

export default Inner;
```

With react, I don't need that `<Suspense>` in `Outer.tsx` to make the code work. But with `preact/compat`, if I comment the `<Suspense>` in `Outer.tsx` like above, the top-level (in `index.tsx`) `<Suspense>` won't disappear after the `Inner.tsx` loaded, and sometimes the page will crash.

I'm not sure if this difference is by design or not since I can't found related info in changelog.

### LINKED ISSUE #1813: [10.0.0-rc.0] Suspense's fallback not being unmounted when it should

andrewiggins:
Reviving #1859 with a test that reproduces the bug. @sventschui also noted that we need to prevent tracking the same promise twice. Looking to see if we need a test for that

Fixes #1813 
Fixes #1738

### DIFF
--- a/demo/index.js
+++ b/demo/index.js
@@ -21,6 +21,7 @@ import { initDebug } from 'preact/debug/src/debug';
 // import Redux from './redux';
 // import TextFields from './textFields';
 // import ReduxBug from './reduxUpdate';
+import SuspenseRouterBug from './suspense-router';
 
 let isBenchmark = /(\/spiral|\/pythagoras|[#&]bench)/g.test(window.location.href);
 if (!isBenchmark) {
@@ -79,6 +80,7 @@ class App extends Component {
 						<Link href="/suspense" activeClassName="active">Suspense / lazy</Link>
 						<Link href="/textfields" activeClassName="active">Textfields</Link>
 						<Link href="/reduxBug/1" activeClassName="active">Redux Bug</Link> */}
+						<Link href="/suspense-router" activeClassName="active">Suspense Router Bug</Link>
 					</nav>
 				</header>
 				<main>
@@ -111,6 +113,7 @@ class App extends Component {
 						<Redux path="/redux" />
 						<TextFields path="/textfields" />
 						<ReduxBug path="/reduxBug/:start" /> */}
+						<SuspenseRouterBug path="/suspense-router" />
 					</Router>
 				</main>
 			</div>
--- a/demo/index.js
+++ b/demo/index.js
@@ -2,25 +2,25 @@ import { createElement, render, Component, Fragment } from 'preact';
 // import renderToString from 'preact-render-to-string';
 import './style.scss';
 import { Router, Link } from 'preact-router';
-import Pythagoras from './pythagoras';
-import Spiral from './spiral';
-import Reorder from './reorder';
-import Todo from './todo';
-import Fragments from './fragments';
-import Context from './context';
+// import Pythagoras from './pythagoras';
+// import Spiral from './spiral';
+// import Reorder from './reorder';
+// import Todo from './todo';
+// import Fragments from './fragments';
+// import Context from './context';
 import installLogger from './logger';
-import ProfilerDemo from './profiler';
-import KeyBug from './key_bug';
-import StateOrderBug from './stateOrderBug';
-import PeopleBrowser from './people';
-import StyledComp from './styled-components';
+// import ProfilerDemo from './profiler';
+// import KeyBug from './key_bug';
+// import StateOrderBug from './stateOrderBug';
+// import PeopleBrowser from './people';
+// import StyledComp from './styled-components';
 import { initDevTools } from 'preact/debug/src/devtools';
 import { initDebug } from 'preact/debug/src/debug';
-import DevtoolsDemo from './devtools';
-import SuspenseDemo from './suspense';
-import Redux from './redux';
-import TextFields from './textFields';
-import ReduxBug from './reduxUpdate';
+// import DevtoolsDemo from './devtools';
+// import SuspenseDemo from './suspense';
+// import Redux from './redux';
+// import TextFields from './textFields';
+// import ReduxBug from './reduxUpdate';
 
 let isBenchmark = /(\/spiral|\/pythagoras|[#&]bench)/g.test(window.location.href);
 if (!isBenchmark) {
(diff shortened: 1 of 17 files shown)


aduh95:
I took the liberty to test this PR using the test case I made in #1865, unfortunately it seems the issue is still there...

I have merged this branch into master, `npm run build`, and then `npm link`ed it; then I run a dev server using this `App.js` file:

```jsx
import { h, Component } from "preact";
import { Suspense, lazy } from "preact/compat";

function Loading() {
  return <div>Loading...</div>;
}

const pause = timeout =>
  new Promise(d => setTimeout(d, timeout), console.log(timeout));

const DropZone = lazy(() =>
  pause(Math.random() * 1000).then(() => import("./aduh95/dropzone.js"))
);
const Editor = lazy(() =>
  pause(Math.random() * 1000).then(() => import("./aduh95/editor.js"))
);
const AddNewComponent = lazy(() =>
  pause(Math.random() * 1000).then(() => import("./aduh95/addnewcomponent.js"))
);
const GenerateComponents = lazy(() =>
  pause(Math.random() * 1000).then(() =>
    import("./aduh95/component-container.js")
  )
);

  return (
      <Suspense fallback={<Loading />}>
        <DropZone appearance={0} />
        <Editor title="APP_TITLE">
          <main>
            <Suspense fallback={<Loading />}>
              <GenerateComponents appearance={1} />
            </Suspense>
            <AddNewComponent appearance={2} />
          </main>
          <aside>
            <section>
              <Suspense fallback={<Loading />}>
                <GenerateComponents appearance={3} />
              </Suspense>
              <AddNewComponent appearance={4} />
            </section>
          </aside>
        </Editor>

        <footer>Footer here</footer>
      </Suspense>
  );
}
```

> The subcomponents can be found [on my fork](https://github.com/aduh95/preact/tree/bugfix/suspense-fallback-mounted-twice/demo/aduh95) in case you want to take a look at it.

Expected result:
```html
  <div>DropZone (component #0)</div>
  <div>
    <main>
      <div>
        GenerateComponents (component #1)
        <div>Lazy loaded sub component</div>
      </div>
      <div>AddNewComponent (component #2)</div>
    </main>
    <aside>
      <section>
        <div>
          GenerateComponents (component #3)
          <div>Lazy loaded sub component</div>
        </div>
        <div>AddNewComponent (component #4)</div>
      </section>
    </aside>
  </div>
  <footer>Footer here</footer>
```

Actual result:
```html
<div>Loading...</div>
<div>DropZone (component #0)</div>
<div class="Editor">
  <main>
    <div>Loading...</div>
    <div>
      GenerateComponents (component #1)
      <div>Lazy loaded sub component</div>
    </div>
    <div>AddNewComponent (component #2)</div>
  </main>
  <aside>
    <section>
      <div>Loading...</div>
      <div>Loading...</div>
      <div>
        GenerateComponents (component #3)
        <div>Lazy loaded sub component</div>
      </div>
      <div>AddNewComponent (component #4)</div>
    </section>
  </aside>
</div>
<footer>Footer here</footer>
```

As you can see, not all the `Loading` components get unmounted, although it clearly depends on which Promise resolves first, there seems to have some race condition.

Anyway, thanks for your work, I hope my use-case is constructive and will help resolve the issue.

andrewiggins:
@aduh95 Thanks for the info! I checked in your demo directly into my branch (I called it "nested-suspense") so I could play with it. However, I couldn't get the problem to resurface. Every time I loaded the page all the "Loading..."s disappeared. Could you have a look and see if I did anything wrong? I copied the source from your repro directly so I don't think there should be any typos...

You mention you `npm link`ed the code from this PR. I wonder if perhaps that didn't work cuz glancing at our demo's webpack.config.js, it seems `preact` is hardcoded to be the built code in the same repository as the demo (it uses relative paths). I don't think it goes through `node_modules` at all. So maybe that's why it didn't fix it?

Thanks for your help!

aduh95:
@andrewiggins you are right, that works perfectly on the demo! Thank you for the good work :100:

demo/index.js
```diff
@@ -2,25 +2,27 @@ import { createElement, render, Component, Fragment } from 'preact';
 // import renderToString from 'preact-render-to-string';
 import './style.scss';
 import { Router, Link } from 'preact-router';
-import Pythagoras from './pythagoras';
-import Spiral from './spiral';
-import Reorder from './reorder';
-import Todo from './todo';
-import Fragments from './fragments';
-import Context from './context';
+// import Pythagoras from './pythagoras';
```
cristianbote: Uh oh, are these needed to be commented out? Were they messing up with your demo?

andrewiggins: Whoops! I commented these out because debugging the 7MB demo bundle brought the my laptop's Chrome DevTools to a standstill (not entirely but it was very difficult to debug). So I commented out the other demos to make it easier to debug. I'll add them back.

In other PR I'd like to code-split all these demos behind lazy and Suspense, but we need a functioning Suspense component first 😜

andrewiggins: Removed the comments

PR #2026, opened 2019-10-20
Remove useless return from eventProxy

jridgewell:
I don't think there's any event handler that expects a return value, so I don't think this is needed.


sventschui:
Although it is bad practise (in my opinion) it is allowed to prevent default browser behaviour with a return false: https://stackoverflow.com/a/128966

For compatibility reasons I’d vote to keep the return

cristianbote:
Hey @jridgewell so with @sventschui addressing the compatibility reason, I think it's fair enough to consider this as a no-go. Compatibility is a big reason people do use `preact`. Our main slogan was `closer to the metal` since we are big on dom compatibility.

Thank you so much for looking into this! 🙇

jridgewell:
@cristianbote: Note that no browsers respect the return value.

cristianbote:
Hey @jridgewell thanks for clarifying that and also @marvinhagemeister for reopening this 😄!

So, the takeaway are these:
* If you add an event listener as a property, `dom.onclick = () => false;` the default behaviour is prevented.
Example:
```html
<form onsubmit="return false">
  <input type="submit" value="click" />
</form>
```
* If you add the event listener via `addEventListener` and return the value that does not prevent the default behaviour. You would either need to call the event methods or set another property on the event itself, named `returnValue`;
Example:
```js
form.addEventListener('submit', e => { e.returnValue = false; });
```

So the return does get respected by the browsers if you set the listener as a property.

Since we're basically never adding the events listeners as properties, I think we _can_ remove the return statement. Thanks for pointing it out and sorry it tooks us a bit of back and forth. Really appreciate the effort! 👍

marvinhagemeister:
@cristianbote Awesome, thank you so much for getting to the bottom of this :raised_hands:

jridgewell:
Thanks for investigating @cristianbote

PR #2030, opened 2019-10-21
Add support for null fallbacks in Suspense (+3 B)

### LINKED ISSUE #1969: Cryptic error when using preact/compat
I just tried to use `preact/compat` on my quite complex project. It was using Preact 8.x before, but when hooks came out we decided to temporarily move back to React. I basically just installed the new preact version and aliased it through `webpack` and then ran my app again. It partially renders and I'm afraid this might have something to do with the partial support for `<Suspense />`, but I'm getting this cryptic error:

```
Uncaught (in promise) TypeError: Cannot read property '__c' of null
    at Object.MOxe.preact__WEBPACK_IMPORTED_MODULE_0__.options.unmount (hooks.module.js:1)
    at D (preact.module.js:1)
    at r (compat.module.js:1)
```

For some reason source mapping also doesn't work for me with `preact/compat` which makes it very hard to debug this by myself.

andrewiggins:
Fixes #1969


compat/src/suspense.js
```diff
@@ -65,7 +65,9 @@ Suspense.prototype._childDidSuspend = function(promise) {
 		c._suspensions.pop();
 
 		if (c._suspensions.length == 0) {
-			unmount(c.props.fallback);
+			if (c.props.fallback) {
```
sventschui: Awesome! You might want to add a comment why this check is needed

PR #2034, opened 2019-10-22
added css types and shims folder(react-is WIP for discussion)

jeremy-coleman:
only significant change

```js
import * as CSS from 'csstype';

export = preact;
export as namespace preact;

import { JSXInternal } from "./jsx";

declare namespace preact {
	export import JSX = JSXInternal;

	//id prefer to put on the JSX namespace , but compat reasons here. both?meh
	export interface CSSProperties extends CSS.Properties<string | number> {}
```

@andrewiggins mentioned in slack it would be a dependency, but i wouldn't consider it as a dep. I copy/pasted it into the shims folder also


mnkhouri:
FYI my PR to add CSSType is a partial duplicate of this one: https://github.com/preactjs/preact/pull/1797

jeremy-coleman:
@mnkhouri closing this in preference of your pull request.

PR #2039, opened 2019-10-25
Remove unused mount check in context Provider (-6 B)

andrewiggins:
The mount check in Provider sCU was redundant because `renderComponent` already checks a component is mounted before attempting to rerender it. Further whenever a consumer is unmounted we remove it from the sub list.

I added some tests as well just to double check.

### DIFF
--- a/hooks/test/browser/useContext.test.js
+++ b/hooks/test/browser/useContext.test.js
@@ -242,4 +242,49 @@ describe('useContext', () => {
 			done();
 		});
 	});
+
+	it('should not rerender consumers that have been unmounted', () => {
+		const context = createContext(0);
+		const Provider = context.Provider;
+
+		const Inner = sinon.spy(() => {
+			const value = useContext(context);
+			return <div>{value}</div>;
+		});
+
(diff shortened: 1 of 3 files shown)


hooks/test/browser/useContext.test.js
```diff
@@ -242,4 +242,49 @@ describe('useContext', () => {
...
+			}
+		}
+
+		render(<App />, scratch);
+		expect(scratch.innerHTML).to.equal('<div><div>0</div></div>');
+		expect(Inner).to.have.been.calledOnce;
+
+		act(() => changeValue(1));
+		expect(scratch.innerHTML).to.equal('<div><div>1</div></div>');
+		expect(Inner).to.have.been.calledTwice;
```
cristianbote: This is interesting. I would hope that this would be called just once. Is it because `Inner` spy needs to be reset?

andrewiggins: @cristianbote Correct. I'm not resetting the spy after each render so the call counts are over the entire test

PR #2040, opened 2019-10-25
Inline `coerceToVNode` inside of `toChildArray` (-21 B)

andrewiggins:
Inline `coerceToVNode` into the one place it was used: `toChildArray`

### DIFF
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -86,21 +86,3 @@ export function Fragment(props) {
  * @returns {vnode is import('./internal').VNode}
  */
 export const isValidElement = vnode => vnode!=null && vnode.constructor === undefined;
-
-/**
- * Coerce an untrusted value into a VNode
- * Specifically, this should be used anywhere a user could provide a boolean, string, or number where
- * a VNode or Component is desired instead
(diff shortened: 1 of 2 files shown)


src/create-element.js
```diff
@@ -86,26 +86,3 @@ export function Fragment(props) {
...
-export function coerceToVNode(possibleVNode) {
-	if (possibleVNode == null || typeof possibleVNode === 'boolean') return null;
-	if (typeof possibleVNode === 'string' || typeof possibleVNode === 'number') {
-		return createVNode(null, possibleVNode, null, null);
-	}
-
-	// Clone vnode if it has already been used. ceviche/#57
-	if (possibleVNode._dom!=null || possibleVNode._component!=null) {
-		let vnode = createVNode(possibleVNode.type, possibleVNode.props, possibleVNode.key, null);
-		vnode._dom = possibleVNode._dom;
```
andrewiggins: Note: I removed this line (added in #1484) when moving this function into `toChildArray`. I think the situation that occurred in #1484 no longer occurs because we don't call `coerceToVNode` twice.

@marvinhagemeister can you double check my thinking here?

The people demo still seems to work so I think this change is okay?

PR #2041, opened 2019-10-26
Improve code coverage and fix lint warnings

andrewiggins:
* Add tests for various sub-modules that modify the options object, verifying that they call previous options


compat/src/index.js
```diff
@@ -389,6 +388,10 @@ function setSafeDescriptor(proto, key) {
 		Object.defineProperty(proto, key, {
 			configurable: false,
 			get() { return this['UNSAFE_' + key]; },
+			// This set is only used if a user sets an a lifecycle like cWU
```
marvinhagemeister: Small typo `a an`

debug/src/debug.js
```diff
@@ -58,6 +60,8 @@ export function initDebug() {
 			Expected a valid HTML node as a second argument to render.
 			Received ${parentNode} instead: render(<${vnode.type.name || vnode.type} />, ${parentNode});
 		`);
+
+		if (oldRoot) oldRoot(vnode, parentNode);
```
marvinhagemeister: Ohh this is a good one! Never thought about options page in case of render errors before 👍💯

debug/src/debug.js
```diff
@@ -221,34 +227,33 @@ export function initDebug() {
...
-			(hooks._list || []).forEach(hook => {
-				if (hook._callback && (!hook._args || !Array.isArray(hook._args))) {
-					/* istanbul ignore next */
-					console.warn(
-						`In ${vnode.type.name || vnode.type} you are calling useMemo/useCallback without passing arguments.\n` +
-						`This is a noop since it will not be able to memoize, it will execute it every render.`
-					);
-				}
-			});
+			if (hooks._list && Array.isArray(hooks._list)) {
```
marvinhagemeister: Wondering if we could just do `Array.isArray(hooks._list)`. In theorie it should bail out for falsy cases too.

compat/src/index.js
andrewiggins: Thanks!

debug/src/debug.js
andrewiggins: 😁

debug/src/debug.js
andrewiggins: Good cal! Will update

andrewiggins:
I'm building on top of this PR so going to go ahead and merge it in. Feel free to leave more comments. Just @-mention me if you do!

PR #2048, opened 2019-10-28
Reduce line lengths in core

andrewiggins:
We've got some pretty gnarly long lines in core 😅 Some of these don't fit on my laptop screen lol. This PR simply re-indents some long lines so that they fit on my laptop screen lol.

Hopefully this is also helpful in the long run as it'll make diffs easier to view since they'll be less line wrapping. Also some of our more gnarly conditions may be easier to understand now that they are broken up onto multiple lines.

I didn't really consistently apply any heuristic here. Just ran through the files and reformatted ones that bothered me. If we think this is useless we can just close this PR.

Verified this does not change the byte size and so doesn't affect our output.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,12 @@
     "*.ts",
     "dist"
   ],
+  "prettier": {
+    "singleQuote": true,
+    "trailingComma": "none",
+    "useTabs": true,
+    "tabWidth": 2
+  },
   "files": [
     "src",
     "dist",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -70,6 +70,12 @@
     "*.ts",
     "dist"
   ],
+  "prettier": {
+    "singleQuote": true,
+    "trailingComma": "none",
+    "useTabs": true,
+    "tabWidth": 2
+  },
```
ForsakenHarmony: Did you mean to include this?

Might make sense to add a dev dependency in that case

marvinhagemeister: +1 for keeping it and adding `prettier` as a `devDependency`. I'm mainly working on 13" laptops and it having shorter lines makes this much more comfortable for me👍

andrewiggins:
Superseded by #2065

PR #2049, opened 2019-10-28
Simplify ternary expressions (-7 B)

andrewiggins:
After working on #2048, I realized we had some gnarly ternary expressions that would a little hard to decipher. I played around with rewriting some into normal if statements hoping terser would properly minify them for us and it turns it did even better and saved us some bytes! A before and after comparison is provided in the comments of the PR for those curious.

Summary:
f6b39b5 Use if statement in setStyle for readability (+0 B)
8ff18fc Simplify className conditionals (-2 B)
f04da34 Simplify setState/_nextState condintional (-2 B)
f638098 Simplify update condition (-1 B)
448dbe7 Simplify nextState assignment in gDSFP handling (-2 B)

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -32,8 +32,11 @@ Component.prototype.setState = function(update, callback) {
 		s = this._nextState = assign({}, this.state);
 	}
 
-	// if update() mutates state in-place, skip the copy:
-	if (typeof update !== 'function' || (update = update(s, this.props))) {
+	if (typeof update == 'function') {
+		update = update(s, this.props);
+	}
+
+	if (update) {
 		assign(s, update);
 	}
 
--- a/src/component.js
+++ b/src/component.js
@@ -25,9 +25,12 @@ export function Component(props, context) {
  */
 Component.prototype.setState = function(update, callback) {
 	// only clone state when copying to nextState the first time.
-	let s =
-		(this._nextState !== this.state && this._nextState) ||
-		(this._nextState = assign({}, this.state));
+	let s;
+	if (this._nextState !== this.state) {
+		s = this._nextState;
+	} else {
+		s = this._nextState = assign({}, this.state);
+	}
 
 	// if update() mutates state in-place, skip the copy:
 	if (typeof update !== 'function' || (update = update(s, this.props))) {
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -85,10 +85,12 @@ export function diff(
 				c._nextState = c.state;
 			}
 			if (newType.getDerivedStateFromProps != null) {
+				if (c._nextState == c.state) {
+					c._nextState = assign({}, c._nextState);
+				}
+
 				assign(
-					c._nextState == c.state
-						? (c._nextState = assign({}, c._nextState))
-						: c._nextState,
+					c._nextState,
 					newType.getDerivedStateFromProps(newProps, c._nextState)
 				);
 			}
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -55,13 +55,13 @@ function setStyle(style, key, value) {
  * @param {boolean} isSvg Whether or not this DOM node is an SVG node or not
  */
 function setProperty(dom, name, value, oldValue, isSvg) {
-	name = isSvg
-		? name === 'className'
-			? 'class'
-			: name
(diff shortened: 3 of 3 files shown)


src/diff/props.js
```diff
@@ -35,13 +35,17 @@ function setStyle(style, key, value) {
...
+		style[key] = '';
+	}
 	else {
-		style[key] =
-			typeof value === 'number' && IS_NON_DIMENSIONAL.test(key) === false
-				? value + 'px'
-				: value == null
-					? ''
-					: value;
+		style[key] = value;
```
andrewiggins: Terser compiles this into the same ternary we had manually written. I just found this to be much easier to read.

src/diff/props.js
```diff
@@ -54,13 +58,14 @@ function setStyle(style, key, value) {
...
-			? 'className'
-			: name;
+	if (isSvg) {
+		if (name === 'className') {
+			name = 'class';
+		}
+	}
+	else if (name === 'class') {
+		name = 'className';
+	}
```
andrewiggins: Original:
```js
l = i ? ("className" === l ? "class" : l) : "class" === l ? "className" : l;
```

Updated:
```js
i ? "className" === l && (l = "class") : "class" === l && (l = "className");
```

The original ternary could be expanded to:
```js
if (isSvg) {
	if (name === 'className') {
		name = 'class';
	}
	else {
		name = name;
	}
}
else {
	if (name === 'class') {
		name = 'className';
	}
	else {
		name = name;
	}
}
```

The new expression simplifies it by removing the redundant `name=name` parts and allowing terser to do its magic.

src/component.js
```diff
@@ -25,12 +25,19 @@ export function Component(props, context) {
...
-	let s =
-		(this._nextState !== this.state && this._nextState) ||
-		(this._nextState = assign({}, this.state));
+	let s;
+	if (this._nextState !== this.state) {
+		s = this._nextState;
+	}
+	else {
+		s = this._nextState = assign({}, this.state);
+	}
```
andrewiggins: Original:
```js
var u = (this.__s !== this.state && this.__s) || (this.__s = s({}, this.state));
```

Updated:
```js
var u;
u = this.__s !== this.state ? this.__s : (this.__s = s({}, this.state));
```

My guess is that since many of our minified functions begin with `var` statements the updated one compresses in gzip better. And perhaps the conversion to the ternary compresses better saving us an additional byte. Brotli stays the same size with this change

src/component.js
```diff
@@ -25,12 +25,19 @@ export function Component(props, context) {
...
+
+	if (typeof update == 'function') {
+		update = update(s, this.props);
+	}
 
-	// if update() mutates state in-place, skip the copy:
-	if (typeof update!=='function' || (update = update(s, this.props))) {
+	if (update) {
```
andrewiggins: Original:
```js
("function" != typeof n || (n = n(u, this.props))) && s(u, n);
```

Updated:
```js
"function" == typeof n && (n = n(u, this.props));
n && s(u, n);
```

This change is probably a fluke as to how it saves us a byte. Perhaps the double use of `&&`? I think it be worth keeping cuz it expresses our intent using idiomatic JS and enables terser to minify this as it sees fit

src/diff/index.js
```diff
@@ -82,10 +82,12 @@ export function diff(
...
 			if (newType.getDerivedStateFromProps!=null) {
+				if (c._nextState == c.state) {
+					c._nextState = assign({}, c._nextState);
+				}
+
 				assign(
-					c._nextState == c.state
-						? (c._nextState = assign({}, c._nextState))
-						: c._nextState,
+					c._nextState,
```
andrewiggins: Original:
```js
s(
  p.__s == p.state ? (p.__s = s({}, p.__s)) : p.__s,
  P.getDerivedStateFromProps(g, p.__s)
)
```
Updated:
```js
p.__s == p.state && (p.__s = s({}, p.__s)),
s(p.__s, P.getDerivedStateFromProps(g, p.__s))
```

src/diff/props.js
cristianbote: Love this!

andrewiggins:
FYI, I rebased this on top of `prettier-2`

PR #2050, opened 2019-10-28
Queue layoutEffects in component render callbacks array (+0 B)

andrewiggins:
Removes 2 array allocations: 1 global array and 1 array per component using hooks! Also, I suspect that this would save us some bytes if we bring hooks into core since we are getting more reuse out of an existing construct instead of introducing new ones :grin:

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -166,7 +166,14 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 	return newVNode._dom;
 }
 
+/**
+ * @param {Array<import('../internal').Component>} commitQueue List of components
+ * which have callbacks to invoke in commitRoot
+ * @param {import('../internal').VNode} root
+ */
 export function commitRoot(commitQueue, root) {
+	if (options._commit) options._commit(root, commitQueue);
+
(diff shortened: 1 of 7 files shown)


src/diff/index.js
```diff
@@ -166,7 +166,14 @@ export function diff(parentDom, newVNode, oldVNode, context, isSvg, excessDomChi
 	return newVNode._dom;
 }
 
+/**
+ * @param {Array<import('../internal').Component>} commitQueue List of components
+ * which have callbacks to invoke in commitRoot
+ * @param {import('../internal').VNode} root
+ */
 export function commitRoot(commitQueue, root) {
+	if (options._commit) options._commit(root, commitQueue);
```
andrewiggins: If we need a post commit option, we should probably call it `_committed` following our existing option naming pattern

marvinhagemeister: I don't know off the top of my head, but this is arguable the one option hook where most of the devtools code is based upon. Looking at the changes here it should be fine. Just posting to make a note to myself here.

andrewiggins: Oh true... Hmm one workaround would be to delay the devtools code after a `Promise.then` callback

Issue #2052, opened 2019-10-28
Functional refs behave different from React


mindplay-dk:
I may have discovered a subtle deviation from React's behavior regarding the dispatch of functional refs.

That is, something like `<p ref={ p => doSomething(p) }>` seems to behave different during updates.

I've created an sandbox that prints out the sequence of events:

https://codesandbox.io/s/happy-morning-jyu67

You can toggle between `react` and `preact` by un/commenting the imports at the top of the script.

If I copy the console output from `react` and `preact` and put them side-by-side (and some empty lines to show the missing events) it looks like this:

![image](https://user-images.githubusercontent.com/103348/67670685-bc2b7d00-f974-11e9-8e12-9ba02428b9d4.png)

The big question for me here is, why is `react` dispatching the `ref` callbacks with `null` for elements that *weren't* removed during the update?

Even if I add `key` attributes to these elements, I get the `null` events.

Now, according to the [React docs](https://reactjs.org/docs/refs-and-the-dom.html#callback-refs), it seems to me maybe `preact` is actually right on this point:

> React will call the ref callback with the DOM element when the component mounts, and call it with `null` when it unmounts.

Of course, this doesn't say it *won't* call the ref callback with `null` between updates, but it certainly seems to imply so?

I went back as far as `react` `16.8.0`, and it behaves like this in every release - unfortunately, older versions than that refuse to load in CodeSandbox, so I can't say if the behavior was always like this.

I've tested with `preact` and `preact/compat`, both of which behave the same.

Anyhow, bottom line is, if people rely on callbacks with `null` as a means of cleaning up side-effects, this could definitely lead to problems, right? (I suspect it might account for some of the reported incompatibilities with various libraries intended for React, but I haven't looked at those reports/libraries in-depth to see if they use functional refs.)

marvinhagemeister:
That's very interesting! I'm with you, that this looks like a bug in `react`. The elements aren't unmounted so it seems wrong to call `ref` with `null`.

cc @sebmarkbage @gaearon @acdlite

sebmarkbage:
It’s intensional because it’s a new function identity we don’t know if it’s a different ref all together.

```
<p ref={
  c ? n => doSomething(n)
  : n => doSomethingElse(n)
} />
```

If it’s the same referential identity it’s not called.

Seems like Preact wouldn’t handle changes to the ref callback without doing this?

JoviDeCroock:
Hey @sebmarkbage 

[...]

PR #2055, opened 2019-10-28
(feat) - cleanup changing refs correctly

JoviDeCroock:
Fixes: https://github.com/preactjs/preact/issues/2052

This differentiates in one aspect and that is that we do
- call with null
- call with value

instead of
nullcalls
valuecalls


src/diff/children.js
```diff
@@ -115,7 +115,8 @@ export function diffChildren(
 					(refs || (refs = [])).push(
```
andrewiggins: What's the size if call `applyRef` twice?
```js
if ((j = childVNode.ref) && oldVNode.ref != j) {
   if (!refs) refs = [];
   if (oldVNode.ref) refs.push(oldVNode.ref, null, childVNode);
   refs.push(j, childVNode._component || ... );
}
```
Or something similar...

The if statement in `diffChildren` already checks if the refs are different so ideally we could reuse that `if` and not duplicate it in `applyRef` (if I'm reading this code correctly).

src/diff/index.js
```diff
@@ -381,9 +381,9 @@ function diffElementNodes(
  * @param {any} value
  * @param {import('../internal').VNode} vnode
  */
-export function applyRef(ref, value, vnode) {
+export function applyRef(ref, value, vnode, oldRef) {
```
sventschui: On my phone so maybe I’m a bit lost here but what’s this argument for?

JoviDeCroock: oops, me being dumb good spot 👍

JoviDeCroock:
> Thanks! LGTM 🎉

No no, Thank YOU for the idea on how to reduce size, hero!

PR #2060, opened 2019-10-29
Add missing `_renderCallbacks` mangle configuration

robertknight:
This broke layout effect hooks in production builds of v10.0.3.

We've now had a couple of broken releases due to mismatched mangling configuration. We need to change the build or add a suitable test so that this doesn't happen again. For now, this fixes the immediate problem.

### DIFF
--- a/hooks/mangle.json
+++ b/hooks/mangle.json
@@ -16,6 +16,7 @@
       "$_diff": "__b",
       "$_commit": "__c",
       "$_render": "__r",
+      "$_renderCallbacks": "__h",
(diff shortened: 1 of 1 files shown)


robertknight:
The next release of microbundle supports [passing a file path](https://github.com/developit/microbundle/commit/1d3f8ec1ed65b690bd31fbf311fa311879c9daae) to the mangle.json file, so perhaps we could use that to achieve a centralized configuration.

developit:
@robertknight absolutely

PR #2063, opened 2019-10-29
Add SuspenseList component

prateekbh:
- [x] - enable way for Suspense boundaries to be controlled externally(new hook).
- [x] - create `SuspenseList` using above hook to control its children's rendering order.
- [x] - create forwards mode for `SuspenseList`.
- [x] - create backwards mode for `SuspenseList`.
- [x] - create together mode for `SuspenseList`.
- [x] - create default mode for `SuspenseList`.
- [x] - create TS definitions `SuspenseList`.
- [x] - enable nested `SuspenseList`.
- [x] - create tests for `SuspenseList`.


compat/src/suspense-list.js
```diff
@@ -0,0 +1,72 @@
+import { Component, options } from 'preact';
+import { Suspense } from './suspense';
+
+// Hook for Suspense boundaries to ask for any extra work before rendering suspended children.
+options.__onSuspensionComplete = (vnode, cb) => {
```
andrewiggins: Hmm do we need this to be an option? Could we simplify and save some bytes and move the `_parent._component._modifySuspense` into Suspense directly?

compat/src/suspense-list.js
```diff
@@ -0,0 +1,72 @@
...
+	this._thrillers = [];
+}
+
+// Things we do here to save some bytes but are not proper JS inheritance:
+// - call `new Component()` as the prototype
+// - do not set `Suspense.prototype.constructor` to `Suspense`
+SuspenseList.prototype = new Component();
+
+SuspenseList.prototype.__modifySuspense = function(vnode, cb) {
+	switch (this.props.revealOrder) {
```
andrewiggins: Does replacing this switch with an `if` statement using the nifty `props.revealOrder > 't'` trick save us some bytes? 

Perhaps we could get away with the default/`else` case being `forwards`. Using `SuspenseList` without an explicit revealOrder seems like bug ([React warns on it](https://github.com/facebook/react/blob/0f3838a01b0fda0ac5fd054c6be13166697a113c/packages/react-reconciler/src/ReactFiberBeginWork.js#L2197)) and I'd guess the most common case would be `forwards` so perhaps we can save bytes and just use `forwards` as the default case?

compat/src/suspense-list.js
```diff
@@ -0,0 +1,72 @@
...
+			 * The direction is controlled in render method while creating `_thrillers` itself.
+			 */
+			if (this._thrillers && this._thrillers[0].vnode === vnode) {
+				cb();
+				this._thrillers.shift();
+				/**
+				 * Find and execute all callbacks in order from 2nd position.
+				 * Breaks as soon as a non resolved(cb===null) suspense found.
+				 */
+				this._thrillers.find(thrill => {
```
andrewiggins: Unfortunately, [`find` isn't in our browser matrix](https://kangax.github.io/compat-table/es5/) 😢 since we don't compile to ES6. However, you should be able to emulate this `find` call with `.some` and use the return value to break out of the loop.

For the others you can use `.some` and modify the first matching object in your callback when you find the match and then return `true` to break.

compat/test/browser/suspense-list.test.js
```diff
@@ -0,0 +1,66 @@
...
+	Suspense,
+	lazy,
+	Fragment
+} from 'preact/compat';
+import { setupScratch, teardown } from '../../../test/_util/helpers';
+
+const h = React.createElement;
+/* eslint-env browser, mocha */
+
+describe('suspense-list', () => {
```
andrewiggins: If you are looking for some inspiration, here are the [React SuspenseList tests](https://github.com/facebook/react/blob/master/packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js). Thought they may be useful just to see what scenarios they are testing, not how they are doing. In other words, what the names of their tests and do we want some similar ones. But perhaps that can come later :)

compat/src/suspense-list.js
JoviDeCroock: I would go for

```
if (revealOrder[0] === 'f') ....
```

compat/src/suspense-list.js
```diff
@@ -0,0 +1,72 @@
...
+					thrill.cb();
+				});
+			} else {
+				this._thrillers.find(thrill => thrill.vnode === vnode).cb = cb;
+			}
+			break;
+		case 'together':
+			this._thrillers.find(thrill => thrill.vnode === vnode).cb = cb;
+			if (this._thrillers.every(thriller => thriller.cb)) {
+				this._thrillers.forEach(thrill => thrill.cb());
```
JoviDeCroock: you can save a few bytes by doing `.some(t => { t() })` --> this way a `return` won't be added here

compat/src/suspense-list.js
```diff
@@ -0,0 +1,72 @@
...
+			}
+			break;
+		default:
+			cb();
+	}
+};
+
+SuspenseList.prototype.render = function(props, state) {
```
JoviDeCroock: state is unused here

compat/src/suspense-list.js
```diff
@@ -0,0 +1,72 @@
...
+	// assuming all suspense woul
+	this._thrillers = props.children
+		.filter(child => child.type.name === Suspense.name)
+		.map(vnode => ({
+			vnode,
+			cb: null
+		}));
+	if (props.revealOrder === 'backwards') {
```
JoviDeCroock: I'd go for the `props.revealOrder[0] === 'b'`

compat/src/suspense-list.js
```diff
@@ -0,0 +1,72 @@
...
+			}
+			break;
+		default:
+			cb();
+	}
+};
+
+SuspenseList.prototype.render = function(props, state) {
+	// assuming all suspense woul
+	this._thrillers = props.children
```
JoviDeCroock: Wondering if it wouldn't be better to put these in a `useEffect` since should this happen on every render?

compat/src/suspense-list.js
cristianbote: Wondering if we should go for @jviide's suggestions though:
```
> q = x => (x < "f") + (x < "t")
> q("forwards")
1
> q("backwards")
2
> q("together")
0
```
🤔  such a great small snippet

compat/src/suspense-list.js
prateekbh: Sadly no.

A. That makes it really coupled, but I get it can be worthy of the savings.
B. Suspense isn't the only one who needs this logic. A `SuspenseList` can itself be a inside other other `SuspenseList`

e.g.
```jsx
<SuspenseList>
  <SuspenseList>...<SuspenseList>
  <Suspense>...<Suspense>
</SuspenseList>
```

compat/src/suspense-list.js
prateekbh: we'll always need an if because the along with these values it can be non existing as well

so in my head it'll end up looking like
`if(props.revealOrder){}else{switch q ....}`

cuz q(null)===0.

Thus to keep the balance between byte savings and readability, i'd vote for `if (revealOrder[0] === 'f') ....`

compat/test/browser/suspense-list.test.js
prateekbh: I have tried to keep them quite close for now(sans the warning tests).
Will add more to have full parity

compat/src/suspense-list.js
prateekbh: done

compat/src/suspense-list.js
prateekbh: I am not sure about this, what if one of the element gets added or removed as a UI logic?

compat/src/suspense-list.js
prateekbh: done

compat/src/suspense-list.js
jviide: I chatted with @prateekbh about a parallel approach. If the `_thrillers` array's size is `n` then the worst-case amount of work is something like O(n^2), and some work could be avoided by using a singly linked list and a `Map`. Here's an incomplete sketch, sorry about that 🙂

```js
export function SuspenseList(props) {
  this._head = null;
  this._vnodes = null;
}

SuspenseList.prototype.render = function(props, state) {
  // Points to the first node of a linked list storing the callbacks,
  // where the first node points to the next callback that should be called.
  this._head = null;
  // Maps vnodes to nodes in the linked list.
  this._vnodes = new Map();

  // Build the linked list by iterating through the children in a reverse order
  // (so that _head points to the first node in the end).
  const children = props.children.filter(
    child => child.type.name === Suspense.name
  );
  if (props.revealOrder === "forwards") {
    children.reverse();
  }
  children.some(vnode => {
    this._vnodes.set(vnode, (this._head = { cb: null, next: this._head }));
  });
  return props.children;
};

SuspenseList.prototype.__modifySuspense = function(vnode, cb) {
  const node = this._vnodes.get(vnode);
  node._vnodes.delete(vnode);
  node.cb = cb;

  const order = this.props.revealOrder;
  if (order === "forwards" || order === "backwards") {
    if (node === this._head) {
      consume(this);
    }
  } else if (order === "together") {
    if (this._vnodes.size === 0) {
      consume(this);
    }
  } else {
    cb();
  }
};

// A helper to walk down and consume the linked list as long as there are resolved callbacks.
const consume = list => {
  for (let node; (node = list._head) && node.cb; list._head = node.next) {
    node.cb();
  }
};
```

compat/src/suspense-list.js
prateekbh: done

compat/src/suspense-list.js
prateekbh: why not `this._thrillers.forEach(thrill => {thrill.cb()});`

compat/src/suspense-list.js
andrewiggins: Whoa! Nested `SuspenseList` 🤯 Gotta think more about that

compat/test/browser/suspense-list.test.js
andrewiggins: Another interesting case to consider: re-suspending under a SuspenseList: https://twitter.com/dan_abramov/status/1192077102490824707?s=09

prateekbh: Thats an excellent test

prateekbh: Thanks

prateekbh:
This Component definitely needs @jviide 's Linked list magic but can be a top up request.

jviide:
I was looking through the code and got a sneaking suspicion that there might be a hidden assumption: That all `<Suspense>` elements have a child that suspends. I'm not too familiar with `<Suspense>`, but there probably are scenarios when this assumption might not hold (?).

Turns out that a test case like this fails (or I messed up the test run somehow 😛):
```
it('should work even when a <Suspense> child does not suspend', async () => {
	const Component = getSuspendableComponent('A');

	render(
		<SuspenseList revealOrder="forwards">
			<Suspense fallback={<span>Loading...</span>}>
				<div />
			</Suspense>
			<Suspense fallback={<span>Loading...</span>}>
				<Component />
			</Suspense>
		</SuspenseList>,
		scratch
	); // Render initial state

	rerender();
	expect(scratch.innerHTML).to.eql(`<div></div><span>Loading...</span>`);

	await Component.resolve();
	rerender();
	expect(scratch.innerHTML).to.eql(`<div></div><span>A</span>`);
});
```

prateekbh:
@jviide the solution in my head is as small as fixing a small Boolean value but in your case we should just log that `no promise was thrown, may be you don't really need suspense`

andrewiggins:
Isn't it possible that a component won't suspend until a button is pressed or something? So I wouldn't assume that not suspending on a render is a bug

prateekbh:
@andrewiggins fixed and it is totally possible to do it. Although wouldn't it be more easier to read if the `<suspense>` itself is also conditionally rendered?

because as soon as something will throw a promise inside a `Suspense` all the static things will also get removed for a loader

e.g
```jsx
<Suspense fallback={<span>Loading...</span>}>
  <ComponentThatWillThrowPromiseLater />
  All the static text present here will also go as soon as the component above will throw the promise...
</Suspense>
```

jviide:
@prateekbh @andrewiggins Another scenario that I guess could be common: A component tries to render some remote content that may or may not be already cached locally. If the content is not cached then the render function throws a Promise that resolves when the content is finally fetched. If the content is cached, then the component can render the cached stuff immediately without throwing.

prateekbh:
Yes. thanks for the explanation. Thats indeed a valid use case.
PR includes your given test case already

marvinhagemeister:
@jviide I just merged #2106 and it lead to merge conflicts here :grimacing:

prateekbh:
yep looking at it

PR #2064, opened 2019-10-29
Adding Intergram to the demos list

idoco:
Hey all,

I have built a somewhat popular (790✰) open-source chat widget using Preact and thought it would be relevant to include it in the demos list. This chat widget is used in hundreds of websites and is being loaded by about ~175k unique visitors a month.

I used Preact because I wanted to make my bundle as small as possible so I would not burden the hosting website - [link](https://github.com/idoco/intergram#initial-footprint).

Would appreciate your feedback  🙂


idoco:
Thanks! 😊

developit:
Wow, this is an awesome project that I had no idea about! Nice work @idoco. FWIW you should also add this to the preactjs.com site - maybe [here](https://github.com/preactjs/preact-www/blob/master/content/en/about/demos-examples.md#full-apps)?

PR #2065, opened 2019-10-29
Add automatic formatting via prettier

marvinhagemeister:
Long time coming :tada: 

- [x] Add prettier
- [x] Format all source code
- [x] Apply formatting via git hooks automatically
- [x] Make sure that it has no effect on size :tada: 
- [x] Update eslint config

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -1,12 +1,39 @@
-import { hydrate, render as preactRender, cloneElement as preactCloneElement, createRef, h, Component, options, toChildArray, createContext, Fragment, _unmount } from 'preact';
-import { useState, useReducer, useEffect, useLayoutEffect, useRef, useImperativeHandle, useMemo, useCallback, useContext, useDebugValue } from 'preact/hooks';
+import {
+	hydrate,
+	render as preactRender,
+	cloneElement as preactCloneElement,
+	createRef,
+	h,
+	Component,
+	options,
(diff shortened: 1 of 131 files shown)


package.json
```diff
@@ -36,10 +36,11 @@
     "lint": "eslint src test debug compat hooks test-utils"
   },
   "eslintConfig": {
-    "extends": "developit",
+    "extends": ["developit", "prettier"],
     "settings": {
       "react": {
-        "pragma": "createElement"
+        "pragma": "createElement",
+        "version": "latest"
```
prateekbh: is this really needed in this PR?

marvinhagemeister: Alright, reverted it :+1:

compat/src/index.js
```diff
@@ -186,8 +186,9 @@ let Children = {
 	},
 	only(children) {
 		children = toChildArray(children);
-		if (children.length !== 1)
+		if (children.length !== 1) {
```
prateekbh: wait doesn't the git pre-commit hook not revert this change for you?

andrewiggins:
Oh! I also confirmed on my machine that this PR doesn't change gzip size for any of our modules

compat/src/index.js
ForsakenHarmony: Seems to keep it if it's there

PR #2070, opened 2019-10-30
Improve ref typings for IntrinsicElements.

lukeshiru:
This change adds the corresponding constructor type for all IntrinsicElements in JSX, improving `ref` types, for example the anchor tag now has the HTMLAnchorElement type in it's `ref`.

This was created with a simple JS script that extracts the constructor.name of every element in the 
```javascript
Object.keys(/* Object with HTMLIntrinsicElements */)
  .map(tag => document.createElement(tag))
  .map(
    ({ constructor, tagName }) =>
      `${tagName.toLowerCase()}: HTMLAttributes<${constructor.name}>;`
  )
  .join("\n");
```

After running tests I spotted an issue with the `preact.tsx` test in the `ts` directory that wasn't detected previously due to the lack of typing in IntrinsicElements. A ref can be an Element or null, but that wasn't being considered in the types of the test.

This is my first PR in here, I hope I'm not messing anything up 😄


lukeshiru:
@marvinhagemeister I did a rebase on my side and applied the prettier formatting. Let me know if anything else is missing and I'll update accordingly 😄

PR #2074, opened 2019-10-31
(chore) Added script to sync mangle files

cristianbote:
Alright, let's do this! 😄 

This takes `mangle.json` from root as the... source of truth, so basically, whenever someone would modify the root mangle.json the other mangle files would get updated with this script.

⚠️  Important: This is temporary until `microbundle@0.12.0` lands, which has a way to receive a mangle.json path.

```sh
npm run sync:mangle
```

Let me know what you guys think.


robertknight:
I think an important thing here is to have a lint check that verifies that the mangle files don't unintentionally get out of sync, as that is what bit us several times recently.

As for synchronizing the `mangle.json` files, perhaps that could be done with a script which simply copies the one from the root to the other directories as part of a pre-commit hook and `git add`s them. If the files are already the same, `git add` will just be a no-op.

sync-mangle.js
```diff
@@ -0,0 +1,43 @@
...
+
+// List the other mangles
+const children = [
+	'./compat/mangle.json',
+	'./debug/mangle.json',
+	'./hooks/mangle.json'
+];
+
+// Go through each file
+children.forEach(childPath => {
```
robertknight: Could we simply use _exactly_ the same mangle file for all packages? Then the sync script could just do an `fs.copyFileSync` for each source/destination path pair.

cristianbote: I like how you're thinking! The only thing that didn't made me do a straight copy is that nested packages do not have all the properties from root mangle. And then there's other properties that might be customised?

But if my thinking is wrong we could totally just copy them in there.

cristianbote:
Hey @robertknight thanks for checking it out! 😃 I think the update might be trivial, it's just not there yet? I think @developit was looking for some added bytes that sneaked into microbundle's output. Maybe it'll be best to adopt it once it's released? I'm really open here.

The script could easily be added as part of `lint` script.

sync-mangle.js
```diff
@@ -0,0 +1,43 @@
+const { resolve } = require('path');
```
andrewiggins: Perhaps we should add this file to a `scripts` directory?

sync-mangle.js
andrewiggins: That's true today, but once the next microbundle is released there will only be one `mangle.json` file that they all will share. So any customization will go away and they will all be in the one root `mangle.json`. So perhaps we should just coalesce the `mangle.json`s now?

sync-mangle.js
cristianbote: I like it, I'll add that

sync-mangle.js
cristianbote: Awesome! Then, that confirms @robertknight suggestion and I'll copy the files directly. Thanks for jumping in!

package.json
```diff
@@ -81,7 +81,8 @@
     "tabWidth": 2
   },
   "lint-staged": {
-    "**/*.{js,jsx,ts,tsx}": [
+    "**/*.{js,jsx,json,ts,tsx}": [
```
andrewiggins: Just to double check - prettier respects the `.editorconfig` file with regards to `.json` formatting?

cristianbote: Sadly it does not 😞 

Anyway, I don't think this is a good solution. `microbundle` keeps modifying the mangle.json and it's gonna be an issue keep them in sync. I think it'll be best to just waited out on `microbundle@0.12.0`. What do you think?

andrewiggins: Dang :/ Yea, perhaps we should wait.

developit:
Just a crazy suggestion: can we use symlinks instead?

```json
{
  "scripts": {
    "prebuild": "ln -fs $PWD/mangle.json compat/mangle.json && ln -fs $PWD/mangle.json hooks/mangle.json && ln -fs $PWD/mangle.json debug/mangle.json"
  }
}
```

ForsakenHarmony:
@developit cross platform support?

cristianbote:
Closing this. The proper solution will be added via microbundle root mangle file

PR #2082, opened 2019-11-01
Rewrite classes in ES5

developit:
Just checking the file size. I'm not sure this is the best idea given we intend to ship a modern version.

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -92,14 +92,13 @@ function render(vnode, parent, callback) {
 	return vnode ? vnode._component : null;
 }
 
-class ContextProvider {
-	getChildContext() {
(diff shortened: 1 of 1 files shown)


compat/src/index.js
```diff
@@ -359,19 +358,12 @@ function findDOMNode(component) {
...
-		this.isPureReactComponent = true;
-	}
-
-	shouldComponentUpdate(props, state) {
-		return (
-			shallowDiffers(this.props, props) || shallowDiffers(this.state, state)
-		);
-	}
-}
+function PureComponent() {}
```
jviide: Maybe this constructor should do the `super(props);` equivalent?

PR #2084, opened 2019-11-02
Add types for currentTarget on event handlers for IntrinsicElem…

lukeshiru:
Currently when doing something like:
```jsx
const [value, setValue] = useState("");

return (
  <input onInput={event=>setValue(event.currentTarget.value)} />
);
```
We get an error because `value` is not defined on type `EventTarget`. With the change introduced in this PR, the currentTarget in this case is of type `HTMLInputElement`, and the same happens with all intrinsic elements.

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -288,217 +288,300 @@ export namespace JSXInternal {
 		d: string;
 	}
 
-	interface EventHandler<E extends Event> {
+	type TargetedEvent<
+		Target extends EventTarget = EventTarget,
+		TypedEvent extends Event = Event
+	> = Omit<TypedEvent, 'currentTarget'> & {
+		readonly currentTarget: Target;
+	};
+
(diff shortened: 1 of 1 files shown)


lukeshiru:
Fixed to work with SVG elements as well.

malte-wessel:
how about the `target` property? I'm still struggling with making this work:
```js
const [value, setValue] = useState("");

return (
  <input onInput={event=>setValue(event.target.value)} />
);
```

marvinhagemeister:
@malte-wessel can you open a new issue?

lukeshiru:
@malte-wessel you need to consider that while `currentTarget` is always the element that added the eventListener, `target` can be a child element, so:
1. Is a good practice to avoid `target` altogether and just use `currentTarget`.
2. If you really need `target` and you are sure a `children` will not trigger that event, then you can typecast that yourself in your use case:
```tsx
<div onClick={event => {
  (event.target as HTMLDivElement).propertyYouWantToAccess
}></div>
```

PR #2085, opened 2019-11-03
Simplify hook scheduling logic (-17 B)

andrewiggins:
Summary:
* aec90a5 Move rAF check to right before rAF call (-6 B)
* 8dc32be Remove unnecessary prevRaf init (-3 B)
* cbc79a5 Hoist all variable declarations to the top of the file (-8 B)
  I'm guessing since we were intermixing reading and writing an object, terser can't safely hoist these for us so manually doing it instead saves some bytes

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -256,7 +256,7 @@ function afterNextFrame(callback) {
 	}
 }
 
-let prevRaf = options.requestAnimationFrame;
+let prevRaf;
 
 // Note: if someone used Component.debounce = requestAnimationFrame,
 // then effects will ALWAYS run on the NEXT frame instead of the current one, incurring a ~16ms delay.
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -260,7 +260,7 @@ function afterNextFrame(callback) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


hooks/src/index.js
```diff
@@ -259,20 +253,27 @@ function afterNextFrame(callback) {
...
 
-/* istanbul ignore else */
-if (typeof window !== 'undefined') {
-	let prevRaf = options.requestAnimationFrame;
-	afterPaint = newQueueLength => {
-		if (newQueueLength === 1 || prevRaf !== options.requestAnimationFrame) {
-			prevRaf = options.requestAnimationFrame;
+// Note: if someone used options.debounceRendering = requestAnimationFrame,
+// then effects will ALWAYS run on the NEXT frame instead of the current one, incurring a ~16ms delay.
+// Perhaps this is not such a big deal.
```
marvinhagemeister: pinging @robertknight because afaik this hook is used in `enzyme-adapter-preact-pure`.

robertknight: The Enzyme adapter uses the `debounceRendering` hook in order to be able to synchronously flush pending state updates at certain points. I can't see an issue with the change here.

PR #2086, opened 2019-11-03
In tests, import code under test by name instead of relative pa…

andrewiggins:
I'd like to explore the possibility of running our test suite against our build output, so our tests more closely match what our users are doing. I think the first step to do that is to change our test imports to import the source code by name (e.g. `import {} from 'preact'`) instead of by relative path (e.g. `import {} from `../../src/'`). This PR enables us to change what our tests import in the karma config (that'll come in a later PR, assuming running tests against minified output works).

To make VSCode intellisense happy, I added a `jsconfig.json` that tells VSCode where the code for imports like `preact` and `preact/compat` live, so it can provide us guidance.

A couple other changes that fall out of trying to do this:

1. All tests now use `createElement` for JSX since the `jsconfig.json` can only specify one `reactNamespace`.
2. I separated out the `serializeVNode` tests into its own file since it has to import `serializeVNode` directly from source because `serializeVNode` isn't exported by `preact/debug`.
3. Same for `getDomSibling` - Since it is directly testing an internal method, it will have to import directly from `src`.
4. Some tests are hardcoded to use `h` so you'll see `const h = createElement;` in some test files. I tried to not modify tests directly in this PR to verify I didn't break any of them, so I did this little workaround instead.
5. I updated the `node` and `shared` tests to point to the root of the repo so they run against minified builds now! It was easy enough so I thought why not?

### DIFF
--- a/test/_util/optionSpies.js
+++ b/test/_util/optionSpies.js
@@ -1,6 +1,6 @@
 import { options as rawOptions } from 'preact';
 
-/** @type {import('../../src/internal').Options} */
+/** @type {import('preact/src/internal').Options} */
 let options = rawOptions;
 
 let oldVNode = options.vnode;
--- a/test/browser/createContext.test.js
+++ b/test/browser/createContext.test.js
@@ -1,15 +1,15 @@
 import { setupRerender, act } from 'preact/test-utils';
 import {
-	createElement as h,
+	createElement,
 	render,
 	Component,
-	createContext
+	createContext,
+	Fragment
 } from '../../src/index';
+import { i as ctxId } from '../../src/create-context'; // TODO: Rewrite tests to remove this internal dependency
 import { setupScratch, teardown } from '../_util/helpers';
-import { Fragment } from '../../src';
-import { i as ctxId } from '../../src/create-context';
 
-/** @jsx h */
+/** @jsx createElement */
 
 describe('createContext', () => {
 	let scratch;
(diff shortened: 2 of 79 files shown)


test/_util/optionSpies.js
```diff
@@ -1,6 +1,6 @@
 import { options as rawOptions } from 'preact';
 
-/** @type {import('../../src/internal').Options} */
+/** @type {import('preact/src/internal').Options} */
```
andrewiggins: Technically with the jsconfig.json, this is now possible, but felt out of scope for this PR.

test/browser/createContext.test.js
```diff
@@ -1,15 +1,15 @@
...
 import {
-	createElement as h,
+	createElement,
 	render,
 	Component,
-	createContext
+	createContext,
+	Fragment
 } from '../../src/index';
+import { i as ctxId } from '../../src/create-context'; // TODO: Rewrite tests to remove this internal dependency
```
andrewiggins: Updates coming in a future PR

test/ts/Component-test.tsx
```diff
@@ -1,11 +1,11 @@
-import "mocha";
```
andrewiggins: I think Prettier got a hold of all the `.ts` files in the `ts` directory . I didn't edit these. My focus is manually on our other tests

PR #2093, opened 2019-11-05
Have the children ref array always initialized (-10B)

cristianbote:
Hey @JoviDeCroock I wanted to add this suggestion on your PR at #2055, and I was on my phone and it seems that it stayed in a pending state rather that adding it. Anyway, if we always keep the `refs = []` initialised we could shave `-10B`.

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -34,7 +34,8 @@ export function diffChildren(
 	oldDom,
 	isHydrating
 ) {
-	let i, j, oldVNode, newDom, sibDom, firstChildDom, refs;
+	let i, j, oldVNode, newDom, sibDom, firstChildDom;
+	let refs = [];
 
 	// This is a compression of oldParentVNode!=null && oldParentVNode != EMPTY_OBJ && oldParentVNode._children || EMPTY_ARR
 	// as EMPTY_OBJ._children should be `undefined`.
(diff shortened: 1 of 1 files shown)


src/diff/children.js
```diff
@@ -34,7 +34,8 @@ export function diffChildren(
 	oldDom,
 	isHydrating
 ) {
-	let i, j, oldVNode, newDom, sibDom, firstChildDom, refs;
+	let i, j, oldVNode, newDom, sibDom, firstChildDom;
+	let refs = [];
```
andrewiggins: My one concern here is that we are adding a new array allocation for each depth of the tree which could add some non-trivial GC churn. Perhaps this would be okay if we think that refs are used through a tree commonly enough such that they'll get allocated anyway but I'm not sure refs are *that* often used. Every app has a couple, but probably not at every level of their app :/

So maybe the extra bytes are worth it in this case? Though I have no data to back up my hypothesis so I could way off lol.

developit: I really want to test this. It should show up in benchmarks as an increase in average render duration or higher stddev.

cristianbote: Really good points. Hmm 🤔 wondering if we move the `refs` outside and just do a `.length = 0` after we apply the refs, would that closure hurt more?

andrewiggins: So I've been playing around with how we do refs for a couple days and I fixed the final issues I was having. Check out #2098 and let me know what you think! There are still some potential byte-savings vs memory usage trade-offs worth looking into in #2098. I left a comment  in that PR describing where we can look.

cristianbote: Awesome stuff @andrewiggins! Will close this one. Thanks!

cristianbote:
Closing this in favor of @andrewiggins version

PR #2094, opened 2019-11-06
Upgrade tests to use Babel 7

andrewiggins:
I'm exploring writing a babel plugin to convert property names to enable our tests to run against the minified builds. In the process of doing that I realized that our tests still used Babel 6. This PR upgrades them to Babel 7.


babel.config.js
```diff
@@ -0,0 +1,24 @@
+module.exports = function(api) {
```
andrewiggins: `babel-loader` in the `karma.config.js` wasn't picking up the configuration in `.babelrc` but did work when I changed it to a `babel.config.js`. 🤷‍♂

test/browser/lifecycles/getDerivedStateFromError.test.js
```diff
@@ -52,8 +52,8 @@ describe('Lifecycle methods', () => {
 			}
 		}
 
-		spyAll(Receiver.prototype);
 		spyAll(Receiver);
+		spyAll(Receiver.prototype);
```
andrewiggins: This tests failed unless I swapped the order here. Perhaps some new class transform broke the old form? 🤷‍♂

karma.conf.js
```diff
@@ -140,28 +140,8 @@ module.exports = function(config) {
...
-													'**/*.test.js',
-													'**/__tests__/**',
-													'**/node_modules/**',
-													// Our custom extension
-													'{debug,hooks,compat,test-utils}/test/**/*'
-												]
-											}
-										]
-								  ]
+								? [['istanbul', { include: '**/src/**/*.js' }]]
```
marvinhagemeister: That's much better :+1:

PR #2096, opened 2019-11-06
Add "funding" field to package.json

developit:
See https://github.com/npm/rfcs/blob/latest/accepted/0017-add-funding-support.md


prateekbh:
TIL. This is awesome

PR #2097, opened 2019-11-06
Do not check for prototype in newType

jonathantneal:
This change removes the check for `prototype` on `newType`, if the check is only present to detect classes.

If I understand correctly:
- By checking `'prototype' in newType` the script assumes `newType` is an object, otherwise this could throw if `newType` were a primitive.
- The purpose of the `if` statement checking `'prototype' in newType` is to detect classes.

Therefore, knowing that all functions have a `prototype`, this check is potentially unnecessary.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -63,7 +63,7 @@ export function diff(
 				clearProcessingException = c._processingException = c._pendingError;
 			} else {
 				// Instantiate the new component
-				if ('prototype' in newType && newType.prototype.render) {
+				if (newType.prototype.render) {
 					newVNode._component = c = new newType(newProps, cctx); // eslint-disable-line new-cap
 				} else {
 					newVNode._component = c = new Component(newProps, cctx);


jonathantneal:
I now realize this check should be necessary for arrow functions. I’m surprised the tests did not throw. Perhaps there is not a test to handle this particular try/catch branch?

Now, I’m only thinking about this because I’m already in here, but could the check still be shortened safely by allowing an undefined `prototype` to create an object? See:

```js
// current source
'prototype' in newType && newType.prototype.render

// proposed change
Object(newType.prototype).render
```

And a loose size comparison:

```js
// current minified source — 36 bytes, up to 49 bytes gzipped
"prototype" in P&&P.prototype.render

// adjusted minified source after the proposal — 26 bytes, up to 46 bytes gzipped
Object(P.prototype).render
```

I would gladly adjust this PR if this would work and there is interest.

andrewiggins:
Hi @jonathantneal, thanks for the PR! Hmm good observation. I suspect our tests didn't catch the arrow function case because our tests are transpiled using Babel before running so I'm guessing all arrow functions get converted to normal functions.

I'm not super familiar with the inner workings or performance characteristics of the `Object` constructor. Do you know if it allocates a new object if given `null` or `undefined`? I wonder if we could save a similar amount by just doing a falsey check for `prototype`: `P.prototype && P.prototype.render`?

andrewiggins:
Actually, looking at the git blame for this line, it looks like we purposefully avoid checking `prototype` for perf reasons: #1835

jonathantneal:
I’ve learned all sorts of things. Closing this, as I don’t see any way to optimize this further! Thanks!

PR #2098, opened 2019-11-07
Move renderCallbacks to VNode and queue refs as renderCallbacks

andrewiggins:
In the spirit of #2011, queue the `applyRef` calls as `_renderCallbacks` so they are invoked after the entire DOM is updated. This change effectively undos 40212ebea from #1658.

I always didn't like how I had to move ref handling to `diffChildren`. That had to be done in #1658 because refs needed to be unmounted first, then the new value applied, for example, if the VNode changed tag type from `div` to `span`. We would need to call the `ref` with `null` when the `div` is unmounted, then call it again with the `span` in that order. This PR accomplishes the same behavior by queuing the setting of the new ref value in `_renderCallbacks` after all unmounts and DOM manipulations are done.


src/create-element.js
```diff
@@ -64,6 +64,7 @@ export function createVNode(type, props, key, ref) {
 		_dom: null,
 		_lastDomChild: null,
 		_component: null,
+		_renderCallbacks: null,
```
andrewiggins: FYI, initializing this to an array saves us about 11 B. However it does add an extra array allocation for every VNode :/ Unsure how that affects performance. Perhaps someone can investigate this and open a new PR if its worth it?

src/diff/children.js
```diff
@@ -111,12 +111,6 @@ export function diffChildren(
 					isHydrating
 				);
 
-				if ((j = childVNode.ref) && oldVNode.ref != j) {
-					if (!refs) refs = [];
```
andrewiggins: Yay! This array allocation is gone! Now if you attach a ref to a component, we re-use that component's `_renderCallbacks` array. However, if you attach a ref to a DOM VNode, we allocate a new array (that VNode's `_renderCallbacks` array) for that ref. But I think the net is still fewer array allocations.

JoviDeCroock:
Hey,

Could we add a test here (can be skipped) to see if the refs are bound in ComponentDidMount?

Specific example:

```
<Parent>
  <Child />
  <Child />
</Parent>
```

The Child for example has a span with ref={ref} and in `componentDidMount` `if (!ref) throw new Error(à`

andrewiggins:
Oh good call! Will do

compat/test/browser/suspense.test.js
```diff
@@ -382,11 +382,7 @@ describe('suspense', () => {
 			expect(scratch.innerHTML).to.eql(`<p>Count: 2</p>`);
 			expect(componentWillMount).to.have.been.calledOnce;
 			expect(componentDidMount).to.have.been.calledOnce;
-			// TODO: This is called thrice since the cDU queued up after the second
-			// increment is never cleared once the component suspends. So when it
-			// resumes and the component is rerendered, we queue up another cDU so
-			// cDU is called an extra time.
-			expect(componentDidUpdate).to.have.been.calledThrice;
```
andrewiggins: Since `_renderCallbacks` is stored on VNode now, they don't persist across renders. If a VNode suspends, it'll never get added to the `commitQueue` array so its `_renderCallbacks` won't get called. And since VNodes are recreated on each render it won't get invoked on the next render since that'll be with a newVNode, and not the one from the previous render which had the `_renderCallback`.

andrewiggins:
Superseded by #3860

PR #2099, opened 2019-11-07
Fix incompatible ref typing with ReactElement (& popular react libraries)

### LINKED ISSUE #1953: styled-components ref type

xiel:
This PR update the type of VNode to be compatible with popular react libarys like **styled-components** and **react-spring**, which are extending React's JSX.Element and therefore React.ReactElement.

**Demo repo showing the errors:**
https://github.com/xiel/preact-typescript-test

**Steps to reproduce:**
- checkout https://github.com/xiel/preact-typescript-test
- run:  `yarn; yarn check-types`

ReactElement as no guaranteed ref, and therefore the types have a mismatch which leads to errors:

**styled-components**
```
TS2605: JSX element type 'ReactElement<Pick<Pick<Pick<DetailedHTMLProps<HTMLAttributes<HTMLHeadingElement>, HTMLHeadingElement>, "hidden" | ... 252 more ... | "onTransitionEndCapture"> & { ...; }, "hidden" | ... 253 more ... | "onTransitionEndCapture"> & Partial<...>, "hidden" | ... 253 more ... | "onTransitionEndCapture"> & { ...; }, string...' is not a constructor function for JSX elements.
  **Property 'ref' is missing in type 'ReactElement**<Pick<Pick<Pick<DetailedHTMLProps<HTMLAttributes<HTMLHeadingElement>, HTMLHeadingElement>, "hidden" | ... 252 more ... | "onTransitionEndCapture"> & { ...; }, "hidden" | ... 253 more ... | "onTransitionEndCapture"> & Partial<...>, "hidden" | ... 253 more ... | "onTransitionEndCapture"> & { ...; }, string...' but required in type 'Element'.

 node_modules/preact/src/index.d.ts:17:3
    17          ref: Ref<any> | null;
                ~~~
    'ref' is declared here.
```

**react-spring**
```
 TS2605: JSX element type 'ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)>) | (new (props: any) => Component<any, any, any>)>' is not a constructor function for JSX elements.
  **Property 'ref' is missing in type 'ReactElement**<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)>) | (new (props: any) => Component<any, any, any>)>' but required in type 'Element'.

 node_modules/preact/src/index.d.ts:17:3
    17          ref: Ref<any> | null;
                ~~~
    'ref' is declared here.
```

fixes #1953 
Was broken with this PR it seems: #1447


src/index.d.ts
```diff
@@ -14,7 +14,7 @@ declare namespace preact {
 		type: ComponentType<P> | string;
 		props: P & { children: ComponentChildren };
 		key: Key;
-		ref: Ref<any> | null;
+		ref?: Ref<any> | null;
```
andrewiggins: Would you mind leaving a comment here on the importance of the `?` here? Since we don't have a test to verify we don't change this in the future, a comment stating why we have the `?` would be helpful

38elements: @andrewiggins 
IMHO,
Changing type definition for removing error is replacing the means and purpose.
This change raises that error becomes normal.
For React compatibility, it is appropriate that `createVNode()` is changed, not type definition.
But, I think it is unnecessary.

xiel: I updated the description and added a test project which demonstrates the error.

@38elements unnecessary? How are end users supposed to fix this issue? Even as a react lib author I would not know how to fix this issue?

victordidenko:
Please, please, merge this 🥺This also affects any CSS-in-JS library, for example, `astroturf` or `@emotion/styled`

38elements:
@jackbravo 
I do not understand your approval.

38elements:
@JoviDeCroock 
I do not understand your approval.

JoviDeCroock:
@38elements more and more libraries are coming with this solution and no one seems to solve it differently, when evaluating `@types/react` this is inline with what they're expecting there as well. Better proposals? I already PR'ed this once before and we did stop it then but if this blocks people I'd rather merge it this time.

38elements:
@jackbravo
I am sorry.
I made mistake.

38elements:
@JoviDeCroock 
IMHO,
I think user should use preact/compat, if user needs compatibility.
https://preactjs.com/guide/v10/switching-to-preact
If typing is changed, chaning `compat/src/index.d.ts` is appropriate.
Or Using `preact/compat` and `@type/react` instead of `compat/src/index.d.ts`.

JoviDeCroock:
@38elements the users are using `preact/compat` the thing is that this exports from `preact` meaning that we have an inconsistency in our types

38elements:
@JoviDeCroock 
`compat/src/index.d.ts` is using `src/index.d.ts`.
It is necessary that `compat/src/index.d.ts` independents `src/index.d.ts`.
#2103

JoviDeCroock:
No, it uses preact VNodes so why would we split up maintenance so we have to do that in two places? The `undefined` is a grounded change since ref is undefined.

38elements:
@JoviDeCroock 
It is only changing VNode type in `compat/src/index.d.ts`.
It is unnecessary to independ.

JoviDeCroock:
That's quite contradictory, in the last post you said we need to extract it to `/compat` and now you're saying this change is grounded too or am I misunderstanding this?

38elements:
@JoviDeCroock 
I changed my opinion.
It is only changing VNode type in `compat/src/index.d.ts`.

JoviDeCroock:
So are you okay with me merging this?

38elements:
No, I am not.
This pull request changes `src/index.d.ts`, not `compat/src/index.d.ts`.
The types of `preact` should not be chnaged.
I think that  chaning only type definition in `compat/src/index.d.ts` is appropriate.

38elements:
`preact/compat` with `@type/react` may resolve.
I did not try it.

victordidenko:
@38elements please, suggest different solution / workaround? Because now issue #1953 is a real blocker

38elements:
read my comment.

victordidenko:
> I did not try it
Well, I tried. Issue is still here.

victordidenko:
@38elements why do you even so sure, that in preact `ref` is always exists?
I checked in the code, and looks like `undefined` is legit value for `ref` in VNode, check it:

According to JSDoc `props` here could be `undefined`:
https://github.com/preactjs/preact/blob/master/src/create-element.js#L11

If props above is `undefined`, then `ref` here will be `undefined` too:
https://github.com/preactjs/preact/blob/master/src/create-element.js#L59

victordidenko:
As well as `key` btw

victordidenko:
@38elements even in your PR #1447 function `createVNode` is called not only from `coerceToVNode`, but from exported `createElement`
https://github.com/preactjs/preact/blob/66aa9e24bbf116fa30a5fddce7ebc21d8ea3199f/src/create-element.js#L35
and `ref` and `key` could be `undefined` here (when `props` is `undefined` or `null`).

But **you think** (quote from your PR) that ref and key are always exist!

38elements:
https://github.com/preactjs/preact/pull/1955#discussion_r329601117

38elements:
@JoviDeCroock @andrewiggins 
Considering  the case that VNode is created in user defined function, not createElement(), I think it may be better `key?`.

```
class Foo extends Component {
  render() {
    return {type: 'p', props: {}}
  }
}
```

victordidenko:
@38elements you are right, but `undefined` value is still legit value for `ref` and `key`, according to the code, so considering this and following two rows:
```typescript
let a: { b?: null } = { b: undefined }
let c: { b: null | undefined } = { b: undefined }
```
what kind of real issues are solved by the second row, which could be missed by the first row?
In my opinion those two lines logically equivalent, because trying to get undefined key in object you will get undefined, either key exists in the object or not.
Of course, if there is no key in object — JavaScript could get it from prototype. But we have TypeScript here, and this case will be caught beforehand.

xiel:
For the people coming to this thread following the same error messages, I can recommend
https://www.npmjs.com/package/patch-package
to fix this error in your project by yourself.

zaninime:
Workaround instructions (inspired by @xiel's comment)

1. install `patch-package` as a dev dependency
2. create a directory called `patches`
3. download https://gist.github.com/zaninime/3fbb44029fb763f480245c44569431d8 in that folder
4. run `yarn patch-package` or the equivalent `npm run`
5. profit.

JoviDeCroock:
My apologies for the long wait everyone.
@xiel my sincerest thank for PR'ing this, this will be available upon next release!

developit:
Hi folks - we're getting this merged for now in order to unblock usage with these libraries.

For those interested in changes to how we type `preact` VS `preact/compat`, let's move further discussion over to #2103.

PR #2102, opened 2019-11-08
use __data on components

JoviDeCroock:
The initial proposal was to add it onto `_vnode` but this is not possible since we explicitly have a case where `_vnode` is not present:  https://github.com/preactjs/preact/blob/master/src/component.js#L46

Moving _parentDom seems to cause an issue with nested suspense, haven't gotten to the core of that yet, my next research will be looking into removing duplicate dom references.


src/component.js
```diff
@@ -13,6 +13,7 @@ import { Fragment } from './create-element';
 export function Component(props, context) {
 	this.props = props;
 	this.context = context;
+	this.__data = {};
```
JoviDeCroock: Todo: find a way to only do this in one spot, ideally with defaults for the stuff we use

porfirioribeiro:
So all the private component data would be moved to the __data object?
Probably later i'll need to update the composition API PR here
https://github.com/preactjs/preact/pull/1923/files#diff-23369f01c5ea98c3654a47ca1abd2de7R28
and move `__compositions` inside `__data`

JoviDeCroock:
@porfirioribeiro this is a proposal to see if we gain performance benefit from this (we anticipate we will but let's see). If it comes to that I'll be happy to make that PR to yours

PR #2104, opened 2019-11-09
Enable better Webpack tree shaking

davidbailey00:
Add `"sideEffects": false` to `package.json` files in `preact`, `preact/hooks` and `preact/compat` to enable improved tree shaking of unused exports, i.e. when using an immediately-invoked function to generate something which is exported (which may perform additional side effects, hence why Webpack doesn't optimize this by default).

-4.2 KB bundle size when importing only `memo` from `preact/compat`; not sure if there is any improvement in `preact` or `preact/hooks`.

### DIFF
--- a/compat/package.json
+++ b/compat/package.json
@@ -7,6 +7,7 @@
   "main": "dist/compat.js",
   "module": "dist/compat.module.js",
   "umd:main": "dist/compat.umd.js",
+  "sideEffects": false,
   "source": "src/index.js",
   "types": "src/index.d.ts",
   "license": "MIT",
--- a/hooks/package.json
+++ b/hooks/package.json
@@ -7,6 +7,7 @@
   "main": "dist/hooks.js",
   "module": "dist/hooks.module.js",
   "umd:main": "dist/hooks.umd.js",
+  "sideEffects": false,
   "source": "src/index.js",
   "license": "MIT",
   "types": "src/index.d.ts",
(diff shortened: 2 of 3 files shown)


compat/package.json
```diff
@@ -7,6 +7,7 @@
   "main": "dist/compat.js",
   "module": "dist/compat.module.js",
   "umd:main": "dist/compat.umd.js",
+  "sideEffects": false,
```
JoviDeCroock: There are side effects here

hooks/package.json
```diff
@@ -7,6 +7,7 @@
   "main": "dist/hooks.js",
   "module": "dist/hooks.module.js",
   "umd:main": "dist/hooks.umd.js",
+  "sideEffects": false,
```
JoviDeCroock: There are side effects here

davidbailey00:
@JoviDeCroock my bad, looks like some more stuff will be needed to make this work - thanks for pointing out those examples!

developit:
It might be possible for us to structure the modules in compat and hooks a bit more to get the same outcome as `sideEffects:false` while still allowing the intentional side-effects in hooks/index.js and compat/index.js.

kurtextrem:
`sideEffects` can exclude files, like so:
``
"sideEffects": [
	"index.js"
]
``

That one could be merged in, while work on restructuring is getting finished.

PR #2107, opened 2019-11-10
Fix wrong DOM order on suspend inside Fragment

### LINKED ISSUE #2106: Suspense inside a Fragment may scramble the DOM order
A certain tree of `<Suspense>` components, some of which are inside `<Fragment>`, getting resolved in a particular order can cause the DOM order to become scrambled.

Here is a test reproducing the issue (using the helpers from the `suspense.test.js` file):

```js
	it('should correctly render Suspense components inside Fragments', async () => {
		const [Lazy1, resolve1] = createLazy();
		const [Lazy2, resolve2] = createLazy();
		const [Lazy3, resolve3] = createLazy();

		const Loading = () => <div>Suspended...</div>;
		const loadingHtml = `<div>Suspended...</div>`;

		render(
			<Fragment>
				<Suspense fallback={<Loading />}>
					<Lazy1 />
				</Suspense>
				<Fragment>
					<Suspense fallback={<Loading />}>
						<Lazy2 />
					</Suspense>
				</Fragment>
				<Suspense fallback={<Loading />}>
					<Lazy3 />
				</Suspense>
			</Fragment>,
			scratch
		);

		rerender();
			`${loadingHtml}${loadingHtml}${loadingHtml}`
		);

		await resolve2(() => <span>2</span>);
		await resolve1(() => <span>1</span>);
		rerender();
			`<span>1</span><span>2</span>${loadingHtml}`
		);

		await resolve3(() => <span>3</span>);
		rerender();
			`<span>1</span><span>2</span><span>3</span>`
		);
	});
```

The second assertion fails, because the HTML output is `<span>2</span><div>Suspended...</div><span>1</span>`, i.e. the first component's DOM gets bumped to the end of the DOM.

Tested on Preact 10.0.5.

jviide:
This pull request fixes issue #2106 and refactors the Suspense implementation for smaller code size.
 * Modify the Suspense lifecycle to be more explicit:
   * On the first suspended child set the `_detachOnNextRender` flag and force a new render with `setState`.
   * On the next render detach the old suspended children from DOM and replace them with an empty list of children.
 * Use a counter to keep track of the unresolved promises instead of an array of the promises.
 * Add a test for issue #2106.
 * Fix Suspense test "should correctly render nested Suspense components", it didn't return a promise.
 * Add a test for children suspending with the same promise.
 * Add a test ensuring that no children will be rendered if any of them suspends.

The changes in this PR bring down the compat.js.gz bundle size by 38 bytes (from 2700 B to 2662 B).

Fixes #2106.

### DIFF
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -1,10 +1,4 @@
-import {
-	Component,
-	createElement,
-	_unmount as unmount,
-	options,
-	cloneElement
-} from 'preact';
+import { Component, createElement, options, Fragment } from 'preact';
 import { removeNode } from '../../src/util';
 
 const oldCatchError = options._catchError;
@@ -30,15 +24,12 @@ options._catchError = function(error, newVNode, oldVNode) {
 	oldCatchError(error, newVNode, oldVNode);
 };
 
-function detachDom(children) {
-	for (let i = 0; i < children.length; i++) {
-		let child = children[i];
-		if (child != null) {
-			if (typeof child.type !== 'function' && child._dom) {
-				removeNode(child._dom);
-			} else if (child._children) {
-				detachDom(child._children);
-			}
+function detachDom(vnode) {
+	if (vnode) {
+		if (typeof vnode.type !== 'function' && vnode._dom) {
+			removeNode(vnode._dom);
+		} else if (vnode._children) {
+			vnode._children.some(detachDom);
 		}
 	}
 }
@@ -47,7 +38,6 @@ function detachDom(children) {
 export function Suspense(props) {
 	// we do not call super here to golf some bytes...
 	this._suspensions = [];
-	this._fallback = props.fallback;
 }
 
 // Things we do here to save some bytes but are not proper JS inheritance:
@@ -61,7 +51,6 @@ Suspense.prototype = new Component();
 Suspense.prototype._childDidSuspend = function(promise) {
 	/** @type {import('./internal').SuspenseComponent} */
 	const c = this;
-	c._suspensions.push(promise);
 
 	const onSuspensionComplete = () => {
 		// From https://twitter.com/Rich_Harris/status/1125850391155965952
@@ -69,32 +58,28 @@ Suspense.prototype._childDidSuspend = function(promise) {
 			c._suspensions[c._suspensions.length - 1];
 		c._suspensions.pop();
 
-		if (c._suspensions.length == 0) {
-			// If fallback is null, don't try to unmount it
-			// `unmount` expects a real VNode, not null values
-			if (c._fallback) {
-				// Unmount current children (should be fallback)
-				unmount(c._fallback);
-			}
-			c._vnode._dom = null;
-
-			c._vnode._children = c.state._parkedChildren;
-			c.setState({ _parkedChildren: null });
+		if (c._suspensions.length === 0) {
+			c._vnode._children[0]._children = c.state._suspended;
+			c.setState({ _suspended: null });
 		}
 	};
 
-	if (c.state._parkedChildren == null) {
-		c._fallback = c._fallback && cloneElement(c._fallback);
-		c.setState({ _parkedChildren: c._vnode._children });
-		detachDom(c._vnode._children);
-		c._vnode._children = [];
+	if (c._suspensions.push(promise) === 1) {
+		detachDom(c._vnode._children[0]);
+		c.setState({ _suspended: c._vnode._children[0]._children }, () => {
+			promise.then(onSuspensionComplete, onSuspensionComplete);
+		});
+		c._vnode._children[0]._children = [];
+	} else {
+		promise.then(onSuspensionComplete, onSuspensionComplete);
 	}
-
-	promise.then(onSuspensionComplete, onSuspensionComplete);
 };
 
 Suspense.prototype.render = function(props, state) {
-	return state._parkedChildren ? this._fallback : props.children;
+	return [
+		createElement(Fragment, null, state._suspended ? [] : props.children),
+		createElement(Fragment, null, state._suspended && props.fallback)
+	];
 };
 
 export function lazy(loader) {
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -1,5 +1,5 @@
 import { Component, createElement, options, Fragment } from 'preact';
-import { removeNode } from '../../src/util';
+import { assign } from '../../src/util';
 
 const oldCatchError = options._catchError;
 options._catchError = function(error, newVNode, oldVNode) {
@@ -24,13 +24,11 @@ options._catchError = function(error, newVNode, oldVNode) {
 	oldCatchError(error, newVNode, oldVNode);
 };
 
-function detachDom(vnode) {
+function detachedClone(vnode) {
 	if (vnode) {
-		if (typeof vnode.type !== 'function' && vnode._dom) {
-			removeNode(vnode._dom);
-		} else if (vnode._children) {
-			vnode._children.forEach(detachDom);
-		}
+		vnode = assign({}, vnode);
+		vnode._component = null;
+		vnode._children = vnode._children && vnode._children.map(detachedClone);
 	}
 	return vnode;
 }
@@ -39,7 +37,7 @@ function detachDom(vnode) {
 export function Suspense(props) {
 	// we do not call super here to golf some bytes...
 	this._suspensions = 0;
-	this._detachOnNextRender = false;
+	this._detachOnNextRender = null;
 }
 
 // Things we do here to save some bytes but are not proper JS inheritance:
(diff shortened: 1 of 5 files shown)


compat/src/suspense.js
```diff
@@ -30,24 +24,20 @@ options._catchError = function(error, newVNode, oldVNode) {
...
+			vnode._children.some(detachDom);
 		}
 	}
 }
 
 // having custom inheritance instead of a class here saves a lot of bytes
 export function Suspense(props) {
 	// we do not call super here to golf some bytes...
-	this._suspensions = [];
-	this._fallback = props.fallback;
```
ForsakenHarmony: was there some reason for storing this on the class?

jviide: I'm not quite sure, but it might have had something to do how Suspense was implemented before this PR.

ForsakenHarmony: https://github.com/preactjs/preact/commit/17f65e4dfd209d643117b0fab507c43eacdb4b96

Yeah probably some "legacy things"™

compat/src/suspense.js
```diff
@@ -61,40 +53,32 @@ Suspense.prototype = new Component();
...
 Suspense.prototype.render = function(props, state) {
-	return state._parkedChildren ? this._fallback : props.children;
+	if (this._detachOnNextRender) {
+		detachDom(this._vnode._children[0])._children = [];
+		this._detachOnNextRender = false;
+	}
+
+	return [
+		createElement(Fragment, null, state._suspended ? null : props.children),
+		createElement(Fragment, null, state._suspended && props.fallback)
```
marvinhagemeister: I'm wondering if we can skip the inner `Fragments` here. Returning an array from `render` will create one and maybe we could just pass the suspensions as direct children.

compat/test/browser/suspense.test.js
```diff
@@ -916,7 +916,7 @@ describe('suspense', () => {
 
 		expect(scratch.innerHTML).to.equal(loadingHtml);
 
-		resolve1(() => <b>1</b>)
+		return resolve1(() => <b>1</b>)
```
marvinhagemeister: 👍

compat/src/suspense.js
sventschui: The fallback is later cloned using `cloneElement` to not re-use the vnode as this is not supported in our current code base. See line 87 prior to this PR

compat/src/suspense.js
```diff
@@ -61,40 +53,32 @@ Suspense.prototype = new Component();
...
+			c.setState({ _suspended: null });
 		}
 	};
 
-	if (c.state._parkedChildren == null) {
-		c._fallback = c._fallback && cloneElement(c._fallback);
-		c.setState({ _parkedChildren: c._vnode._children });
-		detachDom(c._vnode._children);
-		c._vnode._children = [];
+	if (!c._suspensions++) {
```
sventschui: I somewhen did an effort to de-duplicate `c._suspensions` when throwing the same promise multiple times which lead to duplicates in the `_suspensions` array and therefore to never un-suspend. 

Honestly I do not remember where this effort went as it seems to not be present in the current implementation.

We should check that this refactor still works when throwing the same promise multiple times.

compat/src/suspense.js
sventschui: Not sure whether the trick is to have two never un-mounting children rendered by suspense tricking our diff algo, isn't it?

compat/src/suspense.js
sventschui: I guess the cloning is no longer needed as the fallback is now rendered as a child of a Fragment that is then cloned in coerceToVNode anyways if it was already used.

compat/src/suspense.js
jviide: Yes. This was the smallest variant I found that keeps the tree consistent enough for the diff algorithm.

compat/src/suspense.js
jviide: Yeah I suppose so.

compat/src/suspense.js
jviide: Yep, that's a good test. I'll add it to the suite.

I _suspect_ that the original array-based implementation (and this modified counter-based implementation) will survive the scenario even without special dedup logic. If the same promise is thrown from two children then it's added to the array twice (or the counter is incremented twice in the new implementation). When the promise resolves it's removed from the array twice (or the counter in decremented twice in the new implementation), which keeps the internal state consistent.

jviide: I added the test, and it seems that both the original implementation and this modified one survive it just fine 👍

compat/src/suspense.js
jviide: Oh right, and I did test this with @marvinhagemeister's suggestion but the DOM elements started moving around again :)

compat/src/suspense.js
```diff
@@ -61,40 +53,32 @@ Suspense.prototype = new Component();
...
 };
 
 Suspense.prototype.render = function(props, state) {
-	return state._parkedChildren ? this._fallback : props.children;
+	if (this._detachOnNextRender) {
+		detachDom(this._vnode._children[0])._children = [];
+		this._detachOnNextRender = false;
+	}
+
+	return [
```
prateekbh: Any way we can just do a simple conditional based rendering here?

```js
return state._suspended?createElement(Fragment, null, props.fallback):createElement(Fragment, null, props.children)`
```

prateekbh: just makes it much more readable and less magic

jviide: It appears that Preact's diffing algorithm loses track with this variation as well :/

prateekbh: :(

prateekbh: also 

```js
return createElement(Fragment, null, state._suspended?props.children:props.fallback)
```

sorry, just trying to see if it can look less magical

compat/src/suspense.js
andrewiggins: I think the problem was that the old implementation was manually calling `unmount`. So we had to make sure that the VNode that was called with `unmount` was the most up to date VNode. If we re-used a VNode (e.g. if a component re-suspends), `toChildArray` notices that this VNode has been used before (e.g. it has a `_component` property) and clones it. In the old implementation this was a problem because it mean that the reference to the `props.fallback` VNode was no longer in sync with the VNode that the diff was using.

To fix this, we cloned the `fallback` before going into the diff to skip the `toChildArray` clone, guaranteeing our reference always matched the one the diff was using.

Since this PR changes suspense to no longer manually call `unmount` we longer need to worry about VNode references, since we rely on the natural thing that `diff` does!

(This is from my memory of the previous PR, didn't re-verify it so I hope my memory isn't too off base lol)

compat/src/suspense.js
```diff
@@ -61,40 +53,32 @@ Suspense.prototype = new Component();
...
+		createElement(Fragment, null, state._suspended ? null : props.children),
```
andrewiggins: Double checking, this doesn't call the unmount lifecycles for children when a component suspends? Probably need to double check that we have a test that covers this.

compat/src/suspense.js
```diff
@@ -1,10 +1,4 @@
-import {
-	Component,
-	createElement,
-	_unmount as unmount,
```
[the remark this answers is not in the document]
developit: this is a nice win.

compat/src/suspense.js
sventschui: That's actually a very valid point. Pretty sure this will call the unmount lifecycles what is different to react and not really the meaning of suspense

compat/src/suspense.js
jviide: That variation fails the tests as well.

compat/src/suspense.js
jviide: There appears to be tests for such scenarios already, which is awesome:
 * [`should not call lifecycle methods when suspending`](https://github.com/preactjs/preact/blob/70195cec93c2cd6a613e2b23f80b0240ca129e21/compat/test/browser/suspense.test.js#L230-L283)
 * [`should call fallback's lifecycle methods when suspending`](https://github.com/preactjs/preact/blob/70195cec93c2cd6a613e2b23f80b0240ca129e21/compat/test/browser/suspense.test.js#L285-L339)
 * [`should keep state of children when suspending`](https://github.com/preactjs/preact/blob/70195cec93c2cd6a613e2b23f80b0240ca129e21/compat/test/browser/suspense.test.js#L341-L388)

The first one (`should not call lifecycle methods when suspending`) tests that the lifecycle hooks of a component with a sibling that suspends. I added a new test that tests the lifecycle hooks of a component that itself suspends. I assumed that the `componentDidMount` shouldn't be called for such a component when it suspends, as React doesn't seem to do so either (see this [Codesandbox example](https://codesandbox.io/s/componentdidmount-not-called-for-suspended-component-wvh66)).

compat/src/suspense.js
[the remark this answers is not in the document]
jviide: `compat/src/index.js` imports and uses `unmount`, namely the `Portal` implementation.

jviide:
Updated the implementation. Now DOM is not modified manually at all. Instead the suspended component is removed/detached/lifted from the rendered tree to keep its state intact, and added back to the rendered tree when the suspension gets resolved.

The `[<Fragment>{props.children?}</Fragment>, props.fallback?]` structure that `render` returns is there because it makes bookkeeping easier: We always know that the (potentially suspended) children are inside the first child of the Suspense component.

This brings the compat.js.gz bundle size further down by 7 bytes (now 2655 B vs. the 2700 B in the master branch, 45 B reduction in total).

compat/src/suspense.js
sventschui: But how does that first test succeed? I‘d expect the lifecycles being called with this implementation. Will try to free some time to understand this better 🤔

sventschui: Ah now I see that‘s the detachedClone fiddling you do in render 👍👌

marvinhagemeister:
Merging, coveralls is acting up strangely again...

sventschui:
maybe coveralls is right. There are some missed branches in the suspense impl. Might be worth looking at these. I might find some time to do so later today

andrewiggins:
Note, if you click on the "coverage" badge you can see exactly which lines and branches aren't covered

marvinhagemeister:
There were no coverage changes related to any files in this PR. The one were coverage went down was the entry file of `preact/debug` :man_shrugging:

prateekbh:
One problem: Anything relying on `props.children` now gets additional nodes due to the two children.

Anything traversing props.children will probably break. Unless thats an anti pattern for libraries that I am not aware of

debug/test/browser/debug-suspense.test.js
```diff
@@ -9,6 +9,14 @@ import {
 
 /** @jsx createElement */
 
+async function waitResolved(suspense) {
```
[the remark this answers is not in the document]
jviide: Sweet! Sorry that I didn't react to this in time. If this PR caused merge conflicts to your branch I can create a patch to resolve them.

jviide:
@prateekbh `props.children` do not get modified by this code. Could you give an example of a case that you suspect might break?

prateekbh:
You're right I mistakenly added a ghost space in my tests and thought it came because of this

prateekbh:
All good 👍👍👍

PR #2108, opened 2019-11-10
Add dir property possible values

antonk52:
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir


antonk52:
wow that was speedy 🚀

PR #2109, opened 2019-11-10
Don't remove ref from props if it should be forwarded

porfirioribeiro:
At the moment we are removing ref from props on createElement, but we add it back in preact/compat if the component is `_forwarded`
Why not just don't remove `ref` from `props` at all if the component is `forwarded`?

Adds 11B to preact core and removes 18B from preact/compat

This would also allow to remove `options.vnode` from here
https://github.com/preactjs/preact/pull/1923/files#diff-23369f01c5ea98c3654a47ca1abd2de7R7


src/create-element.js
```diff
@@ -30,9 +30,13 @@ export function createElement(type, props, children) {
 			if (props[i] === undefined) props[i] = type.defaultProps[i];
 		}
 	}
-	let ref = props.ref;
+	let ref = null;
+	if (!(type != null && type._forwarded)) {
```
JoviDeCroock: We should not introduce the notion of forwarded in preact-core in my opinion, this is a principle that only exists in /compat

porfirioribeiro: I see, just wanted to remove the overhead of moving ref around unnecessarily.

PR #2114, opened 2019-11-12
Add test for lifecycles and state of a delayed suspending compo…

andrewiggins:
This new test covers the case where a component suspends after the first where some state change causes the suspension (e.g. a button press that asynchronously increments the count when it needs to get more values then it has. Actually, that's a good test case. Lol maybe sometime later I'll add that).

### DIFF
--- a/compat/test/browser/suspense.test.js
+++ b/compat/test/browser/suspense.test.js
@@ -12,6 +12,10 @@ import { setupScratch, teardown } from '../../../test/_util/helpers';
 const h = React.createElement;
 /* eslint-env browser, mocha */
 
+/**
+ * @typedef {import('../../../src').ComponentType} ComponentType
+ * @returns {[typeof Component, (c: ComponentType) => Promise<void>, (c: ComponentType) => void]}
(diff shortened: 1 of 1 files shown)


compat/test/browser/suspense.test.js
```diff
@@ -290,6 +293,104 @@ describe('suspense', () => {
...
+		return resolve().then(() => {
+			rerender();
+
+			expect(scratch.innerHTML).to.eql(`<p>Count: 2</p>`);
+			expect(componentWillMount).to.have.been.calledOnce;
+			expect(componentDidMount).to.have.been.calledOnce;
+			// TODO: This is called thrice since the cDU queued up after the second
+			// increment is never cleared once the component suspends. So when it
+			// resumes and the component is rerendered, we queue up another cDU so
+			// cDU is called an extra time.
```
andrewiggins: Note there is a follow up to this comment in #2098 (specifically [this comment](https://github.com/preactjs/preact/pull/2098/files#r345516634))

PR #2116, opened 2019-11-12
Consolidate VNode compat options (-62 B)

andrewiggins:
Currently the compat src has VNode compat functions spread across a couple different functions making it hard to modify (should I add it `normalizeVNode`? `createElement`? `handleElementVNode`?).

This PR simplifies VNode compat by inline all of those disparate functions into the `options.vnode` hook. Every vnode that is created (including `cloneElement`) goes through this options hook so it guarantees VNode compat is consistently applied.

Also saves some bytes!

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -229,13 +229,27 @@ function createElement(...args) {
 	return normalizeVNode(vnode);
 }
 
+let classNameDescriptor = {
+	configurable: true,
+	get() {
+		return this.class;
+	}
+};
+
 /**
  * Normalize a vnode
  * @param {import('./internal').VNode} vnode
  */
 function normalizeVNode(vnode) {
+	// Alias `class` prop to `className` if available
+	let a = vnode.props;
+	if (a.class || a.className) {
+		classNameDescriptor.enumerable = 'className' in a;
+		if (a.className) a.class = a.className;
+		Object.defineProperty(a, 'className', classNameDescriptor);
+	}
+
 	vnode.preactCompatNormalized = true;
-	applyClassName(vnode);
 	return vnode;
 }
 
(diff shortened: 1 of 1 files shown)


compat/src/index.js
```diff
@@ -470,6 +440,10 @@ options.vnode = vnode => {
 		setSafeDescriptor(type.prototype, 'componentWillUpdate');
 		type._patchedLifecycles = true;
 	}
+
+	// TODO: Removing this saves 16 bytes!
+	vnode.preactCompatNormalized = true;
```
andrewiggins: Anyone know if we need to support this property? Or can we remove it?

JoviDeCroock: I vote for removing it, it's unused and don't really see a use case for it

compat/src/index.js
```diff
@@ -194,51 +177,6 @@ let Children = {
...
-	vnode.preactCompatNormalized = false;
-	return normalizeVNode(vnode);
-}
-
-/**
- * Normalize a vnode
- * @param {import('./internal').VNode} vnode
- */
-function normalizeVNode(vnode) {
-	vnode.preactCompatNormalized = true;
```
andrewiggins: Just calling that I removed this property cuz I don't think it's being used anymore. Holler if you think this is worth the 16 bytes.

compat/src/index.js
andrewiggins: Removed. Since this comment is now outdated (which I think means it'll be collapsed for some), I made a new comment for others to see and discuss.

compat/src/index.js
marvinhagemeister: Iirc this is a leftover from earlier iterations. Previous versions would check for this property and only normalize a `vnode` if it was missing.

cristianbote:
Great work @andrewiggins! 🎉

PR #2117, opened 2019-11-13
Fix internal Suspense-related typings

jviide:
This pull request fixes the Suspense-related types in compat/src/internal.d.ts that Yours Truly broke in #2107.

To sweeten the deal there are a couple of small code golfs that bring compat.js.gz size down by 4 bytes (from 2655 B to 2651 B) 🙂


jviide:
Removed a bit of code that is now unnecessary, shaving off 7 B from compat.js.gz.

Issue #2113, opened 2019-11-12
onTouchStart event handler does not work properly on custom elements with display: block


Nejivoi:
**After upgrading to Preact X** `onTouchStart` event handler stopped working on custom elements with `display: block`.
Here is code sample:
`<custom style={{display: 'block'}} onTouchStart={() => {console.warn('onTouchStart!');}}>Hello, World!</custom>`

In devtools Event Listeners I noticed strange event name `TouchStart` (with capital T and S)
![image](https://user-images.githubusercontent.com/13499048/68684124-01f16380-0579-11ea-82a1-18f35cf70c89.png)

If i change event name to `ontouchstart` it all works fine.
example:
`<custom style={{display: 'block'}} ontouchstart={() => {console.warn('onTouchStart!');}}>Hello, World!</custom>`

What is strange is that such behavior only occurs in Chrome on my desktop pc. On actual touch devices handler works as intended.

So here is **the question**. Is it expected behavior?
And if it is, should I change event name for all my components?

sventschui:
I did some investigations in a codesandbox. It is clear that the issue of adding a handler for `TouchStart` instead of `touchstart` comes from `'ontouchstart' in <any dom element>` results in false on chrome on desktop without touch input device.
> Repro: Open https://codesandbox.io/s/touchstart-ycofm in chrome, it will say that `ontouchstart` is not in `document.body`

Once you go into mobile (touch) emulation mode using the devtools the event handler for `touchstart` will start to fire, the event handler for `TouchStart` or handlers added using `<dom element>.ontouchstart = () => {}` won't.
> Repro: Open https://codesandbox.io/s/touchstart-ycofm in chrome (with devtools closed), open devtools/start mobile (touch) emulation mode and click on the viewport. You'll see the event handle for `touchstart` will fire.

When you load a page with the devtools/mobile (touch) emulation open beforehand, the check for `'ontouchstart' in document.body` will suddenly become true. Also the handlers for `touchstart` and `document.body.ontouchstart = () => {}` will trigger (the `TouchStart` still does not)
> Repro: Open devtools, start mobile (touch) emulation and then open https://codesandbox.io/s/touchstart-ycofm


That means that the issue is related to switching from non-touch-input to touch-input after some content was rendered. For sure adding `TouchStart` instead of `touchstart` makes no sense anyways.

[...]

PR #2120, opened 2019-11-13
Always use lower cased touch events in compat

sventschui:
Always use lower cased touch events in compat. See #2113

### DIFF
--- a/compat/src/events.js
+++ b/compat/src/events.js
@@ -7,7 +7,7 @@ export function applyEventNormalization({ type, props }) {
 	let newProps = {};
 
 	for (let i in props) {
-		if (/^on(Ani|Tra)/.test(i)) {
+		if (/^on(Ani|Tra|Tou)/.test(i)) {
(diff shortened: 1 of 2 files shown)


compat/src/index.js
```diff
@@ -207,7 +207,7 @@ function applyEventNormalization({ type, props }) {
 	let newProps = {};
 
 	for (let i in props) {
-		if (/^on(Ani|Tra)/.test(i)) {
+		if (/^on(Ani|Tra|Tou)/.test(i)) {
```
andrewiggins: #2124 might make this difficult to merge... sorry! FYI, this change should now be made in `events.js` I think

compat/test/browser/index.test.js
```diff
@@ -120,6 +120,60 @@ describe('preact-compat', () => {
 			);
 		});
 
+		it('should support onTouch* events', () => {
```
andrewiggins: Re: merging with #2124, I think this test now goes in `events.test.js`

PR #2125, opened 2019-11-14
Suspended components should rerender through sCU

### LINKED ISSUE #2176: Issue with Suspense + Lazy + @reach/router
Reach Router(https://github.com/reach/router) has an issue where it doesn't seem to trigger re-render once the Suspended Component is ready. In my projects if I navigate away and then come back again or do something to cause the parent to re-render suddenly the content is there. But otherwise it never shows on its own. 

There appears to be no errors in the console. The content just does not appear as expected if Suspense is triggered on render.

I forked the sandbox for a React Router issue(which seems now to be fixed in 10.1.0). This example uses @react/router instead and appears to still be broken:
https://codesandbox.io/s/preact-suspense-lazy-issue-tdfjc

andrewiggins:
Resurrecting the awesome work @sventschui did in #1660 now built on top of @jviide and @prateekbh suspense changes. Just calls `forceUpdate` on the suspended component to guarantee it rerenders when the promise resolves.

Fixes #2176

### DIFF
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -30,7 +30,8 @@ function detachedClone(vnode) {
 // having custom inheritance instead of a class here saves a lot of bytes
 export function Suspense(props) {
 	// we do not call super here to golf some bytes...
-	this._suspensions = 0;
+	this._pendingSuspensionCount = 0;
+	this._suspenders = null;
 	this._detachOnNextRender = null;
 }
 
@@ -46,6 +47,11 @@ Suspense.prototype = new Component();
 Suspense.prototype._childDidSuspend = function(promise, suspendingComponent) {
 	/** @type {import('./internal').SuspenseComponent} */
 	const c = this;
+
+	if (c._suspenders == null) {
+		c._suspenders = [];
+	}
+	c._suspenders.push(suspendingComponent);
 
 	const resolve = suspended(c._vnode);
 
@@ -73,13 +79,18 @@ Suspense.prototype._childDidSuspend = function(promise, suspendingComponent) {
 	};
 
 	const onSuspensionComplete = () => {
-		if (!--c._suspensions) {
+		if (!--c._pendingSuspensionCount) {
 			c._vnode._children[0] = c.state._suspended;
 			c.setState({ _suspended: (c._detachOnNextRender = null) });
+
+			let suspended;
+			while ((suspended = c._suspenders.pop())) {
+				suspended.forceUpdate();
+			}
 		}
 	};
 
-	if (!c._suspensions++) {
+	if (!c._pendingSuspensionCount++) {
 		c.setState({ _suspended: (c._detachOnNextRender = c._vnode._children[0]) });
 	}
 	promise.then(onResolved, onResolved);
--- a/compat/test/browser/suspense.test.js
+++ b/compat/test/browser/suspense.test.js
@@ -129,7 +129,7 @@ describe('suspense', () => {
 	});
 
 	it('should support lazy', () => {
-		const LazyComp = () => <div>Hello from LazyComp</div>;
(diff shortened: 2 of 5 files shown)


debug/src/debug.js
```diff
@@ -376,7 +376,7 @@ Component.prototype.forceUpdate = function(callback) {
 		);
 	} else if (this._parentDom == null) {
 		console.warn(
-			`Can't call "this.setState" on an unmounted component. This is a no-op, ` +
+			`Can't call "this.forceUpdate" on an unmounted component. This is a no-op, ` +
```
marvinhagemeister: Wondering if we should mention both here. The user is probably not aware that we call `forceUpdate` for `setState` under the hood.

compat/src/suspense.js
```diff
@@ -60,6 +61,7 @@ Suspense.prototype._childDidSuspend = function(promise) {
 		if (!--c._suspensions) {
 			c._vnode._children[0] = c.state._suspended;
 			c.setState({ _suspended: (c._detachOnNextRender = null) });
+			suspender.forceUpdate();
```
sventschui: Don't we need to forceUpdate all the children that once suspended here? I.e. if `!--c._suspensions` is false that child needs to be force updated once `!--c._suspensions` goes to through in a later resolved suspension. Or am I missing a point here?

andrewiggins: Oh true! Lemme extend my tests and try again

debug/src/debug.js
andrewiggins: I think since #1939 forceUpdate and setState and fully independent if I'm reading this correctly.

compat/src/suspense.js
andrewiggins: Added tests. Need to think more on this...

sventschui:
Your current approach is currently running into similar issues as #2159 

This makes me rethink our decision to unmount suspended content instead of adding `display="none"` as react does. But from the early days I remember adding a `display="none"` is not that easy

compat/src/suspense.js
```diff
@@ -73,13 +79,18 @@ Suspense.prototype._childDidSuspend = function(promise, suspendingComponent) {
 	};
 
 	const onSuspensionComplete = () => {
-		if (!--c._suspensions) {
+		if (!--c._pendingSuspensionCount) {
 			c._vnode._children[0] = c.state._suspended;
 			c.setState({ _suspended: (c._detachOnNextRender = null) });
+
+			let suspended;
```
andrewiggins: I didn't spend much time golfing this so there might be a smaller alternative

compat/src/suspense.js
```diff
@@ -73,13 +79,18 @@ Suspense.prototype._childDidSuspend = function(promise, suspendingComponent) {
...
+			while ((suspended = c._suspenders.pop())) {
+				suspended.forceUpdate();
```
andrewiggins: Note: I think this is a no-op if no sCU component is between Suspense and Lazy cuz `process()` will sort the render such that Suspense and all its children are diffed first, then when the Lazy forceUpdate are handled, the component will no longer be marked "_dirty" so they will be skipped

compat/test/browser/suspense.test.js
```diff
@@ -55,7 +56,7 @@ export function createSuspender(DefaultComponent) {
 		}
 
 		render(props, state) {
-			return state.Lazy ? h(state.Lazy, {}) : h(DefaultComponent, {});
+			return state.Lazy ? h(state.Lazy, props) : h(DefaultComponent, props);
```
sventschui: Are props really meant to be passed down?

andrewiggins: That was my intention. Mimics what React.lazy allows: https://codesandbox.io/s/reactlazy-prop-passing-b8txb

PR #2130, opened 2019-11-15
Remove unnecessary string concatenation

mhmdanas:


developit:
This is specifically done to avoid the test harness and/or Babel modifying the symbol value.

mhmdanas:
Since this string concatenation has a not so obvious purpose, wouldn't it be good to add a comment for later reference?

marvinhagemeister:
Happy to accept PR's that add a comment here 👍

mhmdanas:
Opened #2369 to add a clarification comment! :tada:

PR #2131, opened 2019-11-15
Rename useMemo _callback to _factory (+0 B)

andrewiggins:
While re-reading this code, the name `_callback` kept tripping me up because I thought it was something that needed to be invoked after some asynchronous event. But then I realized I was just misunderstanding this parameter. So I thought I'd rename it to something that more explicitly states the intention of what this parameter is: a factory that produces a value we remember.

This PR also includes an improvement to useMemo types I copied from DefinitelyTyped (link in comments)

Thoughts?


hooks/src/index.d.ts
```diff
@@ -106,7 +106,8 @@ export function useCallback<T extends Function>(callback: T, inputs: Inputs): T;
  * This optimization helps to avoid expensive calculations on every render.
  * If no array is provided, a new value will be computed whenever a new function instance is passed as the first argument.
  */
-export function useMemo<T>(factory: () => T, inputs?: Inputs): T;
+// for `inputs`, allow undefined, but don't make it optional as that is very likely a mistake
+export function useMemo<T>(factory: () => T, inputs: Inputs | undefined): T;
```
andrewiggins: I copied this from [the react types for useMemo](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/86ba65935681b62875e2d5267e8ae9b409935f89/types/react/index.d.ts#L1032)

PR #2132, opened 2019-11-15
Rewrite some tests to no longer rely on internals

andrewiggins:
This is a stepping stone to allowing us to run tests against minified builds

* Rewrite event tests to not rely on internal _listeners
* Fix lazy ref test to not rely on internals
* Rewrite teardown tests to not rely on internals
* Remove src import from test helpers


test/browser/components.test.js
```diff
@@ -1809,24 +1795,6 @@ describe('Components', () => {
 		expect(child._vnode._dom).to.equalNode(child.base);
 	});
 
-	it('should update old dom on forceUpdate in a lifecycle', () => {
```
andrewiggins: I just moved this test into the `forceUpdate` describe section below.

test/browser/components.test.js
```diff
@@ -5,15 +5,15 @@ import {
...
+	serializeHtml,
+	sortAttributes,
+	spyAll
 } from '../_util/helpers';
 import { div, span, p } from '../_util/dom';
 
 /** @jsx createElement */
 const h = createElement;
 
-let spyAll = obj => Object.keys(obj).forEach(key => sinon.spy(obj, key));
```
andrewiggins: Removed this copy of the `spyAll` function and replaced it with the same function declared in `helpers.js`. Same for `sortAttributes` below

PR #2135, opened 2019-11-18
Optimize createElement()

developit:
Avoiding shape mutation on props seems to give a 20% improvement in benchmarks.

### DIFF
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -1,5 +1,4 @@
 import options from './options';
-import { assign } from './util';
 
 /**
  * Create an virtual node (used for JSX)
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -30,8 +30,9 @@ export function createElement(type, props, children) {
 	// Note: type may be undefined in development, must never error here.
 	if (typeof type === 'function' && type.defaultProps != null) {
 		for (i in type.defaultProps) {
-			if (typeof normalizedProps[i] === 'undefined')
+			if (normalizedProps[i] === undefined) {
 				normalizedProps[i] = type.defaultProps[i];
+			}
 		}
 	}
 
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -10,32 +10,38 @@ import { assign } from './util';
  * @returns {import('./internal').VNode}
  */
 export function createElement(type, props, children) {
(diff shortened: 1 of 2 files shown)


src/create-element.js
```diff
@@ -10,32 +9,38 @@ import { assign } from './util';
  * @returns {import('./internal').VNode}
  */
 export function createElement(type, props, children) {
-	props = assign({}, props);
+	let normalizedProps = {},
+		i;
+	for (i in props) {
```
JoviDeCroock: This is clever

src/create-element.js
```diff
@@ -10,32 +9,38 @@ import { assign } from './util';
...
-	// "type" may be undefined during development. The check is needed so that
-	// we can display a nice error message with our debug helpers
-	if (type != null && type.defaultProps != null) {
-		for (let i in type.defaultProps) {
-			if (props[i] === undefined) props[i] = type.defaultProps[i];
+	// If a Component VNode, check for and apply defaultProps
+	// Note: type may be undefined in development, must never error here.
+	if (typeof type === 'function' && type.defaultProps != null) {
+		for (i in type.defaultProps) {
+			if (typeof normalizedProps[i] === 'undefined')
```
JoviDeCroock: can't we just do === undefined or is typeof cheaper?

developit: I believe in this case we can use === yes, I actually was testing this and forgot to remove it.

test/shared/createElement.test.js
```diff
@@ -44,8 +44,8 @@ describe('createElement(jsx)', () => {
 	});
 
 	it('should set VNode#key property', () => {
-		expect(<div />).to.have.property('key').that.is.undefined;
-		expect(<div a="a" />).to.have.property('key').that.is.undefined;
+		expect(<div />).to.have.property('key').that.is.empty;
+		expect(<div a="a" />).to.have.property('key').that.is.empty;
```
andrewiggins: I think these assertions want to be `.to.not.exist`  which would pass for `null` or `undefined`

developit: whoops, I literally read that in the docs and then wrote `.empty`. Tired me is tired!

PR #2136, opened 2019-11-18
Always use ==/!= for typeof instead of ===/!==

### LINKED ISSUE #2129: Force consistent use of double or triple equal/not equal in typeofs
Reading through the source code, I find places where `==` or `!=` are used for `typeof`, and others where `===` or `!==` are used. I believe we should choose one and stick with it. I'd be happy to open a PR for this once there is agreement on which choice should be used.

**Note**: looking at the produced bundles, it seems that with `typeof`, `===` is compiled to `==`, and `!==` is compiled to `!=`. So, there shouldn't really be any differences in performance and bundle size.

mhmdanas:
Bundle size is not affected since the minifier compiles `===` to `==` and `!==` to `!=` with `typeof`s.
Fixes #2129.

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -13,9 +13,7 @@ Component.prototype.isReactComponent = {};
 
 /* istanbul ignore next */
 export const REACT_ELEMENT_TYPE =
-	(typeof Symbol !== 'undefined' &&
-		Symbol.for &&
-		Symbol.for('react.element')) ||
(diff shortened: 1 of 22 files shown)


developit:
todo: benchmark

marvinhagemeister:
Closing, the PR seems to be abandoned and we can't merge it due to merge conflicts.

PR #2137, opened 2019-11-18
Reuse if-condition (-3 B)

marvinhagemeister:
Just a random golf opportunity I found while skimming our code. By reusing the previous if-condition we can skip a block.

Saves `3 B` :tada:


marvinhagemeister:
Nevermind, that statement is actually needed in case the `excessDomChildren` logic kicks in. In that case `dom` may not be `null` and may need a different code path. Bummer!

Issue #2126, opened 2019-11-14
Automating the release process


marvinhagemeister:
Our [current release process](https://github.com/preactjs/preact/blob/master/CONTRIBUTING.md#releasing-preact) involves a few manual steps which could be automated away. I'm hoping that having less manual tasks will allow us to release more frequently and ship code to users quicker.

Back when I joined the team there was a cycle of roughly 1 release per quarter, which we managed to get down to a release every 1-2 weeks (depending on the amount of changes). This is a substantial improvement and it's something I'm proud of the team that we were able to do that :+1:

Despite that, cutting a release usually takes around 30-60min which is more than I'd like. I'm confident that we make that a lot less time consuming. This is complicated by having 2 release lines (8.x + 10.x), where one needs to be careful to empty the `dist` folders when switching between them. For this reason I have 2 separate checkouts locally, but that's obviously more of a workaround rather than a long term solution.

## Long term goal

Ideally we should have either a button or a git tag trigger a release. On top of that it would be desirable to have a canary pipeline which publishes to `preact@next` and is triggered when a PR is merged into `master`. That way users that are highly dependent on a specific PR in `master` can immediately benefit from the latest changes at the slight risk of running into bugs.

## Current process

Let's look at the status quo, to have a better grasp of the necessary steps:

1. Checkout `master` or `8` branch
2. Ensure all old build artifacts are deleted
3. Run `npm run build`
4. Publish via `npm publish`
5. Grab changelog via `changelogged` (should be automated!!)
6. Create release + tag via GitHub's UI

## Should we stay with TravisCI?

Initially I planned to play around with GitHub's new Action feature over the coming weekend, but there are costs associated with it, if we'd go that route. Right now the free plan has 3000 free minutes and our average build time is 3 min. ~~With a rough approximation this would result in having an average of 10 builds per day. Right now we fall into that category, but I do expect us to continue to grow like we did in the past 2 months. Paying per user would very likely blow past our available budget.~~

~~So overall I'm leaning towards staying with Travis, although I do like the GH Actions UI more.~~

**EDIT:** My math was way off, it's approx. 30 builds per day which is totally fine. 

## Preact user account

We already have a `preact` user account that is theoretically able to publish packages to npm. @developit had a lot of foresight in the earlier days of Preact and registered it. We could leverage those for our automation task.

## Next steps (in no particular order)

- Add option to pass GH token via an argument to `changelogged`
- Figure out the Travis yaml syntax for listening on specific branches for specific tags
[...]

PR #2138, opened 2019-11-18
Switch to GitHub Actions

marvinhagemeister:
See #2126 for more details.


.github/build-test.yml
```diff
@@ -0,0 +1,30 @@
+name: Node CI
```
sventschui: You need to put this file into .github/workflows folder ;)

marvinhagemeister: Whoops! :sweat_smile:

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,31 @@
...
+          user: $SAUCE_USERNAME
+          key: $SAUCE_ACCESS_KEY
+      - name: npm install, build, and test
+        run: |
+          npm ci
+          npm run build --if-present
+          COVERAGE=true npm test
+          ./node_modules/coveralls/bin/coveralls.js < ./coverage/lcov.info;
+        env:
+          CI: true
```
developit: Adding [this](https://github.com/babel/preset-modules/blob/dff636d75ca8760a4882c7b65252fe443546c242/.github/workflows/main.yml#L26-L32) as the last action and removing the current SauceLabs setup should work:

```
    - name: saucelabs
      env:
        CI: true
        SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
        SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
      run: |
        npm run test:karma
```

We will just need to change the `test` script to skip karma tests in CI.

developit:
I'm hoping Actions also gives us nicer ways to put file size on PRs. I looked into doing this a while back but it was more cumbersome than I'd expected.

PR #2141, opened 2019-11-20
Remove _unmount in mangle files

daybrush:
If I used preact 10.0.5 and `createPortal` and change the parent, I have a problem.

스크린샷 2019-11-20 오후 1 57 02
스크린샷 2019-11-20 오후 1 57 17

This function is also exported to index.js.
https://github.com/preactjs/preact/blob/master/src/index.js

스크린샷 2019-11-20 오후 2 21 44

However, the mangle tool renamed `_unmount` to `_e`.

스크린샷 2019-11-20 오후 2 23 21

### DIFF
--- a/compat/mangle.json
+++ b/compat/mangle.json
@@ -24,8 +24,7 @@
       "$_diff": "__b",
       "$_render": "__r",
       "$_hook": "__h",
-      "$_catchError": "__e",
-      "$_unmount": "_e"
+      "$_catchError": "__e"
     }
   }
 }
--- a/hooks/mangle.json
+++ b/hooks/mangle.json
@@ -19,8 +19,7 @@
       "$_renderCallbacks": "__h",
       "$_suspensions": "__u",
       "$_hook": "__h",
-      "$_catchError": "__e",
-      "$_unmount": "_e"
+      "$_catchError": "__e"
     }
   }
 }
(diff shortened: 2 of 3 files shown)


cristianbote:
Hey @daybrush, thanks for looking into this! Is it possible that some imports are messed up? Could you share a codesandbox/repo with the reproducing issue?

daybrush:
@cristianbote 

Beacause of rollup problem, only `preact/compat` is resolved.

https://codesandbox.io/s/festive-booth-s1gc9

https://s1gc9.csb.app/


https://github.com/daybrush/moveable/blob/master/src/InnerMoveable.tsx#L14

cristianbote:
Hey @daybrush, that was deep one 😄 but I think I know what the issue is. So `preact/compat` should be defined as external in here https://github.com/daybrush/moveable/blob/master/rollup.config.js#L18 if not, in your bundle the `preact/compat` package will be inlined. So, the issue you're running into is actually the inlined version from `moveable` is trying to call `_unmount` which is mangled. `_unmount` is mangled inside `preact/compat` as well, so if you're using the external `preact/compat` you shouldn't be running into issues.

I did a dry-run locally, and it seems to be working fine -- but you're more aware of your codebase, so give it a go!

Let me know if this works out for you. Ands thanks for providing more details!

daybrush:
@cristianbote 

You are right. It works well as you say. Thank you.

cristianbote:
Awesome, glad it all worked out nicely! Again, thanks for opening this and giving it a go.

Cheers!

PR #2142, opened 2019-11-20
Move key and ref to src/internal.d.ts

### LINKED ISSUE #1953: styled-components ref type

38elements:
As same as `_parent` property,
It is not necessary that user know about `key` property and `ref` property of VNode.
I think it may be better to change `ref` and `key` to  `_ref` and `_key`.

### DIFF
--- a/src/internal.d.ts
+++ b/src/internal.d.ts
@@ -47,18 +47,20 @@ export interface VNode<P = {}> extends preact.VNode<P> {
 	// Redefine type here using our internal ComponentFactory type
 	type: string | ComponentFactory<P>;
 	props: P & { children: preact.ComponentChildren };
-	_children: Array<VNode<any>> | null;
-	_parent: VNode | null;
-	_depth: number | null;
+	key?: preact.Key;
+	ref?: preact.Ref<any> | null;
+	_children?: Array<VNode<any>> | null;
+	_parent?: VNode | null;
+	_depth?: number | null;
 	/**
 	 * The [first (for Fragments)] DOM child of a VNode
 	 */
-	_dom: PreactElement | Text | null;
+	_dom?: PreactElement | Text | null;
 	/**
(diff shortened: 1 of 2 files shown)


xiel:
Is there a test suite to validate this change against? Or are we just playing guessing game here?

38elements:
Is it a question, not an objection?

src/internal.d.ts
```diff
@@ -47,18 +47,20 @@ export interface VNode<P = {}> extends preact.VNode<P> {
...
 	 */
-	_dom: PreactElement | Text | null;
+	_dom?: PreactElement | Text | null;
 	/**
 	 * The last dom child of a Fragment, or components that return a Fragment
 	 */
-	_lastDomChild: PreactElement | Text | null;
-	_component: Component | null;
+	_lastDomChild?: PreactElement | Text | null;
+	_component?: Component | null;
```
38elements: Considering following code, I think it is necessary to add `?`.

```js
class Foo extends Component {
  render() {
    return {"type":"h1","props":{"children":"bar"}, constructor: undefined }
  }
}
```

38elements:
`vnode.key` and `vnode.ref` is needed at this line.

https://github.com/preactjs/preact/blob/c44ff298b8447d842431db30e65bf40991be1456/src/index.d.ts#L220

38elements:
`vnode.key` and `vnode.ref` is needed at this line.

https://github.com/preactjs/preact/blob/c44ff298b8447d842431db30e65bf40991be1456/src/index.d.ts#L220

Issue #2110, opened 2019-11-11
IE11 throws on input type="date"


marvinhagemeister:
Reported via slack:

>charlie 9:56 PM
going for a good head scratcher.. If I I have <input type="date" /> IE11 just breaks with a "Error: Invalid argument." error (changing the type to text brings it back to life). Fine in 8. While the type isn't supported, I'd expect IE to just gracefully degrade back to a text input.

Had a quick look into it and IE 11 throws when you try to use the setter directly with a value it doesn't support.

```js
// Throws
input.type = "date";

// Works
input.setAttribute("type", "date");
```

Rafi993:
Hi there. I would like to help in this issue. Forgive me for being noob in preact internals. What is this issue ?

marvinhagemeister:
The issue is that IE 11 throws an exception where it should not. If an input type is not supported it should automatically fall back to type text per spec. That's what IE 11 does with `input.setAttribute()` but not with `input.type`. So we need to make sure that we're calling `setAttribute` in our code.

Rafi993:
awesome thanks @marvinhagemeister

Rafi993:
@marvinhagemeister  I don't think input type "date" is supported in IE 11. It seems to fallback to a text input type "text" since IE 11 does not recognize date (in plain html file). https://stackoverflow.com/a/54846005/2445295 so we could do the same and fallback to input type text in IE 11 and show user appropriate warning in development mode.

JoviDeCroock:
Yes but when called directly in type="date" then IE will throw, when using setAttribute it will use the fallback instead of crash

JoviDeCroock:
You can find how we handle o ther of these edge cases in diffProps

JoviDeCroock:
This should be fixed by the next release thanks to #2147

PR #2147, opened 2019-11-23
Setting type attribue with setAttribute function

Rafi993:
This is done to prevent crash in IE11 as mentioned in this issue https://github.com/preactjs/preact/issues/2110

### DIFF
--- a/test/browser/render.test.js
+++ b/test/browser/render.test.js
@@ -91,6 +91,10 @@ describe('render()', () => {
 		expect(scratch.childNodes[0].nodeName).to.equal('SPAN');
 	});
 
+	it('should not throw error in IE11 with type date', () => {
+		expect(() => render(<input type="date" />, scratch)).to.not.throw();
+	});
+
 	it('should support custom tag names', () => {
 		render(<foo />, scratch);
 		expect(scratch.childNodes).to.have.length(1);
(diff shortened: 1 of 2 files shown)


Rafi993:
> we usually add a test to assert if this is done correctly to prevent future regressions

sorry. I'll add test for this too.

JoviDeCroock:
No need to say sorry, just notifying thanks for tackling this

marvinhagemeister:
That's awesome, thank you so much for the PR :+1: :tada: For tests you can take a look at the existing ones. We have a few where we check if an operation throws an error via `expect(() => ...).to.not.throw` or something similar :+1:

test/browser/render.test.js
```diff
@@ -91,6 +91,13 @@ describe('render()', () => {
 		expect(scratch.childNodes[0].nodeName).to.equal('SPAN');
 	});
 
+	it('should create input tag with type date', () => {
+		render(<input type="date" />, scratch);
+		expect(scratch.childNodes).to.have.length(1);
+		expect(scratch.childNodes[0].nodeName).to.equal('INPUT');
+		expect(scratch.childNodes[0].type).to.equal('date');
```
marvinhagemeister: I think we should test something that more closely resembles the original bug report. The fix in this PR specifically addresses the case where setting `input.date` would throw an error in IE11. The test case should reflect only that. On our CI our test suite will be run in various browsers, including IE11. So we only need to check if it throws or not.

```js
it('should not throw in IE11 with type date', () => {
  expect(() => render(<input type="date" />, scratch)).to.not.throw();
});
```

Rafi993:
Thanks for the help @marvinhagemeister  @JoviDeCroock you guys are awesome !!!

PR #2148, opened 2019-11-24
Import preact devtools renderer code

marvinhagemeister:
This is an integration for our [custom devtools](https://github.com/preactjs/preact-devtools) :tada: 

Screenshot from 2019-11-24 22-34-26

**Tasks:**

- [x] Convert tests written in TS to JS
- [x] Fix infinite loop in `jsonify` with some `Suspense` components in demo app
- [x] Fix `createContext` value always parsed as an `object` 
- [x] Code Coverage :grimacing:

### DIFF
--- a/debug/src/devtools/10/IdMapper.js
+++ b/debug/src/devtools/10/IdMapper.js
@@ -1,0 +1,82 @@
+/**
+ * VNode relationships are encoded as simple numbers for the devtools. We use
+ * this function to keep track of existing id's and create new ones if needed.
+ * @returns {import('./types').IdMapper}
+ */
+export function createIdMapper() {
+	/** @type {Map<any, number>} */
+	const instToId = new Map();
+
+	/** @type {Map<number, import('../../internal').VNode>} */
+	const idToVNode = new Map();
+
+	/** @type {Map<number, any>} */
+	const idToInst = new Map();
+
+	let uuid = 1;
+
+	/** @type {import('./types').IdMapper["getVNode"]} */
+	const getVNode = id => idToVNode.get(id) || null;
+
+	/** @type {import('./types').IdMapper["hasId"]} */
+	const hasId = vnode => {
+		if (vnode != null) {
+			return instToId.has(getInstance(vnode));
+		}
+		return false;
+	};
+
+	/** @type {import('./types').IdMapper["getId"]} */
+	const getId = vnode => {
+		if (vnode == null) return -1;
+		const inst = getInstance(vnode);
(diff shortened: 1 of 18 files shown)


marvinhagemeister:
I'm unsure if I should move it to `preact/devtools` with this PR or with a separate one. Leaving it in `preact/debug` for now would have the advantage of keeping the existing imports the same for users.

marvinhagemeister:
I think this PR is ready to review since a few days. If there is anything to make reviewing it easier, please reach out.

debug/src/devtools/10/IdMapper.js
```diff
@@ -0,0 +1,85 @@
...
+	/** @type {Map<any, number>} */
+	const instToId = new Map();
+
+	/** @type {Map<number, import('../../internal').VNode>} */
+	const idToVNode = new Map();
+
+	/** @type {Map<number, any>} */
+	const idToInst = new Map();
+
+	let uuid = 1;
```
JoviDeCroock: NIT: can we call this id instead, `uuid` feels like the long string

debug/src/devtools/10/options.js
```diff
@@ -0,0 +1,68 @@
...
+	let prevBeforeUnmount = options.unmount;
+	let prevBeforeDiff = options._diff;
+	let prevAfterDiff = options.diffed;
+
+	options.vnode = vnode => {
+		// Tiny performance improvement by initializing fields as doubles
+		// from the start. `performance.now()` will always return a double.
+		// See https://github.com/facebook/react/issues/14365
+		// and https://slidr.io/bmeurer/javascript-engine-fundamentals-the-good-the-bad-and-the-ugly
+		vnode.startTime = NaN;
```
JoviDeCroock: Today I learned, thanks!

debug/src/devtools/10/vnode.js
```diff
@@ -0,0 +1,95 @@
+import { Fragment } from 'preact';
+
+/**
+ * Check if a `vnode` is the root of a tree
+ * @param {import('../../internal').VNode} vnode
+ * @returns {boolean}
+ */
+export function isRoot(vnode) {
+	return vnode._parent == null && vnode.type === Fragment;
```
JoviDeCroock: Makes me wonder if we need a special indicator for this node

marvinhagemeister: Yeah, true, we may need something to mark root nodes as roots instead of plain `Fragments` in the future.

PR #2156, opened 2019-11-27
Initialize compat options when relevant functions are called

andrewiggins:
With this change, the only pieces left that [`agadoo`](https://npm.im/agadoo) can't treeshake out of compat are Suspense and SuspenseList because they inherit from Component, which they need to do because they call `setState` :/

Total Change:
compat.js.gz: +24 B
preact.js.gz: +5 B

Related: #1529

### DIFF
--- a/compat/src/PureComponent.js
+++ b/compat/src/PureComponent.js
@@ -1,4 +1,4 @@
-import { Component } from 'preact';
+import { Component } from './Component';
 import { shallowDiffers } from './util';
 
 /**
--- a/compat/src/PureComponent.js
+++ b/compat/src/PureComponent.js
@@ -4,16 +4,33 @@ import { shallowDiffers } from './util';
 /**
  * Component class with a predefined `shouldComponentUpdate` implementation
  */
-export class PureComponent extends Component {
-	constructor(props) {
-		super(props);
-		// Some third-party libraries check if this property is present
-		this.isPureReactComponent = true;
-	}
+// export class PureComponent extends Component {
+// 	constructor(props) {
+// 		super(props);
+// 		// Some third-party libraries check if this property is present
+// 		this.isPureReactComponent = true;
+// 	}
+//
+// 	shouldComponentUpdate(props, state) {
+// 		return (
+// 			shallowDiffers(this.props, props) || shallowDiffers(this.state, state)
+// 		);
+// 	}
+// }
 
-	shouldComponentUpdate(props, state) {
-		return (
-			shallowDiffers(this.props, props) || shallowDiffers(this.state, state)
-		);
-	}
+export function PureComponent(props, context) {
+	this.props = props;
+	this.context = context;
+
+	// Some third-party libraries check if this property is present
+	// this.isReactComponent = {};
+	// this.isPureReactComponent = true;
 }
+
+PureComponent.prototype.isReactComponent = {};
+PureComponent.prototype.isPureReactComponent = true;
+PureComponent.prototype.setState = Component.prototype.setState;
+PureComponent.prototype.forceUpdate = Component.prototype.forceUpdate;
+PureComponent.prototype.shouldComponentUpdate = function(props, state) {
+	return shallowDiffers(this.props, props) || shallowDiffers(this.state, state);
+};
--- a/compat/src/TODO.md
+++ b/compat/src/TODO.md
@@ -1,0 +1,54 @@
+# Tree shaking in compat TODO
+
+## Classes
+
+### The problem
+
+In order to enable tree-shaking, we need to transform our classes into a tree-shakeable form.
+
+If a class doesn't use inheritance, then the standard prototype assignments should be tree-shakeable. However classes that inherit (or manually inherit using `Suspense.prototype = new Component()`) are not tree-shakeable by default. BabelJS compiles classes into function closures with a `/*#__PURE__*/` comment informing tree-shakers that this closure can be safely removed if unused.
+
+However for Preact, since we bundle our code output, rollup sees these classes as used (because they are exported) and terser minifies the output and removes the comments. This means our classes don't get compiled away if a consumer of Preact doesn't use them because of the lack of a `/*#__PURE__*/` comment.
+
+### Possible solutions
+
+One option is to modify terser to add a new option to keep `/*#__PURE__*/` comments in source (probably pretty difficult...).
+
+Another idea would be to manually code the classes to be tree-shakeable and not use inheritance.
+
+Some code I tried out:
+
+```js
+// portals.js
+var ContextProvider = function ContextProvider() {};
+ContextProvider.prototype.getChildContext = function getChildContext() {
+	return this.props.context;
+};
+ContextProvider.prototype.render = function render(props) {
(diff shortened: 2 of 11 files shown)


compat/src/TODO.md
```diff
@@ -0,0 +1,54 @@
...
+
+```js
+// portals.js
+var ContextProvider = function ContextProvider() {};
+ContextProvider.prototype.getChildContext = function getChildContext() {
+	return this.props.context;
+};
+ContextProvider.prototype.render = function render(props) {
+	return props.children;
+};
```
andrewiggins: Defining ContextProvider like this saves us 6 bytes

compat/src/TODO.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+```js
+// PureComponent.js
+export function PureComponent() {
+	this.isPureReactComponent = true;
+}
+PureComponent.prototype.shouldComponentUpdate = function(props, state) {
+	return shallowDiffers(this.props, props) || shallowDiffers(this.state, state);
+};
```
andrewiggins: Defining PureComponent like this saves us 11 bytes on top of the 6 saved from rewriting ContextProvider

compat/src/TODO.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+```js
+// suspense.js
+
+// Comment out the following line to remove a side-effect
+// Suspense.prototype = new Component();
```
andrewiggins: Looking at this again, I think I will just go all the way and remove the inheritance from Suspense and SuspenseList. If that's important for integrating with 3rd party ecosystem libraries, can we recommend using the full compat (i.e. using `render` & `createElement` from `preact/compat` which installs the full ReactCompat option suite)?

compat/src/forwardRef.js
```diff
@@ -1,16 +1,5 @@
...
 
-let oldVNodeHook = options.vnode;
-options.vnode = vnode => {
-	if (vnode.type && vnode.type._forwarded && vnode.ref) {
-		vnode.props.ref = vnode.ref;
-		vnode.ref = null;
-	}
-
-	if (oldVNodeHook) oldVNodeHook(vnode);
-};
```
andrewiggins: Moving this code from compat to core shaves 47 bytes out of compat and only adds 6 bytes to core so I thought it was worth it.

andrewiggins: And if we didn't install react compat in Component's constructor this entire PR would be free for compat - no bytes added lol. So close!

src/create-element.js
```diff
@@ -10,9 +10,19 @@ import options from './options';
...
 export function createElement(type, props, children) {
 	let normalizedProps = {},
-		i;
+		i,
+		ref;
+
 	for (i in props) {
-		if (i !== 'key' && i !== 'ref') normalizedProps[i] = props[i];
+		if (i == 'ref') {
+			if (type && type._forwarded) {
```
andrewiggins: See comment in `forwardRef.js` for why this was moved to core

compat/src/TODO.md
andrewiggins: Err just realized PureComponent has to inherit from Component so consumers have access to `setState` so scratch this idea

compat/src/TODO.md
```diff
@@ -0,0 +1,54 @@
+# Tree shaking in compat TODO
```
andrewiggins: I'll remove this file before checking in. Just left it here for discussion

src/create-element.js
porfirioribeiro: I have this exact same code in a local branch
So i could use the same mechanism on the composition API
And get rid of options.vnode also

JoviDeCroock: I kinda dislike the notion of mixing concerns, preact-core has no notion of forwarding props so I kinda dislike putting that here, if I'm the only one however feel free to ignore that

andrewiggins: Yea, I agree that I'd prefer this live in compat but the byte savings were too good to pass up I thought lol

andrewiggins:
Todo: look at using the regex comments option in Terser to preserve PURE comments: https://github.com/terser/terser/blob/master/README.md#output-options

Thanks @sventschui for pointing this out to me!

andrewiggins:
@sventschui Seems like the `comments` option won't work. Terser always [removes the __PURE__ annotation even if you set it to keep all comments](https://git.io/Je16f). I also tried locally with a regex (`/#__PURE__/`) and it still removes `#__PURE__`

JoviDeCroock:
Yes Pure is removed when doing high level minification. Try building like this once: https://github.com/JoviDeCroock/hooked-form/blob/master/package.json#L12

sventschui:
Should be fixed in microbundle@next: https://github.com/developit/microbundle/commit/445fde2ff0bad178db978326c511ebf3691f6129 but the build crashes. Will investigate

andrewiggins:
Todo: per @developit suggestion, exploring using `enqueueRender` directly to remove need to call `setState` in compat components. However, probably won't work for `PureComponent` which consumers inherit from and expect full Component interface. Perhaps copy it from `Component` onto `PureComponent`? That may be cheaper than exposing `enqueueRender` from others as well.

src/create-element.js
marvinhagemeister: I'm really a fan of these savings, but I'm with @JoviDeCroock here. For as long we have the `core` vs `compat` separation it feels wrong to put code into `core` that is only needed for `compat`.

marvinhagemeister (commented):
I really like the direction this is going! Having a more tree-shakeable `compat` is a net benefit for everyone :+1: :100:

compat/src/PureComponent.js
```diff
@@ -1,19 +1,36 @@
...
+export function PureComponent(props, context) {
+	this.props = props;
+	this.context = context;
+
+	// Some third-party libraries check if this property is present
+	// this.isReactComponent = {};
+	// this.isPureReactComponent = true;
 }
+
+PureComponent.prototype.isReactComponent = {};
```
developit: a thought: could we use ES5 inheritance here?

```js
export function PureComponent(props, context) {
	// is this smaller?
	//Component.call(this, props, context);
	this.props = props;
	this.context = context;
}
PureComponent.prototype = new Component();
PureComponent.prototype.constructor = PureComponent;
PureComponent.prototype.isReactComponent = {};
PureComponent.prototype.isPureReactComponent = true;
PureComponent.prototype.shouldComponentUpdate = function(props, state) {
	return shallowDiffers(this.props, props) || shallowDiffers(this.state, state);
};
```

developit: Ah, nevermind, just saw your explanation.

developit:
Only thought I had: maybe we don't try to make Component tree-shakeable?  It's much more likely to be used than Suspense/PureComponent/etc, and it seems like a lot of the complexity could be saved if we could use `.prototype=new Component`.

JoviDeCroock:
Closing: targeted the old agadoo tree-shaking setup; v11 restructured compat with an ESM-only modern build, and the diff no longer applies.

PR #2158, opened 2019-11-28
Update perf tests for Preact X and enable on master

andrewiggins:
* Move performance test into bench folder
* Update performance testing thresholds to match Preact X
* Enable running perf tests on all builds to ensure we don't regress

If we start seeing builds flake out, we should probably adjust these thresholds or disable specific flakey tests


test/benchmarks/performance.test.js
```diff
@@ -234,7 +236,7 @@ describe('performance', function() {
 			},
 			({ ticks, message }) => {
 				console.log(`PERF: repeat diff: ${message}`);
-				expect(ticks).to.be.below(2000 * MULTIPLIER);
+				expect(ticks).to.be.below(3000 * MULTIPLIER);
```
andrewiggins: Per @developit, X is a little slower here cuz we no longer recycle DOM nodes

test/benchmarks/performance.test.js
```diff
@@ -298,7 +300,7 @@ describe('performance', function() {
 			},
 			({ ticks, message }) => {
 				console.log(`PERF: large VTree: ${message}`);
-				expect(ticks).to.be.below(2000 * MULTIPLIER);
+				expect(ticks).to.be.below(200 * MULTIPLIER);
```
andrewiggins: Both 8 and X are well below 2000 ticks here so adjusting to a more reasonable value

PR #2159, opened 2019-11-28
Suspense should support updating state of children while suspended

sventschui:
Suspense should support updating state of children while suspended.

Could not yet come up with a solution

### DIFF
--- a/compat/test/browser/suspense.test.js
+++ b/compat/test/browser/suspense.test.js
@@ -1218,4 +1218,55 @@ describe('suspense', () => {
 			expect(scratch.innerHTML).to.equal(`<div>A</div><div>B</div>`);
 		});
 	});
+
+	it('should support updating state while suspended', () => {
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
FYI: We can always merge the test when it's skipped until we find a solution. That's how we did it in the early days of X :)

Inviz:
I think I've stumbled upon the same issue - when the component that has Suspense in it updates while lazy component is being loaded, preact crashes on:

```const onSuspensionComplete = () => {
		if (!--c._suspensions) {
			c._vnode._children[0] = c.state._suspended; // < crashes here as _children is undefined
			c.setState({ _suspended: (c._detachOnNextRender = null) });
		}
	};```

JoviDeCroock:
Closing: test-only draft; the author noted no solution was found. compat Suspense was rewritten in v11.

PR #2179, opened 2019-12-11
Add component stack in debug warnings

### LINKED ISSUE #2177: Error from preact/debug for code wich doesnt exist?
Hello,

since Im using preact X with the debugger I get the following error:

> Calling "this.setState" inside the constructor of a component is a no-op and might be a bug in your application. Instead, set "this.state = {}" directly. 

Ive searched a lot in my codebase and didnt find any this.setState in a constructor neither online any solution for this error message.,

It would be nice if the debug mode would give me the name of the component wich causes the error.

marvinhagemeister:
This PR adds a component stack trace to errors and warning messages in `preact/debug`. See the following screenshot for an example.

Screenshot from 2019-12-11 22-48-04

Fixes #2177

*Tasks:*

- [x] Fix failing tests
- [x] Check if we could support proper component owners instead of parents
  - Note: `@babel/plugin-transform-react-jsx-self` doesn't support arrow functions, so it's a deal breaker

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,3 +1,5 @@
+const DEV = process.env.NODE_ENV !== 'production';
+
 module.exports = function(api) {
 	api.cache(true);
 
@@ -16,9 +18,10 @@ module.exports = function(api) {
 		],
 		plugins: [
 			'@babel/plugin-proposal-object-rest-spread',
+			DEV && '@babel/plugin-transform-react-jsx-source',
 			'@babel/plugin-transform-react-jsx',
 			'babel-plugin-transform-async-to-promises'
-		],
+		].filter(Boolean),
 		ignore: ['./dist']
 	};
 };
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,5 +1,4 @@
-const DEV = process.env.NODE_ENV !== 'production';
-
+// Only used for mocha tests. For karma, see karma.config.js
 module.exports = function(api) {
 	api.cache(true);
 
@@ -18,10 +17,9 @@ module.exports = function(api) {
 		],
 		plugins: [
 			'@babel/plugin-proposal-object-rest-spread',
-			DEV && '@babel/plugin-transform-react-jsx-source',
 			'@babel/plugin-transform-react-jsx',
 			'babel-plugin-transform-async-to-promises'
-		].filter(Boolean),
+		],
 		ignore: ['./dist']
 	};
 };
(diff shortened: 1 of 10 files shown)


babel.config.js
```diff
@@ -16,9 +18,10 @@ module.exports = function(api) {
 		],
 		plugins: [
 			'@babel/plugin-proposal-object-rest-spread',
+			DEV && '@babel/plugin-transform-react-jsx-source',
```
ForsakenHarmony: should probably mention somewhere that this plugin can help

debug/src/component-stack.js
```diff
@@ -0,0 +1,50 @@
...
+ * Return the component stack that was captured up to this point.
+ */
+export function getComponentStack() {
+	return stack.reduce((acc, vnode) => {
+		acc += `  in ${getDisplayName(vnode)}`;
+
+		const source = vnode.__source;
+		if (source) {
+			acc += ` (at ${source.fileName}:${source.lineNumber})`;
+		}
```
sventschui:
```suggestion
		} else if (!babelSourcePluginNoteWirtten) {
			babelSourcePluginNoteWirtten = true;
			console.warn('Add @babel/plugin-transform-react-jsx-source to get a more detailed component stack. Note that you should not add it to production builds of your App for bundle size reasons.');
		}
```

sventschui:
Just had a very short discussion with @marvinhagemeister on slack.

My idea is to collect the stack during `options.render` instead of `options._diff` to get the "owner" component.

Marvin pointed out the `@babel/plugin-transform-react-jsx-source` plugin which adds `_self={this}` property to JSX calls. This will also help us to determine the owner component as `this` should reference the component creating this vnode if I'm not mistaken.

There are some things we need to investigate though.

- As outlined by @marvinhagemeister we need to investigate what `this` references to in arrow function components.
- I'd personally prefer the `options.render` approach as this would also work when (a) not having the babel plugin installed and (b) not using JSX but `h(FooComponent)` directly

debug/src/component-stack.js
marvinhagemeister: Love this!

marvinhagemeister:
@sventschui Turns out that you were right all a long. Using `options._render` in tandem with `options.vnode` allows us to capture the correct owner stack. Now I feel bad for shooting down that idea way to quickly. Really glad that you nudged me in the right direction! :raised_hands:

marvinhagemeister:
Looks like coveralls has troubles connecting to Travis. I'll retry the CI in a couple of hours.

PR #2186, opened 2019-12-16
Fix componentWillReceiveProps when calling setState/forceUpdate

### LINKED ISSUE #2185: componentWillReceiveProps not being called after setState/forceUpdate
In Preact X, `componentWillReceiveProps` doesn't get called when calling `setState` or `forceUpdate` and receiving props, eg:

```js
state = { count: this.props.count };

componentWillReceiveProps(nextProps) {
  // componentWillReceiveProps never gets called
  if (nextProps.count) {
    this.setState({ count: nextProps.count });
  }
}

onChange = event => {
  // Call setState or forceUpdate
  this.setState({ unrelated: 'Update' });

  // Call onChange to update our parent which will give this component new props
  this.props.onChange(this.state.count + 1);
};
```

**Current Behaviour:**
Preact X Codesandbox:
https://codesandbox.io/s/preact-x-cwrp-8x1hi

**Expected Behaviour:**
Preact 8 Codesandbox:
https://codesandbox.io/s/preact-8-cwrp-7lh4j
React 16 Codesandbox:
https://codesandbox.io/s/react-16-cwrp-q8pu2

For some context on how I found this, when updating to Preact X, a third-party library [`rc-input-number`](https://www.npmjs.com/package/rc-input-number) was breaking in our codebase, I noticed that the problem was in the [`onChange` handler](https://github.com/react-component/input-number/blob/master/src/index.js#L274)

jamesb3ll:
Fixes #2185

`componentWillReceiveProps` doesn't get called when calling `setState` or `forceUpdate` and receiving props.

`componentWillReceiveProps` never gets called as `this.setState/forceUpdate` will set `component._force = false/true`, so `_force` will not be `null` or `undefined` as expected in [`diff/index.js`](https://github.com/preactjs/preact/blob/master/src/diff/index.js):
```js
if (
  newType.getDerivedStateFromProps == null &&
  c._force == null &&
  c.componentWillReceiveProps != null
) {
  c.componentWillReceiveProps(newProps, cctx);
}
```

Instead of using the `_force` flag, I've decided to check the equality of `oldProps` and `newProps` to check whether props have changed. I've added a test for this case too. It would be great to know if I'm missing an edge case with this fix?

This fix also drops the size by `-6B` 🎉

### DIFF
--- a/test/browser/lifecycles/componentWillReceiveProps.test.js
+++ b/test/browser/lifecycles/componentWillReceiveProps.test.js
@@ -183,6 +183,59 @@ describe('Lifecycle methods', () => {
 			expect(Inner.prototype.componentWillReceiveProps).to.have.been.called;
 		});
 
+		it('should be called when rerender with new props from parent even with setState/forceUpdate in child', () => {
+			let setStateAndUpdateProps;
+			let forceUpdateAndUpdateProps;
+			let cWRPSpy = sinon.spy();
+
+			class Outer extends Component {
+				constructor(p, c) {
+					super(p, c);
+					this.state = { i: 0 };
+					this.update = this.update.bind(this);
+				}
+				update() {
+					this.setState({ i: this.state.i + 1 });
+				}
+				render(props, { i }) {
+					return <Inner i={i} update={this.update} />;
+				}
+			}
+			class Inner extends Component {
+				componentDidMount() {
+					expect(this.props.i).to.be.equal(0);
+
+					setStateAndUpdateProps = () => {
+						this.setState({});
+						this.props.update();
+					};
+					forceUpdateAndUpdateProps = () => {
+						this.forceUpdate();
+						this.props.update();
+					};
+				}
+				componentWillReceiveProps(nextProps) {
+					cWRPSpy(nextProps.i);
+				}
+				render() {
+					return <div />;
+				}
+			}
+			// Initial render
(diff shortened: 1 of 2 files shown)


marvinhagemeister:
@JoviDeCroock This is not always true. For the component where `setState/forceUpdate` will be called on, the `props` will be referentially equal. The whole subtree will have newly created props objects due to the line you linked to tough.

jamesb3ll:
Thanks @JoviDeCroock 🙂I was also confused with that, but with the debugger I saw that the parent component (setting state or forceUpdate) kept the same equality for `newProps` and `oldProps` and that way we can avoid that component calling `componentWillReceiveProps`

I've been playing around in this codesandbox here: https://codesandbox.io/s/preact-x-cwrp-b1inz

test/browser/lifecycles/componentWillReceiveProps.test.js
```diff
@@ -183,6 +183,59 @@ describe('Lifecycle methods', () => {
 			expect(Inner.prototype.componentWillReceiveProps).to.have.been.called;
 		});
 
+		it('should be called when rerender with new props from parent even with setState/forceUpdate in child', () => {
```
marvinhagemeister: This test is so good 😍

marvinhagemeister:
FYI: We'll likely publish a new release in a couple of hours :+1:

Issue #2192, opened 2019-12-18
throwing in useEffect causes subsequent effects to queue and never fire


tyler-graham-ck:
After causing an exception in an effect an app will continue to render, however, subsequent `useEffect` calls queue but do not fire.

With a little investigation it appears that the [`afterPaint` function](https://github.com/preactjs/preact/blob/master/hooks/src/index.js#L270-L277) is getting into a weird state. The effects continue to get queued up so the length is never one and thus [`flushAfterPaintEffects`](https://github.com/preactjs/preact/blob/master/hooks/src/index.js#L228-L237) is not called (while tracing in my actual code the raf always appeared to be undefined).

[Here's a codesandbox reproducing the issue](https://codesandbox.io/s/zealous-snow-54q6c) and has some further exposition.

*Note: codesandbox seems to have issues resolving preact/compat sometimes. Hard reload the sandbox when this happens*

tyler-graham-ck:
[Here's the same code example using React instead](https://codesandbox.io/s/heuristic-bush-yk7nm).

It performs the way I would expect. Without an error boundary the whole app crashes, but with one it recovers and subsequent effects fire properly.

PR #2193, opened 2019-12-18
bailout when hook throws an error

JoviDeCroock:
fixes: https://github.com/preactjs/preact/issues/2192

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -53,7 +53,8 @@ options._commit = (vnode, commitQueue) => {
 			);
 		} catch (e) {
 			options._catchError(e, component._vnode);
-			return true;
+			if (oldCommit) oldCommit(vnode, commitQueue);
+			throw e;
 		}
 	});
 	if (oldCommit) oldCommit(vnode, commitQueue);
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -246,9 +246,14 @@ export function useErrorBoundary(cb) {
 function flushAfterPaintEffects() {
 	afterPaintEffects.some(component => {
 		if (component._parentDom) {
-			component.__hooks._pendingEffects.forEach(invokeCleanup);
-			component.__hooks._pendingEffects.forEach(invokeEffect);
-			component.__hooks._pendingEffects = [];
+			try {
+				component.__hooks._pendingEffects.forEach(invokeCleanup);
+				component.__hooks._pendingEffects.forEach(invokeEffect);
+				component.__hooks._pendingEffects = [];
+			} catch (e) {
+				options._catchError(e, component._vnode);
+				return true;
+			}
 		}
 	});
 	afterPaintEffects = [];
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -52,11 +52,14 @@ options._commit = (vnode, commitQueue) => {
 				cb._value ? invokeEffect(cb) : true
 			);
 		} catch (e) {
+			commitQueue.some(c => {
+				if (c._renderCallbacks) c._renderCallbacks = [];
+			});
+			commitQueue = [];
 			options._catchError(e, component._vnode);
-			if (oldCommit) oldCommit(vnode, commitQueue);
-			throw e;
 		}
 	});
+
 	if (oldCommit) oldCommit(vnode, commitQueue);
 };
 
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -46,12 +46,16 @@ options.diffed = vnode => {
 
 options._commit = (vnode, commitQueue) => {
 	commitQueue.some(component => {
-		component._renderCallbacks.forEach(invokeCleanup);
-		component._renderCallbacks = component._renderCallbacks.filter(cb =>
-			cb._value ? invokeEffect(cb) : true
-		);
+		try {
+			component._renderCallbacks.forEach(invokeCleanup);
+			component._renderCallbacks = component._renderCallbacks.filter(cb =>
+				cb._value ? invokeEffect(cb) : true
+			);
+		} catch (e) {
+			options._catchError(e, component._vnode);
+			return true;
+		}
 	});
-
 	if (oldCommit) oldCommit(vnode, commitQueue);
 };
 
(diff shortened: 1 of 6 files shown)


hooks/src/index.js
```diff
@@ -228,9 +228,14 @@ export function useDebugValue(value, formatter) {
...
-			component.__hooks._pendingEffects.forEach(invokeCleanup);
-			component.__hooks._pendingEffects.forEach(invokeEffect);
-			component.__hooks._pendingEffects = [];
+			try {
+				component.__hooks._pendingEffects.forEach(invokeCleanup);
+				component.__hooks._pendingEffects.forEach(invokeEffect);
+				component.__hooks._pendingEffects = [];
+			} catch (e) {
+				options._catchError(e, component._vnode);
+				return true;
```
ForsakenHarmony: react bails as well?

JoviDeCroock: Yes, the error is thrown up so no execution will continue. This return is mainly so we just clean up our global variable `afterPaintEffects`

tyler-graham-ck:
oh snap, thanks, was gonna clone after lunch 😉 

does it make sense to catch in [`__render`](https://github.com/preactjs/preact/blob/master/hooks/src/index.js#L26-L30), [`unmount`](https://github.com/preactjs/preact/blob/master/hooks/src/index.js#L65-L67), etc too?

JoviDeCroock:
No, those are part of the diffing algorithm it are only the ones scheduled for delayed execution that bypass diff and don't get picked up by the parentComponents.

tyler-graham-ck:
ahh. if you get a chance can you link to where that happens?

JoviDeCroock:
I can surely do so

For the first case we do a synchronous cleanup for the effects still pending for a component this can happen with many subsequent renders, here we use the [_render option](https://github.com/preactjs/preact/blob/master/src/diff/index.js#L159) this is within the diffing loop so if a throw occurs it will catch it just as a normal throw in render would.

For the next case we have `unmount`, something that can only occur during diffing when we are diffing children and we see that one isn't there anymore we call [unmount](https://github.com/preactjs/preact/blob/master/src/diff/children.js#L204) which links through to [the option](https://github.com/preactjs/preact/blob/master/src/diff/index.js#L406) we are still in our diffing loop since `diffChildren` is called from the regular `diff` so if it happens to throw we'll and up in the same catch as the one mentioned above.

tyler-graham-ck:
ah ok, so then [`diffed`](https://github.com/preactjs/preact/blob/master/src/diff/index.js#L216) is just a callback to be executed when the diffing is done?

hooks/src/index.js
```diff
@@ -228,9 +228,14 @@ export function useDebugValue(value, formatter) {
 function flushAfterPaintEffects() {
 	afterPaintEffects.some(component => {
 		if (component._parentDom) {
-			component.__hooks._pendingEffects.forEach(invokeCleanup);
-			component.__hooks._pendingEffects.forEach(invokeEffect);
-			component.__hooks._pendingEffects = [];
+			try {
+				component.__hooks._pendingEffects.forEach(invokeCleanup);
+				component.__hooks._pendingEffects.forEach(invokeEffect);
```
robertknight: These callbacks are also invoked in the `_render` hook but in that case the exceptions are not caught.

JoviDeCroock: > For the first case we do a synchronous cleanup for the effects still pending for a component this can happen with many subsequent renders, here we use the _render option this is within the diffing loop so if a throw occurs it will catch it just as a normal throw in render would.

tyler-graham-ck: as I understand it `_render` is handled [here](https://github.com/preactjs/preact/blob/master/src/diff/index.js#L217-L219). the main issue isn't that there is an exception, but that `afterPaintEffects` is needs to be mutated here for the next flush to happen. the other invocations don't suffer from the same affliction

JoviDeCroock:
One last scenario is throwing in `useLayoutEffect` this can error out because we are reusing the `commitQueue` with objects (x._value()) instead of x(). So if a useLayoutEffectt throws that is not removed from the commitQueue and will just run the commitQueue afterwards (since this is part of the options hook)

This PR: https://github.com/preactjs/preact/pull/2203 can solve that more easily

hooks/src/index.js
ForsakenHarmony: why do we not have the try catch wrapping the loop itself?

JoviDeCroock: Then we lack the component in question or we cast that to an outer-scope variable

ForsakenHarmony: oh right, sorry

src/diff/index.js
```diff
@@ -227,19 +227,21 @@ export function diff(
...
-			c._renderCallbacks = [];
-			commitQueue.some(cb => {
-				cb.call(c);
-			});
-		} catch (e) {
-			options._catchError(e, c._vnode);
-		}
-	});
+	try {
+		if (options._commit) options._commit(root, commitQueue);
```
ForsakenHarmony: does it make sense to skip all of them for one error? sorry not super familiar with internals right now

JoviDeCroock: It does make sense but we need the ref to our current vnode, will refactor to use one try block

src/diff/index.js
```diff
@@ -228,7 +228,6 @@ export function diff(
  */
 export function commitRoot(commitQueue, root) {
 	if (options._commit) options._commit(root, commitQueue);
-
 	commitQueue.some(c => {
```
marvinhagemeister: The whitespace removal above here is probably unintentional :slightly_smiling_face:

cristianbote: I thought that prettier did this 🤔

JoviDeCroock: Oh yes sorry I did not notice this let me fix this

marvinhagemeister:
@robertknight with the new changes in this PR errors in `_commit` will be caught. Just checked with @JoviDeCroock regarding `_render` and imo it's "edge-casy" enough that we can move that to another PR. We only flush effects there on synchronous rendering. Example from our test suite:

```jsx
it('works with closure effect callbacks capturing props', () => {
  const values = [];

  function Comp(props) {
    useEffect(() => values.push(props.value));
    return null;
  }

  render(<Comp value={1} />, scratch);
  render(<Comp value={2} />, scratch);
  return scheduleEffectAssert(() => expect(values).to.deep.equal([1, 2]));
);
```

Issue #2168, opened 2019-12-05
Unmount not called for nested children on render (since preact X)


timon-witt:
I encountered a behaviour, that I think can be considered a bug, unless it was intentionally introduced with preact X.

This is a common way to force unmount of components in a specific container:
```js 
render(<div />, container, container.firstElementChild)
```

Using this on the following HTML with preact 8, results in componentWillUnmount of the component being properly called. (Note that the component can be as deeply nested inside the container as wished)
```html
<div id="container">
  <div>
    <MyComponent />
  </div>
</div>
```

With preact X however, the component is not unmounted, which is really bad for my understanding. There now is no way, that I know of, to achieve emptying a container and properly unmouting all components inside. This is specifically bad for single page applications, like mine, that do this frequently.

If the component is not nested, but a direct child of the container, it even works with preact X.
```html
<div id="container">
  <MyComponent />
</div>
```

Here I have two exactly similar codepens showing this behaviour, only different in the preact version. One runs with 8.5.2 and one with 10.0.5. 

preact 8: https://codepen.io/timonwitt/pen/NWPqwxO
preact X: https://codepen.io/timonwitt/pen/BayNwzW

timon-witt:
I made a pull request with unit tests that check this behaviour. One shows that it is successful as long as there is no nesting. The other one shows the failure if there is nesting.

JoviDeCroock:
I've commented on why this is correct behavior

timon-witt:
@JoviDeCroock So you are saying, that there is no way to clear a site (or any other container) of all preact components, if it is not entirely rendered with preact?

Maybe let me try to explain my particular problem: The project I'm working on uses many smaller preact components, that are spread all over the site. There is no preact root or a single large preact tree. And since it is a SPA, it relies on clearing the DOM, and all preact components inside of it, on page change. Because of that, this bug is hindering us to upgrade to preact X.

Imo, as long as preact's use case is not limited to rendering the whole page with it, there must be a way to properly unmount your components, like in preact 8.

JoviDeCroock:
Why can't you do `render(null, target)`? `ReplaceNode` is more or less meant to cover cases like [these](https://medium.com/@luke_schmuke/how-we-achieved-the-best-web-performance-with-partial-hydration-20fab9c808d5)

timon-witt:
Because `render(null, renderRoot)` doesn't work either.
In the example I only included `ReplaceNode` because it was necessary in preact 8.

[...]

PR #2195, opened 2019-12-19
#2168: Added unit tests to check proper component unmounting

timon-witt:
I added two tests that show the bug: https://github.com/preactjs/preact/issues/2168

### DIFF
--- a/test/browser/render.test.js
+++ b/test/browser/render.test.js
@@ -1379,6 +1379,74 @@ describe('render()', () => {
 			newScratch.parentNode.removeChild(newScratch);
 		});
 
+		it.only('should unmount existing components in prerendered HTML', () => {
+			const newScratch = setupScratch();
+			const unmount = sinon.spy();
+			const mount = sinon.spy();
+			class App extends Component {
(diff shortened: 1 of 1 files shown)


test/browser/render.test.js
```diff
@@ -1379,6 +1379,74 @@ describe('render()', () => {
...
+				componentWillUnmount() {
+					unmount();
+				}
+
+				render() {
+					return <span>App</span>;
+				}
+			}
+
+			// The only difference to the previous test is the extra div here
```
JoviDeCroock: Yes but this `<div>` is not part of your vnode tree since you start underneath and then target it with `newScratch.firstElementChild` did this work with Preact 8? This isn't supported in Preact X atleast since this falls outside your render root

JoviDeCroock:
I don't mind merging in the first test, the second has been discussed on your issue right? Do you think we could do more to help out in this case?

timon-witt:
Yes, the issue has been discussed and there is a reasonable workaround for now. Keeping the first test would still be great like you said. Thank's for the support!

PR #2196, opened 2019-12-19
[WIP] Bugfix: Component's default render() should render nothing

developit:
Currently it uses Fragment as its implementation to reduce size, but Fragment renders `props.children`. This means that components inheriting from Component without providing their own render() method render their children by default.

### DIFF
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -1,4 +1,4 @@
-import { Component, createElement, options } from 'preact';
+import { Component, createElement, options, Fragment } from 'preact';
 import { assign } from './util';
 
 const oldCatchError = options._catchError;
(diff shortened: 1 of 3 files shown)


JoviDeCroock (approved):

cristianbote (approved):

sventschui:
Can we add a test for that so we don‘t regress when golfing? I might find an hour or two today to add one

developit:
Tests are failing, and I think it's because Suspense assumed `Component.prototype.render` would render children.

JoviDeCroock:
Is this still WIP? Maybe we can merge it in

developit:
I think we can merge. Only thing I was wondering is whether we wanted to hoist the function up as a constant (`EMPTY_FUNC`). Potentially useful for null callbacks?

JoviDeCroock:
> I think we can merge. Only thing I was wondering is whether we wanted to hoist the function up as a constant (`EMPTY_FUNC`). Potentially useful for null callbacks?

Feels good to me

cristianbote:
@developit do you still wanna add the `EMPTY_FUNC` in this PR? Or is it ready? Can be added later as well.

JoviDeCroock:
Closing this as stale

PR #2203, opened 2019-12-22
Move hooks to core

JoviDeCroock:
Before: 
- hooks: 891
- core: 3629

Total: 4520B

After: 
- hooks: 171
- core: 4308

Total: 4479B

After (useDebug and useImperative to /hooks):
- hooks: 251
- core: 4229

Total: 4480B

> Note: this has to be refactored for `useErrorBoundary` but I'm gong to wait for more comments on this. If we decide to expand functionality of hooks we should be wary of this (since size++)


src/hooks.js
```diff
@@ -0,0 +1,251 @@
...
+
+/**
+ * @param {object} ref
+ * @param {() => object} createHandle
+ * @param {any[]} args
+ */
+export function useImperativeHandle(ref, createHandle, args) {
+	useLayoutEffect(
+		() => {
+			if (typeof ref === 'function') ref(createHandle());
```
porfirioribeiro: Would it be possible to reuse the logic from here:
https://github.com/preactjs/preact/blob/master/src/diff/index.js#L387

JoviDeCroock:
> In other hand, it brings hooks logic into core, that might not be used in some cases

This should be close to a noop (maybe 50B at max) since when unused that whole file can be tree-shaken

porfirioribeiro:
> > In other hand, it brings hooks logic into core, that might not be used in some cases
> 
> This should be close to a noop (maybe 50B at max) since when unused that whole file can be tree-shaken

Ohh yeah, that's great then!
Maybe i will even be able to use some of that logic on composition side

PR #2205, opened 2019-12-23
Proposal for error boundary hook

JoviDeCroock:
This is in no-way final, I would like to light the discussion on this. Would love some external opinions as well.

This PR can facilitate moving components to their own package since it does not rely on a lifecycle or something similar.

In the long term I would like to use the line from this PR: https://github.com/preactjs/preact/pull/2200/files#diff-b5e3843a998fbd0144a6ac5b1b7536a4R21

To tranfsorm it into

```
hasCaught = options._catch()
```

Where options._catch could listen for a lifecycle or whatever is needed in other api's like composition

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -222,23 +222,20 @@ export function useDebugValue(value, formatter) {
 	}
 }
 
-export function errorBoundary(fn, callback) {
-	function Boundary(props) {
-		const errState = useState();
-
-		if (!currentComponent.componentDidCatch) {
-			currentComponent.componentDidCatch = err => {
-				if (callback) callback(err);
-				errState[1](err);
-			};
-		}
-
-		return fn(props, errState[0]);
+export function useErrorBoundary(cb) {
+	const errState = useState();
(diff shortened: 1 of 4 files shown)


hooks/src/index.js
```diff
@@ -222,6 +222,25 @@ export function useDebugValue(value, formatter) {
 	}
 }
 
+export function errorBoundary(fn, callback) {
+	let setErr;
```
developit: I think this breaks if the component is rendered in two different places, since they would both assign to setErr. Maybe a decent argument for not having this produce its own state? (callback-only)

JoviDeCroock: Problem being that we can't set the state in the callback (second argument) so we can't render a fallback which could result in endless loops

marvinhagemeister:
The more I look at it, the more I like it. I think it's a genius way to support catching errors. Great work :raised_hands: :100:

JoviDeCroock:
All right, in case we merge it I'll make a docs entry for it

PR #2206, opened 2019-12-24
fix compat hydration

JoviDeCroock:
In current master we unmount the entire tree when hydrating because `x._children` is `undefined` that's why we need a separate method for hydration


shakyShane:
@JoviDeCroock @developit thanks for this!

Is there a planned patch release for this?

I only ask/nag since the current compat+hydrate situation is so fundamentally broken, thanks again :)

marvinhagemeister:
We'll cut a release soon :tada:

Issue #2004, opened 2019-10-12
Render / replaceNode unexpected behavior?


richardhinkamp:
I'm having some issues with `render()`, followup to #2002
I have an app which is mostly not preact based, but uses a custom made system to generate HTML. We are migrating to preact, so different parts of the app are now created in preact. We currently use preact 8, and use this pattern:

```
const container = document.getElementById('someId');
let rendered = preact.render(<SomeComponent />, container);

// based on some events an update of the component may be required
preact.render(<SomeComponent />, container, rendered);

// base on another event it must be removed
preact.render("", container, component);
rendered = undefined;
```

This works like a charm.

Now I'm trying to migrate to Preact X and I'm facing some issues, because render changed. Maybe I do not understand the desired approach for this, but I found some strange effects using `render()`:

https://codepen.io/rhinkamp/pen/abbdBgE
The rendered component is not appended to container, but added before the h1.

https://codepen.io/rhinkamp/pen/ExxPNOv
A existing div in the HTML is replaced by the component, but it's position changed to before the h1.

https://codepen.io/rhinkamp/pen/WNNroVQ
Using the replaceNode parameter it's in the correct place, but the nodes look merged, since it's the component and the id of the original node still present.

https://codepen.io/rhinkamp/pen/eYYJgGV
If the replaceNode is a different type (span opposed to div) it is not merged?

https://codepen.io/rhinkamp/pen/gOOPLVJ
A button to re-render the component with a different text suddenly drops the "Bar:" text on the first click. On the second click "Bar:" is back, but "allo" is missing.

https://codepen.io/rhinkamp/pen/XWWXpgE
Using an empty container node and no replaceNode option does work correct.

https://codepen.io/rhinkamp/pen/OJJMWJL
A button to remove the element and a button to add/update. This seems to work, but after removing and adding, clicking the add/update button again will remove all text from the component?

Are these cases correct but am I expecting the wrong thing, or are these bugs? 

Is there a desired approach for adding preact component to an existing app? Especially adding it somewhere into the existing DOM tree.

I think I can create container elements to render components in, so I don't have to use replaceNode, I guess that will fix most if not all issues, but the replaceNode does result in some unexpected behavior?

richardhinkamp:
https://codepen.io/rhinkamp/pen/OJJMWzM
Rendering 2 components in 1 containerNode will result in 1 rendered component instead of 2, is that supposed to happen?

JoviDeCroock:
[...]

PR #2210, opened 2019-12-27
[DRAFT] Handle subtree replaced via replaceNode

developit:
I really don't like the solution I came up with here. This is a case where `excessDomChildren` should end up with `[<a></a>]` in it which should remove the orphaned root element. I tried for quite a while to figure out why this was not the case, and installed a bunch of logging into these tests to figure it out. In the end I could get the tests passing by tweaking `excessDomChildren` to flush for known Fragment roots rendered using `replaceNode`, but doing so broke all the Portals tests.

### DIFF
--- a/compat/test/browser/portals.test.js
+++ b/compat/test/browser/portals.test.js
@@ -188,7 +188,7 @@ describe('Portal', () => {
 		}
 
 		render(<App />, root);
-		expect(dialog.childNodes.length).to.equal(2);
+		expect(dialog.childNodes.length).to.equal(1);
 		render(null, root);
 		expect(dialog.childNodes.length).to.equal(0);
 	});
--- a/src/render.js
+++ b/src/render.js
@@ -2,6 +2,7 @@ import { EMPTY_OBJ, EMPTY_ARR } from './constants';
 import { commitRoot, diff } from './diff/index';
 import { createElement, Fragment } from './create-element';
 import options from './options';
+import { removeNode } from './util';
 
 const IS_HYDRATE = EMPTY_OBJ;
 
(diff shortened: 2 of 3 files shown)


JoviDeCroock (approved):

developit:
Just a note from me: the reason I don't think we should merge this as-is comes down to it being a patch. This directly address the issue from #2004 by removing an unmatched root after rendering, but doing so fails to unmount that root's VDOM tree. The reason it cannot currently unmount the previous tree is that we don't have a reference to it: render() was called on an element somewhere within a VDOM tree, but we can't go from Node to VNode without crawling all the way back up to the root looking for the tree, then performing a BFS on the VDOM tree to find the outermost VNode with a _dom reference to the element we're updating. Personally I think that's far too much complexity for this since it's not technically a supported use-case.

What I would like to have as a long-term solution for this case would be to drop the idea of unmounting that VDOM - invoking render() on a Node without giving Preact VDOM tree information logically shouldn't try to infer that information. Instead, we can acknowledge that we'll _remove_ unmatched Elements but not unmount unrelated VDOM trees. If we can commit to that, fixing the issue here becomes more compelling:  I would like to see if there is a way for `diffChildren` or 
`diffElementNodes` to identify when they are handling a mismatched root with no `excessDomChildren` available to search through, and in those cases use `.replaceNode()` where they would otherwise have used `.insertBefore()`.

cristianbote (approved):
👍 awesome!

JoviDeCroock:
Closing: `replaceNode` (the 3rd `render()` argument) was removed in v11, so this no longer applies.

PR #2211, opened 2019-12-27
Track original VNodes to allow referential equality diff bypass

developit:
This improves scores on one of the benchmarks by more than 30%.

### DIFF
--- a/compat/mangle.json
+++ b/compat/mangle.json
@@ -10,6 +10,7 @@
       "$_nextState": "__s",
       "$_renderCallbacks": "__h",
       "$_vnode": "__v",
+      "$_original": "__v",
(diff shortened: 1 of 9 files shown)


developit:
Suspense tests are failing badly here, but nothing else.

developit:
Paging @prateekbh - do you know of anything that would cause the Suspense implementation to rely on the diff always running against referentially-equal VNodes?

cristianbote:
@developit do you feel this is ready? Are there any Suspense related issues left?

prateekbh:
The failure in tests is consistent. I have no idea what is causing them

developit:
Superseded by @JoviDeCroock's #2386.

PR #2213, opened 2019-12-27
Skip setProperty for known bypass cases (key, children)

developit:
This should reduce average stack depth and improve performance.


marvinhagemeister:
@developit is this ready to merge or are there further things you'd like to play with?

PR #2214, opened 2019-12-27
Experimental support for resuming suspended hydration

developit:
This PR adds two metadata properties to VNodes which are used to store the `isHydrating` flag and current DOM element (`oldDom`) when bailing out of diffing as a result of a suspension (or error).

This is particularly useful during hydration, as it means we can flip _back_ into hydration mode and pick up the previously-attempted DOM element when we attempt to render the previously-suspended tree's root VNode.

The implementation here is not great in terms of filesize, and I haven't added dedicated tests for it yet.  I have verified that this works with a prototype Suspense-based `AsyncComponent` implementation for Preact CLI (included below) that can be tested via `npm link` [here](https://github.com/developit/preact-cli-async).

<details>
<summary>experimental Suspense-based <code>async-component.js</code> for preact cli</summary>

The main thing worth noting here is that there is no use of internals.

How it works: the generated wrapper component (`AsyncComponent`) renders a child component that it knows will throw a given value (an object for simple referential equality). It then checks for that value having been thrown in `componentDidCatch` and absorbs the error, which has the effect of dropping its subtree's rendering on the floor. This preserves the DOM tree in-tact during the period where the lazy-loaded component is not yet available.

It perfom this throw routing for all renders up until its payload (whatever is passed to the callback given to `load`) becomes available. Once available, it "unsuspends" by triggering a re-render using `setState({})`. This is where the magic happens:

Since Preact knows the original "owner" VNode behind this component was suspended (in hydrate mode!) and has maintained a reference to its associated DOM element at time of suspension, it will immediately jump back into hydrate mode in the `diff()` call originating from setState and re-use that stored DOM element.

```js
import { h, Component } from 'preact';

const PENDING = {};

function Pending() {
	throw PENDING;
}

export default function async (load) {
	let component;
	function AsyncComponent() {
		Component.call(this);
		if (!component) {
			this.componentWillMount = (() => {
				load(mod => {
					component = mod && mod.default || mod;
					this.componentDidCatch = null;
					this.setState({});
				});
			}
			),
			this.componentDidCatch = err => {
				if (err !== PENDING) throw err;
			};
			this.shouldComponentUpdate = () => component != null;
		}
		this.render = props => h(component || Pending, props);
	}
	AsyncComponent.preload = load;
	(AsyncComponent.prototype = new Component).constructor = AsyncComponent;
	return AsyncComponent;
}
```

</details>


src/diff/index.js
```diff
@@ -215,6 +225,11 @@ export function diff(
 
 		if ((tmp = options.diffed)) tmp(newVNode);
 	} catch (e) {
+		// Before bailing out, mark the current VNode with the DOM element and hydration state.
+		// We can use this information if we return here to render later on.
+		newVNode._dom = oldDom;
```
JoviDeCroock: Can be reduced to:

```suggestion
		oldVNode._hydrateDom = newVNode._dom = oldDom;
```

This removes the need for the next line (multi-line suggestions aren't available)

src/diff/index.js
```diff
@@ -40,6 +40,16 @@ export function diff(
 	// constructor as undefined. This to prevent JSON-injection.
 	if (newVNode.constructor !== undefined) return null;
 
+	// If this was the innermost VNode at a point where the tree suspended,
+	// pick up diffing where we left off using the saved DOM element and hydration state.
+	if (oldVNode._hydrateDom && excessDomChildren == null) {
+		oldDom = oldVNode._hydrateDom;
```
```suggestion
		newVNode._dom = oldDom = oldVNode._hydrateDom;
```

JoviDeCroock (approved):
This looks really good to me

prateekbh:
I tried reproducing the broken test in a real world app but couldn't everything worked fine there

prateekbh:
@developit am i supposed to see this with your given instructions?

Screen Shot 2020-01-01 at 11 33 21 PM

JoviDeCroock:
Closing: the VNode-metadata approach to resuming suspended hydration is superseded by the v11 `MODE_HYDRATE`/`MODE_SUSPENDED` flags and excess-resume logic.

Issue #2194, opened 2019-12-19
Strange ref assign order


ajihyf:
I'm trying to use same `ref` in two conditional rendering components like this.

```javascript
function App() {
    ...
    return state ? (
        <div key="a">
            <div ref={ref} />
        </div>
    ) : (
        <div key="b">
            <div ref={ref} />
        </div>
    );
}
```

When state changes from `true` to `false`, it seems that the first unmounting div's ref assignment happens after the second mounting div, which makes `ref.current = null`.

I've created a sandbox to reproduce this issue. https://codesandbox.io/s/preact-ref-bug-esfv7

marvinhagemeister:
Can confirm that this is a bug. Here is the [React version](https://codesandbox.io/s/react-conditional-ref-7qgzg) which correctly sets the `ref`.

PR #2217, opened 2019-12-29
don't assign null when the ref isn't assigned to the current vn…

JoviDeCroock:
This scenario occurs when a parent unmounts a vnode and assigns null to the ref, making the already mounted newChild have an invalid ref.

Trying to see if I'm missing any edge cases (note that this doesn't fix for functions yet) I think this is pretty inherent to how we diff atm

So with this it will still do fail for the `ref={arg => x}` case where the unmounting component will still be called with null so if assigning to the same ref will fail

Possible solution: move applying of refs for the updated/created tree to `commitRoot` and separate ref removal and ref appliance from eachother that way we can do the `null` calls first and then apply refs

Fixes: https://github.com/preactjs/preact/issues/2194

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -406,7 +406,7 @@ export function unmount(vnode, parentVNode, skipRemove) {
 	if (options.unmount) options.unmount(vnode);
 
 	if ((r = vnode.ref)) {
-		if (!vnode.ref.current || vnode.ref.current === vnode._dom)
+		if (!r.current || r.current === vnode._dom)
 			applyRef(r, null, parentVNode);
 	}
 
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -406,7 +406,8 @@ export function unmount(vnode, parentVNode, skipRemove) {
 	if (options.unmount) options.unmount(vnode);
 
 	if ((r = vnode.ref)) {
-		applyRef(r, null, parentVNode);
+		if (!vnode.ref.current || vnode.ref.current === vnode._dom)
+			applyRef(r, null, parentVNode);
 	}
 
 	let dom;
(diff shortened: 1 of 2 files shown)


src/diff/index.js
```diff
@@ -406,7 +406,8 @@ export function unmount(vnode, parentVNode, skipRemove) {
 	if (options.unmount) options.unmount(vnode);
 
 	if ((r = vnode.ref)) {
-		applyRef(r, null, parentVNode);
+		if (!vnode.ref.current || vnode.ref.current === vnode._dom)
+			applyRef(r, null, parentVNode);
```
developit: It seems like we can either drop the `r` assignment altogether, or use it for L409 as well?

Also - it seems like we can skip nulling the ref if it's already null? (when `!vnode.ref.current`)

JoviDeCroock: Do you think we should move to assigning `refs` during `commit` to also fix this for functional refs or would you prefer only fixing objects (the former will add a lot more bytes)?

developit: I think for now let's keep this in-place and normalize the objects - if there's not a reason to avoid using `r` here we can just use `r`. Longer term this seems like a potential case for the queue, but we'll need to check what the timing requirements are for that - eg, should refs be populated/called prior to effect hooks, or after?

JoviDeCroock: I applied your changes

src/diff/index.js
```diff
@@ -406,7 +406,8 @@ export function unmount(vnode, parentVNode, skipRemove) {
 	if (options.unmount) options.unmount(vnode);
 
 	if ((r = vnode.ref)) {
-		applyRef(r, null, parentVNode);
+		if (!r.current || r.current === vnode._dom)
```
developit: Would this work? Seems like we're wanting to null the ref if it's not already null.

```suggestion
		if (r.current != null)
```

JoviDeCroock: We only want to nul lthe ref if it's assigned to the node that's unmounting else we could result in a false positive (see issue #2194)

PR #2218, opened 2019-12-29
Add Potentially missing bookkeeping for excessDomChildren

developit:
Just putting up a WIP of what I am thinking is a potential fix for a few issues, by adding in bookkeeping that appears to be missing for `excessDomChildren`.

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -123,6 +123,11 @@ export function diffChildren(
 						firstChildDom = newDom;
 					}
 
+					if (excessDomChildren != null) {
+						const index = excessDomChildren.indexOf(newDom);
+						if (index !== -1) excessDomChildren[index] = null;
+					}
+
 					if (childVNode._lastDomChild != null) {
 						// Only Fragments or components that return Fragment like VNodes will
 						// have a non-null _lastDomChild. Continue the diff from the end of
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -301,11 +301,12 @@ function diffElementNodes(
 		excessDomChildren = null;
 	}
 
-	if (newVNode.type === null) {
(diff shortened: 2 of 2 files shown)


JoviDeCroock:
Do these issues relate to the one where you have this PR for: https://github.com/preactjs/preact/pull/2210 since this actually looks very good to me

developit:
@JoviDeCroock my hope was that this would make #2210 unnecessary.

marvinhagemeister:
I can't decide which approach is better, because it's an area I haven't dabbled much with. I do like the test cases in #2210 though! And from your description @developit it seems like this one here may catch more problems than #2210, right?

JoviDeCroock (approved):
@marvinhagemeister yes, those tests would be great to have here. If those work I'd choose to merge this one over the other.

JoviDeCroock:
I've tried adding the tests from #2210 but the tests added there fail

prateekbh:
@JoviDeCroock @developit is the change in Portal tests in #2210 even required while bringing it here?

src/diff/children.js
```diff
@@ -123,6 +123,11 @@ export function diffChildren(
 						firstChildDom = newDom;
 					}
 
+					if (excessDomChildren != null) {
+						const index = excessDomChildren.indexOf(newDom);
+						if (index !== -1) excessDomChildren[index] = null;
```
cristianbote: Do you think we would run into issue -- deopting and such, if we always set the index, whatever the value? 🤔 
```suggestion
						excessDomChildren[excessDomChildren.indexOf(newDom)] = null;
```

JoviDeCroock:
Closing: draft against the 2019 excess-DOM-children bookkeeping, which was entirely rewritten in the v11 children diff.

PR #2220, opened 2019-12-30
fix renderer tests by using sinon global

JoviDeCroock:
Tests were failing on master IE11 due to the sinon import

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -22,16 +22,16 @@ var sauceLabsLaunchers = {
 		browserName: 'firefox',
 		platform: 'Windows 10'
 	},
+	sl_edge: {
+		base: 'SauceLabs',
+		browserName: 'MicrosoftEdge',
+		platform: 'Windows 10'
+	},
 	sl_ie_11: {
 		base: 'SauceLabs',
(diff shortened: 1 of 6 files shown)


karma.conf.js
```diff
@@ -88,6 +88,8 @@ const babelOptions = (options = {}) => {
 	};
 };
 
+sauceLabs = true;
```
developit: what's this for?

JoviDeCroock: Make sauceLabs run on these tests so I can see if the IE11  tests are going well, hence the `(remove this when verified)` in said commit

developit: ah, missed the commit. Seems like our travis setup needs to be tweaked - we should be running saucelabs for branches in the main trepo, just not forks.

JoviDeCroock:
@marvinhagemeister do we merge it with less coverage or leave it failing for now?

marvinhagemeister:
@JoviDeCroock I'm fine with merging this, because we'll run into the same issue sooner or later when the next version of devtools comes out. Just remove the `sauceLabs = true` flag and than it's good to go on my end :+1:

PR #2223, opened 2020-01-03
Proof of concept for useProvider

JoviDeCroock:
Current issue is that `<Context.Consumer>` won't pick this up yet, the issue is displayed in `useContext` this PR is mostly to gather some opinions

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -209,7 +209,7 @@ export function useContext(context) {
 		state._value = true;
 		provider.sub(currentComponent);
 	}
-	return provider.props.value;
+	return provider._isHookProvider ? provider._value : provider.props.value;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


ForsakenHarmony:
would it be possible for the hook to add a parent element (the provider)?

like an implicit hoc

JoviDeCroock:
> would it be possible for the hook to add a parent element (the provider)?
> 
> like an implicit hoc

This was my first thought as well but my initial thought was the biggest perf-win would be to allow Preact consumers to reduce tree-depth since this will be our biggest bottleneck. I do agree that this immediately solves the issue of having to support `Consumer` so I agree with it from that side

yisar:
It seems like a good idea, but I want to know what's the purpose of this?

In my opinion, `useContext` is supported to be compatible with react, but what is the purpose of adding `useProvider`?

JoviDeCroock:
Look at the point above, the goal is to allow users to minimise tree-depth. Tree-depth is generally one of the bottlenecks for any diffing algo.

With the coming of createContext we see more trivial use of Providers and this deepend the tree so this can be considered convenience API surface.

porfirioribeiro:
Nice work, i'm really looking into this!
Making it more easy to add providers without bloating up the tree!

remcohaszing:
Why has this been abandoned? It still seems very useful to me.

PR #2232, opened 2020-01-06
modernize our README

JoviDeCroock:
This mainly removes a lot of overhead in favor of  our dedicated docs page and preact-awesome list

@developit do you have the link to your gist with the size trial? I'd like to include it here as well to point to if needed

TODO: compare `preact-awesome` with removed links

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@
 [![travis](https://travis-ci.org/preactjs/preact.svg?branch=master)](https://travis-ci.org/preactjs/preact)
 [![coveralls](https://img.shields.io/coveralls/preactjs/preact/master.svg)](https://coveralls.io/github/preactjs/preact)
 [![gzip size](http://img.badgesize.io/https://unpkg.com/preact/dist/preact.min.js?compression=gzip)](https://unpkg.com/preact/dist/preact.min.js)
+[![brotli size](http://img.badgesize.io/https://unpkg.com/preact/dist/preact.min.js?compression=brotli)](https://unpkg.com/preact/dist/preact.min.js)
 
 Preact supports modern browsers and IE11+:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -171,344 +54,31 @@ The following guide assumes you have some sort of ES2015 build set up using babe
 You can also start with [preact-boilerplate] or a [CodePen Template](http://codepen.io/developit/pen/pgaROe?editors=0010).
```
38elements: I think it is better to replace `preact-boilerplate` to `preact-cli`.

JoviDeCroock: Preact-cli is mentioned above and underneath

reznord: A link to `codesandbox` instead of `codepen` would be great here.

Codepen is too old :P

38elements: `preact-boilerplate` recommends to use `preact-cli`.
I think it may be better to remove this line.
https://github.com/developit/preact-boilerplate#preact-boilerplate--starter-kit--

JoviDeCroock: Good spot!

README.md
```diff
@@ -53,110 +32,14 @@
 [![gzip size](http://img.badgesize.io/https://unpkg.com/preact/dist/preact.min.js?compression=gzip)](https://unpkg.com/preact/dist/preact.min.js)
 [![install size](https://packagephobia.now.sh/badge?p=preact)](https://packagephobia.now.sh/result?p=preact)
```
JoviDeCroock: Can I remove this badge, it's not that informative

developit: oh, yeah, definitely.

README.md
```diff
@@ -21,142 +21,24 @@
...
 
 [![Browsers](https://saucelabs.com/browser-matrix/preact.svg)](https://saucelabs.com/u/preact)
 
 
 ---
 
-
-## Demos
-
-#### Real-World Apps
```
JoviDeCroock: Will PR these to `awesome-preact`

README.md
```diff
@@ -166,349 +49,31 @@ Preact supports modern browsers and IE9+:
...
-You can also start with [preact-boilerplate] or a [CodePen Template](http://codepen.io/developit/pen/pgaROe?editors=0010).
-
-
-### Import what you need
+### Minimal example
 
-The `preact` module provides both named and default exports, so you can either import everything under a namespace of your choosing, or just what you need as locals:
-
-##### Named:
+Here we'll have a small example
```
developit: @JoviDeCroock  I think we need a little bit more progression here. Not certain what I have below works, but I'd like to have this be a quick "zero to preact" tutorial that ends in your stateful component example.

````suggestion
Preact lets you build user interfaces by assembling trees of components and elements. Components are functions or classes that return a description of what their part of the tree should look like. These descriptions are typically written in [JSX](https://facebook.github.io/jsx/) (shown here), or [HTM](https://github.com/developit/htm) which is similar but uses only standard JavaScript. Both syntaxes let us express trees of elements, each with "props" (like HTML attributes) and children.

To get started using Preact, first we'll take a look at the `render()` function. We can pass a tree description to render() and it will create the structure we described and append it to a parent DOM element we provide. Future render() calls will find our existing tree and update it in-place in the DOM, calculating the difference from our previous render in order to perform as few operations as possible.

```js
// Tells babel to use h for JSX. It's better to configure this globally.
// See https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#usage
/** @jsx h */

// create our tree:
render(<main><h1>Hello</h1></main>, document.body);

// update the tree in-place:
render(<main><h1>Hello World!</h1></main>, document.body);
// ^ only updates the text of our <h1>
```

Great, now we've built a User Interface using JSX and Preact! This approach would be tough to use for a full application though, since we have to re-render our whole app any time there is a change. This is where Components come into play: we can divide up our User Interface into nested Components, and each can update itself wherever it happens to be mounted rather than updating our whole app. Let's take a look at a simple component:
````

File: README.md [hunk 5]
```diff
@@ -166,349 +49,31 @@ Preact supports modern browsers and IE9+:
...
-	}
-
-	render(props, state) {
-		let time = new Date(state.time).toLocaleTimeString();
-		return <span>{ time }</span>;
-	}
+	return (
+		<div>
+			<p>Do you agree to the statement: "Preact is awesome"?</p>
+			<input value={input} onChange={e => setInput(e.target.value)} />
```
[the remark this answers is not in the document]
JoviDeCroock: Well I think it's pretty linear to just exposing them to doing this in a `Component`, no?

README.md
```diff
@@ -166,349 +48,49 @@ Preact supports modern browsers and IE9+:
...
+// create our tree:
+render(<main><h1>Hello</h1></main>, document.body);
 
-// or, as a stateless functional component:
-const Link = ({ children, ...props }) => (
-    <a {...props}>{ children }</a>
-);
+// update the tree in-place:
+render(<main><h1>Hello World!</h1></main>, document.body);
+// ^ only updates the text of our <h1>
```
kristoferbaxter: // ^ this second invocation of `render()` will only use a single DOM call to update the text of the <h1>

README.md
```diff
@@ -166,349 +48,49 @@ Preact supports modern browsers and IE9+:
...
 ```
 
-
-## Extensions
-
-It is likely that some projects based on Preact would wish to extend Component with great new functionality.
-
-Perhaps automatic connection to stores for a Flux-like architecture, or mixed-in context bindings to make it feel more like `React.createClass()`.  Just use ES2015 inheritance:
+Great, now we've built a User Interface using JSX and Preact! This approach would be tough to use for a full application though, since we have to re-render our whole app any time there is a change. This is where Components come into play: we can divide up our User Interface into nested Components, and each can update itself wherever it happens to be mounted rather than updating our whole app. Let's take a look at a simple component:
```
kristoferbaxter: Hooray! `render()` has taken our structure and output a User Interface! This approach demonstrates a simple case, but would be difficult to use as an application grows in complexity. Each change would be forced to calculate the difference between the current and updated stucture for the entire application. Components can help here – by dividing the User Interface into nested Components each can calculate their difference from their mounted point. Here's an example:

README.md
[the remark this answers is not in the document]
kristoferbaxter: Since the community is moving toward Hooks right now, I think making the first example Hooks driven gives additional confidence the author written code will not need to be significantly change or refactored.

README.md
```diff
@@ -9,7 +9,7 @@
```
JoviDeCroock:
```suggestion
<p align="center">Fast <b>3kB</b> alternative to React with a similar API.</p>
```

JoviDeCroock: This is for the line above just a proposition

README.md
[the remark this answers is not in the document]
marvinhagemeister: +1 For a hooks driven example. My gut tells me that users coming from React will perceive Preact as "old" when the first example only shows classes.

Issue #2237, opened 2020-01-07
10.2.0 hydrate strange behavior using JSDOM


jchip:
Hi, I have the following component:

```js
/** @jsx h */
import { h } from "preact"
export default props => <p>hello {props.foo}</p>
```

babel would transpile that to

```js
props => h("p", null, "hello ", props.foo)
```

I have some unit test using JSDOM to run with node.js, like this:

```js
const { h, hydrate } = require("preact");
const { JSDOM } = require("jsdom");

const dom = new JSDOM(`<div id="test"><p>hello bar</p></div>`);
global.window = dom.window;
global.document = dom.window.document;
const element = dom.window.document.getElementById("test");

// component as transpiled by babel
const Component = props => h("p", null, `hello `, props.foo);

hydrate(h(Component, { foo: "bar" }), element);
console.log(element.innerHTML);
```

- With preact 10.2.0, the output from above is `<p>hello barbar</p>`, which is unexpected.

- With preact 10.1.1, the output is `<p>hello bar</p>`

note that if I changed the component to:

```js
const Component = props => h("p", null, `hello ${props.foo}`);
```

then 10.2.0 behaves the same as 10.1.1.

JoviDeCroock:
Hey :wave:,

Thanks for the accurate reproduction, it enabled me to quickly put up a PR with a fix!

marvinhagemeister:
@jchip Just published Preact `10.2.1` which includes the fix for this issue :+1:

PR #2238, opened 2020-01-07
Add dom.data during hydration to avoid redundantly adding text…

JoviDeCroock:
Due to the dom not being available as a node `diffChildren` will double add this text if we don't set the dom.data

I have been looking at solving it in `diffChildren` but the problem persists itself like this:
```
LOG: 'oldDom', hello bar
LOG: 'newChldren', ['hello ', 'bar']
LOG: 'oldChldren', []
LOG: 'excess', [hello bar]
```

So we effectively come in at the first newChild and diff it with oldDom, next up we have no oldDom and we just append it to the text node.

A potential perf improvement would be to check if the `parent.innerHTML === props` this would allow us to skip childDiffing for that text node (if it does not contain any dom nodes).

Fixes https://github.com/preactjs/preact/issues/2237

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -306,7 +306,7 @@ function diffElementNodes(
 			excessDomChildren[excessDomChildren.indexOf(dom)] = null;
 		}
 
-		if (oldProps !== newProps && !isHydrating && dom.data != newProps) {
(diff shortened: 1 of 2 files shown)


jchip:
wow! that was fast!!! thanks.

Issue #2241, opened 2020-01-10
Use Object.assign if available.


laino:
https://github.com/preactjs/preact/pull/1433 re-implemented ``Object.assign`` with a for-loop.

However using a for loop is only roughly half as fast in every browser I tested today (https://jsperf.com/object-assign-vs-for-loop-yaaay). Which isn't surprising.

So utils should check for ``Object.assign`` and export that if available instead of the polyfill.

JoviDeCroock:
Hey thanks for the suggestion, I put up a PR where we implement this where we'll discuss this.

lukeed:
It starts to get close when dealing with Objects with more than 24 keys, but `assign` still edges ahead. Even so, vnode objects aren't dealing with this many keys anyway

<img width="675" alt="Screen Shot 2020-01-10 at 12 51 02 PM" src="https://user-images.githubusercontent.com/5855893/72185590-0040f200-33a8-11ea-809a-d8814db191ec.png">

marvinhagemeister:
Closing, because as demonstrated by @lukeed our custom polyfill is faster for small objects, like we have in our code.

laino:
@marvinhagemeister It's generally not afaik. There was more discussion on the PR: https://github.com/preactjs/preact/pull/2242#issuecomment-573261721

marvinhagemeister:
cc @developit who did tons of benchmarking on that. It seems like `jsperf` leads to different results than `esbench`.

marvinhagemeister:
cc @developit who did tons of benchmarking on that. It seems like `jsperf` leads to different results than `esbench`.

marvinhagemeister:
We can't switch to `Object.assign` in the 10.x release line, because it would be a breaking change as `Object.assign` isn't available in IE11. We removed it for the upcoming v11. Closing this issue, since no further action is required.

PR #2242, opened 2020-01-10
use Object.assign when available

JoviDeCroock:
Adds 13 bytes but performs in browsers with Object.assign 43% better (only IE11 misses this so the perf gain is significant)

Resolves: https://github.com/preactjs/preact/issues/2241


marvinhagemeister:
Looks good to me, but I'd like to here @developit 's opinion on this :+1:

lukeed:
It's muchhh faster to just use the current helper, especially since we already know/control the argument count.

This is still true even with @developit's famous extend within an extend.

https://esbench.com/bench/5ab9bbeaf2949800a0f6197c

JoviDeCroock:
> It's muchhh faster to just use the current helper, especially since we already know/control the argument count.

Could this differ from perf-site to perf-site? https://jsperf.com/object-assign-vs-for-loop-yaaay

lukeed:
@JoviDeCroock No, it shouldn't. I _think_ the problem in that the jsperf you linked has no teardown/reset code?

I recreated the same ESBench on jsperf just in case: https://jsperf.com/assign-vs-object-assign

JoviDeCroock:
All right, thanks @lukeed removing this

laino:
Beating browser built-ins is probably not going to happen in any real world use-case. Not even close.

Use test objects with more than one property:

```js
var tmp = {};
var foo = { a:1, b:2, c:3, d:5, k:[1,2,3], l:{m: 1}, hello: "hi" };
var bar = { b:0, K: [], r: 3, bye: "cya" };
```

And also disable "transpile using babel" on esbench so it doesn't polyfill ``Object.assign``.

Finally see this (Chrome left, Firefox right):

</img>

Or on jsperf (Chrome left, Firefox right):

</img>

jsperf link: https://jsperf.com/assign-vs-object-assign/2

lukeed:
Here is my Chrome running the same `Setup` object you've posted:

Screen Shot 2020-01-10 at 2 11 06 PM

`Object.assign` _barely_ inches ahead... 1%

But when adding 9 additional keys to the `foo` object, tables turn once again:

```js
var tmp = {};
var foo = { a:1, b:2, c:3, d:5, k:[1,2,3], l:{m: 1}, hello: "hi", z1:1, z2:2, z3:3, z4:4, z5:5, z6:6, z7:7, z8:8, z9:9  };
var bar = { b:0, K: [], r: 3, bye: "cya" };
```

Screen Shot 2020-01-10 at 2 12 41 PM

Also, I manually disable "transpile with Babel" every time. The site just has it checked by default when you load the page.

---

There are many, many things faster than native JS APIs. Look at [basically any Array method](https://github.com/lukeed/arr/blob/master/benchmarks/results/node-8.md) (Node, but true of browser too). The benefit using the native APIs is that the JS engines can get better (and they often do). Of course, the dream & eventual destination is that the native APIs will always be the most succinct _and_ the fastest, but I wouldn't say that we're 100% there... yet.

laino:
@lukeed I'm now convinced your test setup isn't representative of basically anything.

Are you running these on some macbook that will run into thermal throttling issues halfway through? Maybe your CPU is missing some instruction sets that would enable some speedups?

I have now tried:

- Firefox 72 on Linux on a Thinkpad
- Chrome 73 on Linux on a Thinkpad
- Chrome 79 on Linux on a Thinkpad
- Chrome 79 on Windows on a Desktop
- Firefox 71 on Windows on a Desktop
- Firefox 72 on Windows on a Desktop
- Firefox 68 Mobile on Android 
- Chrome 79 on Mobile on Android
- Chrome 80 Beta on Mobile on Android
- Chrome 79 on Linux on a Desktop
- Firefox 72 on Linux on a Desktop

And the results were the same on **every**. **single**. **one**. with ``Object.assign`` being the clear winner.

You can view an entire gallery of them here: https://imgur.com/a/XF98A5m

I don't know what you are using for these benchmarks, but you probably shouldn't assume whatever you're benchmarking on that device is representative.

lukeed:
Could very well be the case. Most microbenchmarks fall into this category.

But to answer your question, initial screenshots were from an OSX desktop. No power constraints and didn't think to check CPU throttling since it's a browser tab and no fans spin. Also, the original ESBench is quite old and was made on a different machine

That said, I'm fully aware js engines behave differently and each carry their own fine tunings and trade-offs. At the time, O.a was faster in only the latest FF.

I also don't mind be entirely wrong or outdated. I'd much rather use O.a and I'd hope by the end of this year it's always the winner unequivocally (else maybe spread).

Am out and about right now. Here are two separate runs on my Android (Chrome). Single O.a wins but check out "O.a multiple" on both runs.

Have a great weekend!

Screenshot_20200110-182722

Screenshot_20200110-182455

ForsakenHarmony:
<details>
<summary>On firefox nightly, O.a seems to win by a big margin</summary>


Test



</details>

<details>
<summary>Chrome assign by a tiny margin</summary>



</details>

<details>
<summary>Android Chrome, O.a by a decent margin</summary>



</details>

PR #2246, opened 2020-01-13
Upgrade devtools adapter to support Profiler

marvinhagemeister:
This PR replaces the current devtools adapter with the one that is dynamically injected by the extension. This leads to a stark size reduction of `preact/debug` :wink: 

Instead of just replacing the existing code in `preact/debug`, this PR adds a `preact/devtool` sub-package. The benefit of this is that it allows us to enable the devtools in production without bringing in the weight of `preact/debug`. The new devtools adapter weights just `140 B` although that size will likely be smaller once bundled into the main package.

**New features:**

- Attach devtools to production sites :tada: 
- Record profiling times of a Preact component tree :100:

**Tasks:**

- [x] Release new version of devtools and wait until it's approved before merging this PR. That way users have a smooth upgrade experience.
  - [x] Firefox
  - [x] Chrome

Screenshot from 2020-01-13 09-37-55


devtools/package.json
```diff
@@ -0,0 +1,18 @@
...
+	"amdName": "preactDevtools",
+	"version": "1.0.0",
+	"private": true,
+	"description": "Preact bridge for Preact devtools",
+	"main": "dist/devtools.js",
+	"module": "dist/devtools.module.js",
+	"umd:main": "dist/devtools.umd.js",
+	"source": "src/index.js",
+	"license": "MIT",
+	"mangle": {
```
porfirioribeiro: Isn't mangle already defined in `mangle.json`?

marvinhagemeister: Good catch :+1:

marvinhagemeister:
@JoviDeCroock `FLAKEY` is used for some two tests unrelated to devtools and this PR.

1) https://github.com/preactjs/preact/blob/e3591ccac0cf922da8c43b4f1c80d94121fa99f2/test/browser/lifecycles/lifecycle.test.js#L216
2) https://github.com/preactjs/preact/blob/e3591ccac0cf922da8c43b4f1c80d94121fa99f2/test/browser/render.test.js#L835

Because there unrelated to devtools, I'm a bit hesitant to remove them with this PR.

JoviDeCroock:
@marvinhagemeister, those work without Flakey last time I checked. Let’s push one commit to make sauce run for this PR then we can see (or we do it in a new PR after this one)

developit:
IIRC FLAKEY is only used for cross-browser tests via Saucelabs. Those tests should be running in CI against Chrome already.

PR #2254, opened 2020-01-16
adding the package-lock.json

JoviDeCroock:
This will enable pika to run


cristianbote:
Thinking if we should add a specific node version as well, since different node version would generate different hashes. I remember this would pop-up from time to time. Is still happening?

PR #2258, opened 2020-01-17
fix shouldComponentUpdate getting called on setState after forc…

laino:
See the added test.

``_force`` also had a weird type.


JoviDeCroock:
`shouldComponentUpdate` should be called when you use `setState` and `forceUpdate()` however when you'd use `forceUpdate(true)` then it shoudn't reset the `_force` to `false` within `setState`.

PR #2259, opened 2020-01-17
Gathering @developit's suspense hydration work

prateekbh:
- Compiles #2214 and #2218
- Brings suspense based progressive hydration.
- Enables Suspense based hydration for preact CLI


src/diff/index.js
```diff
@@ -215,6 +224,12 @@ export function diff(
 
 		if ((tmp = options.diffed)) tmp(newVNode);
 	} catch (e) {
+		if (isHydrating) {
```
prateekbh: This is literally the only thing I did in this PR.
Rest everything is @developit's code. 
Would love to send it to those PRs instead if that is preferred.

src/diff/index.js
```diff
@@ -291,6 +306,11 @@ function diffElementNodes(
 	}
 
 	if (dom == null) {
+		if (excessDomChildren != null) {
+			const index = excessDomChildren.indexOf(dom);
+			if (index !== -1) excessDomChildren[index] = null;
```
cristianbote: Maybe shorten the statements as they are in here? https://github.com/preactjs/preact/pull/2259/files#diff-a04f8653407a342eac2cd2ea6a5c3ff6R127

prateekbh: done

prateekbh:
I added the test case for out scenerio.
But in the process I discovered another bug

## Use case 1
```jsx
<div id="test">
  <Suspense>
    <Component />
  </Suspense>
  <p>Hello foo</p>
</div>
```

This is our current use case where `Component` is lazy loaded but its markup is not deleted thus

```html
<div id="test">
  <p>Component</p>
  <p>Hello foo</p>
</div>
```
is the markup while `Component` is being loaded.

## Use Case 2
```jsx
<div id="test">
  <Suspense fallback={<p>fallback</p>}>
    <Component />
  </Suspense>
  <p>Hello foo</p>
</div>
```

Now since the markup is not being deleted the html would be
```html
<div id="test">
  <p>fallback</p>
  <p>Component</p>
  <p>Hello foo</p>
</div>
```

This is where the problem is, both fallback and the actual markup are shown on the screen.

Since most `Suspense` are in practice written with fallback this'd be a problem.

prateekbh:
Either our solution should satisfy both cases or the behavior we're targeting with this PR should be achievable with some hook but should not be the default one.

prateekbh:
IMHO preserving of DOM should not when a promise is thrown should **NOT** be the default behavior as this breaks the `Suspense` rendering and makes it unintuitive for a large number of use cases. But preact should have options/API hooks extensible enough to achieve the same.

prateekbh:
Ok this is now fixed I feel.
Fix: During hydration, suspense components will not render a `fallback` component. It will show the pre-rendered component markup and then hydrate it once the `Suspense` resolves.

If nothing is pre-rendered in place of `Suspense` component then it will just inject the component when the `Suspense` resolves.

compat/src/suspense.js
```diff
@@ -11,7 +11,11 @@ options._catchError = function(error, newVNode, oldVNode) {
 		for (; (vnode = vnode._parent); ) {
 			if ((component = vnode._component) && component._childDidSuspend) {
 				// Don't call oldCatchError if we found a Suspense
-				return component._childDidSuspend(error, newVNode._component);
+				return component._childDidSuspend(
+					error,
+					newVNode._component,
+					oldVNode && oldVNode._hydrateDom && oldVNode._hydrating
```
developit: Just a thought (doesn't change the output or anything) - do you think we'll ever want to access the other properties of `oldVNode` from within `_childDidSuspend`?  If so we could pass `oldVNode` here and then check these properties in the _childDidSuspend implementation.  I don't have strong feelings either way, just occurred to me.

src/diff/index.js
```diff
@@ -215,6 +224,12 @@ export function diff(
 
 		if ((tmp = options.diffed)) tmp(newVNode);
 	} catch (e) {
+		if (isHydrating) {
+			// Before bailing out, mark the current VNode with the DOM element and hydration state.
+			// We can use this information if we return here to render later on.
+			oldVNode._hydrateDom = newVNode._dom = oldDom;
+			oldVNode._hydrating = true;
```
developit: I think if we're merging this with the `if(isHydrating)` condition, we should drop the `_hydrating` prop. It's always `true` when `vnode._hydrateDom != null`.

prateekbh: done

compat/src/suspense.js
prateekbh: done

src/diff/children.js
```diff
@@ -123,6 +123,10 @@ export function diffChildren(
 						firstChildDom = newDom;
 					}
 
+					if (excessDomChildren != null) {
+						excessDomChildren[excessDomChildren.indexOf(newDom)] = null;
```
marvinhagemeister: Just merged #2274 which has a similar fix. Are they the same?

JoviDeCroock: Did not notice this, you can remove the lines in diffElementChildren and use the tests in master to double check.

squidfunk:
This looks promising and I can add another use case: I’m using Suspense without `lazy` to throw a `fetch` promise in a data loading component (in `componentDidMount`). Is it possible to somehow detect that the component was pre rendered and avoid the `fetch` call?

src/diff/children.js
prateekbh: merged and extracted the common `if` condition outside

developit:
@squidfunk Avoiding the fetch call seems difficult - how would you get the data loaded otherwise?

If you're just looking for a way to indefinitely hold off rendering part of your app, this should likely give you that option. All you have to do is throw a pending Promise:

```js
const data = fetch('/data.json')
    .then(r => r.json())
    .then(value => data.value = value);
function MyComponent() {
  // don't render this subtree until we have data
  if (!data.value) throw data;

  return <pre>{JSON.stringify(data.value,0,2)}</pre>
}

// somewhere up the tree:

class MyApp extends Component {
  componentDidCatch(err) {
    if (err instanceof Promise) {
      err.then(() => this.setState({}));  // or forceUpdate
    }
  }
}
```

squidfunk:
@developit thanks for your response! Maybe I'm missing something but I may have found a way making things work. 

What I want to do is: separate article content in a Preact application (i.e. blog) from application logic. I'm using `htmdx` to build Preact's `VNode` intermediate representation and serialize it to JSON. The blog then requests the JSON and passes the nodes dynamically to `h`. Everything works as expected, except with SSR, as the actual DOM nodes are already there. What I want is to skip the `fetch` and just mount the event handlers registered with the components. However, I realize that Preact needs to know what DOM nodes are rendered so the only way to do this is to inline the JSON intermediate representation with the pre-rendered Markup, check if it's available and only fallback to the `fetch` when it is not. Or is there an easier way?

Sorry for hi-jacking this thread but I though it was somehow related. Happy to move the discussion somewhere else if necessary.

developit:
@squidfunk technically you could create a Virtual DOM tree from the existing DOM tree. There's actually a module that does that exact thing [over in preact-markup](https://github.com/developit/preact-markup/blob/master/src/to-vdom.js#L5):

```js
import toVdom from 'preact-markup/src/to-vdom.js';

let cached;
function AbsorbDom() {
    if (cached) return cached;
    return cached = toVdom(
        document.querySelect('article'),  // your article's root element
        vnode => {},  // optional modify the created Virtual DOM elements
        h,  // preact's h()
        { trim: false }
    );
}
```

andrewiggins:
FYI, one of my recent completed PRs conflicted with this one so I fixed for ya

ForsakenHarmony:
rebasing when it's ready makes more sense to me than merging master into it all the time, but idk :woman_shrugging:

prateekbh:
we discovered that there were false passing tests and that this approach doesn't work. @developit is working on an alternate approach in #2290 and #2291. 
Closing this one

PR #2262, opened 2020-01-21
Implement Map-assisted diffing

jviide:
This pull request suggest a modification to the diffing algorithm that uses Maps and linked lists to avoid O(n^2) scenarios.

In the beginning of each diffChildren call we now create two maps that are used to find old and new child vnodes whose keys/types match. There are separate maps for keyed and unkeyed children, because theoretically same value could be used as a key and as a type (it might be worth looking into whether using just one map would still be better).

The values in the map are singly-linked lists, each list containing the old vnodes that have that specific key or type. Before each list is searched they are cleaned up by removing already consumed vnodes from the beginning of the list. This needs to be done separately because nodes may not get consumed in order: the code dealing with vdom "holes" and vnodes with matching keys but different types cause such situations.

Because Maps are not really O(1) time in the worst case and because the key _and_ type have to match for each vnode the algorithm can still potentially take O(n^2) time - but hopefully less often. Also, the linked list item creation and other shenanigans may take more time than they save, so this approach needs benchmarking and tuning to figure out its viability.


jviide:
Modified the strategy:
 * Now just one shared map is created for both unkeyed and keyed children.
 * The map is created & initialized only when there are (suitable old children) AND (new children that require more extensive searching).

jviide:
@JoviDeCroock Yes, that would probably allow golfing down the size, but would that cover cases where `key` is not a string or number (for example for vnodes whose type is a function)?

jviide:
One potential optimization to cut down linked list construction time: Add new properties `_index` and `_next` to vnode objects so they can double as linked list items.

JoviDeCroock:
Doesn't that behave the same in `Map` though? I think it only alters in the case of a `WeakMap`

jviide:
For [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) key equality is based on the sameValueZero algorithm. E.g. for functions it's the same as strict equality.

```js
const f = () => {};
const g = () => {};

const m = new Map()
m.set(f, "hello")
m.get(f) // 'hello'
m.get(g) // undefined

> const o = { [f]: "hello" };
o[f] // 'hello'
o[g] // 'hello'
```

developit:
The bot can't post comments because @jviide's fork is read-only (github is weird), so here's the size output:



**Size Change:** +439 B (1%) 

**Total Size:** 38.6 kB

|:--- |:---:|:---:|:---:|
| `dist/preact.js` | 3.8 kB | +110 B (2%) |  |
| `dist/preact.min.js` | 3.8 kB | +110 B (2%) |  |
| `dist/preact.module.js` | 3.81 kB | +106 B (2%) |  |
| `dist/preact.umd.js` | 3.86 kB | +113 B (2%) |  |

<details><summary><strong>View Unchanged</strong></summary>

|:--- |:---:|:---:|:---:|
| `compat/dist/compat.js` | 2.95 kB | 0 B |  |
| `compat/dist/compat.module.js` | 2.97 kB | 0 B |  |
| `compat/dist/compat.umd.js` | 3 kB | 0 B |  |
| `debug/dist/debug.js` | 3.08 kB | 0 B |  |
| `debug/dist/debug.module.js` | 3.07 kB | 0 B |  |
| `debug/dist/debug.umd.js` | 3.14 kB | 0 B |  |
| `devtools/dist/devtools.js` | 175 B | 0 B |  |
| `devtools/dist/devtools.module.js` | 185 B | 0 B |  |
| `devtools/dist/devtools.umd.js` | 252 B | 0 B |  |
| `hooks/dist/hooks.js` | 1.06 kB | 0 B |  |
| `hooks/dist/hooks.module.js` | 1.08 kB | 0 B |  |
| `hooks/dist/hooks.umd.js` | 1.12 kB | 0 B |  |
| `test-utils/dist/testUtils.js` | 390 B | 0 B |  |
| `test-utils/dist/testUtils.module.js` | 392 B | 0 B |  |
| `test-utils/dist/testUtils.umd.js` | 469 B | 0 B |  |


</details>

PR #2266, opened 2020-01-21
Include support for customised built-in elements

defx:
This PR adds the additional second argument when invoking document.createElement to provide support for [customised built-in elements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#Customized_built-in_elements).

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -294,9 +294,12 @@ function diffElementNodes(
 		if (newVNode.type === null) {
 			return document.createTextNode(newProps);
 		}
+
+		const builtIn = newProps.is;
+
 		dom = isSvg
 			? document.createElementNS('http://www.w3.org/2000/svg', newVNode.type)
-			: document.createElement(newVNode.type);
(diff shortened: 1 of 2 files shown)


src/diff/index.js
```diff
@@ -294,9 +294,12 @@ function diffElementNodes(
 		if (newVNode.type === null) {
 			return document.createTextNode(newProps);
 		}
+
+		const builtIn = newProps.is;
+
 		dom = isSvg
 			? document.createElementNS('http://www.w3.org/2000/svg', newVNode.type)
-			: document.createElement(newVNode.type);
+			: document.createElement(newVNode.type, builtIn && { is: builtIn });
```
marvinhagemeister: Super awesome! Played around with it a bit and we can save 8 bytes by inlining the variable declaration. This presumably works because we're using `newProps` a lot already, so it's already available in the gzip directory:

```js
document.createElement(newVNode.type, newProps.is && { is: newProps.is });
```

defx: Superb! Thanks Marvin

PR #2272, opened 2020-01-22
Fix custom element tests failing in IE

marvinhagemeister:
Verified that by forcefully running saucelabs tests before making this PR.

See: https://travis-ci.org/preactjs/preact/builds/640641910


developit:
good catch.

PR #2273, opened 2020-01-22
Add test coverage for createElement

zubhav:
Added two tests for `createElement`:
- When `children` parameter is passed to the function as `null`
- When `children` parameter is not passed to the function at all

*PS: Newbie to contributing*


developit:
Great first PR! Thanks @zubhav :)

Issue #2264, opened 2020-01-21
Preact 10 rendering into document.body clobbers existing DOM nodes


cgwyllie:
Hello,

I'm seeing an unexpected behaviour when trying to render an app directly into `document.body`.

Maybe this isn't best practice but I'm using Preact in this way to append overlays and modal content to existing web pages, so it's not really a typical SPA where I control the entire page.

What is observed is that on first render an existing DOM node of the same type can be replaced by the rendered content. This does not happen with Preact 8 (although it seems subsequent updates just always append the content again).

What I'd expect is that the Preact app content is appended to the DOM if it was never previously mounted, and after that I'd expect only the 'Preact' nodes to be updated on subsequent renders.

You can see two examples here:

- Preact 8, semi-expected behaviour: https://codepen.io/cgwyllie/pen/YzPdPzW
- Preact 10, unexpected behaviour: https://codepen.io/cgwyllie/pen/QWwzWoQ

I found that if you render into a document fragment, then append the fragment to the body, this works as expected, but it performs poorly and feels a bit like a hack. The DOM nodes are removed from the fragment after initial render although Preact still manages to update them on subsequent renders.

I also looked into the 'replaceNode' parmeter of render, but couldn't achieve the desired effect with it.

Perhaps this isn't a bug, but I'm keen to understand exactly what is going on here, and if there is any more suitable workaround.

The use of Preact 10 is desirable because Fragments allow rendering multiple overlays at the same time without containing them inside another DOM node in the body.

Thanks,
  Chris

cgwyllie:
Hi Jovi,

Thanks for the quick reply!

RE the portal option, yes I have since attempted this and confirm that wrapping `createPortal(<Fragment>...</Fragment>)` now behaves as I would expect.

To explain the use-case a little further tho, and gain a better understanding, the idea is a library for managing certian popover states then draws some UI using Preact.

Not all state exists in the UI component hierarchy, and is passed via props to the 'App' component. The desire to call `render(<App />, document.body)` multiple times is to drive UI updates based on the props managed outside of the UI state.

We did think about managing state inside the App component, but this seemed undersirable as it coupled the state management and UI concerns quite tightly. In our case, the UI is a small piece of the library behaviour and a lot of other state is managed without the UI needing to be aware of it.

Don't know if that helps explain, please ask for any further details!

JoviDeCroock:
I have heard of this use case before for a `modals` library, what they did (and this happens for a lot of UI-libraries):

[...]

Issue #2260, opened 2020-01-18
App breaks after root update if rendered with replaceNode


xkam:
I'm having issue with render replaceNode parameter - the app breaks after I run render the second time with updated properties for the root component. Unfortunately I could not isolate the problem, the simple cases work as intended, so the issue might well be in my code. The first error is indeed in my code, but it looks that it's because something else is broken.

This works fine (1) - https://codepen.io/activewidgets/pen/YzPJebK
```js
render(h(App, initialProps), container);
setTimeout(() => render(h(App, updatedProps), container), 100);
```

This breaks (2) - https://codepen.io/activewidgets/pen/OJPBQKX
```js
render(h(App, initialProps), container, target);
setTimeout(() => render(h(App, updatedProps), container, target), 100);
```

I've found that this one actually works too (3) - https://codepen.io/activewidgets/pen/eYmPVqK
```js
render(h(App, initialProps), container, target);
setTimeout(() => render(h(App, updatedProps), target), 100);
```

Full code (static page) - https://gist.github.com/xkam/098b970d3cbc43427f764491f10dfea1
 
The variant (3) does what I want, so I'm fine if this is an intended use, but from the docs I had the impression that the right one is (2), and that one breaks.

Sorry, the examples have layers of other (less then ideal) code, but I thought that it might be useful anyway.

JoviDeCroock:
I'm sorry I don't know why but codepens always break for me for some reason, so I'm going to do my best to help you out with a previous issue: https://github.com/preactjs/preact/issues/2264

https://codesandbox.io/s/muddy-moon-r9dze

--> This actually uses method 2 so we have to figure out why it goes wrong and I would assume this fails because your dom structure has changed.

Method 3 would use the previously created root from replaceNode so this should work as intended, but method 2 should also work, this might be a bug that we have to trace.

I'm sorry that I can't directly answer your question here, I'll try and run the codepen on my other computer when I get home.

zaninime:
We recently switched some production code from 8 to 10 and had the same unexpected behaviour (we were using method 2). We switched to method 1.-- Francesco Zanini21.01.2020, 14:14, "Jovi De Croock" <notifications@github.com>:I'm sorry I don't know why but codepens always break for me for some reason, so I'm going to do my best to help you out with a previous issue: #2264
https://codesandbox.io/s/muddy-moon-r9dze
--> This actually uses method 2 so we have to figure out why it goes wrong and I would assume this fails because your dom structure has changed.
[...]

PR #2274, opened 2020-01-22
Avoid removing existing dom nodes on subsequent replaceNode cal…

JoviDeCroock:
Fixes: https://github.com/preactjs/preact/issues/2260
Fixes: https://github.com/preactjs/preact/issues/2264

We didn't correctly clean up existing nodes that weren't being replaced

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -315,6 +315,7 @@ function diffElementNodes(
 		}
 	} else if (newVNode !== oldVNode) {
 		if (excessDomChildren != null) {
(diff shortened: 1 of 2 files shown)


src/diff/index.js
```diff
@@ -315,6 +315,7 @@ function diffElementNodes(
 		}
 	} else if (newVNode !== oldVNode) {
 		if (excessDomChildren != null) {
+			excessDomChildren[excessDomChildren.indexOf(dom)] = null;
```
JoviDeCroock: This mutates the array that is still referenced in `diffChildren`

PR #2279, opened 2020-01-25
Fix className patch not applied to props

### LINKED ISSUE #2275: Preact 10.1.0 breaks Emotion styled components
Version 10.1.0 broke my Emotion styled components using `preact/compat`.

CodeSandbox: https://codesandbox.io/s/cocky-stallman-27oh3

Change to version 10.0.5, for ex, and you'll see it working.

Possibly related to recent changes to className?

marvinhagemeister:
This error was caused by our `className` patch being applied to the wrong `props` reference. Whenever a component is in play like with most CSS-in-JS solutions and we attempt to normalize `props` we'll create a new reference.

https://github.com/preactjs/preact/blob/8164289af82e0f463492ce68527d2c5f51e6fccb/compat/src/render.js#L111

A few lines later we attached the `className` patch, but we didn't update the reference of our `props` variable to point to the new object.

Since the order of our normalizations isn't important, the easiest fix is to move the `className` check before the other where `props` will be assigned a new variable.

Fixes #2275

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -84,6 +84,13 @@ options.vnode = vnode => {
 	let type = vnode.type;
 	let props = vnode.props;
 
+	// Alias `class` prop to `className` if available
+	if (props.class || props.className) {
+		classNameDescriptor.enumerable = 'className' in props;
+		if (props.className) props.class = props.className;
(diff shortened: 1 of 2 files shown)


marvinhagemeister:
@JoviDeCroock unfortunately not, but I think I know what's going on.

PR #2287, opened 2020-01-28
Fix Children.map not flattening result

### LINKED ISSUE #2286: The result of a nested React.Children.map is not consistent with React.isValidElement
Something like:
```jsx
const ProblemChild = ({ children }) => {
  return React.Children.map(children, child => {
    return React.Children.map(child.props.children, innerChild => {
      return innerChild;
    });
  }).filter(React.isValidElement)
};

const App = () => {
  return (
    <ProblemChild>
      <div>
        <div>
          1
        </div>
        <div>
          2
        </div>
      </div>
    </ProblemChild>
  )
};
```

returns

```sh
1
2
```

in React, but nothing in preact/compat.

It looks like the nested React.Children.map causes `React.isValidElement` to return false for the `divs`

reproduce with this repo: https://github.com/skvale/preact-react.children.map-issue

marvinhagemeister:
Just checked against the implementation in React and they flatten the result of `Children.map`. We didn't do that which lead to an `array` of children being passed to `isValidElement`. Because `arrays` are not valid children they were filtered out.

Fixes #2286 .

### DIFF
--- a/compat/src/Children.js
+++ b/compat/src/Children.js
@@ -2,7 +2,9 @@ import { toChildArray } from 'preact';
 
 const mapFn = (children, fn) => {
 	if (!children) return null;
-	return toChildArray(children).map(fn);
+	return toChildArray(children)
+		.map(fn)
+		.reduce((acc, value) => acc.concat(value), []);
 };
 
 // This API is completely unnecessary for Preact, so it's basically passthrough.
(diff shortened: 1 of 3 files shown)


sventschui:
Could save a byte when invoking `fn` inside the `reduce` callback or am I missing sth.?

```
	return toChildArray(children)
 		.reduce((acc, value) => acc.concat(fn(value)), []);
```

marvinhagemeister:
@sventschui Super awesome, that is a lot better :100: I had a hunch that the current solution wasn't as optimal as it should be :+1:

PR #2289, opened 2020-01-28
Golfing some bytes - 10b

JoviDeCroock:
Would appreciate a double check on https://github.com/preactjs/preact/commit/9e99ffcb8feccbce5f7918cf933b46b4718b3c71

10 bytes * 200k downloads = 2000k bytes 🎉


src/diff/children.js
```diff
@@ -233,10 +237,10 @@ export function toChildArray(children, callback, flattened) {
 	} else if (!callback) {
 		flattened.push(children);
 	} else if (typeof children === 'string' || typeof children === 'number') {
-		flattened.push(callback(createVNode(null, children, null, null)));
```
JoviDeCroock: I can remove these if this is a perf reduction in modern engines

PR #2291, opened 2020-01-30
Fix Suspense Hydration

developit:
Instead of annotating the VNode at a suspend point (a component that threw during render) with `_hydrateDom`, annotate the VNode at the Suspense boundary (the component that implemented `componentDidCatch()`).  When re-rendering a component, if its owner vnode has a boolean `._hydrating` property value, that means we should attempt to resume from the VNode's current `._dom` reference.

You can try this PR by checking out the [suspense-hydration-testing branch of preactjs/cli-demo](https://github.com/preactjs/cli-demo/tree/suspense-hydration-testing). Clone, then run `npm i && npm link ../preact && npm run build && http-server -s -p 8080 build`

Note: looking at this, I believe I need to change `src/component.js` L129 in order to enable resumption for trees that suspended during non-hydrate rendering:

```js
vnode._hydrating != null ? [oldDom] : null
```

#### To-Do

- [ ] Fix test failures
- [ ] Add MutationObserver-based test from the [cli-demo repo](https://github.com/preactjs/cli-demo/tree/suspense-hydration-testing)
- [ ] Test on `preact-www`


src/diff/index.js
```diff
@@ -223,11 +215,8 @@ export function diff(
 
 		if ((tmp = options.diffed)) tmp(newVNode);
 	} catch (e) {
-		if (isHydrating) {
-			// Before bailing out, mark the current VNode with the DOM element and hydration state.
-			// We can use this information if we return here to render later on.
-			oldVNode._hydrateDom = newVNode._dom = oldDom;
-		}
+		newVNode._dom = oldDom; // is this needed?
```
prateekbh: The Suspense tests pass if we still keep `isHydrating` check

prateekbh: IIRC `newVNode._dom = oldDom;` is needed because the diff remove the DOM otherwise it removes the existing DOM in next render.

developit: I'll try to get this fixed.

compat/src/suspense.js
```diff
@@ -89,7 +89,8 @@ Suspense.prototype._childDidSuspend = function(
 	 * and hydrate it when the suspense actually gets resolved.
 	 * While in non-hydration cases the usual fallbac -> component flow would occour.
 	 */
-	if (!(oldVNode && oldVNode._hydrateDom) && !c._suspensions++) {
+	const wasHydrating = vnode && vnode._hydrating === true;
+	if (!wasHydrating && !c._suspensions++) {
```
JoviDeCroock, line 93:
```suggestion
	if (!(vnode && vnode._hydrating === true) && !c._suspensions++) {
```

src/diff/catch-error.js
```diff
@@ -29,8 +30,13 @@ export function _catchError(error, vnode) {
 					component.componentDidCatch(error);
 				}
 
-				if (hasCaught)
-					return enqueueRender((component._pendingError = component));
+				if (hasCaught) {
+					component._pendingError = component;
+					vnode._hydrating = wasHydrating;
```
JoviDeCroock, line 35:
```suggestion
					vnode._hydrating = vnode._hydrating;
```

src/diff/catch-error.js
```diff
@@ -10,6 +10,7 @@ import { enqueueRender } from '../component';
 export function _catchError(error, vnode) {
 	/** @type {import('../internal').Component} */
 	let component, hasCaught;
+	const wasHydrating = vnode._hydrating;
```
JoviDeCroock, line 13:
```suggestion
```

src/diff/catch-error.js
developit: `wasHydrating` is the `vnode._hydrating` value from the original (innermost) vnode, whereas here `vnode` refers to the current value of the loop on L15.

JoviDeCroock:
Closing: the `_hydrating`-on-boundary approach is superseded — v11 `compat/src/suspense.js` already handles `MODE_HYDRATE` suspension.

PR #2294, opened 2020-01-30
Reduce reads of dom.nextSibling

andrewiggins:
While exploring ways to use `getDomSibling` to replace `nextSibling` in `diffChildren`, I discovered that `_lastDomChild` should really be `_lastDomChildSibling`. The only thing our code uses `_lastDomChild` for is to [determine what DOM node the diff should continue with](https://github.com/preactjs/preact/blob/383443976e418033e448505aa6b4c53a07a65be4/src/diff/children.js#L176) after diffing a Component/Fragment.

However when Components/Fragments are nested (like the example below) each Component reads it's child's `_lastDomChild` property and reads the `nextSibling` property. If the Components are the last child of their parent's (like the example below), then this `nextSibling` read is redundant to the read that it's child already performed. Note: it's only useless when the Component is the last child because the result of `oldDom = newDom.nextSibling` is thrown away. In diffChildren, `oldDom` is only used to determine what DOM the diff should use for siblings. When the Component doesn't have any siblings, the value of `oldDom` is thrown away and `_lastDomChild` is used instead in the parent.

```jsx
<Provider1>
  <Provider2>
    <Child>
      <div />
    </Child>
  </Provider2>
</Provider1>
```

In other words, what we really want to bubble up while diffing Components/Fragments isn't the last DOM child of the Component, but what DOM should the diff continue with. So this PR changes `_lastDomChild` to `_lastDomChildSibling`.

I also used this change to precalculate what `oldDom` should be in situations where we know it without having to read `nextSibling`

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -105,7 +105,7 @@ export function initDebug() {
 					`\n\n${getOwnerStack(vnode)}`
 			);
 		} else if (type != null && typeof type === 'object') {
-			if (type._lastDomChild !== undefined && type._dom !== undefined) {
+			if (type._lastDomChildSibling !== undefined && type._dom !== undefined) {
 				throw new Error(
 					`Invalid type passed to createElement(): ${type}\n\n` +
 						'Did you accidentally pass a JSX literal as JSX twice?\n\n' +
@@ -133,10 +133,10 @@ export function initDebug() {
 			);
 		} else if (
 			type === 'tr' &&
-			(parentVNode.type !== 'thead' &&
-				parentVNode.type !== 'tfoot' &&
-				parentVNode.type !== 'tbody' &&
-				parentVNode.type !== 'table')
+			parentVNode.type !== 'thead' &&
+			parentVNode.type !== 'tfoot' &&
+			parentVNode.type !== 'tbody' &&
+			parentVNode.type !== 'table'
 		) {
 			console.error(
 				'Improper nesting of table. Your <tr> should have a <thead/tbody/tfoot/table> parent.' +
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -105,7 +105,7 @@ export function initDebug() {
 					`\n\n${getOwnerStack(vnode)}`
 			);
 		} else if (type != null && typeof type === 'object') {
-			if (type._lastDomChildSibling !== undefined && type._dom !== undefined) {
+			if (type._children !== undefined && type._dom !== undefined) {
 				throw new Error(
 					`Invalid type passed to createElement(): ${type}\n\n` +
 						'Did you accidentally pass a JSX literal as JSX twice?\n\n' +
--- a/debug/test/browser/debug.test.js
+++ b/debug/test/browser/debug.test.js
@@ -80,7 +80,7 @@ describe('debug', () => {
 	it('should print an error on double jsx conversion', () => {
 		let Foo = <div />;
 		let fn = () => render(h(<Foo />), scratch);
-		expect(fn).to.throw(/createElement/);
+		expect(fn).to.throw(/JSX twice/);
 	});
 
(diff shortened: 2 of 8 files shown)


debug/src/debug.js
```diff
@@ -133,10 +133,10 @@ export function initDebug() {
 			);
 		} else if (
 			type === 'tr' &&
-			(parentVNode.type !== 'thead' &&
-				parentVNode.type !== 'tfoot' &&
-				parentVNode.type !== 'tbody' &&
-				parentVNode.type !== 'table')
+			parentVNode.type !== 'thead' &&
```
andrewiggins: This is how Prettier formatted this for me 🤷‍♂

src/create-element.js
```diff
@@ -68,7 +68,11 @@ export function createVNode(type, props, key, ref) {
 		_parent: null,
 		_depth: 0,
 		_dom: null,
-		_lastDomChild: null,
+		// _lastDomChildSibling must be initialized to undefined b/c it will eventually
+		// be set to dom.nextSibling which can return `null` and it is important
+		// to be able to distinguish between an uninitialized _lastDomChildSibling and
+		// a _lastDomChildSibling that has been set to `null`
+		_lastDomChildSibling: undefined,
```
andrewiggins: Thinking out loud... would `_nextDom` be a better name for this?

src/diff/children.js
```diff
@@ -144,6 +146,7 @@ export function diffChildren(
 
 						outer: if (oldDom == null || oldDom.parentNode !== parentDom) {
 							parentDom.appendChild(newDom);
+							nextDom = null;
```
andrewiggins: Since we are adding `newDom` to the end of `parentDom`, `newDom.nextSibling` will always be `null`. So instead of going through the DOM APIs to get this, I'm precalculating it here so save us a DOM read.

src/diff/children.js
```diff
@@ -156,6 +159,7 @@ export function diffChildren(
 								}
 							}
 							parentDom.insertBefore(newDom, oldDom);
+							nextDom = oldDom;
```
andrewiggins: Since we are inserting `newDom` before `oldDom`, `newDom.nextSibling` will always be `oldDom` so I'm precalculating this now to save us a DOM read.

debug/src/debug.js
```diff
@@ -105,7 +105,7 @@ export function initDebug() {
 					`\n\n${getOwnerStack(vnode)}`
 			);
 		} else if (type != null && typeof type === 'object') {
-			if (type._lastDomChild !== undefined && type._dom !== undefined) {
+			if (type._lastDomChildSibling !== undefined && type._dom !== undefined) {
```
andrewiggins: Hmm I think I might've messed up this condition here... Need to look more at it

andrewiggins: Okay, I think the latest change fixes this by using `_children` as the sentinel property.

debug/test/browser/debug.test.js
```diff
@@ -80,7 +80,7 @@ describe('debug', () => {
 	it('should print an error on double jsx conversion', () => {
 		let Foo = <div />;
 		let fn = () => render(h(<Foo />), scratch);
-		expect(fn).to.throw(/createElement/);
+		expect(fn).to.throw(/JSX twice/);
```
andrewiggins: This test was giving a false positive because my change caused it to throw a different error that contained `createElement` when it should've been throwing this error. Caught this by looking at the code coverage and noticing that the line under the condition I changed was no longer covered lol.

src/create-element.js
developit: tbh that seems preferable.

andrewiggins: I'll PR this

andrewiggins: Renamed in #2297

PR #2295, opened 2020-01-31
Preserve state when using placeholders

andrewiggins:
Add a test and fix a bug where `null`, `false`, and `true` were removed from the render result before going into `diffChildren`. Calling `toChildArray` without a callback removes the values mentioned above, which we were doing on the result of render before going into `diffChildren`.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -1,7 +1,7 @@
 import { EMPTY_OBJ, EMPTY_ARR } from '../constants';
 import { Component } from '../component';
 import { Fragment } from '../create-element';
-import { diffChildren, toChildArray } from './children';
+import { diffChildren } from './children';
 import { diffProps } from './props';
 import { assign, removeNode } from '../util';
 import options from '../options';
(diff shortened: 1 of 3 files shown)


src/diff/index.js
```diff
@@ -165,9 +165,7 @@ export function diff(
 			tmp = c.render(c.props, c.state, c.context);
 			let isTopLevelFragment =
 				tmp != null && tmp.type == Fragment && tmp.key == null;
-			newVNode._children = toChildArray(
-				isTopLevelFragment ? tmp.props.children : tmp
-			);
+			newVNode._children = isTopLevelFragment ? tmp.props.children : tmp;
```
andrewiggins: Should this property be renamed to `_renderResult` since it isn't guaranteed to be an array until after `diffChildren` is complete? I vaguely remember this confusing people before. I was hesitant to do this immediately cuz it adds yet another property to VNode and this one is particularly short lived (it's just a way to pass the result into `diffChildren`)

test/browser/lifecycles/shouldComponentUpdate.test.js
```diff
@@ -707,4 +707,24 @@ describe('Lifecycle methods', () => {
 			expect(scratch.textContent).to.equal('foo');
 		});
 	});
+
+	it('should correctly render when sCU component has null children', () => {
```
andrewiggins: This test was in this branch from a while ago. Not sure what it was trying to do but figured it can't hurt 🤷‍♂

PR #2300, opened 2020-02-03
Run karma tests against build output

### LINKED ISSUE #2062: Suspense/Lazy and debug
Found an issue with using Suspense/Lazy and debug.

I'm getting the following error in the console when debug is being used:
>Uncaught Error: Missing Suspense. The throwing component was: Lazy

If I remove debug the app works correctly.

All of my Lazy routes are nested in a Suspense.

andrewiggins:
Add npm script `test:karma:minify` to run karma tests against the minified output of our build and set that script as the default that is run in CI builds. Because this script runs against the build output, you must `build` before running `test:karma:minify` in order to changes made in `src` to be tested. Running tests this way should help us catch bugs in our mangle.json's sooner.

`test:karma` and `test:karma:watch` behavior is unchanged and they still run against the original, unminified source files. This behavior means a test that passes in `test:karma` many not pass in `test:karma:minify` if there is a minification bug (likely a missing mangle.json prop).

Some notable changes:
* Added some missing mangle.json props, notably `_childDidSuspend` and some hooks props
* Re-synced all of the mangle.json files to be the same (except compress options which should only apply to core)
* Renamed some test files regarding options to all end in `.options.test.js`
* Use `babel-plugin-transform-rename-properties` to "minify" the properties of some of the test files, namely the options test. 
* Replace karma babel options with babel override in root config (more details in comment below)
* Remove unnecessary karma-source-map-support package. Seems to only be necessary for browserify pipelines. Regardless our stack traces without this plugin show both the mapped and unmapped line and filename.

Fixes #2062


karma.conf.js
```diff
@@ -61,40 +62,16 @@ var localLaunchers = {
 	}
 };
 
-const babelOptions = (options = {}) => {
```
andrewiggins: If my understanding is correct, these separate options are needed to accomplish two things:
1. Selectively apply `transform-react-jsx-source` plugin for tests that need it.
2. Add istanbul plugin for code coverage

For (1) I've moved that special override to the babel.config.js file using the override section to selectively add that that plugin only the the relevant test files using the same regex specified in the webpack loader. All the tests still pass, so I'm assuming this works as expected.

For (2) I just inlined that config in the webpack loader since that is the only place it needs to be specified. I'm uncertain if there is a way to do this in the babel.config.js and only have it applied to source files during test. If so, it might be nice to just have one babel config, but this works for now.

karma.conf.js
```diff
@@ -61,40 +62,16 @@ var localLaunchers = {
...
+const subPkgPath = pkgName =>
+	path.join(__dirname, pkgName, !minify ? 'src' : '');
 
 module.exports = function(config) {
 	config.set({
 		browsers: sauceLabs
 			? Object.keys(sauceLabsLaunchers)
 			: Object.keys(localLaunchers),
 
-		frameworks: ['source-map-support', 'mocha', 'chai-sinon'],
```
andrewiggins: See note in description. I don't think this is needed anymore

karma.conf.js
```diff
@@ -61,40 +62,16 @@ var localLaunchers = {
...
-			'@babel/plugin-proposal-object-rest-spread',
-			options.debug && '@babel/plugin-transform-react-jsx-source',
-			'@babel/plugin-transform-react-jsx',
-			'babel-plugin-transform-async-to-promises'
-		].filter(Boolean),
-		ignore: ['./dist']
-	};
-};
+const subPkgPath = pkgName =>
+	path.join(__dirname, pkgName, !minify ? 'src' : '');
```
andrewiggins: Selectively load the minified files or the src files depending on the MINIFY env variable

andrewiggins:
Note: Because we are now running tests against minified builds (which don't have comments such as `/* istanbul ignore next */), code coverage will be slightly lower cuz one statement isn't covered in compat which isn't ignored anymore.

karma.conf.js
marvinhagemeister: yup, I only did that to enable the plugin for certain files and didn't know any better way to do it :+1: Glad there is a better one :100:

babel.config.js
```diff
@@ -20,6 +32,21 @@ module.exports = function(api) {
 			'@babel/plugin-transform-react-jsx',
 			'babel-plugin-transform-async-to-promises'
 		],
-		ignore: ['./dist']
+		include: ['**/src/**/*.js', '**/test/**/*.js'],
+		overrides: [
```
marvinhagemeister: TIL about babel overrides :sweat_smile:

PR #2303, opened 2020-02-03
Code golf compat/suspense-list and diff/props

theZieger:


src/diff/props.js
```diff
@@ -55,16 +55,15 @@ function setStyle(style, key, value) {
  * @param {boolean} isSvg Whether or not this DOM node is an SVG node or not
  */
 function setProperty(dom, name, value, oldValue, isSvg) {
-	if (isSvg) {
-		if (name === 'className') {
-			name = 'class';
-		}
-	} else if (name === 'class') {
+	if (name === 'key' || name === 'children') return;
```
JoviDeCroock: return is a non-minified keyword so this most of the time will introduce extra code compared to making one big if else branch

compat/test/browser/render.test.js
```diff
@@ -190,7 +190,6 @@ describe('compat render', () => {
 		update();
 		rerender();
 
-		console.log(scratch.textContent);
```
JoviDeCroock: Good spot!

compat/src/suspense-list.js
```diff
@@ -30,10 +30,8 @@ const resolve = (list, child, node) => {
 	// callbacks won't get queued in the node anyway.
 	// If revealOrder is 'together' then also do an early exit
 	// if all suspended descendants have not yet been resolved.
-	if (
-		!list.props.revealOrder ||
-		(list.props.revealOrder[0] === 't' && list._map.size)
-	) {
+	const revealOrder = list.props.revealOrder;
```
JoviDeCroock: Most of the time assigning won't save bytes unless used 3+ times (var adds 3bytes)

theZieger:
@JoviDeCroock thanks for the fast response ... been hoping for the bot comment to get the real result on the golfing ... could only compare the output of `npm run build` locally so I was hoping to get a better insight from the bot ... should maybe mark this as WIP.

JoviDeCroock:
You can check it here: https://github.com/preactjs/preact/pull/2303/checks?check_run_id=424170676

theZieger:
What a bummer. is there a reliable way to look this up locally?

JoviDeCroock:
Most of the time I just do `yarn build:<target>` and compare with changes

theZieger:
Bildschirmfoto von 2020-02-03 23-23-44
left: my pr
right: master

At least the changes to compat look still slightly better.
Don't know what I compared regarding preact locally :man_shrugging:
Should maybe revert only those...

theZieger:
Ok it's compat.module is in fact increasing. I'm just to stupid to compare numbers ...

JoviDeCroock:
Your module build is 4b larger on the left

theZieger:
Yeah. Still don't know what I compared locally ... should just pull to terminals up instead of switching back and forth between branches ...
will close this and open one for the console.log in a sec.

PR #2304, opened 2020-02-03
Fix: Resolves Node submodule subpath error

4cm4k1:
After installing v10.3.0 (which implements Node submodule path support), _and_ if you are using Node v13, you will encounter this error:
```shell
(node:75051) ExperimentalWarning: Conditional exports is an experimental feature. This feature could change at any time
(node:75051) UnhandledPromiseRejectionWarning: Error: Package exports for '/Users/4cm4k1/Code/personal-website/node_modules/preact' do not define a './compat/server' subpath
    at applyExports (internal/modules/cjs/loader.js:529:13)
    at resolveExports (internal/modules/cjs/loader.js:549:12)
    at Function.Module._findPath (internal/modules/cjs/loader.js:664:22)
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:967:27)
    at Function.Module._load (internal/modules/cjs/loader.js:862:27)
    at Module.require (internal/modules/cjs/loader.js:1040:19)
    at require (internal/modules/cjs/helpers.js:72:18)
    at Object.<anonymous> (/Users/4cm4k1/Code/personal-website/node_modules/react-dom/server.js:1:18)
    at Module._compile (internal/modules/cjs/loader.js:1151:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1171:10)
(node:75051) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:75051) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
This PR adds `./compat/server.js` as an export path, which is currently only written in `require` syntax, and resolves the error. 😄

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,216 +1,216 @@
 {
-	"name": "preact",
-	"amdName": "preact",
-	"version": "10.3.0",
-	"private": false,
-	"description": "Fast 3kb React-compatible Virtual DOM library.",
-	"main": "dist/preact.js",
-	"module": "dist/preact.module.js",
-	"umd:main": "dist/preact.umd.js",
-	"unpkg": "dist/preact.umd.js",
-	"source": "src/index.js",
-	"exports": {
-		"./compat": {
-			"require": "./compat/dist/compat.js",
-			"import": "./compat/dist/compat.module.js",
-			"browser": "./compat/dist/compat.umd.js"
-		},
-		"./debug": {
-			"require": "./debug/dist/debug.js",
-			"import": "./debug/dist/debug.module.js",
-			"browser": "./debug/dist/debug.umd.js"
-		},
-		"./hooks": {
-			"require": "./hooks/dist/hooks.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,213 +1,216 @@
 {
-  "name": "preact",
```
JoviDeCroock: I think your liner is doing something odd here, can merge after we fix this

4cm4k1: @JoviDeCroock You're right! I'm getting mad at VS Code. It kept auto-formatting spaces to tabs, but I have fixed it now! 👍

PR #2305, opened 2020-02-03
removes warning from tests by removing console.log in render.test.js

theZieger:
Removes a forgotten console.log from the render.test.js that I saw while working on some unsuccessful code golfing


theZieger:
Thank you for your fast response on the other PR! 
Also thanks for the insights in code golfing. Even though they were unsuccessful. 
100% appreciated them.

marvinhagemeister:
Oh no that's on me :see_no_evil: Glad someone caught it!

PR #2307, opened 2020-02-04
move to slice call -2b

JoviDeCroock:
Previous: https://github.com/preactjs/preact/pull/2289


src/diff/props.js
```diff
@@ -93,7 +93,7 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 		}
 	}
 	// Benchmark for comparison: https://esbench.com/bench/574c954bdb965b9a00965ac6
-	else if (name[0] === 'o' && name[1] === 'n') {
+	else if (name.slice(0, 2) === 'on') {
```
cristianbote: How's that linked esbench looking with slice?

PR #2311, opened 2020-02-04
move to umd and make browser a modular export

JoviDeCroock:
This was proposed in: https://github.com/preactjs/preact/issues/2282#issuecomment-578995288

Would like some eyes on this

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
     ".": {
       "require": "./dist/preact.js",
       "import": "./dist/preact.module.js",
-      "browser": "./dist/preact.umd.js"
+      "browser": "./dist/preact.module.js",
+      "umd": "./dist/preact.umd.js"
     },
     "./compat": {
       "require": "./compat/dist/compat.js",
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "./test-utils": {
       "require": "./test-utils/dist/testUtils.js",
       "import": "./test-utils/dist/testUtils.module.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -13,22 +13,26 @@
...
       "import": "./hooks/dist/hooks.module.js",
-      "browser": "./hooks/dist/hooks.umd.js"
+      "browser": "./hooks/dist/hooks.module.js",
+      "umd": "./hooks/dist/hooks.umd.js"
     },
     "./test-utils": {
       "require": "./test-utils/dist/testUtils.js",
       "import": "./test-utils/dist/testUtils.module.js",
-      "browser": "./test-utils/dist/testUtils.umd.js"
+      "browser": "./test-utils/dist/testUtils.umd.js",
```
marvinhagemeister: I think this needs to be an em bundle too?

package.json
```diff
@@ -13,22 +13,26 @@
     "./compat": {
       "require": "./compat/dist/compat.js",
       "import": "./compat/dist/compat.module.js",
-      "browser": "./compat/dist/compat.umd.js"
+      "browser": "./compat/dist/compat.module.js",
+      "umd": "./compat/dist/compat.umd.js"
     },
     "./debug": {
-      "require": "./debug/dist/debug.js",
+      "require": "./compat/dist/debug.js",
```
Kanaye: Is that change from `./debug` to `./compat` correct ?

JoviDeCroock: Thanks!

Issue #2321, opened 2020-02-08
Preact 10.3.0 and up doesn't work with nuka-carousel


mhmdanas:
The following component:
```javascript
import { h, render } from 'preact';
import Carousel from 'nuka-carousel';

function Foo() {
    return (
        <Carousel>
            <img src='https://via.placeholder.com/400/ffffff/000000/&text=Slide1' />
            <img src='https://via.placeholder.com/400/ffffff/000000/&text=Slide2' />
            <img src='https://via.placeholder.com/400/ffffff/000000/&text=Slide3' />
        </Carousel>
    );
}
```
Stopped working correctly on Preact 10.3.0. In Preact 10.2.1 and down, it works correctly. It now starts from slide 3 and the other slides don't even show anything. 

**Note**: I tried to create a CodeSandbox, but I don't have a lot of time and it's my first time using it, so I couldn't do it.

reznord:
It is unable to resolve `preact-compat` relatively.

Here's a sandbox link: https://codesandbox.io/s/billowing-https-18jvu

JoviDeCroock:
Here's a working sandbox showing the issue: https://codesandbox.io/s/sleepy-pond-igsyd

PR #2322, opened 2020-02-08
fix children.map/forEach (missing index)

JoviDeCroock:
Fixes: https://github.com/preactjs/preact/issues/2321

When mapping over the children they were missing the second parameter (currentIndex)


src/diff/index.js
```diff
@@ -163,6 +163,7 @@ export function diff(
 			c._parentDom = parentDom;
 
 			tmp = c.render(c.props, c.state, c.context);
+
```
ForsakenHarmony: 👀

PR #2324, opened 2020-02-09
automated publishing workflow

JoviDeCroock:
This will automatically publish our new version to npm when we publish a release

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,21 @@
+name: Publish
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2-beta
+        with:
+          fetch-depth: 1
+      - uses: actions/setup-node@v1
+        with:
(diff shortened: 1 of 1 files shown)


JoviDeCroock:
Internal talk brought up a security concer, closing

cecilemuller:
Was the concern that using a NPM token to publish lowers the security from two-factors to single-factor ?

JoviDeCroock:
@cecilemuller One of the more important ones yes

marvinhagemeister:
@cecilemuller yep, we require 2FA for everyone to be able to publish to npm and we're reluctant to turn that off. Nonetheless we'd really love to automate more of our release process. Not sure how though 🙈

cecilemuller:
@marvinhagemeister I hear you, I do this step manually too because I refuse to remove the 2FA requirement, but I'm always on the lookout in case a solution emerges (that's even why I got curious when I saw the name of this PR).

cecilemuller:
Now that [Github acquired NPM](https://github.blog/2020-03-16-npm-is-joining-github/), perhaps they will provide a safer way to publish packages in the future.

Issue #2150, opened 2019-11-25
How to properly alias Preact/compat with TypeScript?


talentlessguy:
# Reproducing

1. `git clone https://github.com/talentlessguy/preact-emotion-ts.git`
2. `yarn install`
3. Open vscode, with these settings:

```json
{
    "eslint.packageManager": "yarn",
    "npm.packageManager": "yarn",
    "editor.formatOnSave": true,
    "typescript.updateImportsOnFileMove.enabled": "always",
    "editor.suggestSelection": "first",
    "eslint.autoFixOnSave": true,
    "eslint.alwaysShowStatus": true,
    "prettier.configPath": "/home/v1rtl/.prettierrc.yml",
    "[typescriptreact]": {
        "editor.defaultFormatter": "esbenp.prettier-vscode"
    }
```

And see this error on `<Header>` component:

```
JSX element type 'Header' does not have any construct or call signatures.
```

# Versions

**emotion**: 10.0.22
**preact**: 10.0.5
**eslint**: 6.7
**typescript**: 3.7.2

# Screenshot

![image](https://user-images.githubusercontent.com/35937217/69535027-057ae680-0f8c-11ea-9db4-270127febed1.png)

The weirdiest thing is that `npx eslint` doesn't show anything like this

```
preact-emotion-ts is 📦 v1.0.0 via ⬢ v13.2.0 
❯ npx eslint src/*.tsx --fix
Warning: React version was set to "detect" in eslint-plugin-react settings, but the "react" package is not installed. Assuming latest React version for linting.
```

I don't what the issue is coming from but I suppose it is a problem with typings of preact (?)

JoviDeCroock:
The reason behind `npx eslint src/*.tsx --fix` not reporting this is because it is a `tsx` error, this means that the typescript compiler is throwing this and isn't related to linting.

JoviDeCroock:
This seems to be related to `/compat/src/index.d.ts` missing the deprecated `SFC` shorthand.

talentlessguy:
> The reason behind `npx eslint src/*.tsx --fix` not reporting this is because it is a `tsc` (the process that compiles typescript) error, this means that the typescript compiler is throwing this and isn't related to linting.

okay, thanks for pointing out the issue.

`npx tsc` outputs this:

```
❯ npx tsc
node_modules/@emotion/core/types/index.d.ts:17:8 - error TS2307: Cannot find module 'react'.

17 } from 'react'
          ~~~~~~~

[...]

Issue #1930, opened 2019-09-13
10.0.0-rc.3 preact/compat typescript incompatibilities


bvella:
First off, well done on this excellent library. It boggles my mind how smaller this is from react.

I have some existing react libraries and a react app which I am trying out with preact. It is working fine but I had to do some modifications to the typescript typings as below:

I added the following definitions to compat/src/index.d.ts :
```
export type Component<P = {}, S = {}, SS = any> = preact.Component<P, S>;
export type ReactElement<P = {}, S = any> = preact.VNode<P>;
export import FunctionComponent = preact.FunctionComponent;
export import ComponentClass = preact.ComponentClass;
export import ComponentType = preact.ComponentType;
export import Context = preact.Context;
export import ReactNode = preact.ComponentChildren;
export type CSSProperties = string | {[key: string]: string | number};
export type AllHTMLAttributes<T> = JSX.HTMLAttributes;
```

I also changed jsx to a global namespace in src/jsx.d.ts:
```
...
export = JSX;
		
declare global {
export namespace JSX {
...
}
}
```

I also noticed that there are some subtle types missing, like e.g. the target in the onChange event for an input tag is not typed. I had this jsx line in which the event's target attribute was correctly inferred to be an html input element from the react typings:

```
<input type="text" name="name" onChange={ e => setName(e.target.value) } />
```
that I had to change to the following as the preact event handlers all use the same event definition with target being untyped:
```
<input type="text" name="name" onChange={ e => setName((e.target as HTMLInputElement).value) } />
```

Are these by design or is there a reason I am not aware of for these missing types? I can go through the react typings and copy over some stuff to improve type checking, if you think this is of benefit to preact. These would be typescript typing updates only, requiring no changes in preact but improving greatly both switching to preact from react and fix subtle type checking errors like the event above.

marvinhagemeister:
The namespacing of `JSX` is done one purpose. It was initially started in #1036 and if you follow the git history, you'll find a few PRs that built on that.

Regarding the incorrect `target` types: That's definitely a missing feature in our types. We'd welcome any PRs that improve that :+1: :100:

ericlery:
There's still no type safety for event handling in Preact till this day?

kasvith:
Wow, jumped into the same issue and realized its a bug on preact

alextompkins:
[...]

PR #2329, opened 2020-02-11
Improve react typescript compatibility

bvella:


bvella:
Fixed almost all of #1930

bvella:
For clarity, I am trying to get an existing react project working with preact. I encountered a number of typescript incompatibilities as described in #1930 and attempted to fix them with the proposed changes.

I added the following in my tsconfig.json:

```
{
  "compilerOptions": {
    ...
    "paths": {
      "react": ["node_modules/preact/compat/"],
      "react-dom": ["node_modules/preact/compat/"]
    }
  }
}
```

And in package.json applied a browserify global transform using aliasify to alias react and react-dom to preact/compat.

The only outstanding issue I have is typing of html event targets, which I will attempt to fix in a separate pull request

bvella:
Incorporated https://github.com/preactjs/preact/issues/2150#issuecomment-558622551

ForsakenHarmony:
I know it's been a while, there are some more types missing in #2414 if you want to add those

brookback:
Any chance this could be merged? 👼

compat/src/jsx.d.ts
```diff
@@ -0,0 +1,4 @@
+import { JSXInternal } from '../../src/jsx';
+
+export = JSXInternal;
+export as namespace JSX;
```
ForsakenHarmony, line 4: I think this broke typescript before

https://github.com/preactjs/preact-cli/issues/1030

toniopelo:
When using `@apollo/client`

```
ERROR in ./node_modules/@apollo/client/react/context/ApolloConsumer.d.ts(4,55):
TS2694: Namespace 'React' has no exported member 'ReactChild'.
```

I wasn't using this branch though, but I guess you need to add `ReactChild` too ?

dgreene1:
@bvella could you update this branch with the base branch so it can be re-approved and then released?

compat/src/jsx.d.ts
bvella: In what way? I skimmed through the referenced issue but did not understand how this breaks typescript. This file declares the JSX namespace, without which, typescript does not process jsx. This is also the same way react itself exports the JSX namespace

evenfrost:
Any update on this? Is there a complete workaround with all the changes needed to perform to a Preact project to be able to use typed React libraries in it until it merged?

matthias-ccri:
I would like to contribute to this effort..  I'm not sure how to do it besides creating a separate PR. Is that ok?

Issue #2331, opened 2020-02-11
Select onInput doesn't function in Microsoft Edge


kamilgrzegorczyk:
Hello,

Browser: Microsoft Edge 44
Preact 10.3.1

Way to reproduce:
- visit the url https://preactjs.com/guide/v10/forms#select-input
- run the code in REPL
- the value is not getting updated (when you select any of the options)

Expected:
The value is updated properl and onInput callback is getting fired.

marvinhagemeister:
Can't reproduce the issue in the chromium based versions of edge. Can confirm that the issue occurs on older versions with the Chakra engine.

kamilgrzegorczyk:
Yeah, this is also my observation - new Edge engine does not have that issue. Unfortunately Microsoft Edge 44 was released just few months ago so it's hard to treat it as an unsupported browser.

marvinhagemeister:
We never said that we don't support edge. In fact the oldest version we officially support is Internet Explorer 11. This is definitely a bug in Preact.

kamilgrzegorczyk:
I believe that this might be due to the fact that Microsoft Edge does not support the onInput event on the `<select>` tags. I have used onchange there.

Source:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event
> Not supported on select, checkbox, or radio inputs.

developit:
I think we can find a way to patch this since we're already special-casing onChange.

PR #2333, opened 2020-02-11
alias onInput to onChange for select fields

JoviDeCroock:
fixes: https://github.com/preactjs/preact/issues/2331

### DIFF
--- a/test/browser/select.test.js
+++ b/test/browser/select.test.js
@@ -46,6 +46,14 @@ describe('Select', () => {
 		expect(scratch.firstChild.value).to.equal('B');
 	});
 
+	it('should alias onInput to onChange', () => {
+		const func = () => {};
+		render(<select onInput={func} />, scratch);
+
+		expect(scratch.firstChild._listeners.change).to.not.be.undefined;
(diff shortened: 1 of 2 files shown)


test/browser/select.test.js
```diff
@@ -46,6 +46,21 @@ describe('Select', () => {
...
+		render(
+			<select onInput={func}>
+				<option value="A">A</option>
+				<option value="B">B</option>
+				<option value="C">C</option>
+			</select>,
+			scratch
+		);
+
+		expect(scratch.firstChild._listeners.change).to.not.be.undefined;
```
andrewiggins: I think to get this test to pass, don't rely on the internal properties. Instead, use our `createEvent` helper to fire a `change` event and verify the `onInput` spy is called. See tests in `events.test.js` for an example

marvinhagemeister:
I'm torn a bit here. The spec is not super clear here, but it says that `input` is available for _all_ form controls. This works in every browser except chakra edge and IE to my knowledge. From that point of view we should make the change in core imo.

JoviDeCroock:
Closing due to the concern expressed in the above comment

PR #2335, opened 2020-02-11
Select polyfill alt

developit:
Alternative solution. There are a few properties that exist on `<select>` elements but not other inputs. This isn't ideal from a performance perspective, but the check only runs on `onInput` handlers, so the impact should be minimal. Alternatively, we could plumb `vnode` through prop diffing functions and check `vnode.type`.

Properties audit:

```js
const select = document.createElement('select');
const others = ['input', 'textarea', 'div', 'option', 'optgroup'].map(x=>document.createElement(x));
others.push(document.createTextNode(''));
for (let i in select) {
  if (!others.some(o => i in o)) {
    console.log(`select.${i} = ${select[i]}.`);
  }
}

/*
// results:
select.options = [object HTMLOptionsCollection]
select.selectedOptions = [object HTMLCollection]
select.selectedIndex = -1.
select.item = function item() { [native code] }
select.namedItem = function namedItem() { [native code] }
select.add = function add() { [native code] }
*/
```

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -10,12 +10,12 @@ import options from '../options';
  * @param {boolean} isSvg Whether or not this node is an SVG node
  * @param {boolean} hydrate Whether or not we are in hydration mode
  */
-export function diffProps(dom, newProps, oldProps, isSvg, hydrate) {
+export function diffProps(dom, newProps, oldProps, isSvg, hydrate, type) {
 	let i;
 
 	for (i in oldProps) {
 		if (!(i in newProps)) {
-			setProperty(dom, i, null, oldProps[i], isSvg);
+			setProperty(dom, i, null, oldProps[i], isSvg, type);
 		}
 	}
 
@@ -26,7 +26,7 @@ export function diffProps(dom, newProps, oldProps, isSvg, hydrate) {
 			i !== 'checked' &&
 			oldProps[i] !== newProps[i]
 		) {
-			setProperty(dom, i, newProps[i], oldProps[i], isSvg);
+			setProperty(dom, i, newProps[i], oldProps[i], isSvg, type);
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


src/diff/props.js
```diff
@@ -97,6 +97,12 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 		let useCapture = name !== (name = name.replace(/Capture$/, ''));
 		let nameLower = name.toLowerCase();
 		name = (nameLower in dom ? nameLower : name).slice(2);
+		
+		// Fix onInput not firing on select/radio/checkbox in Edge <= 18.
+		if (name === 'input' && (
+			type === 'select' ||
+			type === 'input' && /rad|check/.test(props.type)
```
cristianbote: 🤔 are `props` defined in here? Couldn't find them. Maybe I'm missing something.

sventschui:
```suggestion
			type === 'input' && /rad|check/.test(type)
```

I guess @cristianbote is right here :)

marvinhagemeister: I think we need to differentiate `vnode.type` and the input type which is usually in `props.type` here. As far as I can tell both are needed to pass this check. So we'd need to pass `props` through the arguments.

cristianbote: I was thinking this is what @developit wanted but, since the first condition checks that `type` is strictly equal to 'input' the next condition will never be true, isn't it? Cause that'll end-up with `/rad|check/.test('input')` which is not true.

ForsakenHarmony: the `type` in scope right now is the vnode type (html tag / component)

we need `props.type`

cristianbote: Oh, that means that `value` should be there instead.

cristianbote:
```suggestion
             if ((name === 'input' || /rad|check/.test(name)) &&
			(type === 'select' || type === 'input')
```
This one makes the tests to pass for radio and checkboxes as well. Is this what you had in mind @developit? 😄

ForsakenHarmony: if you're talking about the `value` in scope, no
that's the event listener function

cristianbote: You are right, I've added a solution below.

PR #2337, opened 2020-02-13
cleanup hooks on unmount

JoviDeCroock:
Making this PR for pika builds to try in the benchmark

Researching the potential memory leak


developit:
Ah interesting. If this ends up being the root of the memory leak in that demo, it'll be nice that we're only taking the size hit on `/hooks`.

JoviDeCroock:
Closing in favor of https://github.com/preactjs/preact/pull/2338

Issue #2336, opened 2020-02-12
Preact causes memory leak on useState?


avkonst:
I have spotted memory growth in a [performance benchmark](https://github.com/krausest/js-framework-benchmark). I have compared the same benchmark app with react. React clears memory when a component with useState is unmounted. Preact holds the state value after unmount:

![image](https://user-images.githubusercontent.com/3971413/74377643-34954b80-4e49-11ea-871c-ea8df9976abf.png)

I am not sure about the cause and if it is expected behaviour, but it definitely spoils performance comparison results for Preact and probably a red flag, which is worth investigating further.

## Here is how to reproduce it:

- Clone https://github.com/avkonst/js-framework-benchmark
- npm install
- npm start (notice webserver started, if it does not start check prerequisites here: https://github.com/avkonst/js-framework-benchmark)
- cd ./frameworks/keyed/preact-hookstate
- npm install
- npm run build-prod
- open crome at http://localhost:8080/frameworks/keyed/preact-hookstate/
- open development tools, memory tab
![image](https://user-images.githubusercontent.com/3971413/74378069-1e3bbf80-4e4a-11ea-89fe-612c366c58b8.png)

- take heap snapshot using chrome devtools
![image](https://user-images.githubusercontent.com/3971413/74378084-28f65480-4e4a-11ea-95ef-9cb2a6ddb7c0.png)

- click "Create 1000 rows"
![image](https://user-images.githubusercontent.com/3971413/74378093-2f84cc00-4e4a-11ea-9ca9-d91a38a9055f.png)

- take heap snapshot using chrome devtools
- click "Clear"
![image](https://user-images.githubusercontent.com/3971413/74378129-3d3a5180-4e4a-11ea-9d44-691a932853a6.png)

- take heap snapshot using chrome devtools

- go to snapshot 2 and setup comparison with snapshot 1
![image](https://user-images.githubusercontent.com/3971413/74378196-62c75b00-4e4a-11ea-9022-1a4383ccbedf.png)

- notice an object with 2001 allocations (this is the Hookstate state link referencing the state variable placed into useState)
- go to snapshot 3 and setup comparison with snapshot 2
![image](https://user-images.githubusercontent.com/3971413/74378314-9ace9e00-4e4a-11ea-8205-1380d590f6ec.png)

- notice only 1 state link deleted, although it is expected 2001 are deleted, at least eventually. 

- retantion tree for the state variable from useState() call is looking strange, I can not interpret much here:
[...]

PR #2338, opened 2020-02-13
[draft] potential fix for memory leak

developit:
Just a thought based on the retaining objects seen in #2336.

Fixing this would be another potential upside we'd gain from moving to a persistent backing tree.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -430,7 +430,7 @@ export function unmount(vnode, parentVNode, skipRemove) {
 			}
 		}
 
-		r.base = r._parentDom = null;
+		r.base = r._parentDom = r._vnode = vnode._component = null;
 	}
 
 	if ((r = vnode._children)) {


src/diff/index.js
```diff
@@ -430,7 +430,7 @@ export function unmount(vnode, parentVNode, skipRemove) {
 			}
 		}
 
-		r.base = r._parentDom = null;
+		r.base = r._parentDom = r._vnode = vnode._component = null;
```
developit: could we switch to `undefined` here?

JoviDeCroock (approved):
This is a much better solution yes, I'm wondering what our best guess is at solving circular references in child-parent? Should we start drawing out some idea's for backing-tree?
This does seem to mess with scheduled effects, I'm wondering how it's possible an unmount has a yet scheduled effect in progress (and certainly a layoutEffect)

src/diff/index.js
andrewiggins: I don't see why not

developit:
@JoviDeCroock perhaps if we null'd the `._hooks` reference when unmounting? The demo in the linked issue continually remounts trees, so it's very likely that this leak happens for unmounted components.

JoviDeCroock:
@developit yes that was my initial suggestion but very hooks specific.... I think we need to clean up our afterPaintEffects on unmount as well

JoviDeCroock:
Closing: draft targeting `r.base = r._parentDom = null`, a line that no longer exists. v11 unmount already nulls `_dom`/`_component`/`_parent`.

PR #2343, opened 2020-02-14
[experiment] Support for JSX2

developit:
Note: I don't necessarily think this is something we want to do. I just wanted to get an implementation up for folks investigating on the compilation side of things.

https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#jsx-transform-changes

### DIFF
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -1,4 +1,5 @@
 import options from './options';
+import { assign } from './util';
 
 /**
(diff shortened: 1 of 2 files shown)


JoviDeCroock:
Closing: JSX2 experiment; this was not a direction to pursue, and v11 ships a real `jsx-runtime` package.

PR #2347, opened 2020-02-14
Add support for false value in aria-attributes

marvinhagemeister:
cc @robertknight

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -133,7 +133,17 @@ function setProperty(dom, name, value, oldValue, isSvg) {
 					value
 				);
 			}
-		} else if (value == null || value === false) {
+		} else if (
+			value == null ||
+			(value === false &&
+				// ARIA-attributes have a different notion of boolean values.
+				// The value `false` is different from the attribute not
+				// existing on the DOM, so we can't remove it. For non-boolean
+				// ARIA-attributes we could treat false as a removal, but the
+				// amount of exceptions would cost us too many bytes. On top of
+				// that other VDOM frameworks also always stringify `false`.
+				!/^ar/.test(name))
+		) {
 			dom.removeAttribute(name);
 		} else {
 			dom.setAttribute(name, value);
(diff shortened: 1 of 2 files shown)


marvinhagemeister:
Please don't merge this yet. There are some things we should check first, mainly I'm worried that setting a falsey value to other aria attributes may not work as expected anymore.

marvinhagemeister:
It looks to be a bit more involved as some aria attributes have a tristate value instead of plain booleans, like [aria-grabbed](https://www.w3.org/WAI/PF/aria-1.1/states_and_properties#aria-grabbed)

| Value | Description |
|---|---|
| false | Indicates that the element supports being dragged. |
| true | Indicates that the element has been "grabbed" for dragging. |
| undefined (default) | Indicates that the element does not support being dragged. |

marvinhagemeister:
Ok, so I went through the spec and checked each attribute for more edge cases. There are none other than the one which lead to this PR where some attributes have three states. Attributes like `aria-checked` treat `false` as unchecked, whereas the attribute not being present is interpreted as `not-checkable`.

I also compared this with the behaviour present in React and they always stringify `false`, even if the aria attribute is not boolean like (e.g. `aria-labelledby`).

Therefor it should be safe for us too, to always stringify `false`.

robertknight:
Wow, thanks for such a quick response on this @marvinhagemeister! The current `/^ar` regex works for [known HTML attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#Attribute_list) although it could cause unexpected results for new attributes in future, custom attributes or embedded markup (eg. SVG or MathML). I believe React always requires string values for unknown attribute names (and warns if you pass a boolean), so this might not be an issue.

marvinhagemeister:
@robertknight I checked the current specs and for now the regex only matches aria attributes. If new ones are added to the spec we'll need to update it, for sure. With Custom Elements it get's more complex as they may theoretically also use something like `<foo-bar aria-custom-foo="whatever">`. I'm not sure how to satisfy everyone here :S

Issue #2227, opened 2020-01-03
Large child array diff performance


t-veor:
Hello,
I was looking at performance of updating large arrays of child components which led me to looking at this for loop here: https://github.com/preactjs/preact/blob/master/src/diff/children.js#L83
(summarised with some pseudocode)
```js
oldVNode = oldChildren[i];

if (/* oldVNode matches childVNode */ ) {
    oldChildren[i] = undefined;
} else {
    for (j = 0; j < oldChildrenLength; j++) {
        oldVNode = oldChildren[j];
        if (/* oldVNode matches childVNode */) {
            oldChildren[j] = undefined;
            break;
        }
        oldVNode = null;
    }
}
```
If I understand this code correctly, when Preact gets an array of children for each one it checks to see if there was a child already there with the same type and key at the same index in the old array, and otherwise it does a linear search to see if any other components match. This means that if every element changes, this code is quadratic in the number of elements.

The problem is that every element or almost every element updating is actually very common (at least, I'm suggesting, having no evidence 😄) in array manipulations. Some examples are:
* Removing an element from the front of an array, like in a scrollback buffer.
* The user reordering a list by moving an element from near the start of a list to near the end.
* Adding new things to the front of an array, like in a news feed.

In these cases most of the elements are unchanged, but a large amount of them will have shifted forward or backwards by 1 in the array, and so the loop above is quadratic.

I think a simple change might be just to change the loop slightly, so there's a much better chance the loop finds the corresponding child earlier:
```js
    for (j = 0; j < oldChildrenLength; j++) {
        const idx = (j + i - 1) % oldChildrenLength;
        oldVNode = oldChildren[idx];
        if (/* oldVNode matches childVNode */) {
            oldChildren[idx] = undefined;
            break;
        }
        oldVNode = null;
    }
```
This way the loop starts looking at `i - 1` first but cycles around to all the elements in the array anyway, so it speeds up those cases since for each it only needs to perform 1-3 comparisons, and falls back to the old quadratic behaviour otherwise. This does make the loop slightly more complicated so might be a pessmisation in real use cases, but again I don't really have any evidence of real usage patterns and any potential performance gains so I'd just like to know if something like this was considered before.

Of course, it's possible to do away with possible quadratic time entirely by pre-computing a look-up table ahead of time on the keys of `oldChildren`, but I'm more wary of this since it needs to allocate a hashmap and do lookups, so in real use cases it could be much slower.

localvoid:
```js
[...]

PR #2348, opened 2020-02-15
implement  suggested perf improvement for child diffing

JoviDeCroock:
Seems like it worked, however the byte size changes is pretty big 😅 

https://github.com/preactjs/preact/issues/2227#issuecomment-570760823 <-- this is still subject to -1 property access and fails the algo

This part: `(j >>> 1) ^ -(j & 1)` fails because when j = 1 we'll get a -1 here and this will influence the rest of the algo

Resolves: https://github.com/preactjs/preact/issues/2227

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -81,10 +81,7 @@ export function diffChildren(
 					// Either oldVNode === undefined or oldChildrenLength > 0,
 					// so after this loop oldVNode == null or oldVNode is a valid value.
 					for (j = 0; j < oldChildrenLength; j++) {
-						const idx =
-							i === 0
-								? (j + i) % oldChildrenLength
-								: (j + i - 1) % oldChildrenLength;
+						const idx = i === 0 ? j : (j + i - 1) % oldChildrenLength;
 						oldVNode = oldChildren[idx];
 						// If childVNode is unkeyed, we only match similarly unkeyed nodes, otherwise we match by key.
 						// We always match by type (in either case).
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -70,12 +70,7 @@ export function diffChildren(
 				// (holes).
 				oldVNode = oldChildren[i];
 
-				if (
-					oldVNode === null ||
-					(oldVNode &&
-						childVNode.key == oldVNode.key &&
-						childVNode.type === oldVNode.type)
-				) {
+				if (oldVNode === null) {
 					oldChildren[i] = undefined;
 				} else {
 					// Either oldVNode === undefined or oldChildrenLength > 0,
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -81,7 +81,8 @@ export function diffChildren(
 					// Either oldVNode === undefined or oldChildrenLength > 0,
 					// so after this loop oldVNode == null or oldVNode is a valid value.
 					for (j = 0; j < oldChildrenLength; j++) {
-						oldVNode = oldChildren[j];
+						const idx = (i + ((j >>> 1) ^ -(j & 1))) % oldChildrenLength;
+						oldVNode = oldChildren[idx];
 						// If childVNode is unkeyed, we only match similarly unkeyed nodes, otherwise we match by key.
 						// We always match by type (in either case).
 						if (
@@ -89,7 +90,7 @@ export function diffChildren(
 							childVNode.key == oldVNode.key &&
 							childVNode.type === oldVNode.type
 						) {
-							oldChildren[j] = undefined;
+							oldChildren[idx] = undefined;
 							break;
 						}
 						oldVNode = null;
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -81,7 +81,9 @@ export function diffChildren(
 					// Either oldVNode === undefined or oldChildrenLength > 0,
 					// so after this loop oldVNode == null or oldVNode is a valid value.
 					for (j = 0; j < oldChildrenLength; j++) {
-						const idx = i === 0 ? j : (j + i - 1) % oldChildrenLength;
+						const idx =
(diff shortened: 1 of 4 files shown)


src/diff/children.js
```diff
@@ -81,15 +81,16 @@ export function diffChildren(
 					// Either oldVNode === undefined or oldChildrenLength > 0,
 					// so after this loop oldVNode == null or oldVNode is a valid value.
 					for (j = 0; j < oldChildrenLength; j++) {
-						oldVNode = oldChildren[j];
+						const idx = i === 0 ? j : (j + i - 1) % oldChildrenLength;
```
JoviDeCroock: We could transform this to

```
const idx = j + i === 0 ? j : (j + i - 1) % oldChildrenLength;
```

But this seems to goof up some focus tests:
- should maintain focus when removing elements around input
- Fragment - should preserve state of children nested at same level (adds stateful upate)
- Fragment - should preserve state with reordering in multiple levels (adds stateful upate)
- keys - should swap children efficiently (makes it more efficient, one operation instead of 3)
- should move keyed children to the end of the list  (makes it more efficient, one operation instead of 3)
- should reverse children effectively more efficient (11 --> 9 operations)

cristianbote: The last 3 items are spot on! 👍 Wondering about the other 3 what's up with those?

developit:
Thoughts on how to test this (maybe I'll get time):

1. make it so that OOB array access throws during testing:

```js
Object.defineProperty(Array.prototype, '-1', {
  get() { throw Error('Attempted to access array[-1]'); },
  set(v) { throw Error('Attempted to set array[-1] = '+v); }
});
```

2. Track mutations and demonstrate that this improves performance by reducing the number of childList mutations:

```js
const list = [1,2,3,4,5].map(v => <li key={v}>{v}</li>);

render(list, scratch);

let added = 0, removed = 0;
new MutationObserver(records => {
  for (const record of records) {
    added += record.addedNodes.length;
    removed += record.removedNodes.length;
  }
}).observe(scratch, {
  subtree: true,
  childList: true
});

// swap the second and fourth items
list = list.slice();
const fourth = list[3];
list[4] = list[1];
list[1] = fourth;
render(list, scratch);

expect(added).to.equal(1);
expect(removed).to.equal(1);
```

davidje13:
Rather than `const idx = i === 0 ? j : (j + i - 1) % oldChildrenLength`, I'd suggest:

```js
const idx = (j + i + oldChildrenLength - 1) % oldChildrenLength
```

Since that will compile to smaller code. That would also let you use the suggested magic algorithm for searching in both directions (makes this bit of code bigger, but means the special case code above for checking the current index can be removed):

```js
const idx = (i + ((j >>> 1) ^ -(j & 1)) + oldChildrenLength) % oldChildrenLength
```

JoviDeCroock:
@davidje13 this will force -1 access making the algo much less efficient

davidje13:
@JoviDeCroock that's what the `+ oldChildrenLength` fixes, unless I've misunderstood your concern.

Point is that `%` works as desired for all positive numbers, so a -1 can only happen if the input is negative, but by adding `oldChildrenLength` before performing a modulo, you ensure it will always be positive, therefore always produce a positive result.

Then, by combining that with the previously suggested algorithm which gives 0, -1, +1, -2, +2, etc. you can get rid of the early special-case for 0 (lines ~71-80), since it will start at 0 offset anyway. Should be quite a simplification when the dust settles, and no risk of negative index access because it will never go lower than `-oldChildrenLength/2 + oldChildrenLength` (which can never be negative).

JoviDeCroock:
Good shout, I misread pardon me for that, thanks a lot

test/browser/render.test.js
```diff
@@ -1198,6 +1198,40 @@ describe('render()', () => {
...
+		});
+
+		// swap the second and fourth items
+		list = list.slice();
+		const fourth = list[3];
+		list[4] = list[1];
+		list[1] = fourth;
+		render(list, scratch);
+		expect(scratch.innerHTML).to.equal(
+			'<li>1</li><li>4</li><li>3</li><li>4</li><li>2</li>'
```
andrewiggins, line 1228: Are there suppose to be two `<li>4</li>`? What happened to `<li>5</li>`?

test/polyfills.js
```diff
@@ -12,6 +12,15 @@ import 'core-js/fn/string/code-point-at';
 import 'core-js/fn/string/from-code-point';
 import 'core-js/fn/string/repeat';
 
+Object.defineProperty(Array.prototype, '-1', {
```
andrewiggins, line 15: This is great! ❤️

test/browser/render.test.js
JoviDeCroock: fixed

test/polyfills.js
marvinhagemeister: Agree, really love this for catching perf issues :+1:

davidje13:
You can delete at least lines 75–77 now to reduce the code size without changing the behaviour.

Also from my limited understanding it appears that with the existing code, if an element is moved to a point which used to be a gap, it will be replaced instead of moved (due to the null check on 74). Perhaps something for a different change, but it feels like there's a chance to somehow reduce the code further there while also making it more consistent. I'd have to get a much better understanding of this block before I could recommend an actual change there though.

davidje13:
@JoviDeCroock did you abandon these changes? I'd be happy to take over if you're done with them; I'm hoping to see this improvement land for one of my own projects, and I think there's still a lot of code that can be dropped.

Is there anything specific which needs to happen for this to be approved?

JoviDeCroock:
I’m going to refactor children in another way, instead of a one side traverse two sided to improve swap and insert ops

PR #2351, opened 2020-02-15
Fix diff against wrong oldDom in diffChildren

### LINKED ISSUE #2350: iframe reloads when parent's sibling returns null
This caught us out in a recent upgrade from preact 8 to 10.
It does not happen with preact 8, nor react.
If a component updates and returns null (having previously been non-null), then any siblings' iframes will be reloaded. I'm not sure if it's expected behaviour in preact, but it took a long time to debug!

### Reproduction
Example here: https://codesandbox.io/s/runtime-silence-no4zx

### Steps to reproduce
In the above example:
1. Type an 'e' in the input, causing the `Label` component to render.
2. Deleting the 'e' triggers `Label` to return null, which unexpectedly(?) causes the `Iframe` component's iframe to be recreated.

### Expected Behavior
The iframe is not reloaded when a separate component returns null.

### Actual Behavior
The iframe is reloaded.
Making the `Label` component return e.g. `<div></div>` instead of null fixes the issue.
Also, moving the `Label` component below the `Iframe` component also fixes it.

marvinhagemeister:
This PR fixes an issue with reconciliation of placeholders in `diffChildren`. We picked the correct `vnode` to diff, but the wrong old dom node in those scenarios.

Only suspense-related tests are failing for now. Need to check what's going on there. In the meantime this is a draft.

Fixes #2350

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -98,6 +98,12 @@ export function diffChildren(
 
 				oldVNode = oldVNode || EMPTY_OBJ;
 
+				// If the previous node was a placeholder we pick a new value
+				// for oldDom
+				if (oldDom === undefined) {
+					oldDom = oldVNode._dom || null;
+				}
+
 				// Morph the old element into the new one, but don't append it to the dom yet
 				newDom = diff(
 					parentDom,
@@ -197,6 +203,9 @@ export function diffChildren(
 
 						newParentVNode._nextDom = oldDom;
 					}
+				} else {
+					// If newDom is null we have a placeholder node
+					oldDom = undefined;
 				}
 			}
 


andrewiggins:
FYI, I was curious about this too and took a shot over in [fix/null-placeholder-sibling-placement](https://github.com/preactjs/preact/tree/fix/null-placeholder-sibling-placement). See ccd505e for where I'm at so far. I've got all but one test passing

marvinhagemeister:
@andrewiggins Love the solution in your PR :+1: The last test case is very interesting as it stumbles on `_children` not being an array. What I got so far is that the reason for that is this section in `diff()`:

```js
// Render often returns a single vnode instead of an array, which is correct
let isTopLevelFragment =

// We assign _children to either an array or non-array value. If it's not an
// array we wait for toChildArray() later to correct that.
newVNode._children = isTopLevelFragment ? tmp.props.children : tmp;
```

Trouble is that `getDomSibling` expects `_children` to always be an array.

Always ensuring that `_children` is set to `null` or an `array` fixes the failing tests and makes all others pass. I'm unsure where the best place to put it is though. Should we fix it add the above snippet or should we handle that in `getDomSibling`?

The fix for `diff()` would look like this without any golfing:

```js
let isTopLevelFragment =
newVNode._children = isTopLevelFragment ? tmp.props.children : tmp;

// Ensure _children is always an array. Kinda defeats the point of
// toChildArray and duplicates its logic
if (newVNode._children != null && !Array.isArray(newVNode._children)) {
  newVNode._children = [newVNode._children];
}
```

Compare this to the fix in `getDomSibling` (again, pre-golfing):

```js
export function getDomSibling(vnode, childIndex) {
	if (childIndex == null) {
		// Use childIndex==null as a signal to resume the search from the vnode's sibling
		return vnode._parent
            // Edge case for single children
			? !Array.isArray(vnode._parent._children)
				? vnode._parent._children._dom
				: getDomSibling(
						vnode._parent,
						vnode._parent._children.indexOf(vnode) + 1
				  )
			: null;
	}
```

Looking at our typing definitions we expect `vnode._children` to always be `null` or an `array`.

marvinhagemeister:
Replaced by #2355

PR #2353, opened 2020-02-16
fix failing test by defaulting to empty array

JoviDeCroock:


andrewiggins:
Just realized I messed up one of the tests in my branch. Rebase on top to see if everything still passes

JoviDeCroock:
@andrewiggins I found another fix which works on the rebased branch

src/diff/index.js
```diff
@@ -165,7 +165,11 @@ export function diff(
 			tmp = c.render(c.props, c.state, c.context);
 			let isTopLevelFragment =
 				tmp != null && tmp.type == Fragment && tmp.key == null;
-			newVNode._children = isTopLevelFragment ? tmp.props.children : tmp;
+			newVNode._children = isTopLevelFragment
+				? tmp.props.children
+				: Array.isArray(tmp)
+				? tmp
+				: [tmp];
```
JoviDeCroock: There are scenario's where we skip an update with a placeholder, in this case tmp returns null with render, this is not an array so we `[null]` it

PR #2354, opened 2020-02-16
Use queueMicrotask for scheduling

marvinhagemeister:
By suggestion from @ mathiasbynens https://mobile.twitter.com/mathias/status/1229142133904478209 🎉 

Size `-11 B` 🙌


JoviDeCroock:
This seems to have low browser compat though https://caniuse.com/#search=queueMicrotask

marvinhagemeister:
Good point. The data seems to be incorrect about edge though. The latest version does haver `queueMicrotask` available.

developit:
Would love to get performance numbers for this change. The size decrease is awesome, but we'll want to be confident about the perf of this (and of setTimeout fallback in Edge <= 18) to merge.

JoviDeCroock (approved):
I've been thinking about this and if `setTimeout` isn't a serious thing (which it shouldn't) I'm up for this.

marvinhagemeister:
We're hesitant to make the switch right now as that would push a lot of browsers over to `setTimeout`. We'll revisit this in the future when the browserlandscape has changed a bit 👍

JoviDeCroock:
Closing: already landed — v11 schedules with `(prevDebounce || queueMicrotask)(process)` in `src/component.js`.

PR #2355, opened 2020-02-17
Improve null placeholder DOM placement

### LINKED ISSUE #2350: iframe reloads when parent's sibling returns null

andrewiggins:
Previously if a Component went from rendering DOM to rendering `null`, our child diffing algorithm would have to re-place all children after the DOM that was removed. This PR improves the diffing algorithm to detect when this has happened and continue diffing from the removed DOM node's sibling.

Also reverts a change I made to a test in #2352 that invalidated the test.

Fixes #2350

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -98,12 +98,6 @@ export function diffChildren(
 
 				oldVNode = oldVNode || EMPTY_OBJ;
 
-				// If the previous node was a placeholder we pick a new value
-				// for oldDom
-				// if (oldDom === undefined) {
-				// 	oldDom = oldVNode._dom || getDomSibling(oldVNode)
-				// }
-
 				// Morph the old element into the new one, but don't append it to the dom yet
 				newDom = diff(
 					parentDom,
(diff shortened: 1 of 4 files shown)


src/diff/children.js
```diff
@@ -197,6 +197,14 @@ export function diffChildren(
 
 						newParentVNode._nextDom = oldDom;
 					}
+				} else if (
+					oldDom &&
+					oldVNode._dom == oldDom &&
+					oldDom.parentNode != parentDom
+				) {
+					// The above condition is handle null placeholders. See test in placeholder.test.js:
```
marvinhagemeister: Nit: I think it should be `to handle` here.

andrewiggins: Oh true, thanks!

PR #2358, opened 2020-02-18
remove warning for use(Layout)Effect

JoviDeCroock:
Due to a conversation on Slack, which rightfully pointed out that libraries often use this for animations, ... I wanted to put this PR up for discussion. We should refer to `eslint-plugin-hooks` instead of implementing this ourselves. eslint will ensure we only evaluate user code

Another good argument for removing it: https://github.com/preactjs/preact/blob/master/hooks/src/index.js#L184

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -83,6 +83,7 @@ export function initDebug() {
 			default:
 				isValid = false;
 		}
+
 		if (!isValid) {
 			let componentName = getDisplayName(vnode);
 			throw new Error(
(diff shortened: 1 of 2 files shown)


xiel:
Thanks for your quick reaction and the PR. Just re-posting my PR comment here:

---

There are valid use cases where useEffect and useLayoutEffect must be used without a dependency array. As these warnings are not configurable, they are a massive deterioration of the developer experience as it makes the console really messy.

In my option there shouldn't be (non-configurable) warnings while using the API as designed. 
The **dependency array is described as optional** for all useEffect, useLayoutEffect and useImperativeHandle in the react docs.
https://reactjs.org/docs/hooks-reference.html#useeffect

eg. my console gets flooded with 100s of warnings like:
```
You should provide an array of arguments as the second argument to the "useLayoutEffect" hook.
Not doing so will invoke this effect on every render.
```

Currently using useImperativeHandle can also trigger a misleading useLayoutEffect warning, as it accepts an empty deps array as well and uses useLayoutEffect in its implementation 
https://github.com/preactjs/preact/blob/master/hooks/src/index.js#L178

**Examples:**
- using refs after render to measure, manipulate focus, animate
- read and write refs in effects
- sync dom nodes with external libs
- eg. see popular libraries like react-spring (useSpring)

These cases are better covered by the react hooks eslint plugin:
https://www.npmjs.com/package/eslint-plugin-react-hooks#installation

JoviDeCroock:
@xiel thanks for the detailed explanation and my apologies for not reporting back on Slack, I completely forgot to check back.

PR #2359, opened 2020-02-18
Fix hook outside of component test

JoviDeCroock:


andrewiggins:
Should we also keep the original test since it still fails I think?

JoviDeCroock:
@andrewiggins I think that'll be a hard one to fix

andrewiggins:
That's fair. On the day we fix it we can add back the test

PR #2251, opened 2020-01-15
extends the CAMEL_PROPS regex to accept svg attribut clipPathUnits

friebe:
issue #2036

PR #2366, opened 2020-02-25
Fixed JSX SVG attribute names not being converted to the correct attribute names in the DOM.

steveharrison:
This fixes [an issue that I found where `clipPath` does not convert correctly to `clip-path`](https://github.com/preactjs/preact-compat/issues/538) (I accidentally raised the issue in the old `preact-compat` repository). Both `clipPath` and `clipRule` were broken in [this fix for `clipPathUnits`](https://github.com/preactjs/preact/pull/2251/files).

However, after looking through the [MDN SVG attributes list](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute), @jamesb3ll and I found other attributes that were not getting converted correctly, so I've also included them in this fix.

Thanks to @jamesb3ll for coming up with a nice regular expression and identifying other SVG attributes that weren't being converted correctly!

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -7,7 +7,7 @@ import {
 } from 'preact';
 import { applyEventNormalization } from './events';
 
-const CAMEL_PROPS = /^(?:accent|alignment|arabic|baseline|cap|color|fill|flood|font|glyph|horiz|marker|overline|paint|stop|strikethrough|stroke|text|underline|unicode|units|v|vector|vert|word|writing|x)[A-Z]/;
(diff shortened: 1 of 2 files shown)


jamesb3ll:
We can probably save a few bytes in the regex by just using the first uppercase character:
```js
const CAMEL_PROPS = /^(?:accent|alignment|arabic|baseline|cap|clip(?!PathU)|color|fill|flood|font|glyph(?!R)|horiz|marker(?!H|W|U)|overline|paint|stop|strikethrough|stroke|text(?!L)|underline|unicode|units|v|vector|vert|word|writing|x(?!C))[A-Z]/;
```

marvinhagemeister:
That's super awesome! I'd love to see the nice size improvements suggested by @jamesb3ll to be included in this PR before merging this 👍 Again really great work 💯

marvinhagemeister:
FYI: We'll probably cut a new release before the end of the week 👍

PR #2380, opened 2020-02-29
Fix event.isPropagationStopped()

38elements:
`event.isPropagationStopped` should return boolean, not function.
https://reactjs.org/docs/events.html
https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -52,6 +52,13 @@ options.event = e => {
 	if (oldEventHook) e = oldEventHook(e);
 	e.persist = () => {};
 	e.isDefaultPrevented = () => e.defaultPrevented;
+	const normalStopPropagation = e.stopPropagation;
+	let stoppedPropagating = false;
+	e.stopPropagation = function() {
+		stoppedPropagating = true;
+		normalStopPropagation.call(this);
+	};
+	e.isPropagationStopped = () => stoppedPropagating;
 	return (e.nativeEvent = e);
 };
 
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -50,18 +50,9 @@ export function hydrate(vnode, parent, callback) {
 let oldEventHook = options.event;
 options.event = e => {
 	if (oldEventHook) e = oldEventHook(e);
-	if (e.nativeEvent === undefined) {
-		e.persist = () => {};
-		e.isDefaultPrevented = () => e.defaultPrevented;
-		const stopPropagation = e.stopPropagation;
-		event.stopPropagation = function() {
-			stopPropagation.call(this);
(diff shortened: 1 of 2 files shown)


compat/src/render.js
```diff
@@ -50,10 +50,18 @@ export function hydrate(vnode, parent, callback) {
...
-	e.persist = () => {};
-	e.isDefaultPrevented = () => e.defaultPrevented;
-	e.isPropagationStopped = () => e.stopPropagation;
-	return (e.nativeEvent = e);
+	if (e.nativeEvent === undefined) {
+		e.persist = () => {};
+		e.isDefaultPrevented = () => e.defaultPrevented;
+		const stopPropagation = e.stopPropagation;
+		event.stopPropagation = function() {
+			stopPropagation.call(this);
```
JoviDeCroock: I merged the last one too fast, I think just keeping the old code and adding some local variable would be better since this is going to add a lot of size

38elements: IMHO,
I think it is better that `event.isDefaultPrevented()` and `event.isPropagationStopped()` are not supported, since it seems they has a little needs .

38elements:
@JoviDeCroock 
I reverted this code.

compat/test/browser/events.test.js
```diff
@@ -34,17 +34,12 @@ describe('preact/compat events', () => {
 
 		expect(spy).to.be.calledOnce;
 		const event = spy.args[0][0];
-		expect(event).to.haveOwnProperty('persist');
```
JoviDeCroock: Please put these tests back

compat/src/render.js
```diff
@@ -52,7 +52,6 @@ options.event = e => {
 	if (oldEventHook) e = oldEventHook(e);
 	e.persist = () => {};
 	e.isDefaultPrevented = () => e.defaultPrevented;
-	e.isPropagationStopped = () => e.stopPropagation;
```
JoviDeCroock: So my suggestion would be:

```
let stoppedPropagating = false;
e.stopPropagation = () => {
  stoppedPropagating = true;
  normalStopPropagation();
}
e.isPropagationStopped = () => stoppedPropagating;
```

GH doesn't allow me to suggest on empty lines

38elements:
@JoviDeCroock 
Thank you for review.
I changed this.

JoviDeCroock:
Awesome, that bit should just be tested and we're good to go! Great work @38elements

38elements:
@JoviDeCroock 
Thank you.
I added test.
So, Test coverage is increased.

PR #2386, opened 2020-03-02
implement strict-equality for vnodes

JoviDeCroock:
This short-circuits diffing of element nodes when they are strictly equal, I've also added two tests that should be used when extending this to functional vnodes. At this point I'm still looking for a proper way of implementing this.

```
const MyProvider = ({ children }) => {
  const [state, setState] = useState({});
  return (
    <MyContext.Provider value={{ state, setState }}>{children}</MyContext.Provider>
  )
} 
```

Basically the children  should never change since that createElement will never get called again if the update happens through setState so I think asserting on props.children  is correct but this could pose an issue with _catchError since when iterating over nested children and propagating up we can encounter this problem especially when we have to skip a boundary :confused:


src/diff/index.js
```diff
@@ -120,10 +120,14 @@ export function diff(
...
-					c.shouldComponentUpdate(newProps, c._nextState, componentContext) ===
-						false
+					(!c._force &&
+						c.shouldComponentUpdate != null &&
+							c.shouldComponentUpdate(
+								newProps,
+								c._nextState,
+								componentContext
+							) === false) ||
+					(newVNode._original === oldVNode._original && !c._processingException)
```
developit: Q: could we hoist this to L39 to apply it to both component and element diffing at once?

src/diff/index.js
```diff
@@ -218,6 +231,7 @@ export function diff(
 
 		if ((tmp = options.diffed)) tmp(newVNode);
 	} catch (e) {
+		newVNode._original = null;
```
developit: Won't this cause suspended vnodes to all be considered strictly equal? (vs `newVNode._original = newVNode`)

src/diff/index.js
JoviDeCroock: I think that will add a bunch off code sinceit will have to look like:

```
	if (newVNode._original === oldVNode._original) {
		const c = newVNode._component;
		if (c && !c._pendingError) {
			c.props = newVNode.props;
			c.state = c._nextState;
			c._dirty = false;
			c._vnode = newVNode;
			if (c._renderCallbacks.length) {
				commitQueue.push(c);
			}

			for (tmp = 0; tmp < newVNode._children.length; tmp++) {
				if (newVNode._children[tmp]) {
					newVNode._children[tmp]._parent = newVNode;
				}
			}

			return newVNode._dom;
		} else if (!c && excessDomChildren == null) {
			newVNode._lastDomChild = oldVNode._lastDomChild;
			return newVNode._dom;
		}
	}
```

src/diff/index.js
JoviDeCroock: No, if we'd do that we'd keep the reference which in case of error could cause erroneous scenario's

kristoferbaxter: Is there a test for this specific item?

JoviDeCroock: Everything in `lifecycles/componentDidCatch` and `lifecycles/getDerivedStateFromError` would be testament to this.

What happens is

```
const App = () => (
<Boundary>
  <StaticThrowingChild />
<Boundary />
);
```

When the `Boundary` catches the error the child won't have changed since the only createElement that child is subject to is the one from `App`.

kristoferbaxter: Absolutely, this makes sense.

However, for authors 6 months from now (including us!) a test that specifically calls out this equality scenario might be prudent.

Wouldn't call it a blocker, more of a nit.

JoviDeCroock: Yes, absolutely will do for sure it's a good shout. Will also document that, this is more of a first view since tests are failing for suspense either way.

compat/src/suspense.js
```diff
@@ -22,13 +22,14 @@ function detachedClone(vnode) {
 	if (vnode) {
 		vnode = assign({}, vnode);
 		vnode._component = null;
+		vnode._original = vnode;
```
JoviDeCroock: TODO: check whether this is needed or not

hooks/test/browser/useMemo.test.js
```diff
@@ -38,4 +38,51 @@ describe('useMemo', () => {
...
+			spy2();
+			return <p>{count}</p>;
+		};
+
+		const App = ({ x }) => {
+			const y = useMemo(() => <Y count={x} />, [x]);
+			return (
+				<div>
+					<X count={x} />
+					{y}
```
JoviDeCroock: This will only change reference once the `props.x` changes and is a strictly equal vnode for that matter

src/clone-element.js
```diff
@@ -16,6 +16,7 @@ export function cloneElement(vnode, props) {
 		vnode.type,
 		props,
 		props.key || vnode.key,
-		props.ref || vnode.ref
+		props.ref || vnode.ref,
+		null
```
JoviDeCroock: We have to reset `_original` when we're cloning

src/component.js
```diff
@@ -120,10 +120,13 @@ function renderComponent(component) {
 
 	if (parentDom) {
 		let commitQueue = [];
+		const nextVNode = assign({}, vnode);
+		nextVNode._original = nextVNode;
```
JoviDeCroock: reset this reference, if we don't `setState` would never execute because `_original` would still be `oldVNode`

src/diff/children.js
```diff
@@ -255,11 +255,19 @@ export function toChildArray(children, callback, flattened) {
 		}
 	} else if (!callback) {
 		flattened.push(children);
-	} else if (typeof children == 'string' || typeof children == 'number') {
-		flattened.push(callback(createVNode(null, children, null, null)));
+	} else if (typeof children === 'string' || typeof children === 'number') {
```
JoviDeCroock: TODO: correct this, must've been a merge conflict

src/diff/children.js
```diff
@@ -255,11 +255,19 @@ export function toChildArray(children, callback, flattened) {
...
+		flattened.push(callback(createVNode(null, children, null, null, children)));
```
JoviDeCroock: we pass the child to make the `_original` accurate

test/browser/createContext.test.js
```diff
@@ -57,6 +57,43 @@ describe('createContext', () => {
...
 
+	// This optimization helps
+	// to prevent a Provider from rerendering the children, this means
+	// we only propagate to children.
+	// Strict equal vnode optimization
+	it('should support wrappers', () => {
+		const { Provider, Consumer } = createContext();
+		let set,
+			renders = 0;
+		const Layout = ({ children }) => {
```
JoviDeCroock: The `Layout` is only a wrapper and for `Provider` this will never change since it's an external child so this will stop a lot of needless renders

src/diff/index.js
```diff
@@ -120,25 +120,47 @@ export function diff(
...
+					 * ref could have remained static:
+					 *
+					 * <div>
+					 *   {children}
+					 * </div>
+					 *
+					 * What if inside this children ref there was another setState then this should not be marked as not dirty yet until
+					 * we have traversed this node.
+					 */
+					if (newVNode._original !== oldVNode._original) c._dirty = false;
```
JoviDeCroock: Imagine in the context of https://github.com/preactjs/preact/pull/2386/files#diff-54682ce380935a717e41b8bfc54737f6R68 that the `Provider` updates internal state but `Layout` does so as well, this would imply that we `diff` from `Provider` end up at `Layout` see the nodes are equal and mark it as clean while it has pending state that could lead to erroneous DOM

src/diff/children.js
cristianbote:
```suggestion
	} else if (typeof children == 'string' || typeof children == 'number') {
```
Like this?

src/component.js
developit: Any idea if it's cheaper/smaller to instead route this through `createVNode`? The assign here is going to create a VNode object with a different constructor function than all other vnodes.

src/diff/index.js
developit: I think this bit will likely be in conflict with the hydration PR, I'll try to get it updated so we can compare. My hope is that the `_original=null` piece ends up being unnecessary as a result.

src/component.js
JoviDeCroock: Oh I'll try and check that out

test/browser/lifecycles/componentDidCatch.test.js
```diff
@@ -69,7 +69,6 @@ describe('Lifecycle methods', () => {
 			};
 			sinon.spy(ThrowErr.prototype, 'componentDidCatch');
 
-			receiver = undefined;
```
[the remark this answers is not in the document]
JoviDeCroock: it doesn't  need clearing but we could do that yes

compat/test/browser/suspense.test.js
```diff
@@ -739,8 +739,8 @@ describe('suspense', () => {
 				expect(scratch.innerHTML).to.eql(
 					`<div>Hello first 2</div><div>Hello second 2</div>`
 				);
-				expect(Suspender1.prototype.render).to.have.been.calledThrice;
-				expect(Suspender2.prototype.render).to.have.been.calledThrice;
+				expect(Suspender1.prototype.render).to.have.been.calledTwice;
```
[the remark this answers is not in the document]
JoviDeCroock: With this optimization the `render` is only called twice since the VNode can be skipped since it never changes. `Suspense` does not insert any extra DOM around children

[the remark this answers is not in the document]
andrewiggins: Yup. I was thinking that the Suspender component was the one that suspended so it would need to get rerendered in order for the `<div>Hello first 2</div>` VNode to get added to the tree. But in actuality Suspender just renders a `Lazy` component, and that `Lazy` component is the component that actually suspends (i.e. throws the promise). So because of #2125 all of the components that suspended (the Lazy components, not the Suspender components) are queued for updates using `forceUpdate`.

That was the piece I was missing. My mental model had forgotten about the Lazy component under Suspender and was thinking Suspender needed to rerender. So perhaps these test should check is that the Lazy component's render is called after all the promises are resolved, not the Suspender component but that's not in scope of this PR. I'm just talking out loud to share my understanding lol 😅

src/create-element.js
```diff
@@ -74,9 +75,11 @@ export function createVNode(type, props, key, ref) {
 		// a _nextDom that has been set to `null`
 		_nextDom: undefined,
 		_component: null,
-		constructor: undefined
+		constructor: undefined,
+		_original: original
```
andrewiggins, line 79: Oh! I forgot to mention this in my previous review - can we add `_original` to the `VNode` interface in `internal.d.ts`?

src/diff/index.js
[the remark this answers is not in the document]
JoviDeCroock: Will do this tonight

src/component.js
JoviDeCroock: The problem with that is that we're missing a lot of `vnode` properties if we call `createVNode`.

yisar:
@JoviDeCroock 
I have two questions:

1. Why do you prefer strict equality to shallow comparison?

Why not make a shallow comparison within the framework core code, such as built-in `Preact.memo`? What's wrong with it?

2. Why not use `usecallback`? The `setstate` here can be cached with usecallback and useMemo

Thank you for your answer.

JoviDeCroock:
Both of those imposs this optimization lower, this means that in the shallow comparison we would have to do this check in Preact and shallow comparison is often ineffective.

Using hooks as a part of diff would impose function rebinding on every render even more and would need to enter that vnode.

Now we never enter this vnode we check whether or not we need to enter it, which is a good perf increase.

Issue #2363, opened 2020-02-21
cloneElement type should include VNode?


Linksku:
Should the first argument and the return value of cloneElement be VNode and not Element? It's Element in the Typescript definition.

ddayguerrero:
👋 I'm interested in looking into this. From what I can investigate, the issue here is a typing one.

```typescript
function cloneElement(
  vnode: JSX.Element,
  props?: any,
  ...children: ComponentChildren[]
): JSX.Element;
```

where `vnode` should return a `VNode`?

As of now, everything seems to work https://codesandbox.io/s/cloneelement-issue-o0x31. If anyone can confirm that would be nice!

JoviDeCroock:
Hey!

Yes, we should be returning VNode and here's why:

- In `cloneElement` we call `createVNode` https://github.com/preactjs/preact/blob/master/src/clone-element.js#L14 
- In `createVNode` we make a `VNode` structure https://github.com/preactjs/preact/blob/master/src/create-element.js#L59
- We use `createVNode` for `createElement` (our jsx constructor) and this is the return type for that last method: https://github.com/preactjs/preact/blob/master/src/index.d.ts#L176

PR #2388, opened 2020-03-03
Update cloneElement typings

ddayguerrero:
Resolves https://github.com/preactjs/preact/issues/2363

Updating `cloneElement` type definitions for the `vnode` param and its returned value.

## What approach did I use?
Mostly following Preact's [contribution guidelines](https://github.com/preactjs/preact/blob/master/CONTRIBUTING.md):

- Updated the function's JSDoc block
- Added a test
- Ran `test:ts` and `karma:test:watch` to make sure tests still pass

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -17,7 +17,7 @@ declare namespace preact {
 		/**
 		 * ref is not guaranteed by React.ReactElement, for compatiblity reasons
 		 * with popular react libs we define it as optional too
-		*/
+		 */
 		ref?: Ref<any> | null;
 		/**
 		 * The time this `vnode` started rendering. Will only be set when
@@ -176,15 +176,15 @@ declare namespace preact {
 	function createElement(
 		type: string,
 		props:
-			| JSXInternal.HTMLAttributes &
+			| (JSXInternal.HTMLAttributes &
 					JSXInternal.SVGAttributes &
-					Record<string, any>
+					Record<string, any>)
 			| null,
 		...children: ComponentChildren[]
 	): VNode<any>;
 	function createElement<P>(
 		type: ComponentType<P>,
-		props: Attributes & P | null,
+		props: (Attributes & P) | null,
 		...children: ComponentChildren[]
 	): VNode<any>;
 	namespace createElement {
@@ -194,15 +194,15 @@ declare namespace preact {
 	function h(
 		type: string,
 		props:
-			| JSXInternal.HTMLAttributes &
+			| (JSXInternal.HTMLAttributes &
 					JSXInternal.SVGAttributes &
-					Record<string, any>
+					Record<string, any>)
 			| null,
 		...children: ComponentChildren[]
 	): VNode<any>;
 	function h<P>(
 		type: ComponentType<P>,
-		props: Attributes & P | null,
+		props: (Attributes & P) | null,
 		...children: ComponentChildren[]
 	): VNode<any>;
 	namespace h {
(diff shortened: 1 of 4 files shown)


src/index.d.ts
```diff
@@ -223,10 +223,10 @@ declare namespace preact {
 		parent: Element | Document | ShadowRoot | DocumentFragment
 	): void;
 	function cloneElement(
```
JoviDeCroock: We could allow a generic here as we use [in createElement](https://github.com/preactjs/preact/pull/2388/files#diff-b739f95ace6bbf2f0fd5e8cf1dab97d8R185)

src/index.d.ts
```diff
@@ -17,7 +17,7 @@ declare namespace preact {
 		/**
 		 * ref is not guaranteed by React.ReactElement, for compatiblity reasons
 		 * with popular react libs we define it as optional too
-		*/
+		 */
```
JoviDeCroock: You might have some global prettier config

test/browser/cloneElement.test.js
```diff
@@ -8,7 +8,6 @@ describe('cloneElement', () => {
 		const instance = <Comp prop1={1}>hello</Comp>;
 		const clone = cloneElement(instance);
 
-		expect(clone.prototype).to.equal(instance.prototype);
```
JoviDeCroock: Why is this deleted, this should be working

test/ts/VNode-test.tsx
```diff
@@ -116,6 +117,19 @@ describe('VNode TS types', () => {
...
 		expect(types).to.deep.equal(['a', '2', 'b', 'c', 'TestComp']);
 	});
+
+	it('component should work with cloneElement', () => {
+		const comp: VNode = (
+			<SimpleComponent>
+				<div>child 1</div>
+			</SimpleComponent>
+		);
+		const clone = cloneElement(comp);
```
JoviDeCroock:
```suggestion
		const clone: VNode = cloneElement(comp);
```

I think this is good to type-check since this would throw if cloneElement would return an `Element`

src/index.d.ts
```diff
@@ -222,11 +222,11 @@ declare namespace preact {
 		vnode: ComponentChild,
 		parent: Element | Document | ShadowRoot | DocumentFragment
 	): void;
-	function cloneElement(
-		vnode: JSX.Element,
+	function cloneElement<P>(
```
JoviDeCroock: Because of the nature of this file (no ts-compiler) you'll have to specify the non-generic version as well (comparable to createElement https://github.com/preactjs/preact/pull/2388/files#diff-b739f95ace6bbf2f0fd5e8cf1dab97d8R176)

ddayguerrero: Thanks for the quick feedback! 😄

src/index.d.ts
ddayguerrero: I'm not sure what is causing this, I would have thought that no other prettier config ran from my workspace after the pre-commit's `lint-staged`.

That said, it seems to be consistent with the other comment blocks.

JoviDeCroock: All right, no worries ;D looks better either way

src/index.d.ts
ddayguerrero: Shall we consider adding a test for the generic version?

JoviDeCroock: Works for me yes! Would be awesome if you did

ddayguerrero:
Thanks! I addressed the remainder of the comments and rebased the commit history.

PR #2389, opened 2020-03-04
Add support for babel jsx/jsxs/jsxDEV

marvinhagemeister:
There is currently undergoing work to create a successor to `createElement`.

Advantages:

- `ref` is not deleted from `props`. This means no need for `forwardRef` anymore as one can just do ```const Foo = props => <div ref={props.ref} />```
- `key` is a separate argument and not part of `props` anymore. This prevents duplicated keys when spreading `props` on different nodes in the same depth.
- Better perf, because we don't have to iterate over `props` anymore.

More information:

- [RFC](https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md) where the changes were proposed
- Work in Progress babel plugin: https://github.com/babel/babel/pull/11154

### DIFF
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -54,9 +54,11 @@ export function createElement(type, props, children) {
  * diffing it against its children
  * @param {import('./internal').VNode["ref"]} ref The ref property that will
  * receive a reference to its created child
+ * @param {import('./internal').VNode["ref"]} source The ref property that will
+ * receive a reference to its created child
  * @returns {import('./internal').VNode}
  */
-export function createVNode(type, props, key, ref) {
+export function createVNode(type, props, key, ref, source, self) {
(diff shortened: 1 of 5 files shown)


src/create-element.js
```diff
@@ -82,6 +82,51 @@ export function createVNode(type, props, key, ref) {
...
+			if (props[i] === undefined) {
+				props[i] = type.defaultProps[i];
+			}
+		}
+	}
+
+	const vnode = createVNode(type, props, key, props && props.ref);
+
+	// TODO: Should this be inlined into `createVNode`?
+	if (source || self) {
```
developit: Could just check for `source` here, they're always defined as a pair.

marvinhagemeister:
Replaced by #2764

PR #2390, opened 2020-03-04
fix stopPropagation override for IE11

JoviDeCroock:
IE11 throws an error when overriding the stopPropagation method on an event

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -52,12 +52,12 @@ options.event = e => {
 	if (oldEventHook) e = oldEventHook(e);
 	e.persist = () => {};
 	e.isDefaultPrevented = () => e.defaultPrevented;
-	const normalStopPropagation = e.stopPropagation;
 	let stoppedPropagating = false;
-	e.stopPropagation = function() {
-		stoppedPropagating = true;
-		normalStopPropagation.call(this);
-	};
+	Object.defineProperty(e, 'stopPropagation', {
+		value() {
+			stoppedPropagating = true;
+		}
+	});
 	e.isPropagationStopped = () => stoppedPropagating;
 	return (e.nativeEvent = e);
 };
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -51,14 +51,23 @@ let oldEventHook = options.event;
 options.event = e => {
 	if (oldEventHook) e = oldEventHook(e);
 	e.persist = () => {};
-	e.isDefaultPrevented = () => e.defaultPrevented;
-	let stoppedPropagating = false;
-	Object.defineProperty(e, 'stopPropagation', {
-		value() {
-			stoppedPropagating = true;
-		}
-	});
+	let stoppedPropagating = false,
+		defaultPrevented = false;
+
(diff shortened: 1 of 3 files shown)


JoviDeCroock:
image

Looks like it works in IE11

compat/src/render.js
```diff
@@ -52,12 +52,12 @@ options.event = e => {
...
 	e.isDefaultPrevented = () => e.defaultPrevented;
-	const normalStopPropagation = e.stopPropagation;
 	let stoppedPropagating = false;
-	e.stopPropagation = function() {
-		stoppedPropagating = true;
-		normalStopPropagation.call(this);
-	};
+	Object.defineProperty(e, 'stopPropagation', {
+		value() {
+			stoppedPropagating = true;
```
marvinhagemeister: Looking at [the page in mdn](https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation) I'm wondering if we still need to call the original method somehow.

JoviDeCroock: Why, we just add an invocation listener with this?

38elements:
I may be misunderstanding .
I used following code on IE11.
This code worked normally.

```html
<!DOCTYPE html>
<html>
<head>
<meta charset="UTF-8" />
<title>Test</title>
</head>
<body>
<div id="bar">
  <div id="foo">foo<br>foo<br>foo<br></div>
</div>
<script type="text/javascript">
  var foo = document.getElementById('foo')
  foo.addEventListener('click', function(e) {
    var stopPropagation = e.stopPropagation
    e.stopPropagation = function() {stopPropagation.call(this)};
    e.stopPropagation();
    console.log('Foo');
  }, false);
  var bar = document.getElementById('bar')
    bar.addEventListener('click', function(e) {
    console.log('Bar');
  }, false);
</script>
</body>
</html>
```

JoviDeCroock:
@38elements look at the IE11 tests on master, or something weird happens to the closure on IE11 or the override doesn't take effect.

These tests only run on master, https://travis-ci.org/preactjs/preact/builds/656971648?utm_source=github_status&utm_medium=notification

We can enable them by enabling `sauceLabs` on our karma config

38elements:
@JoviDeCroock 
Thank you for reply.
I think that `event.defaultPrevented` may cause a problem.

```    
Error
   at AssertionError (node_modules/chai/chai.js:9449:7)
   at Assertion.prototype.assert (node_modules/chai/chai.js:239:7)
   at Anonymous function (node_modules/chai/chai.js:1021:5)
   at propertyGetter (node_modules/chai/chai.js:7899:9)
   at Anonymous function (webpack:///compat/test/browser/events.test.js:51:2 <- compat/test/browser/events.test.js:7244:5)
```

https://github.com/preactjs/preact/blob/0aaa85516ae64799ae0c842ea5cd5a33cc9c0830/compat/test/browser/events.test.js#L51

JoviDeCroock:
Seems like IE11 on Windows 7 was indeed bugging out, thanks @38elements I must've missed the line-number. You're a hero!

Issue #2391, opened 2020-03-04
[BUG] Uncontrolled input acts as controlled (issue on defaultValue)


aralroca:
In order to reproduce the issue:

* Use a `defaultValue` prop on a `<input>` element
* Causes a render on the `onChange` event (also it happens in `onInput` event).

Expected:

* Updating the input element writing whatever inside. The `defaultValue` only takes the first value, then is updated.

What is happening:

* After writing whatever on the input element; nothing happened. The `defaultValue` overwrites the updated value.

_Comparison React / Preact_

![issue](https://user-images.githubusercontent.com/13313058/75923417-52146e80-5e65-11ea-9349-a7b97859774f.gif)

Code: https://stackblitz.com/edit/preact-bug-defaultvalue

I reproduced in Preact 10.3.3.

PR #2392, opened 2020-03-04
Fix defaultValue with re-render

aralroca:
Fixes https://github.com/preactjs/preact/issues/2391

It's my first contribution to Preact. This should fix the issue that I reported. However please, review carefully because I'm not very comfortable with the preact code yet!

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -106,10 +106,9 @@ options.vnode = vnode => {
 	if (typeof type != 'function') {
 		// Apply defaultValue to value
 		if (props.defaultValue) {
-			if (!props.value && props.value !== 0) {
+			if (props.value == null && props.value !== 0) {
 				props.value = props.defaultValue;
 			}
-			delete props.defaultValue;
 		}
 
 		// Add support for array select values: <select value={[]} />
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -106,7 +106,7 @@ options.vnode = vnode => {
 	if (typeof type != 'function') {
 		// Apply defaultValue to value
 		if (props.defaultValue) {
-			if (props.value == null && props.value !== 0) {
+			if (!props.value && props.value !== 0) {
 				props.value = props.defaultValue;
 			}
 		}
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -105,10 +105,11 @@ options.vnode = vnode => {
 	// Apply DOM VNode compat
 	if (typeof type != 'function') {
 		// Apply defaultValue to value
-		if (props.defaultValue) {
+		if (props.defaultValue && props.value !== undefined) {
 			if (!props.value && props.value !== 0) {
 				props.value = props.defaultValue;
 			}
+			delete props.defaultValue;
 		}
 
 		// Add support for array select values: <select value={[]} />
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -368,6 +368,7 @@ function diffElementNodes(
 			if (
 				'value' in newProps &&
 				newProps.value !== undefined &&
+				newProps.value !== oldProps.value &&
 				newProps.value !== dom.value
 			) {
 				dom.value = newProps.value == null ? '' : newProps.value;
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -368,10 +368,11 @@ function diffElementNodes(
 			if (
 				'value' in newProps &&
 				newProps.value !== undefined &&
-				newProps.value !== oldProps.value &&
-				newProps.value !== dom.value
+				newProps.value !== dom.value &&
+				newProps.defaultValue !== newProps.value
 			) {
 				dom.value = newProps.value == null ? '' : newProps.value;
+				delete newProps.defaultValue; // is this needed?
 			}
 			if (
 				'checked' in newProps &&
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -368,8 +368,7 @@ function diffElementNodes(
 			if (
 				'value' in newProps &&
 				newProps.value !== undefined &&
-				newProps.value !== dom.value &&
-				newProps.defaultValue !== newProps.value
+				newProps.value !== dom.value
 			) {
 				dom.value = newProps.value == null ? '' : newProps.value;
 			}
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -372,7 +372,6 @@ function diffElementNodes(
 				newProps.defaultValue !== newProps.value
 			) {
 				dom.value = newProps.value == null ? '' : newProps.value;
-				delete newProps.defaultValue; // is this needed?
 			}
 			if (
 				'checked' in newProps &&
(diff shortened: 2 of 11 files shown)


src/diff/index.js
```diff
@@ -368,9 +368,11 @@ function diffElementNodes(
 			if (
 				'value' in newProps &&
 				newProps.value !== undefined &&
-				newProps.value !== dom.value
+				newProps.value !== dom.value &&
+				newProps.defaultValue !== newProps.value
```
JoviDeCroock: I think these aren't needed

aralroca: Thanks for the feedback @JoviDeCroock ! Without that extra check, the issue stills happening. The `newProps.value` is taking the value of the `defaultValue` when the originary props.value are `undefined`. Perhaps the issue is here, however I thought this is in this way on purpose so is why I decided to check `newProps.defaultValue !== newProps.value`.


I tried to do a failing test, but I failed in that... I can reproduce it manually in the demo folder, or here: https://stackblitz.com/edit/preact-bug-defaultvalue, typing something in the input text.

This is my attempt to do the failing test:

```js
it('should keep value of uncontrolled inputs using defaultValue', () => {
	// See https://github.com/preactjs/preact/issues/2391
		
	class Input extends Component {
		render() {
			return <input type="text" defaultValue="bar" onChange={() => this.forceUpdate()} />
		}
	}

	render(<Input />, scratch);
	expect(scratch.firstChild.value).to.equal('bar');
	scratch.firstChild.focus();
	scratch.firstChild.value = 'foo';
	rerender();
	expect(scratch.firstChild.value).to.equal('foo');
});
```

However this is always passing, because after  `scratch.firstChild.value = 'foo';` line is not firing the onChange...

src/diff/index.js
```diff
@@ -368,9 +368,11 @@ function diffElementNodes(
...
 			) {
 				dom.value = newProps.value == null ? '' : newProps.value;
+				delete newProps.defaultValue; // is this needed?
```
JoviDeCroock: This shouldn't be needed since there are new props instances every time either way

compat/src/render.js
```diff
@@ -106,10 +106,9 @@ options.vnode = vnode => {
 	if (typeof type != 'function') {
 		// Apply defaultValue to value
 		if (props.defaultValue) {
-			if (!props.value && props.value !== 0) {
+			if (props.value == null && props.value !== 0) {
```
JoviDeCroock: Okay sorry about this I was really confused on how to support uncontrolled in this case, so what we want to do is apply `defaultValue` when `props.value` is falsy but not undefined, right?
In the case of us seeing it is controlled (value & onChange present) we apply defaultValue to `props.value` (and delete props.defaultValue)
In the case of uncontrolled we shouldn't enter this branch at all and go for full native dom behavior

aralroca: Sorry to don't explain better about what I did. The issue comes when:

* `defaultValue` -> defined, some string.
* `value` -> undefined
* `onChange` -> defined + is forcing a render (ex: forceUpdate or setState)

In this case, the `props.value` is taking always the value of `defaultValue`, instead of only setting on mounting the component. This causes that although the `value` is undefined, it is acting as controlled... Without the possibility to type anything inside the input text.

This check here doesn't help a lot. Maybe I can change it again to `!props.value`. The approach that I take is on the other file, `src/diff/index.js`. Removing below the `delete props.defaultValue;` then it is possible to check in `src/diff/index.js` that if the user types something different to the `defaultValue`, the `dom.value` can be updated. 

However, I guess that is not the root of the problem and is only a workaround. If you can help me with the familiarity of the correct behavior maybe we can found the root of the problem.

I'm experimenting this without this workaround:

image

After type a "d", the `props.value` is taking the value of `defaultValue`. Then, in the `src/diff/index.js`, the correct `dom.value` is replaced with the `props.value`, when the `props.value` has "bar" (the `defaultValue`).

image

So, the final result is that although I typed a "d", the `dom.value` is resetting again to the defaultValue:

image

**After this fix:**

The first part is doing just the same that before (maybe here is the root of the problem that should be fixed):

image

However, now is not removing `props.defaultValue` with the idea to check then in `src/diff/index.js` if it has defaultValue.

And in `src/diff/index.js` is not updating the `dom.value` because `props.value` is equal to `props.defaultValue` (props.value updated before). And the `dom.value` already has the correct value because is uncontrolled.

And the result is that the input value now is updated correctly:

image

JoviDeCroock: Yes that’s what I’m saying, when there is ni value property on props we can assume uncontrolled this means that we can skip this whole condition and rely on the native dom behavior, am I mistaken?

aralroca: Well, I guess that it should enter the condition the first time, when the component is mount. Then I guess that is not necessary.

JoviDeCroock: No, never. Uncontrolled should only reflect to the dom-node not to a js-value that’s where we went wrong

aralroca: But entering the first time, then `props.value` is defined, and is entering again...

aralroca: Ahh ok, I am mistaken then! Thank you so much for your time.

JoviDeCroock: Hey @aralroca,

so when we talk about controlled and uncontrolled we are mainly talking about JS vs just DOM, we can see that we are talking about DOM when we aren't passing a `value` to an `<input />`. In this case we can rely on the semantics of how `defaultValue` works since under the hood `defaultValue` is reflected correctly on the DOM.
The measure we are referring to here makes it so that this `defaultValue` is also reflected on the JS-prop, for instance this is an `uncontrolled input` in native Preact, where this measure isn't available, as you can see it just works. https://codesandbox.io/s/keen-breeze-39fdh

From this we can conclude that with `compat` we do too much work, we apply `defaultValue` even when it's an uncontrolled input, making it controlled (why your value keeps changing), this should be fixed by checking on line 108 whether we have a defaultValue and if `value` exists as a property on `newProps` this can be done cheaply by saying `!== undefined` or somewhat more expensive by `hasProperty`

src/diff/index.js
aralroca: You are right. Is not need it :)

compat/src/render.js
aralroca: Thank you so much for your explanation! Your proposal works pretty well. I did a commit in order to fix it!

JoviDeCroock:
hey @aralroca 

Let's mark this as ready for review and get this into master, are you okay with this? :)

aralroca:
@JoviDeCroock  I'm okay with this. I would like to add the failing test. But I failed with that as I commented here: https://github.com/preactjs/preact/pull/2392#discussion_r388206876 

I'm going to change to ready to review. Then I'm going to try again to do the failing test. However, I can add it to a different PR.

JoviDeCroock:
All right, I'll think about it as well. Thanks!

src/diff/index.js
Sasha-Sorokin: > … because after scratch.firstChild.value = 'foo'; line is not firing the onChange...

@aralroca Hey, try dispatching event manually with `scratch.firstChild.dispatchEvent(new InputEvent('input'))`. I think it should work

aralroca: Perfect! Thanks @Sasha-Sorokin

Issue #2393, opened 2020-03-05
createContext is not fully React compatible


pauldijou:
After migrating an app from React to Preact 10, we noticed that some of our components were unmounting with the incorrect props.

### Reproduction

Here is a reproduction of the issue on simplified codebase.

Preact: https://codesandbox.io/s/upbeat-pasteur-r7soh
React: https://codesandbox.io/s/wonderful-worker-u3uus

### Steps to reproduce

So, the idea is that because of the `key` prop assigned to `Parent`, when `id` changes, it should unmount it with the previous id and then mount it with the new one. That's my expectation at least and it works like that in React. Here is the console output when clicking on the text.

```
// React
render, id 0
mount, id 0
render, id 1
unmount, id 0
mount, id 1
```

But for Preact, I'm not fully sure, but it seems that nested consumers will update first, meaning that `Child` will be updated with the new id and only then will `Parent` realize that its `key` has changed and will unmount. Here are the console logs for the same click:

```
// Preact
render, id 0
mount, id 0
render, id 1
updated from 0 to 1
render, id 1
unmount, id 1
mount, id 1
```

As we can see, there is an `update` lifecycle now and the unmount is done with `props.id` at `1` rather than `0`.

I wonder if either React is maintaining a more optimized stack for consumers or, even better, batch all updates to all consumers in one render.

### Expected Behavior

`Child` component should not update and unmount directly with `id: 0`.

### Actual Behavior

`Child` component is updated to `id: 1` and unmounted with that value.

JoviDeCroock:
Hey,

this isn't really related to context, this is related to what order children are diffed. The performance implication of this is really minimal since this doesn't induce two cycles of diffing it just has an update before the parent realises the key changed, this is because of the 1-pass vs 2-pass rendering approach.

pauldijou:
Thanks for the response. I'm not too concerned about performance.

The issue is that we use `key` to ensure `Parent` components are properly mounted / unmounted when the `id` is changed (which happen properly on both framework) but `Children` will cleanup some resources when unmounting based on their `id` prop and now they are trying to cleanup the wrong ones because they have been updated before the unmount.

For this simple example, the solution would be for `Parent` to set the `id` of its `Child` rather than having another consumer. But in our app, `Store` is a Redux store and the tree is way more complex.

I don't have much knowledge about how those rendering pass work to be honest, but I guess having only 1 pass prevent batching like React can do in the `unstable_batchedUpdates` function?

[...]

PR #2395, opened 2020-03-05
Context plumbing top --> bottom

JoviDeCroock:
Fixes https://github.com/preactjs/preact/issues/2393

This sorts the updateQueue for the context subscribers from top to bottom to avoid inconsistent trees


JoviDeCroock:
Closing in favor of #2396

PR #2396, opened 2020-03-06
Sort new renderqueue items

### LINKED ISSUE #2393: createContext is not fully React compatible

jviide:
This pull request attempts to fix #2393 by modifying how the global `renderQueue` is processed. 

Currently `renderQueue` is sorted by vnode depth before its items are processed, but if new items get queued during processing (and `renderQueue.length > 1`) then those new items are also processed _without_ them getting sorted by depth first.

This pull attempts fixing that. When `renderQueue` is processed, it's item count is checked. If there are no items then we can stop. Otherwise the queue is sorted by depth and stored into a temporary variable while `renderQueue` itself is reinitialized with a new, _empty_ array. Now the "old" queued items get processed and any new items get added to the new `renderQueue` list.

After the all old items have been processed the new `renderQueue` is checked & sorted again, and the cycle starts anew.

A great test from @JoviDeCroock's alternative PR #2395 is also included here.

Fixes #2393

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -212,7 +212,7 @@ function process() {
 		// process() calls won't get scheduled.
 		rerenderCount = 1;
 		queue.some(c => {
-			if (c && c._dirty) renderComponent(c);
+			if (c._dirty) renderComponent(c);
 		});
 	}
 }
--- a/src/component.js
+++ b/src/component.js
@@ -208,9 +208,8 @@ function process() {
 	while ((rerenderCount = rerenderQueue.length)) {
 		queue = rerenderQueue.sort((a, b) => a._vnode._depth - b._vnode._depth);
 		rerenderQueue = [];
-		// Pretend momentarily that renderQueue has items in it so unnecessary
-		// process() calls won't get scheduled.
-		rerenderCount = 1;
+		// Don't update `renderCount` yet. Keep its value non-zero to prevent unnecessary
+		// process() calls from getting scheduled while `queue` is still being consumed.
 		queue.some(c => {
 			if (c._dirty) renderComponent(c);
 		});
--- a/src/component.js
+++ b/src/component.js
@@ -161,6 +161,7 @@ function updateParentDomPointers(vnode) {
  * @type {Array<import('./internal').Component>}
  */
 let rerenderQueue = [];
+let rerenderCount = 0;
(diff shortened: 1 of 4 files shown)


src/component.js
```diff
@@ -190,7 +191,10 @@ let prevDebounce;
  */
 export function enqueueRender(c) {
 	if (
-		(!c._dirty && (c._dirty = true) && rerenderQueue.push(c) === 1) ||
+		(!c._dirty &&
+			(c._dirty = true) &&
+			rerenderQueue.push(c) &&
+			!rerenderCount++) ||
```
marvinhagemeister: We might be lacking a test case for this change. All tests keep passing when I remove the `rerenderCount` check.

jviide: Good point. I added a test to browser/render.test.js that passes with this PR, but fails on master or if you remove the `rerenderCount` check.

Issue #2249, opened 2020-01-14
Material-UI 4: popover loose positioning


rubas:
We are using Gatsbyjs and switched out React with Preact X. Everything works fine, except all the [popover](https://material-ui.com/components/popover/) loose their positioning and open in the top let corner (x: 0, y:0).

```
preact@10.2.1
@material-ui/core@4.8.3
@material-ui/styles@^4.8.2
```

As I haven't found any information, that would suggest, that these wouldn't work with Preact. 

===

Edit: Added an [Example](https://github.com/preactjs/preact/issues/2249#issuecomment-576024118).

porfirioribeiro:
Could you create a small reproduction demo on CodeSandbox?
Would be easier to understand the problem and try to fix it!

tky5622:
I prepare this reproduction instead.
please click TOGGLE MENU GROW and compare normal work.

https://codesandbox.io/s/exciting-worker-i67ug?fontsize=14&hidenavigation=1&theme=dark

https://material-ui.com/components/menus/

porfirioribeiro:
It seems to be working fine, maybe i misunderstood something
 
![Screenshot_20200118_151528](https://user-images.githubusercontent.com/499246/72665964-657a9000-3a05-11ea-86e4-3a10e9c44c9f.png)

marvinhagemeister:
I get the same result as @porfirioribeiro .

rubas:
Thanks. Sorry for the delay, I try to provide an isolated sample later today.

rubas:
Example: https://codesandbox.io/s/relaxed-kepler-gn2sj?fontsize=14&hidenavigation=1&theme=dark

**Bug** Dropdown is opening at `0,0`.


```
import { render, h } from 'preact';
import { PureComponent } from 'preact/compat';
import { FormControl, Select, MenuItem } from "@material-ui/core"

class Foo extends PureComponent {
  render() {
    return (

        <FormControl style={{position: 'absolute',top: 120, left: 120}}>
          <Select value={0}>
            <MenuItem value="0">Test 0</MenuItem>
            <MenuItem value="1">Test 1</MenuItem>
            <MenuItem value="3">Test 3</MenuItem>
          </Select>
        </FormControl>
 
    );
  }
}

const dom = document.getElementById('app');
render(<Foo />, dom);
```

antonbashir:
Same issue with Menu component and anchorEl.
[...]

PR #2403, opened 2020-03-10
Fix mui Popover integration

JoviDeCroock:
## The problem

In `mui` there's a concept of `forking` a ref, this means that we combine the passed ref from the top with the one used internally inside a component to manage presentation.

An example: https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Grow/Grow.js#L32

This becomes problematic in the following scenario

```
const Wrapper = forwardRef((props, ref) => <div ref={ref}>I am wrapped</div>);
const App = () => {
  const ref = useRef()
  return (
    <Grow>
       <Wrapper ref={ref} />
    </Grow>
  )
}
```

At this point we have evaluated `Wrapper` and moved the `vnode.ref` to `vnode.props.ref` due to `options.vnode` being called (`createVNode`). This makes the `forking` of the `Grow` ref fail since we talk to `children.ref` which in this case is `Wrapper.ref`. That's why we move the `forwarding` of refs to the point where we are going to diff the `forwarded` component.

Fixes: https://github.com/preactjs/preact/issues/2249

### DIFF
--- a/compat/test/browser/forwardRef.test.js
+++ b/compat/test/browser/forwardRef.test.js
@@ -393,7 +393,7 @@ describe('forwardRef', () => {
 		const Transition = ({ children }) => {
 			const state = useState(0);
 			forceTransition = state[1];
-			expect(children.ref).to.exist;
+			expect(children.ref).to.not.be.undefined;
 			if (state[0] === 0) expect(children.props.ref).to.be.undefined;
 			return children;
 		};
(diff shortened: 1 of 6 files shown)


compat/test/browser/forwardRef.test.js
```diff
@@ -381,4 +381,32 @@ describe('forwardRef', () => {
...
 		expect(_ref.current).to.equal(scratch.firstChild);
 	});
+
+	it('should forward at diff time instead vnode-creation.', () => {
+		let ref, set;
+		const Wrapper = forwardRef((_props, ref) => <div ref={ref}>Wrapper</div>);
+		const Transition = ({ children }) => {
+			const state = useState(0);
+			set = state[1];
+			expect(children.ref).to.not.be.undefined;
```
developit: Wondering if it'd be worth also guarding against it being null here

```suggestion
			expect(children.ref).to.exist;
```

PR #2408, opened 2020-03-12
Add _patchedLifecycles to compat/src/internal.d.ts

38elements:
This adds ` _patchedLifecycles` to `compat/src/internal.d.ts`.


compat/src/internal.d.ts
```diff
@@ -12,6 +12,7 @@ export { PreactElement } from '../../src/internal';
 export interface Component<P = {}, S = {}> extends PreactComponent<P, S> {
 	isReactComponent?: object;
 	isPureReactComponent?: true;
+	_patchedLifecycles?: true;
```
38elements: https://github.com/preactjs/preact/blob/658cb96a89782e5e84e1b92812196c4cf5cc39e1/compat/src/render.js#L156-L162

PR #2409, opened 2020-03-12
Use double equal with typeof

### LINKED ISSUE #2129: Force consistent use of double or triple equal/not equal in typeofs

polemius:
Fix #2129 

Use double equal with `typeof`

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -13,9 +13,7 @@ const CAMEL_PROPS = /^(?:accent|alignment|arabic|baseline|cap|clip(?!PathU)|colo
 Component.prototype.isReactComponent = {};
 
 export const REACT_ELEMENT_TYPE =
-	(typeof Symbol !== 'undefined' &&
-		Symbol.for &&
-		Symbol.for('react.element')) ||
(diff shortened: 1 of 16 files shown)


marvinhagemeister:
@polemius Ugh seems like something went wrong when I rebased it against the latest changes in `master`. Can you correct that locally and resolve the conflict? Happy to merge this PR once the CI is green again :tada: And sorry again for the trouble!

polemius:
@marvinhagemeister Thank you for noticing me know about rebase conflict. I've rebased my branch against very new master and push it with `--force`. I hope CI will be green soon 🤞

polemius:
Thank you!

PR #2412, opened 2020-03-12
try out sauceLabs on preactjs member PR's

JoviDeCroock:
This doesn't work https://github.com/preactjs/preact/runs/504430767?check_suite_focus=true#step:6:158

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,5 +1,5 @@
 /*eslint no-var:0, object-shorthand:0 */
-var isFork = !!process.env.GITHUB_HEAD_REF;
+var isFork = process.env.GITHUB_REPOSITORY !== 'preactjs/preact';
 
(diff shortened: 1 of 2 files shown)


karma.conf.js
```diff
@@ -106,7 +110,7 @@ module.exports = function(config) {
 				`local${require('./package.json').version}`,
 			connectLocationForSERelay: 'localhost',
 			connectPortForSERelay: 4445,
-			startConnect: false
+			startConnect: true
```
developit: just to check: this doesn't need to be conditional because we disable saucelabs itself in cases where we don't want it to run, so the proxy never starts?

JoviDeCroock: I put it to `true` just for testing it's `!!sauceLabs` now

PR #2423, opened 2020-03-19
Keep calling render until state isn't dirty

JoviDeCroock:
This will empty the render queue until it's empty for every single vnode, this prevents us from going down the diff too much.

### DIFF
--- a/test/browser/components.test.js
+++ b/test/browser/components.test.js
@@ -2575,5 +2575,25 @@ describe('Components', () => {
 
 			expect(scratch.innerHTML).to.equal('<div>bar</div>');
 		});
+
+		it('should call render again when state got dirtied', () => {
+			class Persons extends Component {
+				constructor(props) {
+					super(props);
+					this.state = { x: false };
+				}
+
+				render() {
+					if (this.state.x === false) {
+						this.setState({ x: true });
+					}
+
+					return <p>{this.state.x === true ? 'hi' : 'bye'}</p>;
+				}
+			}
+
+			render(<Persons />, scratch);
+			expect(scratch.innerHTML).to.equal('<p>hi</p>');
+		});
 	});
 });
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -164,6 +164,12 @@ export function diff(
 			c._parentDom = parentDom;
 
 			tmp = c.render(c.props, c.state, c.context);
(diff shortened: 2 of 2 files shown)


test/browser/components.test.js
```diff
@@ -2575,5 +2575,25 @@ describe('Components', () => {
...
+		it('should call render again when state got dirtied', () => {
+			class Persons extends Component {
+				constructor(props) {
+					super(props);
+					this.state = { x: false };
+				}
+
+				render() {
+					if (this.state.x === false) {
+						this.setState({ x: true });
```
marvinhagemeister: Curious, since the test is likely a stripped down version of the original problem: What was the scenario which lead to this PR? :slightly_smiling_face: Running effects in render seems like a known to be bad thing.

developit: When exiting a `render()` or functional component invocation, if state got dirtied for that component _during the execution of that function_, we can immediately re-invoke the function and use the updated result. This bypasses re-entering that update into the render queue and skips the intermediary render entirely.

```js
const Persons = (props) =>  {
  // say we need state, to allow for local mutation:
  const [checked, set] = useState(props.checked)

  const toggle = useCallback(() => {
    set(checked);
  }, [checked]);

  // when receiving a new props.checked value, update state:
  useEffect(() => {
    // here's the issue: we first render fully (to DOM) with the old value,
    // then immediately render again with the new value. Invisible but expensive.
    set(props.checked);
  }, [props.checked]);

  return <input type="checkbox" checked={checked} onClick={toggle} />
}

render(<Persons checked={false} />);
```

developit: @JoviDeCroock just pointed out that this actually doesn't get fixed by this PR and the suggested approach, because `useEffect()` has not yet run when we exit render.

andrewiggins: Also curious what scenario this is used for. At least in the above example (if theoretically this PR solved it), I typically try to avoid just copying props into local state since it duplicates the state between two components making it harder to understand who owns it and keeping each copy in sync. Though maybe it's useful for delaying value changes up the tree? perhaps for animations or debouncing?

JoviDeCroock:
Closing due to misunderstanding, https://github.com/preactjs/preact/pull/2423#discussion_r395798267

Issue #2432, opened 2020-03-31
Exception thrown during `act` callback breaks future effects


robertknight:
From the [code for `act`](https://github.com/preactjs/preact/blob/ca42c53210db925fe14d8f0f9122eb3bf43556d7/test-utils/src/index.js#L68) it is clear that exceptions thrown by the callback passed to `act` are not handled. If an exception is thrown, the cleanup code that flushes effects and renders before `act` returns is not run. As a result, the pending render/effect queues are left non-empty and subsequent calls to state updaters / `useEffect` don't schedule flushes of these queues.

A result that we've observed in our test suite is that one test failing with an error thrown inside a call to `act` ends up causing a cascade of spurious failures of other tests because any subsequent `useEffect` effects stop working.

### Expected Behavior

If an exception occurs in `act`'s callback, this should not affect future renders or effects.

### Actual Behavior

If an exception occurs during an `act` callback, the queue of pending renders and effects may be left non-empty. As a result, future state updates and effects don't schedule flushing of these queues.

marvinhagemeister:
I'll see to cut a release shortly. Mainly I want to get #2386 in before making one 🙂

PR #2433, opened 2020-03-31
Improve handling of exceptions thrown during `act` callback

robertknight:
This PR fixes https://github.com/preactjs/preact/issues/2432 by ensuring that effect and pending render queues are flushed if a callback passed to `act` throws an exception. Previously `useEffect` effects and state updates would stop working after a call to `act` aborted with an exception. The result was that a single failure in a test suite could trigger a cascade of failures in unrelated tests, which was pretty confusing.

In the process, I discovered an issue that if an effect threw an exception, the effect would remain in the component's effect queue and would end up running again the next time that component's effects were flushed.

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -258,10 +258,11 @@ export function useErrorBoundary(cb) {
 function flushAfterPaintEffects() {
 	afterPaintEffects.some(component => {
 		if (component._parentDom) {
+			const effects = component.__hooks._pendingEffects;
+			component.__hooks._pendingEffects = [];
 			try {
-				component.__hooks._pendingEffects.forEach(invokeCleanup);
-				component.__hooks._pendingEffects.forEach(invokeEffect);
-				component.__hooks._pendingEffects = [];
+				effects.forEach(invokeCleanup);
+				effects.forEach(invokeEffect);
 			} catch (e) {
 				options._catchError(e, component._vnode);
 				return true;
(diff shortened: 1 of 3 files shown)


hooks/src/index.js
```diff
@@ -258,10 +258,11 @@ export function useErrorBoundary(cb) {
 function flushAfterPaintEffects() {
 	afterPaintEffects.some(component => {
 		if (component._parentDom) {
+			const effects = component.__hooks._pendingEffects;
+			component.__hooks._pendingEffects = [];
```
JoviDeCroock: can't we move this to `catch` and its original position?

robertknight: Do you mean putting `component.__hooks._pendingEffects = []` in _both_ the catch block and its original position, or something else?

JoviDeCroock: I mean that yes, will save bytes since the allocation will probably be the expensive part here

robertknight: Are you referring to minified size or gzipped size? If the latter I'm not a fan of gzip-golf but I'll defer to whatever the team normally prioritises.

JoviDeCroock: The community aims at minzip size so we are prioritizing that as well

JoviDeCroock: I mean even minified it's smaller

robertknight: OK, I've updated this accordingly.

robertknight:
Thanks for the quick review!

marvinhagemeister:
Great work! That's a very nice improvement!!

PR #2434, opened 2020-04-01
Fix error messages logged by failed prop type checks

robertknight:
The error message logged when a value of an incorrect type is passed as
the prop `propName` for a component `Widget` should read:

```
Failed prop type: Invalid prop `propName` supplied to `Widget`
```

Instead it read:

```
Failed Widget type: Invalid Widget `propName` supplied to `Widget`
```

The issue was that the wrong parameters were passed to `checkPropType`.

### DIFF
--- a/debug/src/check-props.js
+++ b/debug/src/check-props.js
@@ -2,6 +2,25 @@ const ReactPropTypesSecret = 'SECRET_DO_NOT_PASS_THIS_OR_YOU_WILL_BE_FIRED';
 
 let loggedTypeFailures = {};
 
+/**
+ * Reset the history of which prop type warnings have been logged.
+ */
+export function resetPropWarnings() {
+	loggedTypeFailures = {};
+}
+
+/**
+ * Assert that the values match with the type specs.
+ * Error messages are memorized and will only be shown once.
+ *
+ * Adapted from https://github.com/facebook/prop-types/blob/master/checkPropTypes.js
+ *
+ * @param {object} typeSpecs Map of name to a ReactPropType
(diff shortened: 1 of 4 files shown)


debug/src/check-props.js
```diff
@@ -2,6 +2,25 @@ const ReactPropTypesSecret = 'SECRET_DO_NOT_PASS_THIS_OR_YOU_WILL_BE_FIRED';
 
 let loggedTypeFailures = {};
 
+/**
+ * Reset the history of which prop type warnings have been logged.
+ */
+export function resetPropWarnings() {
+	loggedTypeFailures = {};
+}
```
robertknight, line 10: I added this helper for use in tests to reset the history of logged warnings. This is currently re-exported from the `preact/debug` package, so this is a minor API addition.

Is there a preferred alternative way of creating this kind of helper for use in tests?

debug/src/debug.js
```diff
@@ -214,8 +214,8 @@ export function initDebug() {
 			checkPropTypes(
 				vnode.type.propTypes,
 				vnode.props,
-				getDisplayName(vnode),
-				serializeVNode(vnode)
+				'prop',
```
robertknight, line 217: TIL that "prop types" can also be used to check things that are not props, such as context. See https://github.com/facebook/prop-types/blob/1c4077b7455e037bd8f81f48e9c51d60c972f8e9/checkPropTypes.js#L37.

debug/src/check-props.js
marvinhagemeister: Sounds good to me 👍

Issue #2425, opened 2020-03-20
RFC: selective context propagation


JoviDeCroock:
# Selective context propagation

I see the comparison of [Redux](https://redux.js.org/) to [Context](https://preactjs.com/guide/v10/context/) quite often lately and wanted to act on this.

## Differences

Redux is performant by nature, it uses the concept of selectors to memoize values and rerender selectively, context on the other hand will propagate through to every  `Consumer` subscribed to the whole context.

The clear difference between these two we can notice is that when x amount of components are subscribed to individual pieces of state, for redux this will rerender the amount of times the state has changed while with the current form of context this will be x.

## Prerequisites

One of the big prerequisites we're lacking out for this concept to succeed is strict-equality on vnodes, this means that if the `children` did not change from render 1 --> render 2 we won't have to evaluate these children. This is a very common pattern when it comes to contextProviders.

Let's consider the following scenario:

```jsx
const App = () => (
  <MyProvider>
    <MyRoutes />
  </MyProvider>
);
```

When `<App />` updates this whole tree will rerender since the function will execute and reevaluate all these children, let's for clarity look at a hypothetical `<MyProvider />` implementation.

```jsx
const { Provider } = createContext();
const MyProvider = ({ children }) => {
  const [state, setState] = useState();
  return (
    <Provider value={{ state, setState }}>
      {children}
    </Provider>
  )
}
```

As we can see there's a prop injected named `children`, if this doesn't change the vnode is equal and shouldn't be diffed. This means that if a `Consumer` calls `setState` the `MyProvider` component will alter its internal state but the `children`, injected by `<App />` (which hasn't reexecuted), will still be the same so we can safely decide to not diff this part.

This means that instead of walking the tree downwards we'll only trigger the `contextConsumers`.

[Current PR](https://github.com/preactjs/preact/pull/2386)

## Implementation

To come back to our selective context propagation, we have two options to support this effectively.

### Option 1

We can supply a rerender function similar to `shouldComponentUpdate`, `useContext`/others could allow a second argument (or a new hook) that allows the user to selectively compare the old and new context and decide wether this hook needs an emit. This would always return the full context (important note when reading the next proposition).

```jsx
const MyName = () => {
  const { name } = useContext(userContext, (prev, curr) => prev.name !== curr.name);
  return <p>Welcome {name}</p>;
}
```

### Option 2

[...]

PR #2435, opened 2020-04-01
POC for selective context option 2

JoviDeCroock:
This still relies on https://github.com/preactjs/preact/pull/2386 to bail out of Provider children
Implements a POC for option 2  https://github.com/preactjs/preact/issues/2425

This isn't aimed at `contextType` on class components since this can naturally benefit from `shouldComponentUpdate` instead. This is aimed at `Consumer` and `useContext`.

Will add tests for this when #2386 is merged since this is quite important.

Will make a second PR implementing option 1

TODO:
- types
- docs
- test Consumer

### DIFF
--- a/src/create-context.js
+++ b/src/create-context.js
@@ -22,17 +22,20 @@ export function createContext(defaultValue) {
 				this.shouldComponentUpdate = _props => {
 					if (this.props.value !== _props.value) {
 						subs.some(c => {
-							c.context = _props.value;
-							enqueueRender(c);
+							if (!c[1] || c[1](_props.value, this.props.value)) {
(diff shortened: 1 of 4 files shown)


src/create-context.js
```diff
@@ -9,7 +9,7 @@ export function createContext(defaultValue) {
 		_id: '__cC' + i++,
 		_defaultValue: defaultValue,
 		Consumer(props, context) {
-			return props.children(context);
+			return props.children(props.selector(context));
```
porfirioribeiro, line 12: We might not have a selector prop in the consumer

JoviDeCroock: Oops :D fixed

PR #2437, opened 2020-04-02
[actions] Upload dist

developit:


.github/workflows/ci-pr.yml
```diff
@@ -60,3 +60,8 @@ jobs:
       uses: coverallsapp/github-action@v1.0.1
       with:
         github-token: ${{ secrets.GITHUB_TOKEN }}
+    - name: Upload Dist
+      uses: actions/upload-artifact@v1
+      with:
+        name: dist
+        path: dist
```
marvinhagemeister, line 67: Does this apply to all sub-packages too, or only core?

developit: just to core, unfortunately. Also I believe the URLs expire after 5 minutes :/

developit:
I think a nicer option here might be to have `compressed-size-action` create a GitHub "check" with full diffs of each matched output file inside of `<details>` elements. That would make it possible to inspect the built differences of any PR without having to check it out locally.

andrewiggins:
I think this is superseded by #2579 which includes as an artifact all the build output from all sub-packages, so I'mma close this. Obviously, if you think this needs to stay open, please do so.

PR #2438, opened 2020-04-03
hydrate fallback

JoviDeCroock:
When we see a newly inserted tree, we fallback to non-hydration mode for that specific subtree. This was brought up in a slack issue.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -298,6 +298,7 @@ function diffElementNodes(
 	}
 
 	if (dom == null) {
+		isHydrating = false;
 		if (newVNode.type === null) {
 			return document.createTextNode(newProps);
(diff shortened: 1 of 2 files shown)


src/diff/index.js
```diff
@@ -298,6 +298,7 @@ function diffElementNodes(
 	}
 
 	if (dom == null) {
+		isHydrating = false;
```
JoviDeCroock, line 301: Might be cheaper to reuse the `null` allocation of line 313 and set `isHydrating` to null instead WDYT?

developit: I think we could just move this to 212 without changing its type to `null`. Terser will be able to merge the statements better there.

Issue #2441, opened 2020-04-06
Typescript definition for act does not account for promises


davidje13:
The [type definition for `act()`](https://github.com/preactjs/preact/blob/master/test-utils/src/index.d.ts#L2) implies it is always synchronous, but it actually [returns a promise in all cases](https://github.com/preactjs/preact/blob/master/test-utils/src/index.js#L27) (even if not given a promise-generating function).

### Reproduction

```ts
import { act } from 'preact/test-utils';

// Both of the following are valid, but rejected in TypeScript:

// calling .then()
// Property 'then' does not exist on type 'void'.
act(foo).then(() => {});

// or a linter rule:
// Unexpected `await` of a non-Promise (non-"Thenable") value (@typescript-eslint/await-thenable)
await act(async () => {
  // ...
});
```

### Expected Behavior

`act` should be marked as returning `Promise<void>` (and ideally as consuming `() => (Promise<void> | void)`, though that is more just for documentation.

ddayguerrero:
Hey @davidje13, hope you don't mind that I took a stab at this (fairly straightforward)! I can mark it as ready if we're good 🙇‍♂️

PR #2444, opened 2020-04-07
Update act type signature

ddayguerrero:
Resolves https://github.com/preactjs/preact/issues/2441

Updating `act`'s typings
1- Callback return type can still be function or explicitly, return a promise (if async ideally)
2- Return type to be `Promise<void>`

## What approach did I use?
🔍 Digging Preact's code base to learn

- TIL act supports async

📚 Mostly following Preact's [contribution guidelines](https://github.com/preactjs/preact/blob/master/CONTRIBUTING.md):

### DIFF
--- a/test-utils/src/index.d.ts
+++ b/test-utils/src/index.d.ts
@@ -1,3 +1,3 @@
 export function setupRerender(): () => void;
-export function act(callback: () => void): void;
+export function act(callback: () => void | Promise<void>): Promise<void>;
 export function teardown(): void;


davidje13:
Looks good. Should anything be added to the [typescript tests](https://github.com/preactjs/preact/tree/master/test/ts) for this? (there doesn't seem to be much so far so maybe the answer is no)

ddayguerrero:
> I'm going to make a note so we don't forget about TS next time 
Sounds like a good idea!

PR #2451, opened 2020-04-08
Support node13

JoviDeCroock:
In node v13 we either need to end our modular file in `.mjs` or have a `module: true` entry in our package json to resolve as a module

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,43 +9,42 @@
   "umd:main": "dist/preact.umd.js",
   "unpkg": "dist/preact.umd.js",
   "source": "src/index.js",
-  "type": "module",
   "exports": {
     ".": {
       "browser": "./dist/preact.module.js",
       "umd": "./dist/preact.umd.js",
-      "import": "./dist/preact.module.js",
+      "import": "./dist/preact.mjs",
       "require": "./dist/preact.js"
     },
     "./compat": {
       "browser": "./compat/dist/compat.module.js",
       "umd": "./compat/dist/compat.umd.js",
       "require": "./compat/dist/compat.js",
-      "import": "./compat/dist/compat.module.js"
+      "import": "./compat/dist/compat.mjs"
     },
     "./debug": {
       "browser": "./debug/dist/debug.module.js",
       "umd": "./debug/dist/debug.umd.js",
       "require": "./debug/dist/debug.js",
-      "import": "./debug/dist/debug.module.js"
+      "import": "./debug/dist/debug.mjs"
     },
     "./devtools": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -54,12 +60,12 @@
   "types": "src/index.d.ts",
   "scripts": {
     "build": "npm-run-all --parallel build:* && cp dist/preact.js dist/preact.min.js",
-    "build:core": "microbundle build --raw",
-    "build:debug": "microbundle build --raw --cwd debug",
-    "build:devtools": "microbundle build --raw --cwd devtools",
-    "build:hooks": "microbundle build --raw --cwd hooks",
-    "build:test-utils": "microbundle build --raw --cwd test-utils",
-    "build:compat": "microbundle build --raw --cwd compat --globals 'preact/hooks=preactHooks'",
+    "build:core": "microbundle build --raw && cp ./dist/preact.module.js ./dist/preact.mjs",
```
kristoferbaxter, line 63: Might want to convert these to `npm-run-all` like "build" is so they work on Windows based systems.

andrewiggins: [&& works on Windows](https://ss64.com/nt/syntax-redirection.html). `cp` does not. Might need to use https://npm.im/cpy-cli

developit: +1 for cpy

developit:
I think I would prefer to try `"type":"module"` rather than the `.mjs` extension. Anyone else feel that way?

JoviDeCroock:
> I think I would prefer to try `"type":"module"` rather than the `.mjs` extension. Anyone else feel that way?

I wouldn't mind trialing it

nsivertsen:
> I think I would prefer to try "type":"module" rather than the .mjs extension. Anyone else feel that way?

That route also feels more right to me but iirc that broke CJS consumers when I tried it, because Node will import the CJS entry point but try to parse it as an ES Module.

package.json
```diff
@@ -28,6 +29,12 @@
       "require": "./debug/dist/debug.js",
       "import": "./debug/dist/debug.module.js"
     },
+    "./devtools": {
+      "browser": "./devtools/dist/devtools.module.js",
```
marvinhagemeister, line 33: Good catch about the missing devtools package 👍 💯

developit:
@nsivertsen I think we could potentially get around that by renaming the commonjs entry modules to `.cjs`. Though that may have ill-effects on tools that do file extension / filename-based mapping

nsivertsen:
@developit Yeah, I feared that might cause similar problems for tooling as using `.mjs`, but I don't have any research to back that up. I guess either way this may cause some breakage

PR #2452, opened 2020-04-08
POC for selective context option 1

JoviDeCroock:
This still relies on https://github.com/preactjs/preact/pull/2386 to bail out of Provider children
Implements a POC for option 1 https://github.com/preactjs/preact/issues/2425

This isn't aimed at `contextType` on class components since this can naturally benefit from `shouldComponentUpdate` instead. This is aimed at `Consumer` and `useContext`.


src/create-context.js
```diff
@@ -22,17 +22,20 @@ export function createContext(defaultValue) {
...
+								enqueueRender(c[0]);
+							}
 						});
 					}
 				};
 
-				this.sub = c => {
-					subs.push(c);
+				this.sub = (c, shouldUpdate) => {
+					const entry = [c, shouldUpdate];
```
developit, line 34: I think `const entry = { _component: c, _shouldUpdate: shouldUpdate };` ends up being smaller?

JoviDeCroock: This adds 1 byte

src/create-context.js
```diff
@@ -21,18 +21,21 @@ export function createContext(defaultValue) {
...
+								enqueueRender(payload.c);
+							}
 						});
 					}
 				};
 
-				this.sub = c => {
-					subs.push(c);
+				this.sub = (c, shouldUpdate) => {
+					const entry = { c, u: shouldUpdate };
```
JoviDeCroock, line 34: Wondering if we want to allow for this function to be updated, my gut sais no. If we want to we could do it with a mutable ref?

developit: wouldn't that require unsubscribing and resubscribing right now? Only way to do it without that would be to pass an object, which is sortof a strange API signature.

JoviDeCroock: Agreed, unless we do like `x.sub({ current: shouldUpdate })` in Consumer and useContext

JoviDeCroock:
Coveralls is being a nuisance on a line that isn't touched 😓

src/create-context.js
```diff
@@ -21,18 +21,21 @@ export function createContext(defaultValue) {
 
 				this.shouldComponentUpdate = _props => {
 					if (this.props.value !== _props.value) {
-						subs.some(c => {
-							c.context = _props.value;
-							enqueueRender(c);
+						subs.some(payload => {
+							payload[0].context = _props.value;
+							if (!payload[1] || payload[1](_props.value, this.props.value)) {
```
gnoff, line 26: payload[1] can be a closure over props from a pervious render. if the author isn't careful the shouldUpdate function may throw since the potentially soon to be stale props won't be compatible with the newly updating value.

I'm not super familiar with the preact codebase so apologies if this is just a distraction and error recovery is handled elsewhere that would protect against this case

JoviDeCroock: That's an issue that can only occur in 2phase renderers if I'm not mistaking

JoviDeCroock: Hey, I've added tests for this

https://github.com/preactjs/preact/pull/2452/files#diff-507a7cc69be8471ba7e6c38f43a7cd4eR437

gnoff: the update looks good and solves the more severe case of stale props (basically capturing it on first use and never updating it looks like) but I was actually more thinking of this edge case

Imagine this component
```
function MyComponent({ someKey }) {
  // update when name changes from dynamically referenced context property
  let ctx = useContext(MyContext, (next, prev) => { next[someKey].name !== prev[someKey] .name})
  return ... // use ctx[someKey].name presumably
}
```

```
let things = {
  aThing: { name: 'foo' }
};

//render
<MyContext.Provider value={things}>
  <MyComponent someKey={Object.keys(things)[0]} /> //get the first thing key
</MyContext.Provider>

then change things
things = {
  aDifferentThing: { name: 'bar' }
};

//render again
// when Provider calls subs the captured update will do things[aThing].name and will throw
<MyContext.Provider value={things}>
  <MyComponent someKey={Object.keys(things)[0]} />
</MyContext.Provider>
```

src/create-context.js
```diff
@@ -21,18 +21,24 @@ export function createContext(defaultValue) {
...
-						subs.some(c => {
-							c.context = _props.value;
-							enqueueRender(c);
+						subs.some(payload => {
+							payload[0].context = _props.value;
+							if (
+								!payload[1] ||
+								payload[1]._update(_props.value, this.props.value)
+							) {
+								enqueueRender(payload[0]);
```
gnoff, line 30: if a component reads from `MyContext1` and then in later render reads from `MyContext2` the sub for `MyContext1` will hang around and you will run the updater meant for `MyContext2` whenever `MyContext1` provides a new value, possibly leading to errors or unnecessary render queuing.

This is actually already true, minus the updater, b/c even without the updater the sub isn't cleaned up until unmount it seems. so the part that is new is surprising errors running a shouldUpdate intended for a new context using the old contexts values

JoviDeCroock: Is this allowed? Hmm TIL, will make a quick PR for that 👍

JoviDeCroock:
Pausing due to https://github.com/preactjs/preact/issues/2425#issuecomment-618724038 need to look into this.

src/create-context.js
gnoff: yes, I don't think users often take advantage of it but the React API does not prohibit it. their model encodes the dependency on a context on the dependent component not on the provider presumably for this very reason. tracking which components need to be notified in a provider is hard when that dependency may not be uniform over time. without selectors it is mostly just a deop where you render some things more than needed but with the selector (or updater in this case) you introduce potential for runtime errors so more consequential. I will say that with 2 phase render React has a much harder job doing to the dependency tracking and with a 1 phase it seems much more possible so I wouldn't necessarily adopt their approach

Issue #2455, opened 2020-04-09
dangerouslySetInnerHTML combined with implicit element content causes TypeError


enterit:
### Reproduction

The issue only happens in a rare combination, when:
- Preact is bundled in strict JS mode
- JSX element uses dangerouslySetInnerHTML
- JSX element has implicit content

Then, on element unmount, this leads to an error:
`Uncaught (in promise) TypeError: Cannot create property '_nextDom' on string ' '`

The issue is reproducible in Preact 10.4.0.

### Steps to reproduce

https://codepen.io/enterit/pen/KKdKzmo

```javascript
/** @jsx h */

const { h, render } = preact
const { useState, useEffect } = preactHooks

const TextDiv = () => (<div dangerouslySetInnerHTML={{ __html: '' }}>some text</div>)
const App = () => {
  const [textDiv, setTextDiv] = useState(true)
  useEffect(() => { setTextDiv(false) }, [])
  return (
    textDiv && <TextDiv/>
  )
}

render(<App/>, document.body)
```

### Expected Behavior

The element should unmount without errors.

### Actual Behavior

TypeError is thrown on element unmount.

enterit:
Ah, you found the root cause, and fixed it, and unit-tested - all of that in the blink of an eye!
Super-cool, thanks a lot!

PR #2456, opened 2020-04-09
Prevent setting of nextDom on textnodes

JoviDeCroock:
Fixes: https://github.com/preactjs/preact/issues/2455

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -441,9 +441,7 @@ export function unmount(vnode, parentVNode, skipRemove) {
 
 	// Must be set to `undefined` to properly clean up `_nextDom`
 	// for which `null` is a valid value. See comment in `create-element.js`
-	if (vnode.type) {
-		vnode._dom = vnode._nextDom = undefined;
-	}
+	vnode._dom = vnode._nextDom = undefined;
 
 	if ((r = vnode._component) != null) {
 		if (r.componentWillUnmount) {
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -439,7 +439,9 @@ export function unmount(vnode, parentVNode, skipRemove) {
 
 	// Must be set to `undefined` to properly clean up `_nextDom`
 	// for which `null` is a valid value. See comment in `create-element.js`
-	vnode._dom = vnode._nextDom = undefined;
+	if (vnode.type) {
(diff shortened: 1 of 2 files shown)


src/diff/index.js
```diff
@@ -439,7 +439,9 @@ export function unmount(vnode, parentVNode, skipRemove) {
 
 	// Must be set to `undefined` to properly clean up `_nextDom`
 	// for which `null` is a valid value. See comment in `create-element.js`
-	vnode._dom = vnode._nextDom = undefined;
+	if (vnode.type) {
```
JoviDeCroock, line 442: Drafted this since I want to see if there's a better solution

developit:
The interesting thing is that we should never have an un-normalized VNode in `unmount()`. Somehow we're getting a String, which means somewhere we're setting `._children = .props.children` or something...

JoviDeCroock:
> The interesting thing is that we should never have an un-normalized VNode in `unmount()`. Somehow we're getting a String, which means somewhere we're setting `._children = .props.children` or something...

Yes, due to `dangerouslySetInnerHTML` I guess

developit:
ah! yeah so this line actually sets `vnode._children` to the wrong value (un-normalized):
https://github.com/preactjs/preact/blob/master/src/diff/index.js#L365

in all cases _except_ dangerouslySetInnerHTML, we end up resetting it here back to the correct value:
https://github.com/preactjs/preact/blob/master/src/diff/children.js#L60

JoviDeCroock:
Do we want to make `dangerouslySetInnerHTML` a vnode? or does this low size fix work for you?

developit:
Alright I think I found the spot. We are intentionally overloading `_children` as "temporary new children" any time we jump into `diffChildren` - the issue here is that we aren't calling diffChildren, so it never gets renormalized.

https://github.com/preactjs/preact/pull/2458

src/diff/index.js
```diff
@@ -439,7 +441,9 @@ export function unmount(vnode, parentVNode, skipRemove) {
...
+		vnode._dom = vnode._nextDom = undefined;
+	}
```
developit, line 446:
```suggestion
	vnode._dom = vnode._nextDom = undefined;
```

developit:
Testing now with only the normalization change and no patch change.

PR #2462, opened 2020-04-10
Tachometer benchmarks

andrewiggins:
Initial infrastructure to run [Polymer/tachometer](https://github.com/polymer/tachometer) based benchmarks. Would love to get some initial feedback to see if people like this direction. The readme includes the basic instructions for running the benchmarks.

I was lazy and didn't setup a build pipeline. I just opted to write `createElement` calls directly so the src was runnable directly in the browser (actually I use [sucrase](https://sucrase.io/) to manually translate JSX and copied it over). Besides the js-framework-benchmark app, I suspect most of our benchmarks will be small in size, so maybe we can get by with this for now? Thoughts?

~~Note: This PR is dependent on polymer/tachometer#156 getting released for comparing different versions of preact to work properly (and on polymer/tachometer#157 for Windows support). I suspect these will get released in the next couple of days~~ 🎉 Upgraded to 0.4.18


andrewiggins:
Oh! I should mention that this is currently running the benchmarks against 3 versions of Preact: 8, 10, and local. We could also do what's in master on GitHub. What combination do you think is most useful?

andrewiggins:
@developit Feel free to still leave a review if you have thoughts/questions. Tag me and I'll respond and open another PR if need be :)

PR #2463, opened 2020-04-10
(fix) - circular vnode ref

JoviDeCroock:
We've been hearing some issues with circular refs from testing frameworks and `react-fast-compare`. This is an attempt to solving these

### DIFF
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -80,7 +80,6 @@ export function createVNode(type, props, key, ref, original) {
 
 	Object.defineProperty(vnode, '_original', {
 		value: original || vnode,
-		enumerable: false,
 		writable: true
 	});
 	if (options.vnode) options.vnode(vnode);
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -78,6 +78,10 @@ export function createVNode(type, props, key, ref, original) {
 		constructor: undefined
 	};
 
+	// We're setting this as non-enumberable (+writeable)
+	// because we need to overwrite this value sometimes and
+	// because this can induce circular references on _children and
+	// for testing frameworks traversing the vnodes.
 	Object.defineProperty(vnode, '_original', {
 		value: original || vnode,
 		writable: true
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -80,10 +80,8 @@ export function createVNode(type, props, key, ref, original) {
 
 	Object.defineProperty(vnode, '_original', {
 		value: original || vnode,
-		enumerable: false,
-		writable: true
+		enumerable: false
 	});
-
 	if (options.vnode) options.vnode(vnode);
 
 	return vnode;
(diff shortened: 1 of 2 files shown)


src/create-element.js
```diff
@@ -75,11 +75,14 @@ export function createVNode(type, props, key, ref, original) {
...
 		_component: null,
-		constructor: undefined,
-		_original: original
+		constructor: undefined
 	};
 
-	if (original == null) vnode._original = vnode;
+	Object.defineProperty(vnode, '_original', {
+		value: original || vnode,
+		enumerable: false,
```
developit, line 83: This is the default value for enumerable.

src/create-element.js
```diff
@@ -75,11 +75,14 @@ export function createVNode(type, props, key, ref, original) {
 		// a _nextDom that has been set to `null`
 		_nextDom: undefined,
 		_component: null,
-		constructor: undefined,
-		_original: original
+		constructor: undefined
 	};
 
-	if (original == null) vnode._original = vnode;
+	Object.defineProperty(vnode, '_original', {
```
developit, line 81: This likely bypasses mangle.json config.

One way I thought of for getting around that was to use `Object.defineProperties`:

```js
Object.defineProperties(vnode, {
  _original: {
    value: original == null ? vnode : original
  }
});
```

JoviDeCroock: From the compiled code:

```
return Object.defineProperty(o,"__v",{value:r||o,enumerable:!1,writable:!0})
```

src/create-element.js
```diff
@@ -75,11 +75,13 @@ export function createVNode(type, props, key, ref, original) {
...
```
marvinhagemeister, line 81: I'd love a quick comment in the code about the reason for using `Object.defineProperty` here. For people new to our codebase this may be a helpful thing to do :slightly_smiling_face:

JoviDeCroock: Will do, good shout Marvin

marvinhagemeister:
Very interesting! I'm wondering if we can get away with just renaming the property from `_original` to `_owner` and add a fake `$$typeof` property. Not sure if that is smaller than the current PR.

Looking at the source of `react-fast-compare` [they explicitly have a case that handles that](
https://github.com/FormidableLabs/react-fast-compare/blob/8253fc884bd4c5f88e251ea2b80d33be0547376c/index.js#L93-L98) and maybe others do too? :crossed_fingers: 

```js
if (keys[i] === '_owner' && a.$$typeof) {
  // React-specific: avoid traversing React elements' _owner.
  //  _owner contains circular references
  // and is not needed when comparing the actual elements (and not their owners)
  // .$$typeof and ._store on just reasonable markers of a react element
  continue;
}
```

JoviDeCroock:
@marvinhagemeister my concern is having to implement that in every preact-specific test framework then as well

JoviDeCroock:
Closing due to upstream fix https://github.com/FormidableLabs/react-fast-compare/pull/67

Issue #2468, opened 2020-04-12
cloneElement leaves visible ref="[object Object]"


vaneenige:
Hi there! Found something odd when using cloneElement, it's not a breaking issue, but it doesn't look pretty in the DOM.

**Reproduction:**

Here's a sandbox: https://codesandbox.io/s/epic-bird-1571e w/ preact@10.4.0

In this demo you see two elements being rendered:

- The first element is using cloneElement to pass a ref to it's children
- The second element uses the children callback to pass it down

If you look at the DOM you can see the first element receiving ref="[object Object]" which looks odd. In my project I've also seen that this doesn't happen when you hydrate or render over SSR'ed DOM. So in my case I noticed it after switching pages.

**Issue:** Normally refs are stripped from the DOM and with cloneElement it was visible.

daniele-orlando:
I still can reproduce this bug with Preact **10.5.7**.
Of course no problem if I use React.

![ref](https://user-images.githubusercontent.com/504386/102650696-9ad6b300-416b-11eb-964a-23ba7051591b.png)

PR #2469, opened 2020-04-12
(fix) - normalize props on cloneElement

JoviDeCroock:
Fixes: https://github.com/preactjs/preact/issues/2468

This normalizes the props object that can optionally be passed as a second argument to `cloneElement` just like we do on the regular `createElement`.

CC @vaneenige

### DIFF
--- a/test/browser/cloneElement.test.js
+++ b/test/browser/cloneElement.test.js
@@ -1,4 +1,4 @@
-import { createElement, cloneElement } from 'preact';
+import { createElement, cloneElement, createRef } from 'preact';
 
 /** @jsx createElement */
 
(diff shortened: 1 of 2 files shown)


test/browser/cloneElement.test.js
```diff
@@ -61,4 +61,10 @@ describe('cloneElement', () => {
 		clone = cloneElement(instance, { ref: b });
 		expect(clone.ref).to.equal(b);
 	});
+
+	it('should normalize props', () => {
+		const div = <div>hello</div>;
+		const clone = cloneElement(div, { ref: createRef() });
```
marvinhagemeister, line 67: Let's add a small test for `key` too :+1:

PR #2470, opened 2020-04-13
Update props.js

developit:


JoviDeCroock (approved):

marvinhagemeister:
@developit A sentence about the reason for the change would be awesome :+1: Is it performance? A bug you encountered somewhere? :upside_down_face:

developit:
Ah nah - I was just guessing this would be smaller and wasn't near a computer so I abused the size bot to check.

JoviDeCroock:
Closing: already done — v11 uses the same ARIA fast-path check (character-based, not the `/^ar/` regex) in `src/diff/props.js`.

Issue #2464, opened 2020-04-11
'checked' is filtered and not assigned to DOM


dntzhang:
Vue, React and no framework are in line with expectations, but preact(v10.1.0) is not.

https://github.com/Tencent/omi/tree/master/components/o-checkbox/demos/preact-demo

```jsx
import './style';
import { Component } from 'preact';
import '@omiu/checkbox';

export default class App extends Component {
  onChange = (evt) => {
    console.log(evt.detail)
  }

  render() {
    return (
      <div>
        <o-checkbox label="Label"></o-checkbox>
        <br />

        <o-checkbox checked={true} label="checked"></o-checkbox>
        <br />

        <o-checkbox onChange={this.onChange} indeterminate label="indeterminate"></o-checkbox>
        <br />

        <o-checkbox disabled label="disabled"></o-checkbox>
        <br />
        <o-checkbox disabled checked label="disabled checked"></o-checkbox>
      </div>
    );
  }
}
```

JoviDeCroock:
Hey @dntzhang 

I've made an attempt at fixing this since the problem isn't entirely clear to met yet, when [this build](https://github.com/preactjs/preact/pull/2465/checks?check_run_id=578921490) finishes would you mind trying it out on your example?

Thanks in advance

dntzhang:
Now it's ok. But when i build the demo, i met a new problem:

```bash
> preact@0.0.0 build /Users/dnt/project/omi/components/o-checkbox/demos/preact-demo
> preact build

 Build  [=================== ] 93% (5.6s) after chunk asset optimization
 ssr-bundle.c3928.css ⏤  167 B (+167 B)
        ssr-bundle.js ⏤  11.8 kB (+11.8 kB)

ℹ INFO ⚛️ No custom sw.js detected: compiling default Service Worker.
 Build  [==================  ] 92% (9.0s) additional asset processingRe-using existing @babel/preset-modules configuration.
 Build  [=================== ] 93% (12.1s) after chunk asset optimization
   ssr-bundle.c3928.css ⏤  0 B (-167 B)
          ssr-bundle.js ⏤  0 B (-11.8 kB)
              sw-esm.js ⏤  741 B (+741 B)
                  sw.js ⏤  738 B (+738 B)
       bundle.c3928.css ⏤  108 B (+108 B)
    bundle.516a0.esm.js ⏤  14.8 kB (+14.8 kB)
 polyfills.d9a55.esm.js ⏤  2 kB (+2 kB)
[...]

PR #2472, opened 2020-04-14
Use "setProperty" to set "value" and "checked" properties

andrewiggins:
Follow up to #2464

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -387,7 +387,7 @@ function diffElementNodes(
 				(i = newProps.value) !== undefined &&
 				i !== dom.value
 			) {
-				setProperty(dom, 'value', i == null ? '' : i, oldProps.value, false);
+				setProperty(dom, 'value', i, oldProps.value, false);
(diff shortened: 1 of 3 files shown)


src/diff/index.js
```diff
@@ -385,17 +384,17 @@ function diffElementNodes(
 		if (!isHydrating) {
 			if (
 				'value' in newProps &&
-				newProps.value !== undefined &&
-				newProps.value !== dom.value
+				(i = newProps.value) !== undefined &&
+				i !== dom.value
 			) {
-				dom.value = newProps.value == null ? '' : newProps.value;
+				setProperty(dom, 'value', i == null ? '' : i, oldProps.value, false);
```
JoviDeCroock, line 390:
```suggestion
				setProperty(dom, 'value', i, oldProps.value, false);
```

We do this in `setProperty`

andrewiggins: Oh true! Good call - thanks :)

PR #2473, opened 2020-04-15
Add npm prepare script

andrewiggins:
This `prepare` scripts all Preact to be installed directly from github. In order for this to work on Windows I needed to replace the Linux only `cp` command with a cross platform alternative. I just went with [`cpy-cli`](https://npm.im/cpy-cli) from sindresorhus. Open to other suggestions.

### DIFF
--- a/config/node-13-exports.js
+++ b/config/node-13-exports.js
@@ -1,0 +1,32 @@
+const fs = require('fs');
+
+const subRepositories = ['compat', 'debug', 'devtools', 'hooks', 'test-utils'];
+
+const snakeCaseToCamelCase = str =>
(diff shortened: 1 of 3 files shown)


andrewiggins:
@JoviDeCroock I forgot about #2451. Do you think you could pull out that copy script into a separate PR so I can build on it here?

ForsakenHarmony:
I thought we discussed not using `cpy-cli` because it expands to 181 dependencies 👀

andrewiggins:
@ForsakenHarmony Yea, that's right. I forgot about our discussion related to #2451 😅 - replacing it with Jovi's script now

PR #2479, opened 2020-04-18
Golfing

JoviDeCroock:
This moves a lot of the `!= null` checks to just truthy checks when they were happening on objects rather than primitive values.

This offsets the bytes added in selective context 😅

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -60,7 +60,7 @@ export function diffChildren(
 	newParentVNode._children = toChildArray(
 		newParentVNode._children,
 		childVNode => {
-			if (childVNode != null) {
+			if (childVNode) {
 				childVNode._parent = newParentVNode;
 				childVNode._depth = newParentVNode._depth + 1;
 
@@ -71,7 +71,7 @@ export function diffChildren(
 				oldVNode = oldChildren[i];
 
 				if (
-					oldVNode === null ||
+					(oldVNode = oldChildren[i]) === null ||
 					(oldVNode &&
 						childVNode.key == oldVNode.key &&
 						childVNode.type === oldVNode.type)
@@ -251,7 +251,7 @@ export function diffChildren(
 export function toChildArray(children, callback, flattened) {
 	if (!flattened) flattened = [];
 
-	if (!children || typeof children == 'boolean') {
+	if (children != null || typeof children == 'boolean') {
 		if (callback) flattened.push(callback(null));
 	} else if (Array.isArray(children)) {
 		for (let i = 0; i < children.length; i++) {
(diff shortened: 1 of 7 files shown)


sventschui:
Afaik explicit null checks are faster, but don't know how much. See https://github.com/preactjs/preact/pull/1646#discussion_r287089946

AlexGalays:
> Afaik explicit null checks are faster, but don't know how much. See [#1646 (comment)](https://github.com/preactjs/preact/pull/1646#discussion_r287089946)

Even when using loose equality? that would be weird.

sventschui:
I hacked together a small ESBench (and as I'm super noob in doing this take these statements with caution). Can't share the bench as ESBench doesn't want to let me save (Internal Server Error).

It seems that in Safari it depends™:

- If clause on an empty object variable: `obj != null` is faster than `obj` is faster than `obj !== null`
- If clause on a null variable: `nullVar` is faster than `nullVar !== null` is faster than `nullVar != null`
- If clause on a string variable: `str !== null` is faster than `str` is faster than `str != null`

If one of you could create a proper bench that would be great :)

JoviDeCroock:
Perf https://jsperf.com/null-condition-perf

src/diff/children.js
```diff
@@ -71,7 +71,7 @@ export function diffChildren(
 				oldVNode = oldChildren[i];
 
 				if (
-					oldVNode === null ||
+					(oldVNode = oldChildren[i]) === null ||
```
aralroca, line 74: Is this `oldVNode = oldChildren[i]` necessary? There is the same declaration on line 71.

aralroca: @JoviDeCroock you are amazingly fast! 👏

developit:
Alright I constructed a comprehensive benchmark to illustrate the differences between the various "existence check" approaches.

Just to clarify:

| name | operator | meaning
|--|--|--|
| **null** | `x!=null` | not `null` or `undefined`
| **strict** | `x!==null && x!==undefined` | not `null` or `undefined`
| **loose** | `if(x)` | not `0`, `false`, `""`, `null` or `undefined`

Here's a benchmark comparing the above three approaches, used in a little VDOM approximation. It uses three different h() implementations to construct nested trees with varying vnode types & shapes. Each comparison operator is tested in the three different IC scenarios: monomorphic, polymorphic and megamorphic. In general, accessing `props` is megamorphic, accessing `vnode.*` is monomorphic, and accessing `children` is polymorphic or megamorphic. I've kept them separated out so we can see the effects of each.

https://esbench.com/bench/5ec67fe712464000a01e462f

Based on this benchmark, I'm not sure my original reasoning for consolidating on `==null` checks holds anymore. In Chrome and Firefox, `null` and `loose` seem to be equivalent.

gianpaj:
Does it mean that this PR can be closed because of the performance deterioration?

JoviDeCroock:
Closing due to being superseded by other PR's

PR #2480, opened 2020-04-19
Add support for hooks inspection via devtools

marvinhagemeister:
This PR makes it possible to inspect a components hook state which is needed for the devtools (https://github.com/preactjs/preact-devtools/pull/143).

The way the devtools work is that they inspect a node lazily when the user selects it in the elements panel and not during rendering. So we need to guess which piece of state belongs to which hook at a later point. To do that I've introduced numbers as unique ids for each hook function.

Some of our hooks leverage other hooks to keep our size small and this PR will detect that and will pick the topmost "native" hook. So if `useState` calls `useReducer` internally, the hook will be identified as `useState` no matter what.

Originally I played around with storing that information on the `_hooks` property of a component, but that turns out to be a lot bigger than just passing it as an additional argument to `options._hook`. Besides the type, the callsite index of the current hooks is very relevant for the devtools to display them in the correct order.

To retrieve any custom hooks we need to do some awkward re-rendering of the inspected component. Custom hooks are just plain JavaScript function and we can be only made aware of those by inspecting the stack trace via the `Error` object.

But simply re-rendering poses another issue in that we need to make sure that it doesn't change the state of the current app in any way. We assume that the render function is pure and leverage a new `options._skipEffects` flag to bypass all scheduled effects.

For a longer description of the approach used in the devtools check this comment out: https://github.com/preactjs/preact-devtools/issues/52#issuecomment-615066128

Screenshot from 2020-04-19 09-33-13

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -229,9 +229,13 @@ export function useCallback(callback, args) {
  */
 export function useContext(context) {
 	const provider = currentComponent.context[context._id];
-	// We could skip this, but we need to call getHookState here, make
-	// the devtools aware of the hook.
+	// We could skip this call here, but than we'd not call
+	// `options._hook`. We need to do that in order to make
+	// the devtools aware of this hook.
 	const state = getHookState(currentIndex++, 9);
+	// The devtools needs access to the context object to
+	// be able to pull of the default value when no provider
+	// is present in the tree.
 	state._context = context;
 	if (!provider) return context._defaultValue;
 	// This is probably not safe to convert to "!"
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -6,6 +6,9 @@ let currentIndex;
 /** @type {import('./internal').Component} */
 let currentComponent;
 
+/** @type {number} */
+let currentHook = 0;
+
 /** @type {Array<import('./internal').Component>} */
 let afterPaintEffects = [];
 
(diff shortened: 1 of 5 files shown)


marvinhagemeister:
I seem to run into a few wrongly mangled properties. Need to investigate...

hooks/src/index.js
```diff
@@ -6,6 +6,9 @@ let currentIndex;
 /** @type {import('./internal').Component} */
 let currentComponent;
 
+/** @type {number} */
+let currentHook = 0;
```
developit, line 10: I'm wondering about the index association here: would it be feasible to instead set+pass a reference to the current hook function?

Preact/debug could then `import * from hooks` and do a reverse lookup to get the hook name:

```js
import { * as hooks } from 'preact/hooks';

options._hook = (comp, index, type) = {
  let hookName;
  for (let i in hooks) if (hooks[i]===type) {
    hookName = i;
    break;
  }
  // Etc
};
```

marvinhagemeister: In the current architecture the whole options handling is injected by the devtools. Previously we did this in `preact/debug`, but it made us less flexible regarding bug fixes. Whenever there is a bug in the devtools bridge we had to tell users to upgrade their Preact version which feels wrong to me.

Because it's all handled by the extension in a central place we can easily roll out fixes for all supported Preact versions.

(For everyone reading this: We discussed various alternatives in Slack but came to the conclusion that the approach in this PR is the best one so far. We'll move forward with this)

MaxmaxmaximusGitHub:
> Some of our hooks leverage other hooks to keep our size small and this PR will detect that and will pick the topmost "native" hook. So if useState calls useReducer internally, the hook will be identified as useState no matter what.

epic

PR #2481, opened 2020-04-19
Optimizing

jeroenouw:
* Reduced huge amount of npm vulnerabilities (by running npm audit fix to go from 288 to only 10 vulnerabilities)
* Reduced file size slightly by re-writing some if statements


JoviDeCroock:
Removing {} doesn’t reduce size the compiler takes care of this. Will check the updated deps but in theory they can’t hurt since they’re devdeps

marvinhagemeister:
Thanks for taking the time to make a PR :tada: As @JoviDeCroock said: The minifer (terser in our case) rewrites the if statements for us automatically, so there is no need for us to minify the source code :+1:

PR #2478, opened 2020-04-18
Child traversal update

JoviDeCroock:
This PR updates and golfs the way we traverse children. At the moment looking into how we can make our `child-diffing` smarter by reducing the amount of dom-operations needed for insertions and swaps.

Supersedes: https://github.com/preactjs/preact/pull/2348 (less git-history and more golfing)
Fixes: https://github.com/preactjs/preact/issues/2227

PR #2482, opened 2020-04-19
Correctly handle child insertions

JoviDeCroock:
There is atm one failing test since we removeone "stateful update" so one dom mutation less, this does handle some of the bugs we've been facing with focus loss and too much dom insertions.

Linearly this could check if we are dealing with a removed index, .... I  think this opens the possibility to remove some code.

Improves perf more than https://github.com/preactjs/preact/pull/2478

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -34,7 +34,13 @@ export function diffChildren(
 	oldDom,
 	isHydrating
 ) {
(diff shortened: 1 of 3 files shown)


JoviDeCroock:
Superseded by https://github.com/preactjs/preact/pull/2507

PR #2483, opened 2020-04-20
Remove processingException check

JoviDeCroock:
Due to the addition of

```
// More info about this here: https://gist.github.com/JoviDeCroock/bec5f2ce93544d2e6070ef8e0036e4e8
if (newVNode._original !== oldVNode._original) c._dirty = false;
```

We can safely ignore `processingException` since we won't stop diffing if the children are still potentially dirty

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -127,7 +127,7 @@ export function diff(
 							c._nextState,
 							componentContext
 						) === false) ||
-					(newVNode._original === oldVNode._original && !c._processingException)
+					newVNode._original === oldVNode._original
 				) {
 					c.props = newProps;
 					c.state = c._nextState;


developit:
Nice win! I'm trying to think of cases where this could be triggered but all I could come up with was something like this, where an intermediary component sets state but a descendant triggers suspend:

```js
class A {
  componentDidCatch() {}
  render() { return <B /> }
}
function B() {
  const [, invalidate] = useState();
  useLayoutEffect(invalidate, []);  // make dirty on mount
  return <C />
}
const C = () => throw new Promise(()=>{});
```

JoviDeCroock:
@developit this should just work since referential equality doesn't come up there because `createElement(B)` is always called

PR #2489, opened 2020-04-20
Add update benchmark from js-framework-benchmark

andrewiggins:
This implementation doesn't slowdown the CPU by 16x like js-framework-benchmark but I assume the bench is still useful/meaningful

### DIFF
--- a/benches/src/util.js
+++ b/benches/src/util.js
@@ -17,6 +17,27 @@ export function afterFrameAsync() {
 	return promise;
 }
 
+export function getRowIdSel(index) {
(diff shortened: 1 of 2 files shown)


benches/src/util.js
```diff
@@ -17,6 +17,27 @@ export function afterFrameAsync() {
 	return promise;
 }
 
+export function getRowIdSel(index) {
```
andrewiggins, line 20: I put these `getRowIdSel` and `getRowLinkSel`  functions in the util file cuz if we add more benches from js-framework-benchmark, they'll be used by multiple benches

Issue #2175, opened 2019-12-10
Error with undefined __hooks property


apfelbox:
Hi,

sorry for this confusing issue, but I haven't found anything so far:

I have the issue that `__hooks` is suddenly undefined:
![2019-12-10 at 17 19](https://user-images.githubusercontent.com/1032411/70547519-4f112700-1b71-11ea-9198-ecef24dd3581.png)

Both of the `forEach` before succeed, but here `__hooks` is suddenly undefined (ie it must be through a side effect of one the `invokeEffect` calls):

Error in console:
![2019-12-10 at 17 20](https://user-images.githubusercontent.com/1032411/70547602-75cf5d80-1b71-11ea-9bb4-e3853e5a5e8b.png)

My `hooks.module.js` (extended by some `console.log()` calls):
![2019-12-10 at 17 20](https://user-images.githubusercontent.com/1032411/70547649-8a135a80-1b71-11ea-8012-91f51651a86c.png)

I highlighted the one that is right before the error.

The component that is triggering this error seems innocent enough (it is not yet finished 😉):
<details>
<summary>Show screenshot</summary>

![2019-12-10 at 17 23](https://user-images.githubusercontent.com/1032411/70547847-e5454d00-1b71-11ea-8302-df016b5a06fe.png)

</details>

Happy to provide any additional info, if you could give me some pointers on where to look.

JoviDeCroock:
I tried reproducing your issue here but can't seem to figure out as to why it's failing https://codesandbox.io/s/brave-panini-i3iho

apfelbox:
Hi @JoviDeCroock,

thank you for taking a look!

I narrowed it down to this case: if I call `fetch.request()` in my effect it errors out, if I don't call it it works.

So, works:
![2019-12-10 at 19 00](https://user-images.githubusercontent.com/1032411/70555499-6bb45b80-1b7f-11ea-8be6-a9868cc9b967.png)

Doesn't work:
![2019-12-10 at 19 00](https://user-images.githubusercontent.com/1032411/70555476-5d663f80-1b7f-11ea-8b1e-2e53ad92710c.png)

I have no idea on how the hooks are implemented in preact, but do you know of anything that might lead to issues in combination with hooks – even if my callback doesn't even return anything (so just some side effects)?

marvinhagemeister:
I'm unable to create a reproducible scenario. Is there any way you can narrow it down further to a specific component? We'd love to help but without access to the quote we're shooting in the dark :S

[...]

PR #2493, opened 2020-04-21
Fix creating multiple roots from useEffect

JoviDeCroock:
Fixes: https://github.com/preactjs/preact/issues/2175

When we dispatch a new `render` from within an effect we have to restore our `currentComponent`.


developit:
You somehow managed to make this smaller lol

PR #2495, opened 2020-04-21
Added fix in preact/hooks. Failing on preact-render-to-string when us…

gunins:
When use  preact-render-to-string in service workers as `iife` Getting errors cannot set property __hooks of undefined.

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -21,9 +21,7 @@ const RAF_TIMEOUT = 100;
 let prevRaf;
 
 options._render = vnode => {
-	if (oldBeforeRender) {
-		oldBeforeRender(vnode);
-	}
+	if (oldBeforeRender) oldBeforeRender(vnode);
 
 	currentComponent = vnode._component;
 	currentIndex = 0;
@@ -36,14 +34,10 @@ options._render = vnode => {
 };
 
 options.diffed = vnode => {
-	if (oldAfterDiff) {
-		oldAfterDiff(vnode);
-	}
+	if (oldAfterDiff) oldAfterDiff(vnode);
 
 	const c = vnode._component;
-	if (!c) {
-		return;
-	}
+	if (!c) return;
 
 	const hooks = c.__hooks;
 	if (hooks) {
@@ -62,29 +56,21 @@ options._commit = (vnode, commitQueue) => {
 			);
 		} catch (e) {
 			commitQueue.some(c => {
-				if (c._renderCallbacks) {
-					c._renderCallbacks = [];
-				}
+				if (c._renderCallbacks) c._renderCallbacks = [];
 			});
 			commitQueue = [];
 			options._catchError(e, component._vnode);
 		}
 	});
 
-	if (oldCommit) {
-		oldCommit(vnode, commitQueue);
-	}
+	if (oldCommit) oldCommit(vnode, commitQueue);
 };
 
 options.unmount = vnode => {
-	if (oldBeforeUnmount) {
-		oldBeforeUnmount(vnode);
-	}
+	if (oldBeforeUnmount) oldBeforeUnmount(vnode);
 
 	const c = vnode._component;
-	if (!c) {
-		return;
-	}
+	if (!c) return;
 
 	const hooks = c.__hooks;
 	if (hooks) {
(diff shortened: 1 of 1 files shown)


JoviDeCroock:
There are a lot of unrelated changes in here

gunins:
Appologies, auto formating.

JoviDeCroock:
I'm not entirely sure how this is possible, what version of `preact-render-to-string` are you using? And what version of `Preact`?

gunins:
Latest

gunins:
Me too, was very surprised, nodejs with require was fine, but in service worker rollup, got error.

JoviDeCroock:
We'll have to see a reproduction since this might not be the best place to fix this

gunins:
Agree, this fix was working for me, but not sure if the best way to fix. But I really appreciate, if `preact-render-to-string` will work in browser too. Like I said working in nodejs environment.

JoviDeCroock:
It does work in browser, this isn't a fix relating to making it work in a browser. This is an underlying or usage issue I think.

developit:
Hiya. If your finding a situation where hooks don't have a value for `currentComponent`, there's no change in Preact that can fix that. This happens when a hook function is called by a function that isn't in the render path.

In the case you described, what's happening is that you're using `preact/hooks` in a way where it can't see the copy of Preact you're using with render-to-string.

Can you post a sample of how you are loading preact, preact/hooks and preact-render-to-string?

gunins:
That’s was in my mind. I try to create basic example, and publish in github. Or option two, try to figure out reason for error, and publish too.

gunins:
Hi! I figured out to create same issue in example. https://github.com/gunins/preact-test this one is compiled file https://github.com/gunins/preact-test/blob/master/dist/index.js
Just run this cile in browser, and check `console.log`
Also I have to admit, this could be edge case, because I solve by using `rollup-plugin-node-resolve` instead of `rollup-plugin-includepaths`,  but there is some issue, why direct include treating differently than compiled.

JoviDeCroock:
Why are you not calling createElement on App in your render initiation?

gunins:
Basically, I did different tests, in examples, and readme not mentioned, createElement is required. But added createElement https://github.com/gunins/preact-test/blob/master/src/index.ts, same error.

hooks/src/index.js
```diff
@@ -26,7 +26,7 @@ options._render = vnode => {
 	currentComponent = vnode._component;
 	currentIndex = 0;
 
-	if (currentComponent.__hooks) {
+	if (currentComponent !== undefined && currentComponent.__hooks) {
```
contedenis, line 29: You don't need to compare with undefined.
```
if (currentComponent && currentComponent.__hooks)
```

hooks/src/index.js
```diff
@@ -129,7 +130,7 @@ export function useState(initialState) {
 export function useReducer(reducer, initialState, init) {
 	/** @type {import('./internal').ReducerHookState} */
 	const hookState = getHookState(currentIndex++, 2);
-	if (!hookState._component) {
+	if (hookState !== undefined && !hookState._component) {
```
contedenis, line 133: same

gunins: Changed.

hooks/src/index.js
gunins: Changed

JoviDeCroock:
@contedenis I found the issue you are using our source from preact instead of the build output. This meant that the options hook wasn't getting called replacing

```
const include = {
	'preact-render-to-string': './node_modules/preact-render-to-string',
	'preact':       './node_modules/preact',
	'preact/hooks': './node_modules/preact/hooks'
}
```

Fixes it

contedenis:
> @contedenis I found the issue you are using our source from preact instead of the build output. This meant that the options hook wasn't getting called replacing
> 
> ```
> const include = {
> 	'preact-render-to-string': './node_modules/preact-render-to-string',
> 	'preact':       './node_modules/preact',
> 	'preact/hooks': './node_modules/preact/hooks'
> }
> ```
> 
> Fixes it

The issue was reported by @gunins, I just commented on his PR.

gunins:
> > @contedenis I found the issue you are using our source from preact instead of the build output. This meant that the options hook wasn't getting called replacing
> > ```
> > const include = {
> > 	'preact-render-to-string': './node_modules/preact-render-to-string',
> > 	'preact':       './node_modules/preact',
> > 	'preact/hooks': './node_modules/preact/hooks'
> > }
> > ```
> > 
> > 
> > Fixes it
> 
> The issue was reported by @gunins, I just commented on his PR.
Actually this fix still breaking build, but I get the point, building by using source will fail.
Thanks!

JoviDeCroock:
@gunins can we close this PR or are you still seeing issues?

gunins:
Yes you can close. Will be helpful, if in `preact-render-to-string` documentation put note, doesn't work from source, need to use compiled version. Thanks again!

JoviDeCroock:
@gunins see linked PR https://github.com/preactjs/preact/pull/2497

PR #2497, opened 2020-04-22
Remove source code from npm-tarball

JoviDeCroock:
This removes the source code from our npm-publishes since this on one hand confuses typescript when auto-importing in for instance vscode and when used can confuse Preact (or lead to two instances of Preact)

Resolves the issue of https://github.com/preactjs/preact/pull/2495


package.json
```diff
@@ -151,24 +150,29 @@
     }
   },
   "files": [
-    "src",
+    "src/index.d.ts",
+    "src/internal.d.ts",
```
andrewiggins, line 154: Should we leave out the internal types to avoid encouraging relying on internals that may break? Further all the internal property names are different than these types define

developit: if nothing currently uses them that seems good.

package.json
```diff
@@ -151,24 +150,29 @@
...
+    "src/jsx.d.ts",
     "dist",
+    "compat/src/index.d.ts",
```
developit, line 157: `"files"` supports glob matches, so I think we can do:

```js
  "files": [
    "**/*.d.ts",
    "**/dist",
    "**/package.json",
    "compat/server.js"
  ],
```

andrewiggins:
When I run `npm pack` in this branch it includes all the demo node_modules folder which is strange... Might need to restrict the globs a bit more? Perhaps exclude the `demo` and `bench` folders or something?

JoviDeCroock:
Superseded by https://github.com/preactjs/preact/pull/2575

PR #2503, opened 2020-04-25
Devtools: Check if hook is present in parent window

marvinhagemeister:
This PR completes support for debugging Preact apps that are rendered inside an `iframe`. If we don't find the devtools hook in the current frame, we check for it on the parent window. If a window does not have a parent, its parent property is a reference to itself.

Screenshot from 2020-04-25 09-59-51

### DIFF
--- a/devtools/src/devtools.js
+++ b/devtools/src/devtools.js
@@ -1,8 +1,12 @@
 import { options, Fragment, Component } from 'preact';
 
 export function initDevTools() {
-	if (typeof window != 'undefined' && window.__PREACT_DEVTOOLS__) {
-		window.__PREACT_DEVTOOLS__.attachPreact('10.4.1', options, {
+	const hook =
+		typeof window != 'undefined' &&
+		(window.__PREACT_DEVTOOLS__ || window.parent.__PREACT_DEVTOOLS__);
+
+	if (hook) {
+		hook.attachPreact('10.4.1', options, {
 			Fragment,
 			Component
 		});


devtools/src/devtools.js
```diff
@@ -1,8 +1,12 @@
...
 
 export function initDevTools() {
-	if (typeof window != 'undefined' && window.__PREACT_DEVTOOLS__) {
-		window.__PREACT_DEVTOOLS__.attachPreact('10.4.1', options, {
+	const hook =
+		typeof window != 'undefined' &&
+		(window.__PREACT_DEVTOOLS__ || window.parent.__PREACT_DEVTOOLS__);
+
+	if (hook) {
+		hook.attachPreact('10.4.1', options, {
```
ForsakenHarmony, line 9: kinda wondering if we could import the version from `package.json` and then inline during build, but not relevant for now

devtools/src/devtools.js
```diff
@@ -1,8 +1,12 @@
 import { options, Fragment, Component } from 'preact';
 
 export function initDevTools() {
-	if (typeof window != 'undefined' && window.__PREACT_DEVTOOLS__) {
-		window.__PREACT_DEVTOOLS__.attachPreact('10.4.1', options, {
+	const hook =
+		typeof window != 'undefined' &&
+		(window.__PREACT_DEVTOOLS__ || window.parent.__PREACT_DEVTOOLS__);
```
developit, line 6: Think it's be worth also checking window.top here? Not sure how common it is to load doubly-nested iframes.

marvinhagemeister:
Marking as draft again until I've figured out how to make it work with Codesandbox. There the `iframe` is boxed off and can't get access to the parent window to attach to the extension listener.

developit:
@marvinhagemeister I wonder if it would be worth trying to have the browser plugin inject it's init code into each frame? I seem to recall an API for this from ages ago.

marvinhagemeister:
@developit agree, that's a much better solution! Added proper support for iframes (and codesandbox :100: ) in https://github.com/preactjs/preact-devtools/pull/209

Closing this PR as it is not necessary anymore.

PR #2507, opened 2020-04-27
Iteratively diff children and convert array children to Fragments

### LINKED ISSUE #2446: Focus displacement when components are added to a list above
### Reproduction

I've created a small sample that appears to me to have a focus handling bug.

https://codesandbox.io/s/preact-focus-boxes-m15qd

```ts
import { useState, useCallback, useRef } from "preact/hooks";

function App() {
  const boxNum = useRef(1);
  const [boxes, setBoxes] = useState([1]);

  const boxComponents = boxes.map(num => {
    return <div>box</div>;
  });

  const onAddBoxToEnd = useCallback(() => {
    boxNum.current++;
    setBoxes(boxes => [...boxes, boxNum.current]);
  }, []);
  const onAddBoxToFront = useCallback(() => {
    boxNum.current++;
    setBoxes(boxes => [boxNum.current, ...boxes]);
  }, []);

  return (
    <div>
      {boxComponents}
      <br />
      <button onClick={onAddBoxToFront}>Add box to front</button>
      <button onClick={onAddBoxToEnd}>Add box to end</button>

      <p>
        Place focus on either button and press space. Focus leaves the button,
        appearing to be offset due to the element changes above.
      </p>
    </div>
  );
}

if (typeof window !== "undefined") {
  render(<App />, document.getElementById("root"));
}
```

### Steps to reproduce

In the demo above:

1. Place focus on the second button.
2. Press spacebar.

### Expected Behavior

Expected focus to remain on the button that was activated, since it did not unmount.

### Actual Behavior

Focus moved to the prior button. It appears focus is offset by the number of elements inserted above (1) in the example.

The first button also has the issue, but it's less clear to see where focus goes, because there's no prior elements with focus indication.

andrewiggins:
At the suggestion from @developit from forever ago, this PR no longer uses `toChildArray` in diffChildren to convert `props.children` to an array. It also no longer flattens nested arrays when diffing children. It now converts them to Fragments.

I think this change will help us improve diff perf in the future and it also fixes #2446.

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -63,6 +63,18 @@ export function diffChildren(
 		// in a `if (childVNode) { ... } condition
 		if (childVNode == null) continue;
 
+		// If this newVNode is being reused (e.g. <div>{reuse}{reuse}</div>) in the same diff,
+		// copy it so it can have it's own DOM & etc. pointers
+		if (childVNode._dom != null || childVNode._component != null) {
+			childVNode = newParentVNode._children[i] = createVNode(
+				childVNode.type,
+				childVNode.props,
+				childVNode.key,
+				null,
+				childVNode._original
+			);
+		}
+
 		childVNode._parent = newParentVNode;
 		childVNode._depth = newParentVNode._depth + 1;
 
@@ -253,16 +265,6 @@ export function toChildArray(children, internal, flattened) {
 		flattened.push(children);
 	} else if (typeof children == 'string' || typeof children == 'number') {
 		flattened.push(createVNode(null, children, null, null, children));
-	} else if (children._dom != null || children._component != null) {
-		flattened.push(
-			createVNode(
-				children.type,
-				children.props,
-				children.key,
-				null,
-				children._original
-			)
-		);
 	} else {
 		flattened.push(children);
 	}
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -275,7 +275,6 @@ export function diffChildren(
  * Flatten and loop through the children of a virtual node
  * @param {import('../index').ComponentChildren} children The unflattened
  * children of a virtual node
- * @param {Array<import('../internal').VNode | string | number>} [flattened] An flat array of children to modify
  * @returns {import('../internal').VNode[]}
  */
 export function toChildArray(children) {
@@ -283,7 +282,7 @@ export function toChildArray(children) {
 		return [];
 	} else if (Array.isArray(children)) {
 		return EMPTY_ARR.concat.apply([], children.map(toChildArray));
-	} else {
-		return [children];
 	}
+
+	return [children];
 }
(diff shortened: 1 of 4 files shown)


src/diff/children.js
```diff
@@ -56,162 +58,196 @@ export function diffChildren(
...
+	newParentVNode._children = [];
+	for (i = 0; i < renderResult.length; i++) {
+		childVNode = renderResult[i];
 
-				// Check if we find a corresponding element in oldChildren.
-				// If found, delete the array item by setting to `undefined`.
-				// We use `undefined`, as `null` is reserved for empty placeholders
-				// (holes).
-				oldVNode = oldChildren[i];
+		if (childVNode == null || typeof childVNode == 'boolean') {
```
andrewiggins, line 65: This big if statement is the internal VNode coercion that `toChildArray` did. Terser converts into a massive ternary for us :)

src/diff/children.js
```diff
@@ -239,39 +275,15 @@ export function diffChildren(
...
  * children of a virtual node
- * @param {(vnode: import('../internal').VNode) => import('../internal').VNode} [callback]
- * A function to invoke for each child before it is added to the flattened list.
  * @param {Array<import('../internal').VNode | string | number>} [flattened] An flat array of children to modify
  * @returns {import('../internal').VNode[]}
  */
-export function toChildArray(children, callback, flattened) {
-	if (flattened == null) flattened = [];
-
+export function toChildArray(children) {
```
andrewiggins, line 281: Since this function is no longer used in the diff hot path, I rewrote it to optimize for size instead of perf/memory. Let me know what you think. The rewrite saved us about 40 bytes

src/diff/children.js
```diff
@@ -56,162 +58,196 @@ export function diffChildren(
...
+		// it's own DOM & etc. pointers
+		else if (typeof childVNode == 'string' || typeof childVNode == 'number') {
+			childVNode = newParentVNode._children[i] = createVNode(
+				null,
+				childVNode,
+				null,
+				null,
+				childVNode
+			);
+		} else if (Array.isArray(childVNode)) {
```
andrewiggins, line 79: Putting Array.isArray into a utility function added a byte. Using a `.pop` property check saves us some bytes (~10) but I worry that the property access will quickly de-opt and become megamorphic since it'll sometimes be a VNode and sometimes by an array. So I think imma leave this for now. Can revisit later

marvinhagemeister: Good point! iirc checking for `.pop()` is much slower than `Array.isArray()` 👍

developit: Yep. It used to be the other way around. Not sure about other engines.  

https://esbench.com/bench/57f1327e330ab09900a1a1d5

ForsakenHarmony: firefox

Jane486:
Спасибо! Но мне интересен только перевод этих всех страниц....
--
Отправлено из Mail.ru для Android среда, 29 апреля 2020г., 21:47 +03:00 от Marvin Hagemeister  notifications@github.com :

Issue #2512, opened 2020-05-02
render fails in iOS 7 Safari - TypeError: 'undefined' is not a function (evaluating 'Object.setPrototypeOf(n,b)')


kenfoo:
### Reproduction

A simple code of `render(<div>hello</div>, document.body)` will fail on iOS 7 Safari with Preact 10.4.1. I believe I wasn't having this issue with an earlier build of Preact X which I tested some time back.

The error that I see from Safari remote debugging is
"TypeError: 'undefined' is not a function (evaluating 'Object.setPrototypeOf(n,b)')"

Seems to work fine in newer iOS Safari.

kenfoo:
https://github.com/preactjs/preact/commit/3c6cb62b551c427dedc887fb02b63a7548e8ca88 seems to be the cause

marvinhagemeister:
Looks like this version of Safari doesn't support `Object.setPrototypeOf`. Does this little polyfill resolve it?

```js
Object.setPrototypeOf = Object.setPrototypeOf || function (obj, proto) {
  obj.__proto__ = proto;
  return obj;
}
```

kenfoo:
Yes, that fixes it. Thanks!

PR #2514, opened 2020-05-02
polyfill prototype setter in debug

JoviDeCroock:
Fixes https://github.com/preactjs/preact/issues/2512


debug/src/debug.js
```diff
@@ -14,6 +14,14 @@ import {
 
 const isWeakMapSupported = typeof WeakMap == 'function';
 
+Object.setPrototypeOf =
```
developit, line 17: Might be safer to ponyfill this?

Issue #2515, opened 2020-05-02
Circular references resulting from h


edemaine:
### Reproduction

```js
require('preact').h('img')
```

results in

```
{
  type: 'img',
  props: {},
  key: undefined,
  ref: undefined,
  __k: null,
  __: null,
  __b: 0,
  __e: null,
  __d: undefined,
  __c: null,
  constructor: undefined,
  __v: [Circular]
}
```

This structure has a circular pointer (I guess made by `createVNode` in [this line](https://github.com/preactjs/preact/blob/master/src/create-element.js#L82)), which makes it not JSON serializable.

This used to work fine on Preact 10.1.1, but has a circular dependency in Preact 10.4.1.

### Expected Behavior

It would be really nice if VNodes were JSON serializable.  I use it in [this project](https://github.com/edemaine/svgtiler/) to detect equality among many different data types (strings, VNodes, etc.).

I imagine this circular reference could also lead to memory leaks...

developit:
Just a note: Preact's VNodes are technically not JSON-serializable, since they're required to have an undefined `constructor` value. That being said, it sounds like you've got a case where that's not applicable - you're only serializing for the purposes of comparison and aren't then deserializing.

There's an ongoing discussion of what to do about `_original` / `__v`. It provides some important performance optimizations, though it may be possible for us to achieve those without using an full self-reference.

edemaine:
@developit Thanks for the heads up! I'm not sure what performance optimization that enables, but I'm not very familiar with Preact's codebase. Without knowing the motivation, circular references seem like they ought to be avoided... at least when just calling `h`. (In my context, all I'm doing is using `preact-render-to-string` to render directly to HTML, or in my case SVG, and not using an actual living VDOM.)

But assuming circular references really are optimal: I guess it's not really crucial that I can serialize; in my setting, I could use `preact-render-to-string` and compare those strings. Perhaps there should be some official API for destroying a VNode, though, to enable garbage collection?

developit:
All current JS engines (including the last few IE versions) use mark-and-sweep GC, which handles circular references without leaks ([more info](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Memory_Management#Cycles_are_no_longer_a_problem)).

[...]

PR #2517, opened 2020-05-02
remove circular ref and golf implementation

JoviDeCroock:
Fixes https://github.com/preactjs/preact/issues/2515


src/diff/index.js
```diff
@@ -208,12 +208,6 @@ export function diff(
 			}
 
 			c._force = false;
-		} else if (
-			excessDomChildren == null &&
-			newVNode._original === oldVNode._original
```
JoviDeCroock, line 213: This condition should never occur for non-function children

src/create-element.js
```diff
@@ -79,7 +79,7 @@ export function createVNode(type, props, key, ref, original) {
 		_original: original
 	};
 
-	if (original == null) vnode._original = vnode;
+	if (original == null) vnode._original = {};
```
andrewiggins, line 82: Could we use `EMPTY_OBJ` here?

JoviDeCroock: no then every vnode would be equal and bail out of every render

src/diff/index.js
andrewiggins: This might be useful for people who want to declare a chunk of static markup outside of render once (perhaps using that static vnode hoisting babel plugin?) and just re-use the vnodes, but perhaps this is better done by things like htm flags, etc. Either I don't have a strong preference atm.

marvinhagemeister (approved):

src/diff/index.js
developit: FWIW the JSX hoisting thing is super common, it's enabled by default in `@babel/preset-react`.

developit:
Can we try using an integer counter here instead of the objects? That would get rid of the 7% memory usage increase.

jsx-runtime/src/index.js
```diff
@@ -35,11 +35,10 @@ function createVNode(type, props, key, __source, __self) {
 		_component: null,
 		_hydrating: null,
 		constructor: undefined,
-		_original: undefined,
+		_original: Math.random(),
```
developit, line 38: This will transition the field from SMI

JoviDeCroock: What do you mean?

mangle.json
```diff
@@ -25,6 +25,7 @@
   "props": {
     "cname": 6,
     "props": {
+      "$_vnodeId": "__i",
```
developit, line 28: A thought: we could use `.__v` since it's not used on options currently, and then the property would nicely match `._original`.

JoviDeCroock:
moving it to options seems to have hurt perf quite a bit 😅

jsx-runtime/src/index.js
marvinhagemeister: SMI = small integer, it's a data type inside v8 https://v8.dev/blog/elements-kinds

PR #2534, opened 2020-05-14
Allow cloneElement to create copies of `ComponentChild`

### LINKED ISSUE #2428: TS error cloneElement children
### Reproduction

https://codesandbox.io/s/strange-wilson-w6k4p

### Steps to reproduce

- Use `preact@10.3.4`
- Pass `props.children` to `cloneElement`.

### Expected Behavior

As per the [documentation example](https://preactjs.com/guide/v10/api-reference/#cloneelement), this should be valid. Thus, no TS errors should occur.

### Actual Behavior

TS reports the following error:

```
  Overload 1 of 2, '(vnode: VNode<any>, props?: any, ...children: ComponentChildren[]): VNode<any>', gave the following error.
    Argument of type 'ComponentChildren' is not assignable to parameter of type 'VNode<any>'.
      Type 'undefined' is not assignable to type 'VNode<any>'.
  Overload 2 of 2, '(vnode: VNode<any>, props?: any, ...children: ComponentChildren[]): VNode<any>', gave the following error.
    Argument of type 'ComponentChildren' is not assignable to parameter of type 'VNode<any>'.
      Type 'undefined' is not assignable to type 'VNode<any>'.ts(2769)
```

ddayguerrero:
Fixes https://github.com/preactjs/preact/issues/2428

Update typings for `cloneElement` to create copies of `ComponentChild`, namely `VNode` and primitives.

## What approach did I use?

Most of these changes involve type definitions. Given `ComponentChild` can be:

```typescript
type ComponentChild = VNode<any> | object | string | number | boolean | null | undefined;
```

- We overload `cloneElement`'s function signature in accept `ComponentChild`
- If the child is singular and not a VNode, we simple return create and return its copy.

On a side note, React's [ReactChild](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L237) is defined as follows:
`type ReactChild = ReactElement | ReactText;`. `cloneElement` should work with `string` and `number` at the very least.

Also, I think this still feels a bit weird — I'm wondering if using type inference / casting [as mentioned](https://github.com/preactjs/preact/issues/2428#issuecomment-606556737) in the issue should be enough to fix this issue.

### DIFF
--- a/src/clone-element.js
+++ b/src/clone-element.js
@@ -7,9 +7,13 @@ import { createVNode } from './create-element';
  * @param {import('./internal').VNode} vnode The virtual DOM element to clone
  * @param {object} props Attributes/props to add when cloning
  * @param {Array<import('./index').ComponentChildren>} rest Any additional arguments will be used as replacement children.
- * @returns {import('./internal').VNode}
+ * @returns {import('./internal').VNode|import('./index').ComponentChildren}
  */
 export function cloneElement(vnode, props) {
+	if (!vnode.type && !vnode._children) {
+		const clone = vnode;
+		return clone;
+	}
 	props = assign(assign({}, vnode.props), props);
 	if (arguments.length > 2) props.children = EMPTY_ARR.slice.call(arguments, 2);
 	let normalizedProps = {};
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -228,6 +228,11 @@ declare namespace preact {
 		parent: Element | Document | ShadowRoot | DocumentFragment
 	): void;
 	function cloneElement(
+		vnode: preact.ComponentChild,
+		props?: any,
+		...children: ComponentChildren[]
+	): VNode<any>;
+	function cloneElement(
 		vnode: VNode<any>,
 		props?: any,
 		...children: ComponentChildren[]
--- a/test/ts/VNode-test.tsx
+++ b/test/ts/VNode-test.tsx
@@ -9,6 +9,7 @@ import {
 	ComponentFactory,
 	VNode,
 	ComponentChildren,
+	ComponentChild,
 	cloneElement
 } from '../../src';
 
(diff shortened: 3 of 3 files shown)


src/clone-element.js
```diff
@@ -7,9 +7,13 @@ import { createVNode } from './create-element';
  * @param {import('./internal').VNode} vnode The virtual DOM element to clone
  * @param {object} props Attributes/props to add when cloning
  * @param {Array<import('./index').ComponentChildren>} rest Any additional arguments will be used as replacement children.
- * @returns {import('./internal').VNode}
+ * @returns {import('./internal').VNode|import('./index').ComponentChildren}
  */
 export function cloneElement(vnode, props) {
+	if (!vnode.type && !vnode._children) {
```
JoviDeCroock, line 13: Is this meant for supporting `text` vnodes? What's making this fail in the current scenario? Since this code isn't actually cloning it's still returning an equal reference making this node never render

src/index.d.ts
```diff
@@ -227,6 +227,11 @@ declare namespace preact {
 		vnode: ComponentChild,
 		parent: Element | Document | ShadowRoot | DocumentFragment
 	): void;
+	function cloneElement(
+		vnode: preact.ComponentChild,
```
38elements, line 231: Is it possible to set `element` argument of `cloneElement()` in React to `string` or `number` or `boolean` or `null` or `undefined`?

src/clone-element.js
ddayguerrero: It would be meant for handling cases where `vnode` is of type `object | string | number | boolean | null | undefined`. 

Not complete sure... prior to this, I think passing children of those types would create an incomplete `vnode`.

```jsx
import { render, cloneElement } from 'preact';

const Foo = (props) => <div>{props.children}</div>;
const myFoo = <Foo>test</Foo>;
const clonedFoo = cloneElement(myFoo.props.children);
console.log(clonedFoo);
// type: undefined, props: {…}, key: undefined, ref: undefined, __k: null, …}
```

and you're right we would want to return the reference 😅

JoviDeCroock: This feels correct `// type: undefined, props: {…}, key: undefined, ref: undefined, __k: null, …}` because that's a textual vnode

src/index.d.ts
ddayguerrero: Haven't tried it in React but in React + TS, I think one is able to pass in anything to `cloneElement(element)` given that `element` is `props.children`. 

Although directly cloning one of those primitives is not possible (e.g. `cloneElement("my string")`)

src/clone-element.js
ddayguerrero: Thanks for the explanation, if I understand correctly (please correct me haha), then should changes should be done inside `createVNode`? or should we change how we reassign `props` within `cloneElement`?

src/index.d.ts
38elements: `createElement()` returns only `VNode`. 
But, this `cloneElement()` handles `VNode` and others. 
This is inconsistance of function name.

src/clone-element.js
JoviDeCroock: The latter sounds most correct to me, when we see the `type = null` we can just do `props = prevProps`

JoviDeCroock:
Closing: draft `cloneElement`/`ComponentChild` typing against the old d.ts overloads. The underlying request (#2428) stays open; better to retype fresh on v11.

PR #2537, opened 2020-05-16
chore: Add preact-render-to-string as peerdep

wardpeet:
I might have missed something but `preact-render-to-string` is used in `preact/compat/server.js` it should be added to interdependencies if it can't be a direct dependency.

https://github.com/preactjs/preact/blob/da382e13d9377a53056e4cb0fd741f6e0aadf1c1/compat/server.js#L3-L9

### DIFF
--- a/compat/package.json
+++ b/compat/package.json
@@ -14,6 +14,7 @@
     "regex": "^_"
   },
   "peerDependencies": {
-    "preact": "^10.0.0"
+    "preact": "^10.0.0",
+    "preact-render-to-string": "^5.1.8"
   }
 }


compat/package.json
```diff
@@ -14,6 +14,7 @@
     "regex": "^_"
   },
   "peerDependencies": {
-    "preact": "^10.0.0"
+    "preact": "^10.0.0",
+    "preact-render-to-string": "^5.1.8"
```
wardpeet, line 18: Unsure if I should lower to 5.0.0
```suggestion
    "preact-render-to-string": "^5.0.0"
```

wardpeet: Maybe an optionalDependency is better? I don't know :)

developit:
the package.json file at `preact/compat/package.json` never gets used for installation because it's not in the package root. I think if there's now a consistent way to denote `optional` dependencies we could add `"preact-render-to-string":"*"` to the root package.json. I recall Yarn and npm finally agreed on a syntax for optional.

wardpeet:
Oh yeah stupid me, I'll move it to optional in the root

wardpeet:
Optional dependencies still get installed, if it fails it won't crash but that doesn't seem what we want.

PR #2546, opened 2020-05-21
3kB -> 4kB

brunolemos:
4kB is still pretty good 👌

[image](https://bundlephobia.com/result?p=preact)

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 </a>
 </p>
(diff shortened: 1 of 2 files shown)


porfirioribeiro:
Well, one can say it's still 3Kb, as it does not passed the 3.something size

brunolemos:
If the main marketing point is the size and the real number is 30% more than advertised, it's a misleading message.

But like I said, 4kB is still pretty good. No harm in saying it's a 4kB React alternative.

PR #2548, opened 2020-05-21
Use "babel-plugin-transform-rename-properties" to consistently mangle properties

andrewiggins:
Use @jviide's `babel-plugin-transform-rename-properties` plugin to consistently mangle property names instead of terser. The benefit of doings this is that it enables us to produce unminified builds whose mangled properties are consistent with the minified builds. Another benefit is that now all builds share one mapping of property names to mangle names.

I manually compared the CommonJS (CJS) output of the builds of master and this branch to verify no properties changed. Only local variables and local function names changed for core and hooks. This check actually lead me to find a bug (as well as the test failing):

The babel plugin only mangles property names, not import/export names. At the time babel runs, the import/export statements are still import/export statements so the babel plugin does not change them. However when terser runs, the import/export statements become `exports._unmount = ...` and `preact._unmount(...)` and so terser could mangle these. This difference tripped up the one mangled export we have, `_unmount` in core which `compat` consumes for portals. To fix this, I manually exported and imported the desired mangled name.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -30,19 +30,11 @@ module.exports = function(api) {
 		plugins: [
 			'@babel/plugin-proposal-object-rest-spread',
 			'@babel/plugin-transform-react-jsx',
-			'babel-plugin-transform-async-to-promises'
+			'babel-plugin-transform-async-to-promises',
+			['babel-plugin-transform-rename-properties', { rename }]
 		],
 		include: ['**/src/**/*.js', '**/test/**/*.js'],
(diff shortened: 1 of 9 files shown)


developit:
This is a really good improvement, and I think we should consider upstreaming it to Microbundle itself.

FWIW I actually prefer the always-mangled import change (_unmount -> __u). To me that clarifies that when we export mangled names, our API is explicitly the mangled version, never the unmangled version.

Issue #2527, opened 2020-05-10
Bug with Fragment diff


webschik:
### Reproduction

CodeSandbox: https://codesandbox.io/s/preact-frament-diff-bug-g4u0h?file=/src/index.jsx

### Steps to reproduce

1. Go to the [CodeSandbox link](https://codesandbox.io/s/preact-frament-diff-bug-g4u0h?file=/src/index.jsx)

2. Initially, you will see a red-bordered panel with a label "bottom panel".
Click on the "Toggle panel position" button several times.

### Expected Behavior

The red panel is rendered before or after `props.children` depending on the state.

### Actual Behavior

After several state changes, the "bottom" panel is rendered inside the child tree instead of being rendered after.
It's reproducible in `preact` and `preact/compat`

developit:
Looks like this is a bug introduced in `10.4.0` - it works correctly in `10.3.3`.

PR #2551, opened 2020-05-24
Fragment diff issue

JoviDeCroock:
Fixes https://github.com/preactjs/preact/issues/2527
The issue at hand: when updating components with children who are strictly equal or implement SCU they could have reordered due to the parent rendering. This works when the children are just DOM-nodes but becomes an issue when these children are Fragments or nested Fragments. We need to update `_nextDom` so the parents `diffChildren` can know where to continue after it has diffed this child, so for instance when we see the following:  

```jsx
const Child = () => (
  <Fragment>
    <div>1</div>
    <div>2</div>
  </Fragment>
);

const Parent = () => (
  <div>
    <Child />
    <div />
  </div>
);
```

  The `_nextDom` property should signal to `diffChildren(Parent)` that after the diff of `children.Child` it has to place the `<div />` after the last dom-property wrapped by the `Fragment` in this case `<div>2</div>`, but what happens if our `Child` bails out of render, and returns `props.children`? These children could be reordered, … by the Parent component. This means that our dom-reference could be stale.  When we aren’t bailing out of render the `placeChild` function handles this by tracking `_nextDom` and reordering everything however when we bail out there is no such logic. For this we abstracted out the `placeChild` logic and reused it in `SCU/strict-equality bailout`, this deeply traverses the DOM when it sees that it isn’t equal and does the placement. This is still an optimized path as compared to non-scu/non-bailout.

I guess we could introduce better bailouts but up to now I'm not entirely sure how yet.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -19,39 +19,38 @@ const reorderChildren = (newVNode, oldDom, parentDom) => {
 		lastVNodeChild &&
 		lastVNodeChild.type &&
 		lastVNodeChild._dom &&
-		lastVNodeChild._dom.isConnected
+		lastVNodeChild._dom.isConnected &&
+		newVNode._nextDom !== newVNode._dom
 	) {
-		if (newVNode._nextDom !== newVNode._dom) {
-			newVNode._children.forEach(vnode => {
-				if (vnode) {
-					if (typeof vnode.type == 'function') {
-						reorderChildren(vnode, oldDom, parentDom);
-					}
-
-					oldDom = placeChild(
-						newVNode,
-						parentDom,
-						vnode,
-						vnode,
-						newVNode._children,
-						null,
-						vnode._dom,
-						oldDom
-					);
-
-					if (typeof newVNode.type == 'function') {
-						// Because the newParentVNode is Fragment-like, we need to set it's
-						// _nextDom property to the nextSibling of its last child DOM node.
-						//
-						// `oldDom` contains the correct value here because if the last child
-						// is a Fragment-like, then oldDom has already been set to that child's _nextDom.
-						// If the last child is a DOM VNode, then oldDom will be set to that DOM
-						// node's nextSibling.
-						newVNode._nextDom = oldDom;
-					}
+		newVNode._children.forEach(vnode => {
+			if (vnode) {
+				if (typeof vnode.type == 'function') {
+					reorderChildren(vnode, oldDom, parentDom);
 				}
-			});
-		}
+
+				oldDom = placeChild(
+					newVNode,
+					parentDom,
+					vnode,
+					vnode,
+					newVNode._children,
+					null,
+					vnode._dom,
+					oldDom
+				);
+
+				if (typeof newVNode.type == 'function') {
+					// Because the newParentVNode is Fragment-like, we need to set it's
+					// _nextDom property to the nextSibling of its last child DOM node.
+					//
+					// `oldDom` contains the correct value here because if the last child
+					// is a Fragment-like, then oldDom has already been set to that child's _nextDom.
+					// If the last child is a DOM VNode, then oldDom will be set to that DOM
+					// node's nextSibling.
+					newVNode._nextDom = oldDom;
+				}
+			}
+		});
 	}
 };
 
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -21,7 +21,6 @@ const reorderChildren = (newVNode, oldDom, parentDom) => {
 		lastVNodeChild._dom &&
 		lastVNodeChild._dom.isConnected
 	) {
-		// newVNode._nextDom = getDomSibling(lastVNodeChild);
 		if (newVNode._nextDom !== newVNode._dom) {
 			newVNode._children.forEach(vnode => {
 				if (vnode) {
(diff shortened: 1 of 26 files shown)


src/diff/index.js
```diff
@@ -140,9 +140,23 @@ export function diff(
...
+
+					if (
+						lastVNodeChild &&
+						lastVNodeChild.type &&
+						lastVNodeChild._dom &&
+						lastVNodeChild._dom.isConnected
+					) {
+						newVNode._nextDom = getDomSibling(lastVNodeChild);
+						if (newVNode._nextDom !== newVNode._dom) {
+							// TODO: traverse children
```
JoviDeCroock: As researched by @andrewiggins wee should traverse the children here and apply the logic for `diffChildren` to assess if we're ordering correctly

src/diff/index.js
```diff
@@ -140,9 +140,25 @@ export function diff(
...
+					if (
+						lastVNodeChild &&
+						lastVNodeChild.type &&
+						lastVNodeChild._dom &&
+						lastVNodeChild._dom.isConnected
+					) {
+						newVNode._nextDom = getDomSibling(lastVNodeChild);
+						if (newVNode._nextDom !== newVNode._dom) {
+							newVNode._children.forEach(() => {
+								// Reorder if needed
```
JoviDeCroock, line 160: As researched by @andrewiggins wee should traverse the children here and apply the logic for diffChildren to assess if we're ordering correctly

test/browser/lifecycles/shouldComponentUpdate.test.js
```diff
@@ -764,4 +764,65 @@ describe('Lifecycle methods', () => {
 		rerender();
 		expect(scratch.innerHTML).to.equal('<p>bar</p>');
 	});
+
+	it('should reorder non-updating nested Fragment children', () => {
```
JoviDeCroock, line 768: Will fail until we find a performant solution to https://github.com/preactjs/preact/pull/2551/files#r439372897

src/diff/index.js
```diff
@@ -1,11 +1,40 @@
...
+			newVNode._children[tmp]._parent = newVNode;
+			lastVNodeChild = newVNode._children[tmp];
+		}
+	}
+
+	if (
+		lastVNodeChild &&
+		lastVNodeChild.type &&
+		lastVNodeChild._dom &&
+		lastVNodeChild._dom.isConnected
```
developit, line 22: What's the reason for checking `isConnected` here? I worry because I think folks sometimes call render() with a parent element that is not yet connected, so the whole tree would have `isConnected=false`.

src/diff/index.js
```diff
@@ -1,11 +1,40 @@
...
+	) {
+		newVNode._nextDom = getDomSibling(lastVNodeChild);
+		if (newVNode._nextDom !== newVNode._dom) {
+			newVNode._children.forEach(vnode => {
+				if (typeof vnode.type == 'function') {
+					reorderChildren(vnode);
+					if (vnode._dom !== vnode._nextDom) {
+						// TODO: we should now correctly change the ordering for this node.
```
JoviDeCroock, line 30: Here we should look if the `lastVNodeChild` is still accurate, due to `children` ordering, ... This is what's tested [here](https://github.com/preactjs/preact/pull/2551/files#diff-cb37cc3064015f47cc4563ffd3b50aa9R768) this should happen deeply. With the least amount of effort

src/diff/index.js
andrewiggins: I added this check in one of my explorations to fix some failing suspense tests though I don't understand why they were failing and why this fixed them. So all that to say, yes, this line deserves further investigation

andrewiggins:
Nice description - thanks for writing that up!

I think much the size savings come from converting `reorderChildren` to a function declaration. The results of compressed-size-action are the same for [remove const in favor of let](https://github.com/preactjs/preact/runs/776667848) and [rename more const to let](https://github.com/preactjs/preact/runs/776680984) (I'm looking at the "Size Differences" section which I think compares that size of the commit to master).

Should we revert 8fca6c9 if it doesn't actually change the size? Just thinking we shouldn't give contributors the wrong idea about let/const unless it does affect size.

JoviDeCroock:
Yes, we should probably rever that one

src/diff/index.js
```diff
@@ -1,11 +1,39 @@
...
 import { diffProps, setProperty } from './props';
 import { assign, removeNode } from '../util';
 import options from '../options';
 
+function reorderChildren(newVNode, oldDom, parentDom) {
+	for (let tmp = 0; tmp < newVNode._children.length; tmp++) {
+		if (newVNode._children[tmp]) {
+			let vnode = newVNode._children[tmp];
+			vnode._parent = newVNode;
+			if (vnode._dom && vnode._dom.isConnected) {
```
andrewiggins, line 14: I think we can remove this `isConnected` now. I'm not sure why I added it in the first place lol 🤷‍♂️ Tests are passing locally for me without it.

```suggestion
			if (vnode._dom) {
```

developit: Glad we're not relying on this anymore - it isn't supported in IE11 anyway.

PR #2552, opened 2020-05-25
put rerenderCount on the Component prototype

JoviDeCroock:
This allows for dev libs to do

```
options.debounceRendering = (process) => {
  defer(() => {
    try {
      process()
    } catch (e) {
      process_rerenderCount = 0;
      throw e;
    }  
  })
}
```

### DIFF
--- a/mangle.json
+++ b/mangle.json
@@ -65,7 +65,8 @@
       "$_catchError": "__e",
       "$_unmount": "__u",
       "$_owner": "__o",
-      "$_skipEffects": "__s"
+      "$_skipEffects": "__s",
+      "$_rerenderCount": "__r"
     }
   }
 }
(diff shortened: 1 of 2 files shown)


andrewiggins:
What's the cost putting the try/catch in core? Wondering if it should be a feature of core that we allow attempting to re-setState after an unhandled error....

JoviDeCroock:
> What's the cost putting the try/catch in core? Wondering if it should be a feature of core that we allow attempting to re-setState after an unhandled error....

The byte-cost is a bit higher and seems to me that this could be perceived as unwanted since it's automatic recovery. Which now only serves the purpose of the dev-tooling.

andrewiggins:
Another random idea, what about putting `_rerenderCount` on the `process` function that gets passed into `debounceRendering`? Would that be smaller?

In general, I'd love to over time move away from storing private properties on objects exposed to developers where they put their own properties 😅

JoviDeCroock:
@andrewiggins 3b lighter 😅

PR #2558, opened 2020-05-27
Normalize value prop on textareas in compat

### LINKED ISSUE #2556: Textareas are not initialized correctly during hydration
I'm currently using Preact + Next.js with the configuration from [nextjs-preact-demo](https://github.com/developit/nextjs-preact-demo). SSR with textareas seems to be breaking in Preact >=10.4.2.

### Reproduction

[CodeSandbox of last working version (10.4.1)](https://codesandbox.io/s/github/btk5h/bug-nextjs-preact-rendering?file=/pages/index.js)
[CodeSandbox of first broken version (10.4.2)](https://codesandbox.io/s/github/btk5h/bug-nextjs-preact-rendering/tree/10.4.2?file=/pages/index.js)

### Expected Behavior

The textarea should be initialized with a value on page load.

### Actual Behavior

The textarea is not initialized with a value on page load. The pre-rendered HTML seems to be the same between both 10.4.1 and 10.4.2, it just seems that Preact is not properly setting the textarea value during hydration.

btk5h:
Fixes #2556. This PR implements [the textarea normalization that React performs](https://reactjs.org/docs/forms.html#the-textarea-tag) in `preact/compat`.

Thanks @marvinhagemeister for pointing out where to make this change, it's been super helpful for getting to understand the codebase better! 🙌

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -130,6 +130,12 @@ options.vnode = vnode => {
 				delete props.value;
 			}
 
+			// Normalize textarea by setting children to value
+			if (props.value != null && type === 'textarea') {
+				props.children = props.value;
+				delete props.value;
+			}
+
 			// Normalize DOM vnode properties.
 			let shouldSanitize, attrs, i;
 			for (i in props) if ((shouldSanitize = CAMEL_PROPS.test(i))) break;
--- a/compat/test/browser/textarea.test.js
+++ b/compat/test/browser/textarea.test.js
@@ -17,4 +17,10 @@ describe('Textarea', () => {
 
 		expect(scratch.innerHTML).to.equal('<textarea>foo</textarea>');
 	});
+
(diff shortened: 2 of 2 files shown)


compat/src/render.js
```diff
@@ -130,6 +130,12 @@ options.vnode = vnode => {
 				delete props.value;
 			}
 
+			// Normalize textarea by setting children to value
+			if (props.value != null && type === 'textarea') {
```
gustavpursche, line 134: Would this handle `defaultValue` too? [React Docs](https://reactjs.org/docs/uncontrolled-components.html#default-values)

btk5h: It does! This runs after the transformation that renames `props.defaultValue` to `props.value`, so it would apply in that case as well. It does make sense to test that case separately though, so I've gone ahead and added one.

PR #2569, opened 2020-06-03
Get document from parentNode.ownerDocument  

### LINKED ISSUE #2545: WrongDocumentError due to `document.createElement`
It is possible to get a `WrongDocumentError` (as IE calls it) due to the calls to `document.createElement` and other node creation APIs [in diffElementNodes](https://github.com/preactjs/preact/blob/0f8c55c6cfad8d5cc3aec6785c1f6940998b4782/src/diff/index.js#L312).

React [avoids this](https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOMComponent.js#L414) by consistent use of the `ownerDocument` of the render root. This avoids issues because the render root's `ownerDocument` is guaranteed to be the right document, whereas the `document` global may not be when the `preact.js` scripts have loaded in a separate document.

### Reproduction

I've recreated [a unit test from React](https://github.com/facebook/react/blob/ac533fde3dff81023809a92d3f5432df7fc2c418/packages/react-dom/src/__tests__/ReactDOMFiber-test.js#L1187), against Preact APIs, showing that the `ownerDocument` is not being used.

https://codesandbox.io/s/optimistic-herschel-jvn5e

```js

const textContent = "Hello world";
const iframe = document.createElement("iframe");
document.body.appendChild(iframe);
const iframeDocument = iframe.contentDocument;
iframeDocument.write(
  "<!DOCTYPE html><html><head></head><body><div></div></body></html>"
);
iframeDocument.close();
const iframeContainer = iframeDocument.body.firstChild;

let actualDocument;
let textNode;
let appendChildWasCalled = false;

const origAppendChild = iframeContainer.appendChild;
iframeContainer.appendChild = function(node) {
  appendChildWasCalled = true;
  actualDocument = node.ownerDocument;
  textNode = node;
  origAppendChild.apply(iframeContainer, arguments);
};

render(textContent, iframeContainer);

if (!appendChildWasCalled) throw new Error();
if (textNode.textContent !== textContent) throw new Error();
if (actualDocument !== document) throw new Error();
if (actualDocument !== iframeDocument) throw new Error();
```

The exact case where this leads to a `WrongDocumentError` with IE is a bit more obscure to reproduce, involving multiple windows. It's something that I'd like to see fixed upstream in Preact, otherwise I'll need to maintain a patch, as this leads to a high priority issue I must resolve.

### Steps to reproduce

1. Open the CodeSandbox.
2. Observe the output area.

### Expected Behavior

No Error should throw if the correct document is used. Should show Hello World in the iframe.

### Actual Behavior

An error is thrown for one of the assertions.

### Suggested Fix

Ideally we can access the `ownerDocument` of the render root container element, and use that instead of `document`. I don't see an extremely direct way of accessing this, but I suspect it can be passed through in the worst case.

Hydrophobefireman:
instead of relying on global (Fixes #2545)


src/diff/index.js
```diff
@@ -226,7 +229,8 @@ export function diff(
 				isSvg,
 				excessDomChildren,
 				commitQueue,
-				isHydrating
+                		isHydrating,
+                		doc
```
JoviDeCroock, line 233: We can remove Lines 39-41 if we inline this (and save bytes)

```suggestion
                		parentDom.ownerDocument
```

Hydrophobefireman:
Do we have an existing test we could add it to?
Could there  be a better way than getting the ownerDocument from the parent dom on every diff call?

JoviDeCroock:
If one subtree is supposeed to be the same document everytime you could pass it to `diff` from `render` and `renderComponent` (Component.js). Then just keep passing it inside the recursive function.

This will only make you call `.ownerDocument` once. I'd prefer a separate test for this so we can clearly assess regressions.

JoviDeCroock:
I might have misexplained myself, let me  elaborate on the previous comment. We should be able to derive `ownerDocument`  [here](https://github.com/preactjs/preact/blob/master/src/render.js#L36) by means of doing: `parentDom.ownerDocument` and passing this in after `isHydrating`.

Next up we go to the second entry point for `tree-updates` [here](https://github.com/preactjs/preact/blob/master/src/component.js#L134) where we can use `parentDom.ownerDocument` again and add `undefined, parentDom.ownerDocument` to the `diff` call.

Now we need to ensure that we pass this argument around to `diff --> diffChildren --> diff` and `diff --> diffElementNodes --> diffChildren`.

This will reduce the amount of times you access that variable, this is the solution you imply with 

>  could there be a better way than getting the ownerDocument from the parent dom on every diff call, like keeping it as a constant somewhere outside?

I think.

Hydrophobefireman:
I tried that [here](https://github.com/preactjs/preact/pull/2569/commits/799f08ef2a757d04e9c65443f8f9946cc3c5e52a) 
but the tests failed for some reason.

JoviDeCroock:
Because you aren't passing it to `diffChildren` in `diffElementNodes` and from `diffChildren` back to `diff`

Hydrophobefireman:
Yep! that'll do it

src/component.js
```diff
@@ -131,7 +131,9 @@ function renderComponent(component) {
 			parentDom.ownerSVGElement !== undefined,
 			null,
 			commitQueue,
-			oldDom == null ? getDomSibling(vnode) : oldDom
+			oldDom == null ? getDomSibling(vnode) : oldDom,
+			null,
```
JoviDeCroock, line 135:
```suggestion
			undefined,
```

Is how it was before

src/diff/index.js.rej
```diff
@@ -0,0 +1,8 @@
+diff a/src/diff/index.js b/src/diff/index.js	(rejected hunks)
```
JoviDeCroock, line 1: You comitted a new file

src/component.js
andrewiggins: This is the `isHydrating` argument so perhaps `false` would be better? Minifies to `!1` instead `void 0` I think

test/browser/getOwnerDocument.test.js
```diff
@@ -0,0 +1,25 @@
...
+
+	before(() => {
+		scratch = setupScratch();
+	});
+
+	after(() => {
+		teardown(scratch);
+	});
+
+	it('should reference the correct document from the parent node', () => {
```
andrewiggins, line 18: Wouldn't this test pass without your change? I think this effectively asserts that `document.createElement('div').ownerDocument == document`. Can we write a test that mimics the original issue this PR fixes? I think it involves iframes or something.

test/browser/getOwnerDocument.test.js
```diff
@@ -0,0 +1,39 @@
...
+		);
+
+		iframeDoc.close();
+
+		let iframeRootNode = iframeDoc.querySelector('div');
+
+		render('Hello', iframeRootNode);
+
+		expect(iframeRootNode.textContent).to.be.equal('Hello');
+		expect(iframeRootNode.ownerDocument).to.be.equal(iframeDoc);
```
JoviDeCroock, line 36: This has historically been rendered by your `iframeDoc.write` so this would've worked before as well. What wouldn't have worked is checking on `iframeRootNode.firstChild`

Hydrophobefireman: I think i can get away with using `Text.ownerDocument` instead of rendering a div

wnayes:
Yes, it seems like it should fix my issue. Thanks for picking this up!

test/browser/getOwnerDocument.test.js
developit: Yeah, a test for this could spy on `iframeDoc.createElement()` to assert that it is called instead of `document.createElement()`.

developit:
**Size Change:** +71 B (0%) 

**Total Size:** 39.8 kB

|:--- |:---:|:---:|:---:|
| `dist/preact.js` | 3.89 kB | +16 B (0%) |  |
| `dist/preact.min.js` | 3.89 kB | +16 B (0%) |  |
| `dist/preact.module.js` | 3.91 kB | +22 B (0%) |  |
| `dist/preact.umd.js` | 3.95 kB | +17 B (0%) |  |

wnayes:
Really looking forward to getting this change in a new release - any thoughts on when this can start moving again? Or anything I can do to help? Thanks!

Hydrophobefireman:
Sorry, took way longer than I should have to add the test. Kind of got into school related work

test/browser/getOwnerDocument.test.js
wnayes: Looks like the last changes to use a spy on `createTextNode` should resolve this review comment.

Hoping to see this PR merge into an official release soon!

wnayes:
I took a quick look at how this change might affect performance, since I was curious what the impact of using the right document would be. Intuitively, it would give better performance since the browser won't need to do any sort of bookkeeping when nodes go between documents.

I made this code sandbox quick to get a sense of some numbers. (And a basic HTML page, since this doesn't work in IE 11.) The snippet does a bunch of createElement + appendNode + removeNode in an iframe, but varying whether the root document or ownerDocument are used.

https://codesandbox.io/s/ownerdoc-compare-s5bwm-s5bwm

My impression from testing this out a bit is:

- I don't see any difference in Chrome.
- Firefox seems about 3x faster with ownerDocument.
- IE 11 seems about 30% faster with ownerDocument.

This is of course a micro benchmark, so the actual savings in a real app may be negligible. But seems there could be a few milliseconds to gain maybe.

I have been applying this patch manually onto the official release as-is and it has addressed the original issue for me.

PR #2575, opened 2020-06-11
Simplify npm packaging and remove source from package

andrewiggins:
In the spirit of #2497, this PR removes the source from the package as well as moves the location of packaged TypeScript definition files to the `dist` folder. This simplifies the folders we have to define in the `files` field of our package.json. Below is a more detailed summary:

* Simplify how we specify the `files` globs in package.json (it now uses brace expansion)
* Remove source files form our npm package
* Copy TypeScript definition files to `dist` folder to simplify packaging and update package.json pointers
* Update TS tests (and other places) to import using the relevant package's `types` field to validate that the `types` field in the package.json is correct
* Combine all public `compat` TypeScript types into `index.d.ts` so it's easier to package
* Rename `config` folder to `scripts` since it doesn't really contain any config files
* Rename `node-13-exports.js` to `postbuild.js` since it now does more than just copy files for Node 13


JoviDeCroock (approved):
Awesome work!

JoviDeCroock:
Closing: superseded — the v11 modern-build work overhauled packaging/exports/dist and the types layout.

PR #2578, opened 2020-06-11
Add native HTML Attribute.

firatsarlar:
'A' tag have a 'native' dom attribute, there is a use case for this attribute in preact router.

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -682,6 +682,7 @@ export namespace JSXInternal {
 		multiple?: boolean;
 		muted?: boolean;
 		name?: string;
+		native?: boolean;
 		nonce?: string;
 		noValidate?: boolean;
 		open?: boolean;


pmkroeker:
`native` is not a spec-ed attribute on a normal a tag. See the [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Attributes). Preact router should add this itself I think.

firatsarlar:
@pmkroeker , preact router does not implement HTML attributes in its definitions.
https://github.com/preactjs/preact-router/blob/master/match/index.d.ts
For this repo it's a one line thing, but I don't know how to add there -preact router-.

pmkroeker:
IMO the jsx attributes should only contain what is in the spec

firatsarlar:
spec vs. usage in preact router  ...
If spec is a strict requirement then preact router should use an attribute like "data-native" .

pmkroeker:
What I am trying to say is the preact types should only include those that are in the spec. Otherwise people might start throwing native on an a tag expecting behaviour changes.

Preact router should allow native _as a prop_ and not an attribute, as that is more accurately what it is.

marvinhagemeister:
Agree with @pmkroeker here. The HTML spec doesn't list a `native` attribute anywhere on `<a>`-elements. It's something `preact-router` adds and therefore `preact-router` should be the pacakge to extend the native `HTMLAttribute` type.

firatsarlar:
1.Typescript,
2.preact router,
3.routes,
4.routes which shouldn't handled by router ( image links etc. ) - which requires native attribute -

this combination is not working , typescript won't compile if a tag has native prop - attribute , however you name it- .

I'm changing this at preact's code at node_modules for every project,

Two PRs for getting rid of this step ( for those who use similar combination ).

One of them should be closed / rejected eventually , may be this one  ..

PR #2579, opened 2020-06-12
Share build output between action jobs

andrewiggins:
Install and build the main repo once, and share that build output with other jobs (in the same workflow) that need it.

* I've combined the previous "build" and "test" jobs so that we only install our dev dependencies once. If the build and tests pass then the build output is uploaded as an artifact which other jobs can download. This change means that no other job (i.e. benchmarks) will run unless tests pass. Depending on what you want that could be a feature or a bug 😅 Let me know what you think!

* I determine what the build output is by running npm pack to generate our npm tarball. Then the build untars that file and uploads it's contents. I wanted to use `npm pack` to reuse the declaration of our build output from our package.json's `files` property.

* I don't upload the tarball directly cuz then each job that needed the output would have to download the artifact, untar it, and then shuffle the files around into their correct position in the repo. The approach done here means that other jobs that want to re-use the build output just download the artifact into the working directory (which is a clone of the preact repo) and the build files are already correctly placed for them. We could also upload the tarball if that's useful but I figured we just do that once we had a use for it.

* I removed the node_modules cache from the benchmarks jobs cuz it was consistently failing and just added a lot of duplication between the benchmarks. The `npm ci` command usually takes between 20s - 40s so I'm not going to sweat it now. We can look at the cache issue later if that becomes important/if there is interest.

### DIFF
--- a/.github/workflows/ci-master.yml
+++ b/.github/workflows/ci-master.yml
@@ -6,29 +6,8 @@ on:
       - master
 
 jobs:
-  build:
-    name: Build
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
-        with:
-          node-version: '12.x'
-      - name: Cache node modules
-        uses: actions/cache@v1
-        env:
-          cache-name: cache-node-modules
-        with:
-          path: ~/.npm
-          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
-          restore-keys: |
-            ${{ runner.os }}-build-${{ env.cache-name }}-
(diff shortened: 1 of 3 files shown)


marvinhagemeister:
Was GitHub down or something? Curious why the required status' are still marked as pending.

andrewiggins:
@developit (or maybe @marvinhagemeister has permissions?) If this change is something we want to do, we'll need to update the required checks for PRs into main since I've changed them here.

andrewiggins:
I think we just need make build_test (the new combined check) the only required one. I think that maintains parity since the current required ones are Build and Test

marvinhagemeister:
Alright, updated the required checks for PRs accordingly :+1:

PR #2583, opened 2020-06-14
fix(compat): don't leak class prop when using className

### LINKED ISSUE #2582: Preact compat leaks class properties in components
### Reproduction

I made a codesandbox for preact-compat: https://codesandbox.io/s/preact-compat-class-leak-uunxl?file=/src/index.js.

### Steps to reproduce

```
  return (
    <div className="App">
      <MyComponent className="style1234"/>
    </div>
  );
}

export default function MyComponent({ className, ...props }) {
  return (
    <div className={className}>
      <span{...props}>MyText</span>
    </div>
  );
}

```

### Expected Behavior

Span should not have a classname on its dom element

### Actual Behavior

span gets the same class as the div

wardpeet:
When using preact-compat with className it leaks the class prop into a react component and breaks spreading props.

I might be missing some context here so this fix might not be 100% accurate.

Fixes #2582

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -104,7 +104,8 @@ options.vnode = vnode => {
 
 	if (type) {
 		// Alias `class` prop to `className` if available
-		if (props.class != props.className) {
+		// only alias on dom elements
+		if (typeof type === 'string' && props.class != props.className) {
 			classNameDescriptor.enumerable = 'className' in props;
 			if (props.className != null) props.class = props.className;
 			Object.defineProperty(props, 'className', classNameDescriptor);
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -88,10 +88,16 @@ function setSafeDescriptor(proto, key) {
 	}
 }
 
-let classNameDescriptor = {
+const classNameDescriptor = {
 	configurable: true,
 	get() {
 		return this.class;
+	}
+};
+const classDescriptor = {
+	configurable: true,
+	get() {
+		return this.className;
 	}
 };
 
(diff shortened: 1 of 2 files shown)


marvinhagemeister:
Good catch! I think we just need a check around the normalization code that verifies that the current `vnode` is a component :+1:

compat/src/render.js
```diff
@@ -104,7 +104,8 @@ options.vnode = vnode => {
 
 	if (type) {
 		// Alias `class` prop to `className` if available
-		if (props.class != props.className) {
+		// only alias on dom elements
+		if (typeof type === 'string' && props.class != props.className) {
```
wardpeet, line 108: Only aliasing on dom elements breaks the automatic normalization for regular components. So let's say I have a preact component that uses `class` as a prop in my preact-compat app. If I pass the component a className it won't apply it cause it's relying on `class`, not `className`. I think this is valid for preact-compat mode but that's not my decision to make.

developit: I wonder if we could always install descriptors for both `class` and `className`, and simply make the corresponding one `enumerable` based on which was specified?

wardpeet: That's what I did now :)

wardpeet:
I could lower the bytes by using a function for the descriptor but gzip & brotli wise it would be bigger. Unsure what we optimize for here.

compat/test/browser/render.test.js
```diff
@@ -242,6 +242,39 @@ describe('compat render', () => {
 		);
 	});
 
+	// Issue #2582
+	it('should not leak class/className normalisation into props', () => {
+		function Foo({ className, class: cl, ...props }) {
```
developit, line 247: This pulls class out of props so it won't be forwarded on to the `<li>` even if present.

```suggestion
		function Foo({ className, ...props }) {
			const cl = props.class;
```

compat/test/browser/render.test.js
```diff
@@ -242,6 +242,39 @@ describe('compat render', () => {
...
+			return (
+				<ul className={className} data-class={cl}>
+					<li {...props}>hi</li>
+				</ul>
+			);
+		}
+		function Bar({ class: className, className: cl, ...props }) {
```
developit, line 254: Same as above

```suggestion
		function Bar({ class: className, ...props }) {
			const cl = props.className;
```

charrondev:
Does this need anything to keep it moving?

I recently tried to give preact and shot and was very happy with the size I was able to trim off my bundle. This seemed to be a major impediment to moving forward with it though.

JoviDeCroock:
Closing: the class-prop-leak fix landed differently in v11 `compat/src/render.js`; the linked issue #2582 is closed.

PR #2584, opened 2020-06-16
fix hydration issue by unmounting the excessChildren

JoviDeCroock:
Since we can't assess what would've been returned from the `render`, it's safer to unmount the remaining result and default to the regular render process which is achieved by the error boundary.

Fixes https://github.com/preactjs/preact/issues/2539

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -232,6 +232,9 @@ export function diff(
 
 		if ((tmp = options.diffed)) tmp(newVNode);
 	} catch (e) {
+		// When we are hydrating and have excessDomChildren we don't know the _children this VNode
+		// should receive so it's safer to unmount them. Else on the subsequent error-boundary diff,
+		// we won't know the oldDom and insert an additional node instead of replace the prerendered one. (#2539)
 		if (isHydrating && excessDomChildren != null) {
 			for (tmp = excessDomChildren.length; tmp--; ) {
 				if (excessDomChildren[tmp] != null) removeNode(excessDomChildren[tmp]);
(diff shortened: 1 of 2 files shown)


src/diff/index.js
```diff
@@ -232,6 +232,12 @@ export function diff(
 
 		if ((tmp = options.diffed)) tmp(newVNode);
 	} catch (e) {
+		if (isHydrating && excessDomChildren != null) {
```
sventschui, line 235: Want to add a comment here why we do this? I know myself good enough to know I forget tomorrow :D

src/diff/index.js
```diff
@@ -232,6 +232,15 @@ export function diff(
 
 		if ((tmp = options.diffed)) tmp(newVNode);
 	} catch (e) {
+		// When we are hydrating and have excessDomChildren we don't know the _children this VNode
+		// should receive so it's safer to unmount them. Else on the subsequent error-boundary diff,
+		// we won't know the oldDom and insert an additional node instead of replace the prerendered one. (#2539)
+		if (isHydrating && excessDomChildren != null) {
+			for (tmp = excessDomChildren.length; tmp--; ) {
```
JoviDeCroock, line 239: We could pass `excessDomChildren` to `options._catchError` and only do this if an error-boundary catches the error. Not entirely sure if that's better.

developit:
Note: this change probably conflicts with #2291.

JoviDeCroock:
> Note: this change probably conflicts with #2291.

I see, can we consolidate these somehow. Like we shouldn’t remove nodes when this is a promise?

JoviDeCroock:
Closing due to conflicts with progressive hydration

PR #2588, opened 2020-06-21
initial poc for backing tree for Preact X

JoviDeCroock:


src/diff/index.js
```diff
@@ -235,12 +235,6 @@ export function diff(
 			}
 
 			c._force = false;
-		} else if (
```
JoviDeCroock, line 238: TODO: restore this

src/diff/index.js
```diff
@@ -309,8 +303,8 @@ function diffElementNodes(
 	isHydrating
 ) {
 	let i;
-	let oldProps = oldVNode.props;
-	let newProps = newVNode.props;
+	let oldProps = oldVNode.vnode && oldVNode.vnode.props;
```
JoviDeCroock, line 306: This should become _vnode

src/diff/children.js
```diff
@@ -69,38 +69,30 @@ export function diffChildren(
 		// or we are rendering a component (e.g. setState) copy the oldVNodes so it can have
 		// it's own DOM & etc. pointers
 		else if (typeof childVNode == 'string' || typeof childVNode == 'number') {
-			childVNode = newParentVNode._children[i] = createVNode(
```
JoviDeCroock, line 72: Something's wrong here making the tests fail

marvinhagemeister: It looks like we're sometimes passing `VNodes` and sometimes backing nodes here. Need to investigate further.

JoviDeCroock: I must've goofed up somewhere

src/diff/index.js
JoviDeCroock: This doesn't seem possible due to `component._vnode` overriding this to a property that we already have 😅

compat/test/browser/PureComponent.test.js
```diff
@@ -68,38 +68,44 @@ describe('PureComponent', () => {
...
 			}
 		}
 		let spy = sinon.spy(C.prototype, 'render');
 
-		let inst = React.render(<C />, scratch);
+		React.render(<C />, scratch);
 		expect(spy).to.have.been.calledOnce;
 		spy.resetHistory();
 
-		inst = React.render(<C />, scratch);
```
JoviDeCroock, line 82: Not entirely sure why calling `forceUpdate/setState` on this `Fragment` trickled through to the component that's confusing as hell

compat/src/render.js
```diff
@@ -35,7 +35,7 @@ export function render(vnode, parent, callback) {
 	preactRender(vnode, parent);
 	if (typeof callback == 'function') callback();
 
-	return vnode ? vnode._component : null;
+	return vnode ? parent._children._component : null;
```
JoviDeCroock, line 38: I think this is erroneous atm since we'll be returning the fragment

compat/test/browser/forwardRef.test.js
```diff
@@ -400,8 +400,9 @@ describe('forwardRef', () => {
 		const Transition = ({ children }) => {
 			const state = useState(0);
 			forceTransition = state[1];
-			expect(children.ref).to.not.be.undefined;
-			if (state[0] === 0) expect(children.props.ref).to.be.undefined;
+			// TODO: ref exists on the backing node
+			// expect(children.ref).to.not.be.undefined;
```
JoviDeCroock, line 404: We can't really test this anymore since ref and props.ref are at the backing node level now

src/diff/children.js
```diff
@@ -69,38 +69,25 @@ export function diffChildren(
...
-				Fragment,
-				{ children: childVNode },
-				null,
-				null,
+			childVNode = newParentVNode._children[i] = createBackingNode(
+				createVNode(Fragment, { children: childVNode }),
 				null
 			);
-		} else if (childVNode._dom != null || childVNode._component != null) {
```
JoviDeCroock, line 87: It's not possible to see a backing node as a child so we can safely remove this branch

src/create-element.js
```diff
@@ -100,3 +79,35 @@ export function Fragment(props) {
...
+		// a _nextDom that has been set to `null`
+		_nextDom: undefined,
+		_component: null,
+		_original: original
+	};
+
+	if (original == null) backingNode._original = vnode;
+
+	if (vnode.props && vnode.props.ref) {
+		delete vnode.props.ref;
```
JoviDeCroock, line 105: This could possibly be the slow part (together with 109)

src/create-element.js
```diff
@@ -100,3 +79,35 @@ export function Fragment(props) {
  */
 export const isValidElement = vnode =>
 	vnode != null && vnode.constructor === undefined;
+
+export function createBackingNode(vnode, original) {
+	const backingNode = {
+		key: vnode.props ? vnode.props.key : undefined,
+		ref: vnode.props ? vnode.props.ref : undefined,
+		type: vnode.type,
```
JoviDeCroock, line 87: should we duplicate `type` on here? My gut sais yes for easy access

src/diff/index.js
```diff
@@ -67,14 +67,14 @@ export function diff(
 
 	// When passing through createElement it assigns the object
 	// constructor as undefined. This to prevent JSON-injection.
-	if (newVNode.constructor !== undefined) return null;
+	if (newVNode._node && newVNode._node.constructor !== undefined) return null;
```
JoviDeCroock, line 70: When we have a function child `_node` won't be defined

src/diff/index.js
```diff
@@ -208,7 +208,7 @@ export function diff(
 			}
 
 			let isTopLevelFragment =
-				tmp != null && tmp.type == Fragment && tmp.key == null;
+				tmp != null && tmp.type == Fragment && tmp.props.key == null;
```
JoviDeCroock, line 211: Since this isn't normalized on `createElement` we need to access it in props

compat/src/suspense-list.js
```diff
@@ -23,7 +23,7 @@ const resolve = (list, child, node) => {
 		// mark the child as completely resolved by deleting it from ._map.
 		// This is used to figure out when *all* children have been completely
 		// resolved when revealOrder is 'together'.
-		list._map.delete(child);
+		list._map.delete(child._node);
```
JoviDeCroock, line 26: Something is wrong here, since we'll only delete one and the second one isn't present in the _map (wrong ref?)

src/create-element.js
```diff
@@ -57,29 +50,15 @@ export function createElement(type, props, children) {
...
-		// a _nextDom that has been set to `null`
-		_nextDom: undefined,
-		_component: null,
-		constructor: undefined,
-		_original: original
+		constructor: undefined
 	};
 
-	if (original == null) vnode._original = vnode;
 	if (options.vnode) options.vnode(vnode);
```
JoviDeCroock, line 62: We could move this to `createBackingNode` but this would be a lot more lazy

src/create-element.js
```diff
@@ -100,3 +79,35 @@ export function Fragment(props) {
  */
 export const isValidElement = vnode =>
 	vnode != null && vnode.constructor === undefined;
+
+export function createBackingNode(vnode, original) {
+	const backingNode = {
+		key: vnode.props ? vnode.props.key : undefined,
```
JoviDeCroock, line 85: Would using `null` here make it better for v8

src/diff/children.js
```diff
@@ -69,38 +69,25 @@ export function diffChildren(
 		// or we are rendering a component (e.g. setState) copy the oldVNodes so it can have
 		// it's own DOM & etc. pointers
 		else if (typeof childVNode == 'string' || typeof childVNode == 'number') {
-			childVNode = newParentVNode._children[i] = createVNode(
-				null,
-				childVNode,
-				null,
-				null,
+			childVNode = newParentVNode._children[i] = createBackingNode(
```
JoviDeCroock, line 72: We shouldn't always recreate the backing node, we'd have to reuse these from the `oldChildren`

debug/src/component-stack.js
```diff
@@ -76,15 +76,15 @@ function isPossibleOwner(vnode) {
...
+	while (next._node._owner != null) {
+		stack.push(next._node._owner);
+		next = next._node._owner;
 	}
 
 	return stack.reduce((acc, owner) => {
 		acc += `  in ${getDisplayName(owner)}`;
 
-		const source = owner.__source;
+		const source = owner._node.__source;
```
developit, line 87: I wonder if it'd be worth copying these to the backing node?

src/create-element.js
developit: Yes.

src/create-element.js
developit: Yep, avoids reaching into the vnode. I do think there's merit in adding a property (bitfield?) that indicates the vnode's type grouping (text, element, function, class) since that's a decent percentage of the current usage of `vnode.type`.

src/create-element.js
developit: For performance, it's cheaper to create a new object than to mutate the shape of the current one. If we were to copy props to `backingNode.props` here, the copy could also exclude these keys:

```js
let props = {},
    originalProps = vnode.props,
    key,
    ref;

if (originalProps != null) {
    for (let i in originalProps) {
        const value = originalProps[i];
        if (i === 'key') key = value;
        else if (i === 'ref') ref = value;
        else props[i] = value;
    }
}

const backingNode = {
    props,
    key,
    ref,
    // ...
}
```

This way, the `originalProps[i]` access will be polymorphic, but nothing else. Elsewhere in the renderer, `backingNode.props` can be assumed to be an object, whereas currently we have to check if `vnode.props` is non-null. We might even be able to get rid of the current property copying loop in `h()`.

src/diff/children.js
developit: IMO this is the big sell of backing node in current Preact.

src/diff/index.js
developit: If it's easier, this could be moved into `createBackingNode()` since it'll still be invoked from within the diff. It just needs to break, doesn't really matter how it breaks.

developit:
Super interesting read-through 👍

The main thing that I was pondering as I read was whether it'd be faster+smaller overall to copy props to the backing node, since that would avoid all the `vnode._node.props` lookups and we could rely on `backingNode.props` always being an object.

src/create-element.js
marvinhagemeister: @developit Should we aim to touch `vnode` as little as possible in general with the backing tree approach? E.g. copying relevant stuff onto the backing node to minimize `vnode` access?

JoviDeCroock:
Deferred to v11 due to incompatability with our current diffing

PR #2595, opened 2020-06-29
Run benches on linux VMs

andrewiggins:


andrewiggins:
Uh oh, looks like I've been bitten by the issue I was seeing earlier - benches are timing out... :'( Need to dig more into this

andrewiggins:
Proof that something like this should work: https://github.com/andrewiggins/js-framework-tachometer/runs/819819054?check_suite_focus=true

andrewiggins:
Thanks @marvinhagemeister for helping me figure this out!

PR #2598, opened 2020-07-01
Simplify compat defaultValue handling

andrewiggins:
The code removed below was originally added back in preactjs/preact-compat#280. However, in a recent change (maybe #2392?) the `props.value = props.defaultValue` stopped being covered by our tests (see attached). So I wondered if this code was even necessary anymore. I added some additional tests to try and push the boundary of whether this code is needed. Seems like it isn't and we can rely on native defaultValue handling. MDN seems to imply [defaultValue works going back a while](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement) (though I did find one bug with textarea and defaultValue in IE11)

I also adding a warning when using preact/compat and `value` and `defaultValue` on an `input` tag. React warns about this and figured it would be good if we did too. The benefit is to push people to use preact/compat with only React-approved patterns to limit the scope of what preact/compat needs to support. I don't want to support unsupported React behaviors in preact/compat too lol.

[coverage.zip](https://github.com/preactjs/preact/files/4856070/coverage.zip)

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -112,14 +112,6 @@ options.vnode = vnode => {
 
 		// Apply DOM VNode compat
 		if (typeof type != 'function') {
-			// Apply defaultValue to value
-			if (props.defaultValue && props.value !== undefined) {
-				if (!props.value && props.value !== 0) {
-					props.value = props.defaultValue;
-				}
-				delete props.defaultValue;
-			}
-
 			// Add support for array select values: <select value={[]} />
 			if (Array.isArray(props.value) && props.multiple && type === 'select') {
 				toChildArray(props.children).forEach(child => {
@@ -133,6 +125,7 @@ options.vnode = vnode => {
 			// Normalize textarea by setting children to value
 			if (props.value != null && type === 'textarea') {
 				props.children = props.value;
+				delete props.defaultValue;
 				delete props.value;
 			}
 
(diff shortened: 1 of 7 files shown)


andrewiggins:
To be clear, it could also be the case that we are missing some tests, so if you think this is still needed, let's add some tests to justify its existence.

andrewiggins:
Okay, this is probably a terrible idea. Already found one failing test. Should probably understand how defaultValue does [not] affect these other form elements: [HTMLSelectElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement), [HTMLOptionElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLOptionElement), [HTMLButtonElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLButtonElement), [HTMLProgressElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLProgressElement)

monaye:
> though I did find one bug with textarea and defaultValue in IE11

@andrewiggins I think I am seeing this issue. Do you know if this had been fixed or do we have an issue open for this?

andrewiggins:
Hey @monaye, I don't think we have an issue that particular thing yet. Could you go ahead and open a new one describing the issue you are running into? That way others on the team can jump on and perhaps we can come up with a quick fix for you.

JoviDeCroock:
Closing: IE11-era defaultValue simplification; v11 dropped IE11 and restructured compat value/defaultValue normalization.

Issue #2599, opened 2020-07-01
got error after updating preact version from 10.4.4 to 10.4.5


logvi:
After updating version from 10.4.4 to 10.4.5 it fails with this error:

```
Uncaught (in promise) TypeError: Cannot use 'in' operator to search for 'current' in null
    at ForwardRef() (compat.module.js:1)
    at mobxreact.esm.js:413
    at mobxreactlite.esm.js:279
    at trackDerivedFunction (mobx.umd.js:3258)
    at Reaction.track (mobx.umd.js:3457)
    at useObserver (mobxreactlite.esm.js:277)
    at d.Observer [as constructor] (mobxreactlite.esm.js:355)
    at d.L [as render] (preact.module.js:1)
    at T (preact.module.js:1)
    at g (preact.module.js:1)
```

revert to 10.4.4 solve this issue

JoviDeCroock:
That means you are passing null to `ref` somewhere, is this intended?

marvinhagemeister:
@JoviDeCroock I think that's not the user value, but `mobx-react-life`. Iirc #2566 had a similar cause. The mobx plugin pulls the component apart and makes the call itself. Haven't come up with a decent test case for that yet.

PR #2600, opened 2020-07-01
(fix) - passing null to forwardRef

JoviDeCroock:
fixes: https://github.com/preactjs/preact/issues/2599

### DIFF
--- a/compat/src/forwardRef.js
+++ b/compat/src/forwardRef.js
@@ -33,7 +33,7 @@ export function forwardRef(fn) {
 		ref = props.ref || ref;
 		return fn(
 			clone,
-			typeof ref === 'object' && !('current' in ref) ? null : ref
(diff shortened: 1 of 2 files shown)


compat/src/forwardRef.js
```diff
@@ -33,7 +33,7 @@ export function forwardRef(fn) {
 		ref = props.ref || ref;
 		return fn(
 			clone,
-			typeof ref === 'object' && !('current' in ref) ? null : ref
+			!ref || (typeof ref === 'object' && !('current' in ref)) ? null : ref
```
marvinhagemeister, line 36: The test passes even without the change

JoviDeCroock: Yes saw that, trying to figure out why the second argument is defaulting to an empty object

PR #2601, opened 2020-07-01
Run same workflow on PR and master push

andrewiggins:
... and break out sauce lab run into it's own workflow.

This PR combines ci-pr.yml and ci-master.yml into one file: ci.yml. The only significant difference between the two was the inclusion of sauce labs secrets. So instead of having to maintain two different copies of the practically the same file, I've combined these two and broke out the sauce lab run into it's own file.

Also update workflows to Node 14.

Let me know what you think!


andrewiggins:
Note: SauceLabs is currently failing on master so one benefit of this approach is that master will now at least get a local Chrome test run even if SauceLabs times out, etc.

Issue #2604, opened 2020-07-08
Uncontrolled textarea


olabala:
this issue kinda like #1899 
same code, you can not clear the value of textarea under preact, but react can

React - ✅
https://codesandbox.io/s/react-controlled-input-0mfvp?file=/src/index.js

Preact - ❌
https://codesandbox.io/s/preact-controlled-input-xpt53?file=/src/index.js

(not sure if this is raised, if any, feel free to close it)

rjyo:
Stucking in the same situation.

I have tried to change textarea to a normal input, and everything just works.

developit:
This is fixed by #2615 and will be released shortly.

PR #2615, opened 2020-07-10
fix case where textarea doesn't reset state

JoviDeCroock:
Fixes: https://github.com/preactjs/preact/issues/2604

The test doesn't reproduce this yet I think it's related to uncontrolled/controlled which isn't emulated correctly in jsdom, I tried it out in demo and this fixes the issue.

### DIFF
--- a/compat/test/browser/textarea.test.js
+++ b/compat/test/browser/textarea.test.js
@@ -16,13 +16,13 @@ describe('Textarea', () => {
 	it('should alias value to children', () => {
 		render(<textarea value="foo" />, scratch);
 
-		expect(scratch.innerHTML).to.equal('<textarea>foo</textarea>');
+		expect(scratch.firstElementChild.value).to.equal('foo');
 	});
 
 	it('should alias defaultValue to children', () => {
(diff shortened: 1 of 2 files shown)


marvinhagemeister:
Note that we don't use jsdom in our tests. They typically run in a chrome instance via karma. But it could be that the bug doesn't manifest in Chrome as easily as it may in other browsers.

compat/test/browser/textarea.test.js
```diff
@@ -23,4 +24,26 @@ describe('Textarea', () => {
...
+			return <textarea value={state} />;
+		};
+
+		render(<App />, scratch);
+		expect(scratch.innerHTML).to.equal('<textarea></textarea>');
+
+		act(() => {
+			set('hello');
+		});
+		expect(scratch.innerHTML).to.equal('<textarea>hello</textarea>');
```
developit, line 42: I think this will catch the issue:

```suggestion
		expect(scratch.firstElementChild.value).to.equal('hello');
		expect(scratch.innerHTML).to.equal('<textarea>hello</textarea>');
```

PR #2617, opened 2020-07-10
Delete .editorconfig

chandi1988:


marvinhagemeister:
Just curious: what is the motivation for deleting it?

PR #2624, opened 2020-07-14
upgrade microbundle

JoviDeCroock:


marvinhagemeister:
@developit Spent a bit of time digging through the differences between the generated output from `0.11` and `0.12.3`. It's like you said on slack caused by terser. The version of terser included in `0.12.3` does inline a lot more functions. It's not truly inlining, but rather moving them inline and wrapping them via parenthesis':

Before:

```js
function k(n) {
	var l, u;
	if (null != (n = n.__) && null != n.__c) {
		for (n.__e = n.__c.base = null, l = 0; l < n.__k.length; l++)
			if (null != (u = n.__k[l]) && null != u.__e) {
				n.__e = n.__c.base = u.__e;
				break;
			}
		return k(n);
	}
}
```

After:

```js
function w() {
	for (var n; (w.__r = u.length); )
		(n = u.sort(function(n, l) {
			return n.__v.__b - l.__v.__b;
		})),
			(u = []),
			n.some(function(n) {
				var l, u, i, t, r, o, f;
				n.__d &&
					((o = (r = (l = n).__v).__e),
					(f = l.__P) &&
						((u = []),
						((i = s({}, r)).__v = i),
						(t = C(
							f,
							r,
							i,
							l.__n,
							void 0 !== f.ownerSVGElement,
							null,
							u,
							null == o ? _(r) : o
						)),
						N(u, r),
						t != o &&
							(function n(l) {
								var u, i;
								if (null != (l = l.__) && null != l.__c) {
									for (l.__e = l.__c.base = null, u = 0; u < l.__k.length; u++)
										if (null != (i = l.__k[u]) && null != i.__e) {
											l.__e = l.__c.base = i.__e;
											break;
										}
									return n(l);
								}
							})(r)));
			});
}
```

marvinhagemeister:
Seems like this is caused by terser removing support for `reduce_funcs`. See this issue on the terser repo: https://github.com/terser/terser/issues/696

developit:
I actually have a Babel plugin that can do inlining and unwrapping, I might be able to try that out here or in Microbundle as a fix.

JoviDeCroock:
Closing due to size issues

PR #2627, opened 2020-07-15
(v11) - createRoot

marvinhagemeister:
Not sure if this is the final approach we should take, but opening it here to gather feedback from the team.

- [x] Introduce a special `Root` node type
- [x] Rewrite `Portal` to use that
- [x] Rewrite `render` API to not attach the old `vnode` on `dom._children`

In this PR we've deprecated the `replaceNode` parameter.

Workaround: https://gist.github.com/developit/7c1b983dbd2cb68e6cefd367dfcf0ca1
Running verison: https://codesandbox.io/s/preact-htm-selective-hydration-t3d3y?file=/client.js:974-1438


test/browser/replaceNode.test.js
```diff
@@ -135,9 +135,10 @@ describe('replaceNode parameter in render()', () => {
 		const expectedB = '<div id="b">childB</div>';
 		const expectedC = '<div id="c">childC</div>';
 
-		render(<div id="a">childA</div>, scratch, childA);
-		render(<div id="b">childB</div>, scratch, childB);
-		render(<div id="c">childC</div>, scratch, childC);
+		// TODO: text nodes are somehow left out.
+		render(<div id="a">childA</div>, childA);
```
JoviDeCroock, line 139: The issue here is that `oldRoot` will be used for different `replaceNode` calls so it will effectively always overwrite previous results.

It should not preserve `oldRoot` for a different `replaceNode` argument

compat/test/browser/portals.test.js
```diff
@@ -372,7 +372,8 @@ describe('Portal', () => {
 		expect(scratch.innerHTML).to.equal('<div><p>Hello</p></div>');
 	});
 
-	it('should work with removing an element from stacked container to new one', () => {
+	// TODO: Reparenting is an issue in createRoot
+	it.skip('should work with removing an element from stacked container to new one', () => {
```
JoviDeCroock, line 376: We probably need to recreate the `createRoot` of the portal

compat/src/render.js
```diff
@@ -27,20 +21,25 @@ export const REACT_ELEMENT_TYPE =
...
 
 	return vnode ? vnode._component : null;
 }
 
 export function hydrate(vnode, parent, callback) {
-	preactHydrate(vnode, parent);
+	const root = !parent._root
+		? (parent._root = createRoot(parent))
+		: parent._root;
+	root.hydrate(vnode);
```
developit, line 42: Not sure if Terser is already doing this, but:

```suggestion
	const root = parent._root || (parent._root = createRoot(parent));
	root.hydrate(vnode);
```

compat/src/portals.js
```diff
@@ -1,85 +1,10 @@
...
+import { createElement, Root } from 'preact';
 
 /**
  * Create a `Portal` to continue rendering the vnode tree at a different DOM node
  * @param {import('./internal').VNode} vnode The vnode to render
  * @param {import('./internal').PreactElement} container The DOM node to continue rendering in to.
  */
 export function createPortal(vnode, container) {
-	return createElement(Portal, { vnode, container });
+	return createElement(Root, { _parentDom: container }, vnode);
```
developit, line 9: We should just move createPortal into core if it's this tiny. It would mean we don't have to export `Root`.

src/create-element.js
```diff
@@ -93,6 +93,10 @@ export function Fragment(props) {
 	return props.children;
 }
 
+export function Root(props) {
```
developit, line 96: What's the reason for having the separate constructor here, versus just reusing Fragment?

If possible, I would really like to avoid exporting `Root` since it's an implementation detail. I think there might be a way to use a signal property no a Fragment vnode.

src/index.js
```diff
@@ -1,14 +1,14 @@
...
 	Fragment,
+	Root,
 	createRef,
 	isValidElement
 } from './create-element';
 export { Component } from './component';
 export { cloneElement } from './clone-element';
 export { createContext } from './create-context';
 export { toChildArray } from './diff/children';
-export { unmount as __u } from './diff';
```
developit, line 13: very happy about this one.

src/render.js
```diff
@@ -1,68 +1,46 @@
...
- * @param {import('./internal').PreactElement} parentDom The DOM element to
- * update
- */
-export function hydrate(vnode, parentDom) {
-	render(vnode, parentDom, IS_HYDRATE);
+		// Flush all queued effects
+		commitRoot(commitQueue, vnode);
+	}
+	return {
+		hydrate(vnode) {
```
developit, line 41: I'm thinking it might be smaller when gzipped to duplicate `render()` here as hydrate(), with the `isHydrating` checks all inlined?

test/browser/focus.test.js
```diff
@@ -1,5 +1,11 @@
 import { setupRerender } from 'preact/test-utils';
-import { createElement, render, Component, Fragment, hydrate } from 'preact';
+import {
+	createElement,
+	createRoot,
+	Component,
+	Fragment,
+	hydrate
```
developit, line 7: Doesn't exist

```suggestion
	Fragment
```

compat/src/index.js
```diff
@@ -66,7 +66,10 @@ function cloneElement(element) {
...
 function unmountComponentAtNode(container) {
 	if (container._children) {
-		preactRender(null, container);
+		createRoot(container).render(null);
+		return true;
+	} else if (container._root) {
+		container._root.render(null);
 		return true;
 	}
 	return false;
```
developit, line 75: I think this may compile to smaller output:

```js
function unmountComponentAtNode(container) {
	let root = container._root || (container._children && createRoot(container));
	return root && (root.render(null), true);
}
```

developit:
Regarding removing the `Root` export thing, here are some options:

```js
// compat/src/portals.js:
import { Fragment } from 'preact';
export function createPortal(vnode, container) {
	const wrapper = createElement(Fragment, null, vnode);
	wrapper._parentDom = container;
	return wrapper;
}

// src/diff/index.js: (L193)
			if (newVNode._parentDom) {
				parentDom = newVNode._parentDom;
			}

// src/diff/index.js: (L469)
	if (!skipRemove && typeof vnode.type != 'function') {
		skipRemove = (dom = vnode._dom) != null;
	} else if (vnode._parentDom) {
		skipRemove = false;
	}
```

src/render.js
JoviDeCroock: it didn't turn out smaller (+3B)

JoviDeCroock (approved):

Issue #2621, opened 2020-07-12
The road to Preact 11


marvinhagemeister:
With Preact X being a great success there The 10.x release line was all about increasing ecosystem compatibility and the initiative was a complete success. Compared to Preact 8.x there are many more third party libraries that work out of the box with Preact, we saw the addition of Fragments, hooks, a revamped [devtools extension](https://preactjs.github.io/preact-devtools/), [prefresh](https://github.com/JoviDeCroock/prefresh) for native HMR support and much more.

In summary, we learned a ton the past year whilst maintaining Preact X and we want to push Preact even further. So the next question for us is: Where do we go from here? What should a successor look like?

This is a collection of some of our thoughts, but is in no means a concrete feature list for Preact 11. Please keep that in mind.

## Size reductions

Our small (smol?) size has always been one of our strong points and the thing I'm personally the most proud of. With an ever growing feature-set we're currently very close to the 4kB mark and would like to bring that number down again. Whilst we may be able to shave of a couple bytes here and there, I think we won't have substantial changes on that front without thinking critical of each feature.

### Only include what's actually used

The most common talking point is that many projects (including preact-devtools) don't use a single class component. The natural question that arises here is why we need to still pay for the cost the class API regardless. Ideally only the features that are actually used should be shipped to users.

The same is true for `preact/compat` which will always include side-effects even though a user may only needs to import to the `Portal` component for example. There have been some attempts to do that in the past, but to me it seems like some changes in core could make that easier.

The overall motto should be: Only include what's used.

### Move `IS_NON_DIMENSIONAL` to compat

Whilst it sounded cute on paper it has turned into a growing [list of properties](https://github.com/preactjs/preact/blob/493711603c81adc4457415ed63573d823d60e49d/src/constants.js#L3) (see #2608 ) which is something we never wanted to have in core. The reason we kept over from the 8.x line in core, was that it allowed us to keep old codepen demos working. This was especially useful at a time where we weren't officially committing to Preact X and we were more just playing around with various approaches.

I'd even go so far as to consider it harmful as I've seen it confuse new developers first-hand when switching between writing CSS and declaring inline styles. It's a feature we can remove entirely because of React, but moving over to compat should be fine.

## Reconciler performance

[...]

Issue #2607, opened 2020-07-08
IS_NON_DIMENSIONAL has false positives and false negatives


kripod:
Recently, I wrote [some tests](https://github.com/kripod/otion/blob/365743c6f7374355bc8b0032b7b220a24fb3a009/packages/otion/src/propertyMatchers.test.ts) for a regular expression with functionality similar to [`IS_NON_DIMENSIONAL`](https://github.com/preactjs/preact/blob/ff0bf291d7539c43fdecfddd5e0c10d9de87635f/src/constants.js#L3). Out of curiosity, I ran the same tests against Preact's implementation and found the following results:

### False positives

When given a number, it _should_ be postfixed with 'px' for the CSS properties below:

- `grid-auto-columns`
- `grid-auto-rows`
- `grid-template-columns`
- `grid-template-rows`

The properties above _don't accept_ unitless values. However, **`IS_NON_DIMENSIONAL` _matches_ them,** so they're _not postfixed_ with 'px' when given a number.

### False negatives

When given a number, it _should not_ be postfixed with 'px' for the CSS properties below:

- `flex-grow`
- `flex-shrink`
- `font-size-adjust`
- `font-weight`
- `line-height`
- `line-clamp`
- `-webkit-line-clamp`
- `animation`
- `border-image`
- `border-image-outset`
- `border-image-slice`
- `border-image-width`
- `shape-image-threshold`
- `scale`
- `mask-border`
- `mask-border-outset`
- `mask-border-slice`
- `mask-border-width`
- `max-lines`
- `tab-size`
- `columns`
- `column-count`
- `initial-letter`

The properties above _accept_ unitless values. However, **`IS_NON_DIMENSIONAL` _doesn't match_ them,** so they're _mistakenly postfixed_ with 'px' when given a number.

### Details of the test

```
✓ regexp matches unitless-valued 'flex' CSS property (1 ms)
✕ regexp matches unitless-valued 'flex-grow' CSS property (2 ms)
✕ regexp matches unitless-valued 'flex-shrink' CSS property
✕ regexp matches unitless-valued 'font-size-adjust' CSS property (1 ms)
✕ regexp matches unitless-valued 'font-weight' CSS property
✓ regexp matches unitless-valued 'grid-area' CSS property
✓ regexp matches unitless-valued 'grid-column' CSS property
✓ regexp matches unitless-valued 'grid-column-end' CSS property (1 ms)
✓ regexp matches unitless-valued 'grid-column-start' CSS property
[...]

PR #2629, opened 2020-07-15
Move IS_NON_DIMENSIONAL to compat

marvinhagemeister:
This PR moves the `IS_NON_DIMENSIONAL` regex to compat as described here: #2621 

- [x] Move regex to `compat`
- [x] Add migration warning to `preact/debug`
- [x] Integrate enhancements from #2607 
- [x] Go through CSS spec and add tests for every numeric property

### DIFF
--- a/compat/src/util.js
+++ b/compat/src/util.js
@@ -21,3 +21,5 @@ export function shallowDiffers(a, b) {
 	for (let i in b) if (i !== '__source' && a[i] !== b[i]) return true;
 	return false;
 }
+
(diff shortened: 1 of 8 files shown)


compat/src/util.js
```diff
@@ -21,3 +21,5 @@ export function shallowDiffers(a, b) {
 	for (let i in b) if (i !== '__source' && a[i] !== b[i]) return true;
 	return false;
 }
+
+export const IS_NON_DIMENSIONAL = /acit|ex(?:s|g|n|p|$)|rph|grid|ows|mnc|ntw|ine[ch]|zoo|^ord|itera/i;
```
JoviDeCroock, line 25: https://github.com/preactjs/preact/issues/2607 should we transfer to this one?

PR #2655, opened 2020-07-29
allow component to be suspended multiple times

tanhauhau:
Fixes https://github.com/preactjs/preact/issues/2660

there's a `shouldComponentUpdate` or `newVNode._original === oldVNode._original` optimisation to skip diffing the vnode, but suspended vnode is in a weird state where it is the original node, but the `_dom` is null, (because the DOM element is unmounted from DOM when the component is suspended).

setting the `_original = null` can skip this optimisation, and therefore recreate the dom nodes.

### DIFF
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -27,6 +27,14 @@ function detachedClone(vnode) {
 	return vnode;
 }
 
+function removeOriginal(vnode) {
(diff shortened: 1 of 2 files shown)


tanhauhau:
yeah, i worked on #2661 branch on top of this one. let me close this one and keep #2661

PR #2658, opened 2020-07-29
fix (#2657) diffElementNodes-bug

### LINKED ISSUE #2657: setState “state.a” from '' to 0,  the dom will not rerender.
### Reproduction

When setState “state.a” from '' to 0,  the dom will not rerender.

https://codesandbox.io/s/runtime-wind-bp75k?file=/src/index.js

### Steps to reproduce

Just wait 3s.

### Expected Behavior

Render as usual.

### Actual Behavior

The view doesn't change.

perseveringman:
fix #2657 

https://codesandbox.io/s/runtime-wind-bp75k?file=/src/index.js

When setState "state.a" from '' to 0, dom.data is '', newProps is 0,  ('' == 0) is true. so the dom will not change.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -354,7 +354,7 @@ function diffElementNodes(
 	}
 
 	if (newVNode.type === null) {
-		if (oldProps !== newProps && dom.data != newProps) {
+		if (oldProps !== newProps && dom.data !== newProps) {
 			dom.data = newProps;
 		}
 	} else {
--- a/test/browser/render.test.js
+++ b/test/browser/render.test.js
@@ -41,6 +41,14 @@ describe('render()', () => {
 		logCall(Element.prototype, 'remove');
 	});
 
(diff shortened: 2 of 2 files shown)


marvinhagemeister:
Awesome, this is a really good catch 🙌 The test case in the linked codesandbox from the issue is perfect to use as the basis of a unit test. Let's add that 👍

perseveringman:
@JoviDeCroock @marvinhagemeister 
Should I write the test in 'test/browser/replaceNode.test.js' ?
Like this 
preact_test

JoviDeCroock:
No just in render somewhere at the top would be good

perseveringman:
@JoviDeCroock 
Done.
 I wrote the test in 'test/browser/render.test.js'.
Before：
屏幕快照 2020-07-30 下午3 08 09
After：
屏幕快照 2020-07-30 下午3 09 17
屏幕快照 2020-07-30 下午3 09 09

Issue #2660, opened 2020-07-29
suspend and unsuspend multiple times of the same <Suspense> ends up with a weird DOM structure


tanhauhau:
### Reproduction

https://codesandbox.io/s/strange-hooks-w4wyx?file=/src/index.js

```jsx
<Suspense fallback="Suspended">
    <Suspender1 />
    <Suspender2 />
</Suspense>
```

Suspend `<Suspender1 />` and then unsuspend, and suspend `<Suspender2 />` and unsuspend, will only end up with `<Suspender2 />` on the DOM.

### Steps to reproduce

1. Click on "Suspend 1"
1. Click on "Resolve 1"
1. Click on "Suspend 2"
1. Click on "Resolve 2"
1. See the DOM

### Expected Behavior

"Resolve 1" and "Resolve 2" should show

### Actual Behavior

Only "Resolve 2" is shown

Issue #2659, opened 2020-07-29
throw promise immediately under render do not clean up the previous DOM


tanhauhau:
### Reproduction

https://codesandbox.io/s/nifty-tesla-4g653?file=/src/index.js

### Steps to reproduce

1. click "increment" to change the props of `<Lazy />` which will cause it to throw a promise.
1. notice that "suspended" is added after "Lazy 1"

### Expected Behavior

Should hide "Lazy 1" when suspended.

tanhauhau:
Also if you try to unmount the App, will end up with stack overflow.

Issue #2508, opened 2020-04-28
Suspense duplicating content on refresh


ondratra:
Suspense doesn't work as expected when throwing promises multiple times. In such case(s) content gets duplicated - it seems that the old component disconnects from its parent aand stays in the dom.

This is not happening with React's Suspense.

### Reproduction
Example of the problem in Preact:
https://codesandbox.io/s/amazing-jennings-6b21f?file=/src/index.jsx

The same example in React: (elements don't get duplicated)
https://codesandbox.io/s/prod-shape-345d2?file=/src/App.jsx

### Steps to reproduce

Throw a promise multiple times from a component inside of a `Suspense`. 

```
import { Suspense, useState } from "preact/compat";
import { Component, render } from "preact";

export default class App extends Component {
  render(props, { results = [] }) {
    return (
      <Suspense fallback="loading ...">
        <MyWrapper />
      </Suspense>
    );
  }
}

if (typeof window !== "undefined") {
  render(<App />, document.getElementById("root"));
}

function MyWrapper() {
  const [refreshState, onRefreshStateChange] = useState(0);
  const forceRefresh = () => onRefreshStateChange(refreshState + 1);

  return <MyAsyncComponent forceRefresh={forceRefresh} />;
}

let uglyCache = null;
function MyAsyncComponent(props) {
  if (uglyCache) {
    const tmp = uglyCache;
    uglyCache = null;
    return tmp;
  }

  const asyncLoad = async () => {
    await artificialDelay();

    const result = (
      <div>
        My happy component
        <button onClick={props.forceRefresh}>click me</button>
      </div>
    );
    uglyCache = result;
  };

  throw asyncLoad();
}

async function artificialDelay(time) {
  return new Promise(resolve => {
    setTimeout(() => resolve(), time);
  });
}

```

### Expected Behavior

Content will not be duplicated.

### Actual Behavior

Content gets duplicated on each `throw` promise after the first one.

ondratra:
The issue might be related to #2504 and #2488.

sventschui:
Something goes crazy with unmounting/parking the old vnodes. When wrapping the `<MyWrapper />` in a `<div>` it is at least unmounted while the fallback is rendered but the old DOM node is inserted again after the suspension clears. I'll try to dig into this the following days.

Btw. seems we have a test for this case but it doesn't work 😓

PR #2661, opened 2020-07-29
allow the same component to be suspended multiple times

tanhauhau:
Fixes https://github.com/preactjs/preact/issues/2659
Fixes https://github.com/preactjs/preact/issues/2508
Fixes https://github.com/preactjs/preact/issues/2660

build on top of https://github.com/preactjs/preact/pull/2655

Fixes 1
- when suspending a component will rendering, the `newVNode` does not have `_dom` or `_children` yet, which causes it to not be able to unmount from the DOM nicely.

Fixes 2
- suspending the same component multiple times may cause the infinte loop while `suspendingComponent.componentWillUnmount`. Need to reset the `componentWillUnmount` back when the suspend promise resolves.

Fixes 3
- node will be gone after suspend and unsuspend a sibling node
- there's a `shouldComponentUpdate` or `newVNode._original === oldVNode._original` optimisation to skip diffing the vnode, but suspended vnode is in a weird state where it is the original node, but the `_dom` is null, (because the DOM element is unmounted from DOM when the component is suspended).
- setting the `_original = null` can skip this optimisation, and therefore recreate the dom nodes.

### DIFF
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -27,6 +27,14 @@ function detachedClone(vnode) {
 	return vnode;
 }
 
+function removeOriginal(vnode) {
+	if (vnode) {
+		vnode._original = null;
+		vnode._children = vnode._children && vnode._children.map(removeOriginal);
+	}
+	return vnode;
+}
+
 // having custom inheritance instead of a class here saves a lot of bytes
 export function Suspense() {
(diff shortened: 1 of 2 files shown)


sventschui:
Waiting for CI and will merge then (except @JoviDeCroock is faster again :P )

PR #2665, opened 2020-07-30
Golfing: Par 4

developit:


compat/src/Children.js
```diff
@@ -16,11 +13,7 @@ export const Children = {
 		return children ? toChildArray(children).length : 0;
 	},
 	only(children) {
-		children = toChildArray(children);
-		if (children.length !== 1) {
-			throw new Error('Children.only() expects only one child.');
```
marvinhagemeister, line 21: Iirc we had a bug report to add this specific error to match React's behavior.

marvinhagemeister:
This is fantastic! :raised_hands:

cristianbote:
boom! 💥 🔥  😄

compat/src/Children.js
sventschui: Stumbled upon that change this morning too. I really vote for keeping this check as React does does this check in production mode so we should do the check in compat too. I guess here the compatibility is more important than the byte wins. Maybe we could add a minified error message to save some bytes?

compat/src/memo.js
```diff
@@ -27,8 +27,8 @@ export function memo(c, comparer) {
 		this.shouldComponentUpdate = shouldUpdate;
 		return createElement(c, props);
 	}
-	Memoed.prototype.isReactComponent = true;
 	Memoed.displayName = 'Memo(' + (c.displayName || c.name) + ')';
```
sventschui, line 30: Things like this make me wonder whether we should ship a "development" and a "production" version. I remember the discussions about making things more complex bundling wise but couldn't we omit this line in a "production" or "minimal" variant of our build? That would save us some bytes here and there.

Or are we somehow able to move this line to debug? I though about monkey patching the exported function but that only works in CJS and not ESM AFAIK.

compat/src/portals.js
```diff
@@ -1,78 +1,76 @@
...
 }
 
 /**
  * Portal component
  * @param {object | null | undefined} props
+ *
+ * TODO: this could use the "fake root node" trick from the partial hydration demo
  */
 function Portal(props) {
-	let _this = this;
```
sventschui, line 17: What is the reason for dropping this? Would save us 2 bytes keeping the `let _this = this` and using `_this` afterwards

compat/src/Children.js
developit: Yeah I was thinking it must have been. Does it need to be the exact same string though?

compat/src/portals.js
developit: yeah I was 50/50 on this. I think some more refactoring can get the total down, but the `_this` alias was masking some of the other possible changes. I was mostly just trying to switch out arrow functions that had the possibility to cause accidental closures for `this` access.

compat/src/memo.js
developit: I was pondering this. There's no usable way to ship debug variants today, and we get a lot of value out of avoiding pushing that stuff onto users. But I think there may be a compromise solution possible: `preact/debug` can actually set this on its own:

```js
// memo.js
Memoed.displayName = 'Memo(';
```

```js
// forwardRef.js
Forwarded.displayName = 'forwardRef(';
```

```js

options.vnode = (vnode) => {
  if (typeof vnode.type === 'function') {
    // turn "Memo(" into "Memo(ChildName)" the first time we see a constructor
    if (/\(/.test(vnode.type.displayName)) {
      const child = vnode.type.call({}).type;
      vnode.type.displayName += (child.displayName || child.name) + ')';
    }
  }
};
```

marvinhagemeister: It would require us striping `forwardRef` and `memo` out of compat if I'm reading that correctly, right?

compat/src/Children.js
developit: Update: I changed it to throw a string with a very cheap reference to Children.only.

compat/src/memo.js
developit: @marvinhagemeister I'd hoped the above would avoid that - instead of relying on `vnode.type === Memo`, it uses the `(`-suffixed name as an indication that `displayName` should be regenerated to include the child component's displayName.

This is probably a non-zero perf hit in `preact/debug` though, with or without compat enabled. Might not be worth it.

src/diff/props.js
```diff
@@ -57,39 +56,34 @@ function setStyle(style, key, value) {
...
+	let useCapture, nameLower;
 
-	if (isSvg) {
-		if (name === 'className') {
-			name = 'class';
-		}
-	} else if (name === 'class') {
-		name = 'className';
-	}
+	if (isSvg && name == 'className') name = 'class';
```
developit, line 61: I'm curious if anyone is opposed to this change. It means we would be using `setAttribute("class")` for a `class` prop and `.className=` for a `className` prop. They're the same thing, but it's possible there's a performance difference.

developit:
Might be okay to merge this now. Thoughts?

Issue #2678, opened 2020-08-04
Preact removes option values on hydration


kenny-f:
Option values for select tags are removed during hydration

Page source:
<img width="1483" alt="Screenshot 2020-08-04 at 14 28 04" src="https://user-images.githubusercontent.com/1841819/89299570-bf1f7080-d65e-11ea-932a-ff0684186ab8.png">

Elements tab after hydration:
<img width="641" alt="Screenshot 2020-08-04 at 14 15 21" src="https://user-images.githubusercontent.com/1841819/89299523-aca53700-d65e-11ea-94a2-457ec516eb18.png">
 

### Reproduction
Clone this branch 
https://github.com/kenny-f/preact-hydration-repro/tree/hydrate-select-options-preact-only

`npm i`
`npm start`

See that the values are missing in the value output for the options (like screenshots above)

NOTE: `render` does not have this problem and works fine, only `hydrate`

kenny-loveholidays:
Update: This started happening since version `10.4.5`

PR #2679, opened 2020-08-04
avoid touching dom-attributes during hydration

JoviDeCroock:
Fixes https://github.com/preactjs/preact/issues/2678

I broke this in https://github.com/preactjs/preact/pull/2551/files#diff-a04f8653407a342eac2cd2ea6a5c3ff6R186

`render-to-string` already sets these values correctly and we shouldn't double touch it

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -183,7 +183,7 @@ export function diffChildren(
 			//
 			// To fix it we make sure to reset the inferred value, so that our own
 			// value check in `diff()` won't be skipped.
-			if (newParentVNode.type == 'option' && !isHydrating) {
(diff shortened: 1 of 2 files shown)


src/diff/children.js
```diff
@@ -183,7 +183,7 @@ export function diffChildren(
 			//
 			// To fix it we make sure to reset the inferred value, so that our own
 			// value check in `diff()` won't be skipped.
-			if (newParentVNode.type == 'option') {
+			if (newParentVNode.type == 'option' && !isHydrating) {
```
developit, line 186: I think reversing the conditions here so the isHydrating check comes first will reduce the size to ~0.

PR #2680, opened 2020-08-04
Add support for IntrinsicElements in ComponentProps

remcohaszing:
This means users can use for example `ComponentProps<'button'>` or `ComponentProps<'div'>` the same way it can be used with `@types/react`.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -75,10 +75,12 @@ declare namespace preact {
 	type ComponentType<P = {}> = ComponentClass<P> | FunctionComponent<P>;
 	type ComponentFactory<P = {}> = ComponentType<P>;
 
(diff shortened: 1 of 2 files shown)


VottusCode:
I've ran across this today and I cannot seem to use `React.ComponentProps` (`react` is aliased `preact/compat`):

```ts
import * as React from 'react';
declare function ZoomInIcon(props: React.ComponentProps<'svg'>): JSX.Element;
export default ZoomInIcon;
```
results in:

obrazek

marvinhagemeister:
@VottusCode FYI: Best to file a new issue for that. Comments on old PRs are quickly hidden in GitHub's ui.

VottusCode:
Alright.

Issue #2682, opened 2020-08-05
Material-UI's Drawer component causes props are not supported: `ref` error on first open.


joepagan:
Hi, I hope this is enough information, let me know otherwise. Thanks for your work.

This looks like another issue had the identical error in it at one point:
https://github.com/preactjs/preact/issues/2249 

Another with a similar error https://github.com/preactjs/preact/issues/1824

### Reproduction

I tried to set this up in a codesandbox app but kept getting "VNode is undefined" errors with MUI components, so I've created a repo for you instead:
https://github.com/joepagan/preact-mui-ref-issue

### Steps to reproduce

1. `npm run dev` & load http://0.0.0.0:8080
2. click menu icon in top-right
3. this should open the [temporary Drawer component](https://material-ui.com/components/drawers/#temporary-drawer)
4. Drawer opens with console error

### Expected Behavior

no console error

### Actual Behavior

console error on draw open, it does not occur again when closed or re-opened:

```
Failed prop type: The following props are not supported: `ref`. Please remove them. check-props.js:43
    __b check-props.js:43
    __b check-props.js:25
    __b forwardRef.js:6
    T index.js:70
    g children.js:143
    T index.js:187
    g children.js:143
    T index.js:187
    g children.js:143
    T index.js:187
    g children.js:143
    T index.js:187
    g children.js:143
    j index.js:390
    T index.js:237
    g children.js:143
    j index.js:390
    T index.js:237
    g children.js:143
    T index.js:187
    g children.js:143
    T index.js:187
    g children.js:143
    j index.js:390
    T index.js:237
    g children.js:143
    T index.js:187
    g children.js:143
    T index.js:187
    g children.js:143
    T index.js:187
    g children.js:143
    T index.js:187
    g children.js:143
    j index.js:390
    T index.js:237
    g children.js:143
    T index.js:187
    m component.js:216
    m component.js:211
    (Async: promise callback)
    w component.js:194
    setState component.js:30
    setState debug.js:340
    __ index.js:134
    toggleDrawer index.js:61
    N props.js:159
    (Async: EventListener.handleEvent)
    C props.js:90
    A props.js:24
    j index.js:390
    T index.js:237
    g children.js:143
    T index.js:187
    g children.js:143
    T index.js:187
    g children.js:143
    T index.js:187
    g children.js:143
    T index.js:187
    g children.js:143
    T index.js:187
    g children.js:143
    j index.js:390
    T index.js:237
    g children.js:143
    j index.js:390
[...]

PR #2685, opened 2020-08-07
preact/debug: Add component stack to prop type validation error, do not pass ref to prop-types validation (fixes mui incompatbility)

sventschui:
Add the component stack to the prop types validation error message to help debug #2682 

Not sure that inline arrow function is the best solution but I guess it is preferable over `.bind(null, vnode)`

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -11,6 +11,7 @@ import {
 	getCurrentVNode,
 	getDisplayName
 } from './component-stack';
+import { assign } from './util';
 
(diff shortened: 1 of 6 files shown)


debug/src/debug.js
```diff
@@ -215,11 +216,19 @@ export function initDebug() {
 					);
 				}
 			}
+
+			let values = vnode.props;
+			if (vnode.type._forwarded) {
```
marvinhagemeister, line 221: We need to add `_forwarded` to debug's `mangle.json`

PR #2686, opened 2020-08-07
Support nested portals targeting the same container

### LINKED ISSUE #2669: [compat] portals not unmounting
Not really sure if this is on people's radar or if there's an easy workaround I'm not seeing but issue is still open in old compat repository and I've just encountered this bug with latest version.

### Reproduction
https://github.com/preactjs/preact-compat/issues/501

### Steps to reproduce
See the first link.

### Expected Behavior
Portal should unmount if it shouldn't be rendered.

### Actual Behavior
Portal doesn't unmount if e.g `flag === false`.
```jsx
{flag && (
  ReactDOM.createPortal(
    <PortalContent />,
    document.body
  )}
)}
```

sventschui:
Support nested portals targeting the same container. Fixes #2669

### DIFF
--- a/compat/src/portals.js
+++ b/compat/src/portals.js
@@ -36,9 +36,10 @@ function Portal(props) {
 		if (!_this._hasMounted) {
 			// Create a placeholder that we can use to insert into.
 			_this._temp = document.createTextNode('');
+			// temporarily store the current children of the container to restore them after render
+			_this._children = container._children;
 			// Hydrate existing nodes to keep the dom intact, when rendering
 			// wrap into the container.
-			_this._children = container._children;
 			hydrate('', container);
 			// Append to the container (this matches React's behavior)
 			container.appendChild(_this._temp);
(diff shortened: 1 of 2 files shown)


compat/src/portals.js
```diff
@@ -36,6 +36,8 @@ function Portal(props) {
 		if (!_this._hasMounted) {
 			// Create a placeholder that we can use to insert into.
 			_this._temp = document.createTextNode('');
+			// temporarily store the current children of the container to restore them after render
+			_this._children = container._children;
```
marvinhagemeister, line 40: That's genuis! Love this :+1:

Aurailus:
Hey all, I'm still getting this issue in 10.5.7 using functional components. Is it possible this fix is broken?

andrewiggins:
Hey @Aurailus, well that's no good. Could you post a repro in a new issue? It's possible our tests have missed your use case and it'd be great to capture it in test so we can fix it an ensure it stays fixed.

PR #2687, opened 2020-08-07
Do not use defaultValue when value is false

sventschui:
Found this while checking our code coverage report.

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -114,6 +114,9 @@ options.vnode = vnode => {
 		if (typeof type != 'function') {
 			// Apply defaultValue to value
 			if (props.defaultValue && props.value !== undefined) {
+				if (props.value === null) {
+					props.value = props.defaultValue;
+				}
 				delete props.defaultValue;
 			}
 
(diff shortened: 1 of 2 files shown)


compat/src/render.js
```diff
@@ -114,9 +114,6 @@ options.vnode = vnode => {
 		if (typeof type != 'function') {
 			// Apply defaultValue to value
 			if (props.defaultValue && props.value !== undefined) {
```
developit, line 116:
```suggestion
			if (props.defaultValue && props.value != null) {
```

compat/test/browser/render.test.js
```diff
@@ -98,14 +99,77 @@ describe('compat render', () => {
...
 
-	it('should ignore defaultValue when value is 0', () => {
-		render(<input defaultValue={2} value={0} />, scratch);
-		expect(scratch.firstElementChild.value).to.equal('0');
+	it('should support defaultValue on textarea when value is omitted', () => {
+		render(<textarea defaultValue="foo" />, scratch);
+		expect(scratch.firstElementChild).to.have.property('innerHTML', 'foo');
+	});
+
+	[undefined, null].forEach(val => {
```
sventschui, line 112: I'll flatten these tests out when we agree on the implementation

JoviDeCroock:
Closing: superseded — v11 compat handles value/defaultValue via `value == null` checks.

PR #2648, opened 2020-07-24
default useRef to null

JoviDeCroock:
Fixes https://github.com/preactjs/preact/issues/2647

Issue #2647, opened 2020-07-24
Feature request. useRef should accept "undefined"


o-alexandrov:
### Reproduction

preact's version: `10.4.6`

### Steps to reproduce

1. Initialize `useRef` as follows and use it:
```
const refExample = useRef()
```
it will trigger a warning when using it with `preact/compat`, for example with the following hook:
- [useScroll](https://github.com/streamich/react-use/blob/master/src/useScroll.ts)

Warning:
```
`useScroll` expects a single ref argument.
```

### Expected Behavior

`preact` shouldn't trigger a warning and handle initialized `useRef()`

### Actual Behavior

`react` doesn't trigger a warning whereas `preact` does.

PR #2689, opened 2020-08-08
Go back to undefined as default value for useRef

### LINKED ISSUE #2688: useRef behavior mismatches react and typing
Due to #2648, when no parameter passed to useRef hook, `ref.current` will be null instead of undefined, which is not only different from react's `useRef`, but also contradicts to what's defined in .d.ts that is
```ts
type PropRef<T> = { current?: T };
```

marvinhagemeister:
Turns out that library from #2647 (`react-use`) has a bug and prints out the error with React, too. Double checked and React uses `undefined` as the default value instead of `null`.

This partially reverts #2648 .

Fixes #2688 .

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -174,10 +174,7 @@ export function useLayoutEffect(callback, args) {
 
 export function useRef(initialValue) {
 	currentHook = 5;
-	return useMemo(
-		() => ({ current: initialValue === undefined ? null : initialValue }),
-		[]
-	);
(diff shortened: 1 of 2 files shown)


zachasme:
The change to `null` as default value for `useRef` also broke `@preact/urql`.

In [this codesandbox](https://codesandbox.io/s/crazy-ramanujan-x273f?file=/src/index.js), reverting to `1.4.6` works, while `1.4.7` throws.

PR #2701, opened 2020-08-17
Prevent cursor jumps inside contenteditable

### LINKED ISSUE #2691: contenteditable broken with Preact 10
### Reproduction

```
<!DOCTYPE html>
<meta charset="UTF-8">
<style>
	span {
		border: 1px solid #000;
	}
	span:empty::after {
		content: 'placeholder';
		font-size: smaller;
		left: 0;
		opacity: 0.5;
		top: 0;
	}

</style>
<!-- <script src="https://cdnjs.cloudflare.com/ajax/libs/preact/8.5.3/preact.dev.js" integrity="sha512-X4tdt+//Fj3d6uBdK8YH5qHT7g0mDWf4ZOJS5UQj7OJu+QAy53ySvOoAe1iky/oZ8JaUaHLAHI2swH2SGnGzWg==" crossorigin="anonymous"></script> -->
<script src="https://cdnjs.cloudflare.com/ajax/libs/preact/10.4.7/preact.umd.min.js" integrity="sha512-dHedEHbrmayd9MzZJGcsUVzD2to6afhlJiXkDYzUBl3qrNOFE4nK1BsPnzSve6K3vTlMcmQLToPJGxs3gbsyew==" crossorigin="anonymous"></script>
<script>
	const { Component, h, render } = self.preact;

	class Editable extends Component {
		constructor(props) {
			super(props);
			const { value } = props;
			this.setState({ value });

			this.onInput = (event) => {
				const value = event.target.innerText;
				this.setState({ value });
			};
		}

		render(props, { value }) {
			return h('div', {},
				h('span', {
					contenteditable: true,
					onInput: this.onInput,
				}, value),
			);
		}
	}

	document.addEventListener('DOMContentLoaded', () => {
		render(h(Editable, {}), document.body);
	});
</script>
```

### Steps to reproduce
1. Apply the above HTML code (using Preact 10, where Preact 8 is commented).
2. Click inside placeholder box and start typing.

### Expected Behavior
The letters typed should be echoed inside the span as you type them. This used to work correctly in Preact 8 (try commenting Preact 10 and uncommenting Preact 8).

### Actual Behavior
Typed letters are replicated n times, where n is the number of characters typed.
The problem happens when the initial state (derived from props) has an empty `value`.

sventschui:
This prevents cursor jumps inside contenteditable. It seems a bit expensive to check for innerHTML equality but the only way I see to get this fixed.

Fixes #2691


test/browser/render.test.js
```diff
@@ -789,6 +791,50 @@ describe('render()', () => {
...
+		range.selectNodeContents(editable);
+		const sel = window.getSelection();
+		sel.removeAllRanges();
+		sel.addRange(range);
+
+		expect(window.getSelection().getRangeAt(0).endOffset).to.equal(1);
+
+		// dispatch a change event
+		editable.innerHTML = 'World';
+		editable.dispatchEvent(createEvent('input'));
```
andrewiggins, line 824: Hmm I don't think I understand this test. How can a real user fire an input event or change the innerHTML but simultaneously keep text selection. My understanding is that all browsers would clear the text selection the moment the user presses a key while selecting text.

sventschui: Yeah you're right. I forgot to get back to this. I originally wanted to use the range to place the caret (see https://stackoverflow.com/a/6249440/2164347) but failed. I will try again to make the test clear. In the end this just makes sure we do not touch innerHTML.

sventschui: I modified the test to be more realistic. It now ensures the caret stays where it was prior to the `input` event

wbercx:
Awesome work! This seems to take `draft-js` from being _completely unusable_ to actually working.

JoviDeCroock (approved):

marvinhagemeister (approved):
This is amazing 🙌

PR #2705, opened 2020-08-19
Backing nodes v11

JoviDeCroock:


src/create-element.js
```diff
@@ -100,3 +79,32 @@ export function Fragment(props) {
...
+		// _nextDom must be initialized to undefined b/c it will eventually
+		// be set to dom.nextSibling which can return `null` and it is important
+		// to be able to distinguish between an uninitialized _nextDom and
+		// a _nextDom that has been set to `null`
+		_nextDom: undefined,
+		_component: null
+	};
+
+	if (vnode.props && vnode.props.ref) {
+		delete vnode.props.ref;
```
developit, line 102: It's likely much faster to loop over the props and copy them. Does that end up being much larger?

One perk of the loop is that we don't have to have a megamorhpic property access for `props.ref` and `props.key` - they can be local variables that get assigned to if the loop encounters `key` or `ref` keys.

src/diff/children.js
```diff
@@ -137,13 +114,23 @@ export function diffChildren(
 			}
 		}
 
-		oldVNode = oldVNode || EMPTY_OBJ;
+		const old = (oldVNode && oldVNode._node) || EMPTY_OBJ;
+		const oldVNodeDom = oldVNode && oldVNode._dom;
+		const oldRef = oldVNode && oldVNode.ref;
+		if (oldVNode == null || !childVNode._node) {
+			childVNode = createBackingNode(childVNode);
```
developit, line 121: I got stuck on this for a sec -  think we could use a second variable for the backing node so that it's clear which is which? Avoids accidentally trying to access backing node properties on a VNode.

src/diff/index.js
```diff
@@ -310,7 +304,7 @@ function diffElementNodes(
 ) {
 	let i;
 	let oldProps = oldVNode.props;
-	let newProps = newVNode.props;
+	let newProps = (newVNode.props = newVNode._node.props);
```
developit, line 307: Doesn't this skip the prop normalization that is done in createBackingNode?

developit:
Pretty nice size drop...

src/diff/index.js
JoviDeCroock: Yes still figuring it all out, it's pretty hard here since we don't have a sep `create` step

PR #2707, opened 2020-08-19
Remove errorInfo type from componentDidCatch

remcohaszing:
This parameter exists in the React equivalent, but not in Preact.


JoviDeCroock:
This will make our types incompatible which doesn't really feel right, I think we had to add this to avoid type-errors.

rschristian:
Looks like support for this parameter landed in #3452, so I'll close this on out

PR #2717, opened 2020-08-26
Fix state turned readonly in update fn

### LINKED ISSUE #2716: Compatibility issues with Immer?
### Reproduction

Barebones code sandbox link:
https://codesandbox.io/s/bitter-frost-xc5hd?file=/src/index.js

### Steps to reproduce

To reproduce, simply click on the `Change` button twice.

### Expected Behavior

The app should not crash.

### Actual Behavior

The app crashes.

Screen Recording 2020-08-25 at 05 53 31 PM

image

Hi! I believe there is a compatibility issue between `preact` and `immer` when you try to use Immer's curried producer, which is referenced in their code sandbox link here: 

https://immerjs.github.io/immer/docs/example-setstate
https://codesandbox.io/s/m4yp57632j

marvinhagemeister:
Some libraries like `immer` mutate the current state object and mark all properties as `readonly`. This broke our update function.

Fixes #2716.

### DIFF
--- a/test/browser/components.test.js
+++ b/test/browser/components.test.js
@@ -2572,6 +2572,35 @@ describe('Components', () => {
 			expect(scratch.innerHTML).to.equal('baz');
 			expect(callbackState).to.deep.equal({ foo: 'baz' });
 		});
+
+		// #2716
+		it('should work with readonly state', () => {
+			let update;
+			class Foo extends Component {
+				constructor(props) {
+					super(props);
+					this.state = { foo: 'bar' };
+					update = () =>
+						this.setState(prev => {
+							Object.defineProperty(prev, 'foo', {
+								writable: false
+							});
+
+							return prev;
+						});
+				}
+
+				render() {
+					return <div />;
+				}
+			}
+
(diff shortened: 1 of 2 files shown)


test/browser/components.test.js
```diff
@@ -2572,6 +2572,35 @@ describe('Components', () => {
...
+
+		// #2716
+		it('should work with readonly state', () => {
+			let update;
+			class Foo extends Component {
+				constructor(props) {
+					super(props);
+					this.state = { foo: 'bar' };
+					update = () =>
+						this.setState(prev => {
```
sventschui, line 2584: Does it also work when passing an object with read only props to `setState`? I.e. `const foo = {}; Object.defineProperty(....); setState(foo)`

JoviDeCroock (approved):

test/browser/components.test.js
marvinhagemeister: Just checked and that scenario is unaffected, because we create a new state copy immediately inside `setState` 👍

PR #2721, opened 2020-08-27
lazy controlled inputs

developit:
This doesn't work in 100% of cases, like when an event handler is registered on an input but does not trigger rendering.


JoviDeCroock:
Closing: abandoned draft; lazy controlled inputs did not work in all cases, and the v10 diff was rewritten in v11.

Issue #2722, opened 2020-08-28
IE11 throws errors when using the style attribute


charlier:
### Reproduction
In IE11, if you have a style attribute in your application, Preact throws an error.
This looks like a regression introduced in 10.4.8, as reverting to 10.4.7 yields happiness.

### Steps to reproduce
1. Create something with a style attribute
```
const App = () => (
	<div style="display: block;">
		Hello World!
	</div>
)
```
2. Open your app in IE11
3. Check for errors

### Expected Behavior
The application renders correctly, without any errors.

### Actual Behavior
IE throws an error
> Assignment to read-only properties is not allowed in strict mode

A workaround is to just use classes.

PR #2723, opened 2020-08-28
Avoid assigning to readonly style

JoviDeCroock:
Fixes: https://github.com/preactjs/preact/issues/2722

@charlier if you  get the chance you can use our pika-distribution (one of the PR-checks) to see if it works locally for you.


marvinhagemeister (approved):
Good catch! I recall us having assigned to `cssText` earlier in our code base 👍

klh:
this causes IE11 to fail making preact 10.4.7 and 10.4.8 incompatible with IE11.

JoviDeCroock:
@klh no this fixes that

PR #2725, opened 2020-08-28
fix rendering children as zero number

JiLiZART:
React allows pass `0` (number, not string) as children prop and properly render it
React sandbox example
https://codesandbox.io/s/cranky-dust-7tjqw

### DIFF
--- a/compat/src/Children.js
+++ b/compat/src/Children.js
@@ -1,11 +1,11 @@
 import { toChildArray } from 'preact';
 
 const mapFn = (children, fn) => {
-	if (typeof children === 'undefined' || children === null) return null;
+	if (children === undefined || children === null) return null;
 	return toChildArray(toChildArray(children).map(fn));
 };
 
-// This API is completely unnecessary for Preact, so it's basically passthrough.
+// This API is completely unne`cessary for Preact, so it's basically passthrough.
 export const Children = {
 	map: mapFn,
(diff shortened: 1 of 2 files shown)


compat/src/Children.js
```diff
@@ -1,7 +1,7 @@
 import { toChildArray } from 'preact';
 
 const mapFn = (children, fn) => {
-	if (!children) return null;
+	if (typeof children === 'undefined' || children === null) return null;
```
JoviDeCroock, line 4:
```suggestion
	if (children === undefined || children === null) return null;
```

JiLiZART: I think this check more robust, since we can overwrite `undefined`variable.

compat/src/Children.js
```diff
@@ -1,7 +1,7 @@
 import { toChildArray } from 'preact';
 
 const mapFn = (children, fn) => {
-	if (!children) return null;
+	if (children === undefined || children === null) return null;
```
JoviDeCroock, line 4: Found a shorter equivalent, thanks for notifying us about this bug!

```suggestion
	if (children == null) return null;
```

JoviDeCroock (approved):
Awesome, thank you so much! 💯

marvinhagemeister (approved):
Sweet, thank you so much for the PR 🙌

Issue #2706, opened 2020-08-19
Remove irrelevant types from JSX namespace


remcohaszing:
The JSX namespace in TypeScript has a special meaning. There’s a specific set of types that can be defined in a JSX namespace to make JSX behave in a certain way. This is (partially) documented [here](https://www.typescriptlang.org/docs/handbook/jsx.html).

The following properties have a special meaning:

- `Element`
- `ElementClass`
- `ElementAttributesProperty`
- `ElementChildrenAttribute`
- `IntrinsicAttributes`
- `IntrinsicElements`
- `LibraryManagedAttributes`

However, in the Preact typings the following types are also defined on the JSX namespace:

- `SVGAttributes`
- `PathAttributes`
- `DOMAttributes`
- `HTMLAttributes`

**Events and event handlers**

- `TargetedEvent`
- `TargetedAnimationEvent`
- `TargetedClipboardEvent`
- `TargetedCompositionEvent`
- `TargetedDragEvent`
- `TargetedFocusEvent`
- `TargetedKeyboardEvent`
- `TargetedMouseEvent`
- `TargetedPointerEvent`
- `TargetedTouchEvent`
- `TargetedTransitionEvent`
- `TargetedUIEvent`
- `TargetedWheelEvent`
- `EventHandler`
- `AnimationEventHandler`
- `ClipboardEventHandler`
- `CompositionEventHandler`
- `DragEventHandler`
- `FocusEventHandler`
- `GenericEventHandler`
- `KeyboardEventHandler`
- `MouseEventHandler`
- `PointerEventHandler`
- `TouchEventHandler`
- `TransitionEventHandler`
- `UIEventHandler`
- `WheelEventHandler`

The JSX namespace in Preact is called `JSXInternal`. This makes it unclear if these types are supposed to be consumed or not.

Since these types are useful, especially the event handlers, I propose to move all these types to the `preact` namespace instead. The `JSXInternal` namespace can import these types an expose type aliases using the `@deprecated` annotation. They can be removed in the next major release.

marvinhagemeister:
That's awesome, sounds like you spent already a bit of time going through everything. Would love to have a PR for this :+1:

andrewiggins:
Makes sense to me. Seems this [matches what the React types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d48442e687742aceaea97f5b515b20c4b3a18a9c/types/react/index.d.ts#L2941) are doing as well so we are in good company to move these out.

[...]

PR #2727, opened 2020-08-29
Infer JSX.IntrinsicElements from DOM lib

remcohaszing:
HTML elements are inferred from the `HTMLElementTagNameMap` in `dom`, and the SVG elements are inferred from `SVGElementTagNameMap`.

The `HTMLMarqueeElement` exists in `dom` and is no longer used, so it has been removed.

There’s more to the cleanup than I initially anticipated, so I decided to do this gradually.

Refs #2706


marvinhagemeister:
Sorry, for leaving this open for so long. I really want to get this in but haven't had the time to properly review it yet. It's pretty much at the top of my todo list.

rschristian:
I'm going to close this as it's mostly been eclipsed by #4546, though it will be kept in mind.

This mostly does the job but TS DOM typings are a bit slow to show up in some cases and users generally want to import interfaces/types as they do in React, both which make this approach a bit troublesome, though it's far less typing and I wish we had landed this back in 2020 as a stop-gap at least.

Thanks for the suggestion, albeit many years late now! Sorry we didn't get to this sooner.

Issue #2728, opened 2020-08-30
useMemo doesn't correctly detect deps changes


Wroud:
### Reproduction

https://codesandbox.io/s/preact-input-uncontrolled-forked-3pvw7

### Steps to reproduce

If we add/remove the last element of useMemo dependencies (same element) useMemo triggers only on first render and after first change

### Expected Behavior

useMemo triggers whenever deps array length changes

### Actual Behavior

useMemo not triggered when deps array length changes

Wroud:
also, I have checked the same behavior in react, and get some other behavior:
![image](https://user-images.githubusercontent.com/811729/91658693-26e8a000-ead3-11ea-94f8-5132e1022c58.png)

marvinhagemeister:
Keeping this open, because we should add a debug warning too :+1:

JoviDeCroock:
In latest this keeps changing

PR #2729, opened 2020-08-30
Add length check for deps array

JoviDeCroock:
Fixes https://github.com/preactjs/preact/issues/2728


marvinhagemeister (approved):
Sweet :raised_hands:

PR #2732, opened 2020-08-30
Debug: Print error on changing dependency array length between renders

### LINKED ISSUE #2728: useMemo doesn't correctly detect deps changes

marvinhagemeister:
This PR adds a new error log to `preact/debug` that comes into play when the dependency array length of a hook changes between renders.

Fixes #2728 .

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -149,7 +149,10 @@ export function useReducer(reducer, initialState, init) {
 export function useEffect(callback, args) {
 	/** @type {import('./internal').EffectHookState} */
 	const state = getHookState(currentIndex++, 3);
-	if (!options._skipEffects && argsChanged(state._args, args)) {
+	if (
+		!options._skipEffects &&
+		argsChanged(state._args, args, currentComponent)
+	) {
 		state._value = callback;
 		state._args = args;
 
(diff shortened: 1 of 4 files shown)


JoviDeCroock (approved):

hooks/src/index.js
```diff
@@ -346,9 +352,17 @@ function invokeEffect(hook) {
 /**
  * @param {any[]} oldArgs
  * @param {any[]} newArgs
+ * @param {import('./internal').Component} currentComponent
  */
-function argsChanged(oldArgs, newArgs) {
-	return !oldArgs || oldArgs.length !== newArgs.length || newArgs.some((arg, index) => arg !== oldArgs[index]);
+function argsChanged(oldArgs, newArgs, currentComponent) {
+	if (options._argsChanged) {
```
andrewiggins, line 358: Does it make any difference if we extend the `_hook` option to handle this case? Perhaps something like `getHookState(currentIndex++, 7, args)` (we probably only need to pass `args` if the hook has an `args` param)

Looking at it now, should we invoke that `_hook` option after we have created/retrieved the hook state so we can pass it to the hook? Sounds like that would make the `_hook` option a lot more useful/powerful.

Will make the debug code more complex but if it saves bytes in product code, maybe that's worth it

marvinhagemeister: Just did that with the latest commit, but it turns out to be bigger than expected. It's 15B larger compared to the additional option hook. Another argument in favor of keeping `options._hook` as is, is that it's a breaking change, even though it's internals. Both devtools and prefresh are built on it.

JoviDeCroock (approved):

JoviDeCroock:
Closing this as stale

Issue #2733, opened 2020-08-31
Rendering an SVG-element with xmlns:xlink-prop fails


vkentta:
### Reproduction

Try to render an SVG-element with property `xmlns:xlink="http://www.w3.org/1999/xlink"` using preact 10.4.8 and htm 3.0.4
https://codesandbox.io/s/dawn-monad-7lhh8

### Steps to reproduce

No special steps, fails on initial render.
Note that the demo works on previous versions of preact.

### Expected Behavior

SVG renders without issues.

### Actual Behavior

Rendering fails (with different error messages depending on the browser).

PR #2734, opened 2020-08-31
Fix rendering an SVG-element with xmlns:xlink-prop

vkentta:
The current implementation breaks `xmlns:xlink`-prop (chaning it to `xmlns:`).

Fixes https://github.com/preactjs/preact/issues/2733


marvinhagemeister:
Good catch! Let's add a test case for that so that we don't ever regress on this in the future and then it's ready to go in 🎉

vkentta:
@marvinhagemeister Alrighty. Do you think this would be an appropriate place for such a test: https://github.com/preactjs/preact/blob/master/test/browser/render.test.js?

marvinhagemeister:
@vkentta Yup, that's the perfect place to put it 👍

vkentta:
@marvinhagemeister looks like it would require quite some tinkering to get the test suite running on my linux install (seems to require chrome). I do not unfortunately have any more time to spend on this issue at this point.

marvinhagemeister:
@vkentta That's weird it runs fine for me under Arch Linux. What kind of error are you getting?

vkentta:
@marvinhagemeister The test command says it's starting chrome and fails with message requesting CHROME_BIN location.

developit:
`CHROME_BIN=$(which chrome) npm t`

vkentta:
I found some time to work on this bit more (got the tests running by altering `karma.conf.js` `localLaunchers` to use Chromium).

So I tried the following:
I added a `xmlnsXlink="http://www.w3.org/1999/xlink"` on this line: https://github.com/preactjs/preact/blob/master/test/browser/svg.test.js#L48
The prop seems to come through without any modification (like `viewBox`, unlike `xlinkHref`). This is another inconsistency with react it seems like: https://codesandbox.io/s/angry-leakey-g5gjw?file=/src/App.js

I tried to search the code base for `xlinkHref` to figure out what logic makes that a special case, but without success. Any help @marvinhagemeister @developit ?

marvinhagemeister:
@vkentta We don't have any sorts of mapping tables as they have a very negative effect on bundle size. That's why grepping for properties won't turn up any results. Any normalizations in core take place in `src/diff/props.js`, right where the changes of this PR are. That's where `xlinkHref` is normalized.

SVG-Attributes are special because JSX doesn't allow a `:` character in a property name. Therefore we currently strip the `xlink:` namespace prefix, because browsers don't care about that when you invoke `setAttributeNS`.

```js
node.setAttributeNS('http://www.w3.org/1999/xlink', 'xlink:href', 'example.com');
// vs
node.setAttributeNS('http://www.w3.org/1999/xlink', 'href', 'example.com');
```

They do render differently in the Browser's "Elements" panel in devtools though.

It seems like we need to extend that logic to account for the `xmlns` namespace. Haven't come up with a good way to do so far. In either way the fix will likely be around the same line as the changes in this PR.

vkentta:
@marvinhagemeister I added a test and made the functionality in line with react. It ended up being slightly more code, but hopefully not too bloated solution :slightly_smiling_face:

src/diff/props.js
```diff
@@ -149,6 +144,14 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
 	}
 }
 
+function updateAttributeNS(dom, ns, name, value) {
```
marvinhagemeister, line 147: Let's inline this function like before to save ~10 bytes. Function declarations are always somewhat expensive, so we'll only use them when really needed. The only difference between different callsites is the `ns` parameter, so we can just store it in a variable before this section :+1:

marvinhagemeister (approved):
Looking good so far! I'd love to keep the size savings like before by keeping the namespace stuff inlined :+1: 

In total this PR adds about `+63 B`. @developit Do you have more ideas to golf this down?

marvinhagemeister (commented):
Looking good so far! I'd love to keep the size savings like before by keeping the namespace stuff inlined :+1: 

In total this PR adds about `+63 B`. @developit Do you have more ideas to golf this down?

marvinhagemeister (commented):
Looking good so far! I'd love to keep the size savings like before by keeping the namespace stuff inlined :+1: 

In total this PR adds about `+63 B`. @developit Do you have more ideas to golf this down?

src/diff/props.js
vkentta: Changed it back to inlined.

developit:
@marvinhagemeister I thought this was a bug I introduced in my golfing PR that has already been fixed?

vkentta:
@developit might it be that this is the very same PR you recall? :slightly_smiling_face: It was pending for a while because of lack of tests (sorry for that).

So this is where the problem appears on client side: https://codesandbox.io/s/dawn-monad-7lhh8

developit:
It might be! No need to apologize, I was extremely slow getting around to reviewing this because my notifications setup is pretty unwieldy these days. You're right, this PR and the original issue you opened is very much the same PR was I thinking of. I've taken a few minutes to take stock of everything so I can make a more informed commentary (apologies for my terse contextless reply!).

I have two concerns relating to this change that I think make it worth reconsidering. I know it's a after-the-fact, and I apologize for being so slow on the review here.

**(1)** First, the "correct" input here would actually be `<svg xmlnsXlink="...">`. JSX, and by extension Virtual DOM, doesn't directly support authoring elements with namespaced attributes. In Preact, this comes from the fact that we prefer setting values as DOM properties rather than attributes. We do this partly for performance reasons, but also because properties allow complex values and are specified as JavaScript types, whereas attributes require serialization and are defined as HTML-specific types. This all boils down to: `<svg xmlnsXlink="..."> would be the semantically correct way to express `xmlns:xlink` as a prop in VDOM, and doing so renders correctly:

https://codesandbox.io/s/blissful-glade-tmifr?file=/src/logo.js:103-147

Additionally, the `xmlns` attribute is technically not required when rendering on the client. When constructing DOM elements on the client, the XML namespace must be passed to `document.createElementNS` and can't be inferred from an attribute ([here's where Preact does it](https://github.com/preactjs/preact/blob/b17a932342bfdeeaf1dc0fbe4f436c83e258d6c8/src/diff/index.js#L365)).

I know this gets murky because HTM _really looks_ like HTML, but HTM actually implements the semantics of JSX. There are a huge number of edge cases in HTML that are not supported, namespaced attributes is honestly one of the smaller differences (void tags and case sensitivity being the largest). In JSX, it's not possible to specify `<svg xmlns:xlink="..">`, and `xmlnsXlink=".."` would be used instead. (but...)

**(2)** The second thing I wanted to raise is mainly to check my understanding here since I'm by no means an SVG guru. The sandbox you provided works identically when I remove `xmlns:xlink=".."`. Again, I'm not an expert, but I believe the default namespace for `xlink` is already the one you're providing. It's also worth noting that neither `xmlns:xlink` nor `xmlnsXlink` are [supported](https://stackoverflow.com/a/32994814) by React, and `xlink:href` has actually been [removed from SVG](https://css-tricks.com/on-xlinkhref-being-deprecated-in-svg/) and will at some point be removed from browsers entirely.

So I guess for this PR, after having looked into all the context and noted some of the changes since Preact first gained SVG support, I worry that we're adding logic to handle an alternative way of specifying something that Preact already supported. 

Preact's goal has always been to deliver the maximum possible value per byte, and I think in this case the extra logic would be really difficult to justify since it's duplicating the existing authoring format.

vkentta:
Here is another non-SVG-guru :wave: :slightly_smiling_face: 

So to clarify the route that led to me making this PR. There was that SVG (exported from some designer tool in the year X) that we see in codesandbox. The SVG rendered fine as is in the browser and that let me believe I could just make a component out of it's markup with HTM. It would be obviously nice to be able to do so, but I completely understand that preact is not meant to cover all possible corner cases.

developit:
That's a good summary. Now I should say, I totally want this to be a thing that "just works" too. There are two ways we can make that work without changing Preact core: a plugin that does this stuff, or an HTM plugin:

#### Option 1: A global Preact plugin

This just handles `xmlns:` globally so that anything using Preact (HTM, JSX, etc) won't error:

```js
import { options } from "preact";
const old = options.vnode;
options.vnode = (vnode) => {
  for (let i in vnode.props) if (i.match(/^xmlns:/)) vnode.props[i] = undefined;
  if (old) old(vnode);
};
```

Here's a demo of this option: https://codesandbox.io/s/prod-bird-9ibc6?file=/src/logo.js:0-23

#### Option 2: HTM Adapter

HTM has a nice way to do things like this - it lets you process each parsed tag from the string before they get passed into Preact, so we can implement a replacement for `htm/preact` that removes `xmlns:xlink` attributes:

```js
import htm from 'htm';
import { h } from 'preact';

const html = htm.bind(function(type, props) {
  for (let i in props) if (i.match(/^xmlns:/)) props[i] = undefined;
  return h.apply(this, arguments);
});

export const logo = html`
  <svg xmlns="http://www.w3.org/2000/svg" width="100%" height="1.0625em" viewBox="0 0 108 14">
    <!-- etc -->
  </svg>
`;
```

Here's a demo of that option: https://codesandbox.io/s/jolly-pascal-pew39?file=/src/logo.js:38-72

One additional thing I thought of is actually separate, but maybe worth mentioning: SVG files exported from a design tool are usually much larger than they need to be, because they contain metadata and try to preserve the layers+objects as they were authored. You can run them through [SVGOMG](https://jakearchibald.github.io/svgomg/) (or SVGO on the command line) to strip all this out.

vkentta:
Thanks a lot for all the support :slightly_smiling_face: 

The issue was already worked around in the production code just by removing the prop manually that caused the error. Just made this PR to fix this for others that might have the same issue. I will not be offended in the slightest if you see it more beneficial to revert the merge to save the bytes :+1:

marvinhagemeister:
After an discussion with the team we've decided to revert the changes in favor of saving the bytes. The arguments that won the discussion are listed above in @developit's [comment](https://github.com/preactjs/preact/pull/2734#issuecomment-693767193).

Nonetheless I'm super happy about the PR, because it lead to us really taking a closer look and revisit some past decision. This is very much a win for Preact, even if we revert the PR :+1:

PR #2742, opened 2020-09-04
[wip] optimizations for compat

### LINKED ISSUE #1989: onChange not working with onMouseUp/onTouchEnd on input type=range
For input type=range
```
<input
          id={`range-for-${uniqueId}`}
          className={getStyleClass(value)}
          onMouseUp={onDragEnd}
          onTouchEnd={onDragEnd}
          type="range"
          min="0"
          max="10"
          value={value}
          onChange={onChange}
          list="steplist"
        />
```
onChange is not working here. It was working on v10.beta-3 release and not on rc or stable v10.

developit:
This reduces vnode normalization to a single loop over `props` with no megamorphic property access and no shape mutations caused by `delete`. These don't show up in our CI benchmarks because they're only in compat.

Here's the demo+details for the `UNSAFE_` prefix change any why it works:
https://codesandbox.io/s/preact-x-react-table-6103-forked-4q3l8?file=/src/patch.js:185-307

This PR also fixes #1989 - instead of using `onchange` as-is for `<input type=range>`, it uses `oninput` in all browsers except IE11 (which fires continuous `onchange` for range inputs, the original bug).

I've incorporated the relevant bits from #2665 as well (everything outside of `src/render.js`).

It might be worth backporting the prop+event normalization stuff from this into v10 - it's pretty safe, and around 150b savings along with the performance boost.

### DIFF
--- a/compat/src/Children.js
+++ b/compat/src/Children.js
@@ -1,11 +1,8 @@
 import { toChildArray } from 'preact';
 
 const mapFn = (children, fn) => {
-	if (!children) return null;
-	return toChildArray(children).reduce(
-		(acc, value, index) => acc.concat(fn(value, index)),
-		[]
-	);
+	if (children == null) return null;
+	return toChildArray(toChildArray(children).map(fn));
 };
 
 // This API is completely unnecessary for Preact, so it's basically passthrough.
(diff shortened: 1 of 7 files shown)


JoviDeCroock (approved):

developit:
@JoviDeCroock since we merge the backport of this to master, I think it might be best to also merge this PR to v11 now.

PR #2749, opened 2020-09-10
Convert test suite to web-test-runner

marvinhagemeister:
This PR converts our current `karma`-based test suite to `web-test-runner`. Compared to the former it is a lot more modern and has the following advantages:

- Proper console.logs (!!)
- Easier debugging in live browsers (Press "D" to open test in Chrome)
- Quicker watch mode
- Plugin-Architecture based on `rollup`


package.json
```diff
@@ -231,19 +237,11 @@
...
-		"karma-coverage": "^2.0.3",
-		"karma-mocha": "^2.0.1",
-		"karma-mocha-reporter": "^2.2.5",
-		"karma-sauce-launcher": "^4.3.4",
-		"karma-sinon": "^1.0.5",
-		"karma-sourcemap-loader": "^0.3.7",
-		"karma-webpack": "^3.0.5",
 		"lint-staged": "^10.5.2",
 		"lodash": "^4.17.20",
+		"lodash-es": "^4.17.20",
```
andrewiggins, line 244: I missed it - where are we using `lodash-es`?

test/chai.js
```diff
@@ -0,0 +1,55 @@
+/**
+ * In the browser, importing chai/chai.js will register to the window. In webpack it will
+ * be parsed as a commonjs module, so it won't be on the window. In this file we conditionally
+ * export from the window, or a webpack specific module import.
+ */
```
andrewiggins, line 5: So is this a workaround for web-test-runner or just something we did so wtr and karma could work side-by-side? Or something else?

web-test-runner.config.mjs
```diff
@@ -0,0 +1,138 @@
...
+const sauceLabsUser = process.env.SAUCE_USERNAME;
+const sauceLabsKey = process.env.SAUCE_ACCESS_KEY;
+const sauceLabs = sauceLabsUser && sauceLabsKey;
+if (sauceLabs) {
+	const sauceLabsCapabilities = {
+		name: 'unit-tests',
+		// if you are running tests in a CI, the build id might be available as an
+		// environment variable. this is useful for identifying test runs
+		// this is for example the name for github actions
+		build: `CI #${process.env.GITHUB_RUN_NUMBER} (${process.env.GITHUB_RUN_ID})`
```
andrewiggins, line 19: I think you could replace `CI` with `${process.env.GITHUB_WORKFLOW}` ([documentation](https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables))

web-test-runner.config.mjs
```diff
@@ -0,0 +1,138 @@
...
+				importMap: {
+					imports: {
+						chai: '/test/chai.js',
+						'preact/compat': '/compat/src/index.js',
+						'preact/debug': '/debug/src/index.js',
+						'preact/devtools': '/devtools/src/index.js',
+						'preact/hooks': '/hooks/src/index.js',
+						'preact/test-utils': '/test-utils/src/index.js',
+						preact: '/src/index.js',
+						'prop-types': '/node_modules/prop-types/prop-types.js'
```
andrewiggins, line 103: Is it possible to maintain the option of running the tests against the build output? I think we did that in karma to help catch mangle property errors

test/shared/createContext.test.js
```diff
@@ -1,7 +1,6 @@
-import { createElement, createContext } from '../../';
+import { createElement, createContext } from '../../src';
```
andrewiggins, line 1: Oh interesting - does web-test-runner not resolve package.json files?

andrewiggins (commented):
Had a couple thoughts and questions :)

test/chai.js
marvinhagemeister: Yep, web-test-runner expects everything to be packaged as ESM. The `chai` package doesn't have such an entry point, that's why we need this wrapper file.

marvinhagemeister:
I have no clue why our SauceLabs action is failing. It passed in #2885 which was based on this branch. When I merged it into this one SauceLabs broke again.

JoviDeCroock:
Closing: the test suite has since migrated to vitest, so both karma and web-test-runner are obsolete here.

PR #2754, opened 2020-09-14
Hydration Suspend & Resume

developit:
This PR implements suspend & resume for both hydration and new tree construction. If rendering throws during hydration or new tree construction (first render of a component or tree), catching the error up the tree in something like an "Error Boundary" will attach the hydration state and available DOM tree to the boundary's VNode. Triggering a render on this node (via any mechanism) attempts to resume from this state. The throw/catch process can be repeated and will continue to preserve DOM state. Once rendering completes, affected VNodes are un-marked and the behavior no longer occurs.

### DIFF
--- a/compat/src/internal.d.ts
+++ b/compat/src/internal.d.ts
@@ -16,7 +16,8 @@ export interface Component<P = {}, S = {}> extends PreactComponent<P, S> {
 
 	_childDidSuspend?(
 		error: Promise<void>,
-		suspendingComponent: Component<any, any>
+		suspendingComponent: Component<any, any>,
+		oldVNode?: VNode
 	): void;
 	_suspendedComponentWillUnmount?(): void;
 }
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -104,7 +104,14 @@ Suspense.prototype._childDidSuspend = function(promise, suspendingComponent) {
 		}
 	};
 
-	if (!c._pendingSuspensionCount++) {
+	/**
+	 * We do not set `suspended: true` during hydration because we want the actual markup
+	 * to remain on screen and hydrate it when the suspense actually gets resolved.
+	 * While in non-hydration cases the usual fallback -> component flow would occour.
+	 */
+	const vnode = c._vnode;
+	const wasHydrating = vnode && vnode._hydrating === true;
+	if (!wasHydrating && !c._pendingSuspensionCount++) {
 		c.setState({ _suspended: (c._detachOnNextRender = c._vnode._children[0]) });
 	}
 	promise.then(onResolved, onResolved);
@@ -124,9 +131,14 @@ Suspense.prototype.render = function(props, state) {
 		this._detachOnNextRender = null;
 	}
 
+	// Wrap fallback tree in a VNode that prevents itself from being marked as aborting mid-hydration:
+	const fallback =
+		state._suspended && createElement(Fragment, null, props.fallback);
+	if (fallback) fallback._hydrating = null;
+
 	return [
 		createElement(Fragment, null, state._suspended ? null : props.children),
-		state._suspended && props.fallback
+		fallback
 	];
 };
 
--- a/src/internal.d.ts
+++ b/src/internal.d.ts
@@ -47,7 +47,7 @@ export type ComponentFactory<P> =
 	| preact.ComponentClass<P>
 	| FunctionalComponent<P>;
 
-export interface PreactElement extends HTMLElement {
+export interface PreactElement extends HTMLElement, Text {
 	_children?: VNode<any> | null;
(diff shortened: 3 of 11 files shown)


marvinhagemeister (approved):
Love this 👍

compat/src/internal.d.ts
```diff
@@ -16,7 +16,8 @@ export interface Component<P = {}, S = {}> extends PreactComponent<P, S> {
 
 	_childDidSuspend?(
 		error: Promise<void>,
-		suspendingComponent: Component<any, any>
+		suspendingComponent: Component<any, any>,
+		oldVNode?: VNode
```
andrewiggins, line 20: I think we can remove this - I didn't see anywhere where an implementation of _childDidSuspend took a third param

compat/src/suspense.js
```diff
@@ -104,7 +104,14 @@ Suspense.prototype._childDidSuspend = function(promise, suspendingComponent) {
 		}
 	};
 
-	if (!c._pendingSuspensionCount++) {
+	/**
+	 * We do not set `suspended: true` during hydration because we want the actual markup
+	 * to remain on screen and hydrate it when the suspense actually gets resolved.
+	 * While in non-hydration cases the usual fallback -> component flow would occour.
+	 */
+	const vnode = c._vnode;
```
andrewiggins, line 112: I still need to run this code to verify my understanding, but my first reading through makes me think this may need to be `suspendingComponent._vnode` since we set `vnode._hydrating` on the vnode that threw the error (diff/index.js:275), not on `Suspense._vnode` (though we could in Suspense's `_catchError` implementation at the top of this file).

src/internal.d.ts
```diff
@@ -70,12 +70,13 @@ export interface VNode<P = {}> extends preact.VNode<P> {
...
 	 */
-	_dom: PreactElement | Text | null;
+	_dom: PreactElement | null;
 	/**
 	 * The last dom child of a Fragment, or components that return a Fragment
 	 */
-	_nextDom: PreactElement | Text | null;
+	_nextDom: PreactElement | null;
 	_component: Component | null;
+	_hydrating: boolean | null;
```
andrewiggins, line 79: Perhaps this wouldn't compress as well since we deal with nulls way more than booleans, but just leaving a note for myself (cuz I'm curious) to explore pros/cons of just making this a normal boolean that is initialized to `false` and flipped between `true`/`false` as appropriate.

andrewiggins (commented):
I know I'm a little late to the party 😅 but left a couple thoughts from an initial read for a later PR. I suspect when we add some tests we'll have some tweaks to make so things can be addressed then.

src/internal.d.ts
developit: @andrewiggins this needs to be a tri-state though:
- `true`: the vnode has DOM from bailout during hydration
- `false`: the vnode has DOM from bailout during initial render (new tree creation)
- `null`: the vnode has no DOM from bailout

We could use an int though, that's probably much better for perf.

compat/src/suspense.js
developit: You may be right. I need to clarify which vnode(s) we want to set the `_hydrating` bit on. My preference would be to set _hydrating on the vnode for the component that _caught/handled_ the suspend, since technically we bail out of the tree at that point and must resume from that point.

developit: Ah - yeah currently we're setting it on both the bailed-out VNode _and_ the VNode for the component that handled the bailout:
https://github.com/preactjs/preact/pull/2754/files#diff-b5e3843a998fbd0144a6ac5b1b7536a4R33

compat/src/internal.d.ts
developit: I wasn't sure if we needed to add it in order to fix the issue you mentioned about picking the wrong vnode on which to check `_hydrating`.

src/internal.d.ts
andrewiggins: Oh, I missed that `false` case. Makes sense

compat/src/suspense.js
andrewiggins: Yea, that makes sense to me (setting in on the place in the tree where we bail out and resume). Though I think when using compat's Suspense, the `_catchError` code isn't run since `Suspense` ships it's own `_catchError` implementation that doesn't have these changes to set hydrating, but that'd be trivial to add.

Looking back at it, I'm don't remember why we didn't just use `componentDidCatch` to implement Suspense... Maybe we needed something like your `handled` logic but no one tried that at the time?

marvinhagemeister: I think the reason for not using `componentDidCatch` was that some users were confused why the received a `Promise` in their error handlers instead of an `Error` object. We originally implemented it that way iirc, but backpedaled.

PR #2764, opened 2020-09-22
Try at the jsx exports

JoviDeCroock:
This still needs typings for `jsx`, ...


marvinhagemeister (approved):

milky2028:
This is totally for my own curiosity, but has anyone written any benchmarks for how much faster the new JSX runtime actually is?

PR #2767, opened 2020-09-23
Fix jsx-runtime not published

johakr:
`Module not found: Error: Can't resolve 'preact/jsx-runtime'`

I don't know if this is the only / right place this needs to be fixed, though.


marvinhagemeister (approved):
Oh no, this is totally on me. Thank you so much for making a PR :raised_hands:

marvinhagemeister:
I'll cut a new release.

PR #2769, opened 2020-09-23
Fix incorrect package export for jsx-runtime

marvinhagemeister:
Also add a check before publishing that asserts if every file defined in the export map exists. Hopefully this will prevent any future publishing errors related to that.


JoviDeCroock (approved):
Ohhh, this is neat!

cristianbote (approved):
🎉  love the script! 😄

PR #2776, opened 2020-09-25
fix edge cases for thrown promises in Suspense

kitten:
I caught these edge cases while reading `preact-ssr-prepass` and `hooks/src/index.js` since I ran into them in `react-ssr-prepass` as well. While maybe a little more unexpected, some users may expect that they can throw promises in `useMemo` or even in `useState` initialisers.

This means that the hook implementation has to take into account that `useMemo`'s factory function or `useState`'s setup function may throw before setting up the hook state. Minor changes though;

- Flipped assignment in `useReducer` (`hookState._value` before `hookState._component`)
- Flipped assignments in `useState` (`state._value` before `state._args`)


hooks/src/index.js
```diff
@@ -201,9 +201,9 @@ export function useMemo(factory, args) {
 	/** @type {import('./internal').MemoHookState} */
 	const state = getHookState(currentIndex++, 7);
 	if (argsChanged(state._args, args)) {
+		state._value = factory();
 		state._args = args;
 		state._factory = factory;
-		return (state._value = factory());
```
JoviDeCroock, line 206: We'll still need to return state._value

JoviDeCroock (approved):
Nice work 💯

marvinhagemeister (approved):
Love this! Thank you so much for making a PR :raised_hands:

PR #2778, opened 2020-09-27
Suspense bug with nested boundaries

JoviDeCroock:
When we nest several boundaries both with a lazy child it seems to bug out, what happens is the following:
- First lazy component throws
- Outer Suspense boundaries catches it
- Rerenders but somehow the "parked" children are still traversed


compat/test/browser/suspense.test.js
```diff
@@ -1718,6 +1718,38 @@ describe('suspense', () => {
...
+					<Suspender2 />
+				</Suspense>
+			</Suspense>,
+			scratch
+		);
+
+		const [resolve1] = suspend1();
+		const [resolve2] = suspend2();
+		rerender();
+		expect(scratch.innerHTML).to.equal('<div>Suspended...</div>');
```
JoviDeCroock, line 1738: Actual output at this point:

```
<div>Suspended2</div>
<div>Suspended...</div>
```

Outer boundary as second child even

sventschui:
> Rerenders but somehow the "parked" children are still traversed

After adding extensive logging I figured out why. The inner suspension causes a re-render (`enqueueRender`) of the inner Suspense barrier. The Problem is, that this inner `Suspense` still has `parentDom` set and thus renders into this DOM node.

The re-render can be suppressed by setting the `_parentDom` of all children to `null`. I added something in https://github.com/preactjs/preact/commit/bf3b2116db5875fc0d5f5f8a2a5b7381a23a28f7 but this (a) should be recursive as long as `child._parentDom === suspense._parentDom` and (b) breaks 5 other tests :/

sventschui:
Okay this one works better, but still failing tests: https://github.com/preactjs/preact/commit/4fe3eb883bdff94220affa093c05b5016fff3411

PR #2782, opened 2020-09-29
give precedence to className over class

JoviDeCroock:


compat/src/render.js
```diff
@@ -101,19 +101,18 @@ let classNameDescriptor = {
...
 let oldVNodeHook = options.vnode;
 options.vnode = vnode => {
 	let type = vnode.type;
 	let props = vnode.props;
 	let normalizedProps = props;
 
+	if (type && props.class != props.className) {
+		classNameDescriptor.enumerable = 'className' in props;
+		if (props.className != null) props.class = props.className;
+		Object.defineProperty(props, 'className', classNameDescriptor);
```
developit, line 113: I think this needs to mutate `normalizedProps` - we throw away `props`.

marvinhagemeister (approved):

developit:
I should have clarified on my comment (sorry, was on mobile) - the className descriptor has to be apply to `normalizedProps` after it has been assigned and populated. Currently it'll be overwritten for all component nodes.

PR #2786, opened 2020-10-04
Fix classname descriptors on Element vnodes

developit:
This addresses my concern from #2782. I've added a test that illustrates where it matters (and grouped the className tests).


JoviDeCroock (approved):

marvinhagemeister (approved):
Good catch! Love the additional tests and hardened existing ones :+1:

Issue #2792, opened 2020-10-06
TypeScript definition for "memo" breaks application after update


dzhykaiev:
### Reproduction

After updating the Preact from v10.3.4 to v10.5.4 and run the application got a lot of TypeScript errors in the console for all places where used "memo" function.

Error example
```shell
ERROR in /path-to-component/component.tsx(14, 39):
web_1        | TS2344: Type 'Props' does not satisfy the constraint 'FunctionalComponent<any>'.
web_1        |   Type 'Props' provides no match for the signature '(props: any, context?: any): VNode<any>'.
```

In PR for this ticket #2443 was changed types for the "memo" function, and now should change types for all places where used this function for fixing the build.

from:
```jsx
type Props = {
    title: string;
};

const Component = memo(<Props>({ title }) => {
...
}
```

to:
```jsx
import { FunctionalComponent } from 'preact';

type Props = {
    title: string;
};

const Component = memo(<FunctionalComponent<Props>>({ title }) => {
...
}
```
But its a lot of work and code duplication. Remember you should also import in each place where used "memo" function type FunctionalComponent. I offer update ```compat/src/index.d.ts``` for both cases with defaultProps and without.

JoviDeCroock:
Fixed in #2793

PR #2793, opened 2020-10-06
#2792 fix memo fn types definition

dzhykaiev:
#2792


JoviDeCroock (approved):
This looks good to me, thank you so much for this!

Issue #2795, opened 2020-10-09
Question: Suspense for Data Fetching behavior


LXSMNSYC:
It seems that Preact does support Suspense for Data Fetching, but I'm curious for the internal behavior:
- Does Preact clean the hooks when a Suspending component throws? Or does it retain the hook values?
- How different is it than React's Suspense?

JoviDeCroock:
Hey,

I recently made [my own experiment for this](https://github.com/JoviDeCroock/realworld) currently we don't clear stuff since we 'park' the vnodes (similar to React putting everything on hidden in the dom) in memory. We could look at whether or not it's needed to clear these things depending on people's thoughts. We'd probably want to start an RFC for that.

LXSMNSYC:
Thanks for the quick response!

I was creating a library that revolves around the use of Suspense for Data Fetching in Preact, but I just learned that the hooks are not cleared when a component is suspending, thus the old states are being received (in this case, the library sets the state with a valid Resource value which is used for suspending), which causes the component to suspend recursively, unable to receive the new state. I have made a work-around to overcome this issue. The problem doesn't occur in a  1:1 React implementation though, curiously.

JoviDeCroock:
Honestly, if this is a pain-point I think we should address it. During my development on Prefresh I also bumped into some cases where not actively cleaning stuff up could result in erroneous situations, we should be able to pinpoint these.

LXSMNSYC:
I have a hook called `useFreshState`:
https://github.com/LXSMNSYC/scoped-model/blob/master/packages/preact-graph-state/src/hooks/useFreshState.tsx
which is a hook similar to `useState` but resets to the initial state whenever the dependencies changes.

Since Preact doesn't clean nor reset hooks, `useFreshState` doesn't reset natively since the old dependencies are still being kept after the failed render. 

https://github.com/LXSMNSYC/scoped-model/blob/master/packages/preact-graph-state/src/hooks/useGraphNodeValue.tsx#L37-L43

`useGraphNodeValue` is being used by `useGraphNodeResource` (a hook that suspends a component if the data is pending.). Since `useFreshState` doesn't reset natively through cleaning, it is stuck with the first state snapshot it received. Once the promise is thrown, the suspended component attempts to re-render, but since the component still receives the old snapshot, it thinks that the state is still 'pending', resuspending again. This goes on for all eternity.

[...]

PR #2796, opened 2020-10-11
Reset hooks state when suspense triggers

JoviDeCroock:
In React when a component suspends it resets all hooks that are being processed, it also runs cleanups of active effects.

Fixes: https://github.com/preactjs/preact/issues/2795

I also used this opportunity to move `_cleanup` to a reserved property name so I can use it in prefresh https://github.com/JoviDeCroock/prefresh/pull/206/files


marvinhagemeister (approved):
Good catch!

LXSMNSYC:
Thanks for this @JoviDeCroock !

Issue #2801, opened 2020-10-18
Babel ^7.12.0 breaks preact jsx-runtime export


johakr:
Hi,

due to this change in babel https://github.com/babel/babel/pull/12116/commits/d6b0822ee9e95321a417b321a6687048faf1c3a1, the preact jsx runtime can't be imported anymore. Preact exports the jsx-runtime as folder, but babel expects a file with file ending `.js` now (`${source}/jsx-runtime.js` and `${source}/jsx-dev-runtime.js`).

`Module not found: Error: Can't resolve 'preact/jsx-runtime.js'`

Is that something we should address at the babel repo, or do we need to change how the runtime is exported in preact?

nicolo-ribaudo:
Fixed in `@babel/helper-builder-react-jsx-experimental@7.12.4`

rati-dzidziguri:
any ideas on how to fix this I am getting the same error 
`Module not found: Can't resolve 'react/jsx-dev-runtime'` and using preact 10.5.7 with craco.

nicolo-ribaudo:
@rati-dzidziguri 

You need to pass `"importSource": "preact"` to the JSX plugin/preset.

rati-dzidziguri:
@nicolo-ribaudo, thank you very much. Would you please provide a little more details on which plugin to install and how to do this? Highly appreciate this answer.

nicolo-ribaudo:
I'm not sure about your setup (or the default build setup for Preact), but there is _already_ compiling JSX in your build process (either TS, or Babel): you need to find this existing tool and update its config.

rati-dzidziguri:
I understand; thank you very much. here are my package JSON and configs.
![image](https://user-images.githubusercontent.com/52178984/100460225-44340700-30e0-11eb-872c-2544f0961a8d.png)

craco.config.js
![image](https://user-images.githubusercontent.com/52178984/100460270-57df6d80-30e0-11eb-994d-d9ced496002c.png)

I have no other customizations, but this and initially used CRA to create the base skeleton.

developit:
I'm guessing this is something strange "craco" is doing.

PR #2802, opened 2020-10-18
Fix jsx import by moving it to the root

JoviDeCroock:
Fixes: https://github.com/preactjs/preact/issues/2801

Babel requires the two files to be available at the root.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ coverage
 *.log
 package/
 preact-*.tgz
+./jsxRuntime.js
+./jsxRuntime.mjs
--- a/package.json
+++ b/package.json
@@ -53,10 +53,9 @@
       "import": "./jsx-runtime/dist/jsxRuntime.mjs"
     },
     "./jsx-dev-runtime": {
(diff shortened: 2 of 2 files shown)


nicolo-ribaudo:
I don't think this works around the Babel regression. In order to make it work, you don't need to copy the files but you have to add this to `"exports"`:
```json
    "./jsx-runtime.js": {
      "browser": "./jsx-runtime/dist/jsxRuntime.module.js",
      "umd": "./jsx-runtime/dist/jsxRuntime.umd.js",
      "require": "./jsx-runtime/dist/jsxRuntime.js",
      "import": "./jsx-runtime/dist/jsxRuntime.mjs"
    },
    "./jsx-dev-runtime.js": {
      "browser": "./jsx-runtime/dist/jsxRuntime.module.js",
      "umd": "./jsx-runtime/dist/jsxRuntime.umd.js",
      "require": "./jsx-runtime/dist/jsxRuntime.js",
      "import": "./jsx-runtime/dist/jsxRuntime.mjs"
    },
```

PR #2803, opened 2020-10-18
Make ref types' current property non-optional

ivantm:
This change aligns Preact's ref types with [those of React](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e47aff8a3a1331d0c540550db79dd5065ab735e5/types/react/index.d.ts#L88-L90), where the `current` property is not optional.

### DIFF
--- a/hooks/src/index.d.ts
+++ b/hooks/src/index.d.ts
@@ -40,7 +40,7 @@ export function useReducer<S, A, I>(
 	init: (arg: I) => S
 ): [S, (action: A) => void];
 
-type PropRef<T> = { current?: T };
+type PropRef<T> = { current: T };
 type Ref<T> = { current: T };
 
 /**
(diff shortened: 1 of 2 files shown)


hooks/src/index.d.ts
```diff
@@ -123,10 +123,7 @@ export function useContext<T>(context: PreactContext<T>): T;
  * @param value Custom hook name or object that is passed to formatter
  * @param formatter Formatter to modify value before sending it to the devtools
  */
-export function useDebugValue<T>(
-	value: T,
-	formatter?: (value: T) => any
-): void;
+export function useDebugValue<T>(value: T, formatter?: (value: T) => any): void;
```
ivantm, line 126: This was prettier's doing.

marvinhagemeister (approved):
Awesome, thanks you so much for the PR 🙌

PR #2805, opened 2020-10-22
make jsx-runtime work when using an alias

JoviDeCroock:


developit (approved):
whoosh! totally hadn't thought of this one. we'll also need to update `@preactjs/compat`.

ForsakenHarmony (approved):

rayriffy:
When this will be realesed in `preact`? I would love to use it with Next and React 17. 🧑‍🔬

matthewkwong2:
> When this will be realesed in `preact`? I would love to use it with Next and React 17. 🧑‍🔬

Same, I'm wondering when will a stable release be available.

shilangyu:
@JoviDeCroock does this actually work? This file does not seem to be included when installing preact:

image

So aliasing `react` to `preact/compat` makes `react/jsx-runtime` point to `preact/compat/jsx-runtime` which does not exist.

PS. yes, I am on the newly released version:


image

marvinhagemeister:
@shilangyu Looks like we forgot to add them to our publish step. Made https://github.com/preactjs/preact/pull/2827 to address this. We'll cut a new release once that's merged.

PR #2813, opened 2020-11-01
Move child reodering to diffChildren

JoviDeCroock:
We used to reorder within the scope of a single bailed vnode diff, we're in a better position when we are currently inside of a `diffChildren` to correctly hand off `oldDom` to the next node.

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -163,15 +163,20 @@ export function diffChildren(
 				firstChildDom = newDom;
 			}
 
-			oldDom = placeChild(
-				parentDom,
-				childVNode,
-				oldVNode,
-				oldChildren,
-				excessDomChildren,
-				newDom,
-				oldDom
-			);
+			if (childVNode._bailed) {
+				oldDom = reorderChildren(childVNode, oldDom, parentDom);
+				childVNode._bailed = false;
+			} else {
+				oldDom = placeChild(
+					parentDom,
+					childVNode,
+					oldVNode,
+					oldChildren,
+					excessDomChildren,
+					newDom,
+					oldDom
+				);
+			}
 
 			// Browsers will infer an option's `value` from `textContent` when
 			// no value is present. This essentially bypasses our code to set it
@@ -194,11 +199,6 @@ export function diffChildren(
 				// If the last child is a DOM VNode, then oldDom will be set to that DOM
 				// node's nextSibling.
 				newParentVNode._nextDom = oldDom;
-			}
-
-			if (childVNode._bailed && !isHydrating) {
-				oldDom = reorderChildren(childVNode, oldDom, parentDom);
-				childVNode._bailed = false;
 			}
 		} else if (
 			oldDom &&
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -4,6 +4,22 @@ import { EMPTY_OBJ, EMPTY_ARR } from '../constants';
 import { removeNode } from '../util';
 import { getDomSibling } from '../component';
 
+export function getLastDomChild(vnode) {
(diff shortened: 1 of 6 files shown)


test/browser/lifecycles/shouldComponentUpdate.test.js
```diff
@@ -778,7 +778,7 @@ describe('Lifecycle methods', () => {
 		expect(scratch.innerHTML).to.equal('<p>bar</p>');
 	});
 
-	it('should reorder non-updating nested Fragment children', () => {
+	it.skip('should reorder non-updating nested Fragment children', () => {
```
JoviDeCroock, line 781: Temp skipped, wanted to view the benchmark diff

JoviDeCroock: The issue here is that we currently reorder correctly but only the first child. we have 3 children under every Fragment and the other two children are clobbered to the back

src/diff/children.js
```diff
@@ -228,6 +233,35 @@ export function diffChildren(
...
 }
 
+function reorderChildren(childVNode, oldDom, parentDom) {
+	for (let tmp = 0; tmp < childVNode._children.length; tmp++) {
+		const vnode = childVNode._children[tmp];
+		if (vnode) {
+			vnode._parent = childVNode;
+
+			if (typeof vnode.type == 'function') {
+				reorderChildren(vnode, oldDom, parentDom);
```
JoviDeCroock, line 243: We'd need to find a way to tell whether or not we have to traverse further:

```
const Cell = () => (
  <Fragment>
    <div>x</div>
    <Fragment><div>y</div></Fragment>
  </Fragment>
)
```

When this Cell bails we need to ensure we move `y` as well as `x` when this is moved from the top. Before this change we would forget about `y` within `diffChildren` but move it from `diff.shouldComponentUpdate = false`.

src/diff/children.js
```diff
@@ -163,15 +163,24 @@ export function diffChildren(
...
-			oldDom = placeChild(
-				parentDom,
-				childVNode,
-				oldVNode,
-				oldChildren,
-				excessDomChildren,
-				newDom,
-				oldDom
-			);
+			if (childVNode._bailed) {
```
JoviDeCroock, line 166: It would be better if we were able to somehow move this away from the `vnode` and report it from the `diff` result, this would involve OR not returning `newDom` OR attaching it to the returned dom-node.

Scenario 1 makes the code above this one harder:
- looking for `firstChildDom`
- applying refs

developit: QQ: couldn't we check `childVNode === oldVNode` here instead of adding the _bailed flag?

JoviDeCroock: Found out we can use new_children === old_children thanks to https://github.com/preactjs/preact/blob/master/src/diff/index.js#L176

src/diff/children.js
```diff
@@ -163,15 +163,26 @@ export function diffChildren(
...
-				childVNode,
-				oldVNode,
-				oldChildren,
-				excessDomChildren,
-				newDom,
-				oldDom
-			);
+			if (
+				typeof childVNode.type == 'function' &&
+				childVNode._children === oldVNode._children
```
JoviDeCroock, line 168: This signals that we bailed out either due to strict equality or SCU false

marvinhagemeister (approved):
This looks really great! In the longterm I think we have to go with moving the placing of children into `diffChildren` either way 👍 

Does this change catch a new bugs where we'd need a test for or is it mostly byte savings?

JoviDeCroock:
@marvinhagemeister mainly savings and readability

PR #2815, opened 2020-11-04
Controlled input

JoviDeCroock:


src/diff/props.js
```diff
@@ -156,6 +156,7 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
  */
 function eventProxy(e) {
 	this._listeners[e.type + false](options.event ? options.event(e) : e);
+	if (e.type === 'input' || e.type === 'change') this.value = this._prevValue || '';
```
JoviDeCroock, line 159: We'll need the initialValue here, since '' isn't the default this should be the initial value. Also we should check whether or not the input node has a value since we can just have an onChange on an uncontrolled component as well.

src/diff/props.js
```diff
@@ -156,6 +156,9 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
  */
 function eventProxy(e) {
 	this._listeners[e.type + false](options.event ? options.event(e) : e);
+	if (e.type === 'input' || e.type === 'change') {
+		this.value = this._prevValue || this.value;
+	}
```
developit, line 161:
```suggestion
	if ((e.type === 'input' || e.type === 'change') && this._prevValue) {
		this.value = this._prevValue;
	}
```

PR #2816, opened 2020-11-05
Throw when hook is invoked outside of render

### LINKED ISSUE #2574: Lazy loading styled-components component triggers warning message incorrectly
### Reproduction

https://github.com/TLadd/preact-styled-components-lazy-load-warning. 
The relevant files are 
- https://github.com/TLadd/preact-styled-components-lazy-load-warning/blob/master/src/index.js
- https://github.com/TLadd/preact-styled-components-lazy-load-warning/blob/master/src/StyledDiv.js

preact@10.4.4

### Steps to reproduce

Load the app and observe the following warning message is printed:

> The component styled.div with the id of "sc-AxjAm" has been created dynamically.
> You may see this warning because you've called styled inside another component.
> To resolve this only create new StyledComponents outside of any render method and function component.

This warning occurs only when lazy loading a styled component or lazy loading something that renders a styled component with preact. The warning message is not printed if a dynamic import is not used or if using React instead of preact.

Looking at styled-components, it's triggering this warning message if [calling `useRef` doesn't throw](https://github.com/styled-components/styled-components/blob/3390244224d357dc5f25a021ad4c58a99dd1261e/packages/styled-components/src/utils/checkDynamicCreation.js#L20) when declaring the component.

### Expected Behavior

The warning message should not be printed.

### Actual Behavior

The warning message is printed.

marvinhagemeister:
Looks like we only threw when a hook was called from outside of a component for the very first component that was rendered. This was caused by `currentComponent` already being set at that point. See the previous code comment in the changeset.

This PR corrects that so that we keep track of valid hook positions and always throw if it's invalid.

Fixes #2574

### DIFF
--- a/debug/test/browser/debug-hooks.test.js
+++ b/debug/test/browser/debug-hooks.test.js
@@ -25,19 +25,18 @@ describe('debug with hooks', () => {
 		teardown(scratch);
 	});
 
-	// TODO: Fix this test. It only passed before because App was the first component
-	// into render so currentComponent in hooks/index.js wasn't set yet. However,
-	// any children under App wouldn't have thrown the error if they did what App
-	// did because currentComponent would be set to App.
-	// In other words, hooks never clear currentComponent so once it is set, it won't
-	// be unset
-	it.skip('should throw an error when using a hook outside a render', () => {
-		const Foo = props => props.children;
-		class App extends Component {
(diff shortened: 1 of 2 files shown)


cristianbote (approved):
👍

JoviDeCroock (approved):
LGTM :shipit:

PR #2817, opened 2020-11-05
Ensure preact/compat is not converting onchange to oninput for input type="range" in IE11

gcraftyg:
When using preact/compat, an input of `type="range"` converts an `onchange` to `oninput` except in IE11. This mimics react's implementation as `oninput` will not fire on an `<input type="range" />` in IE11.

I came across a situation where I was not seeing the `onChange` prop firing in IE11. Looking at it closer, preact/compat detects IE11 by checking `typeof Symbol != 'undefined'`. This works well as long as `Symbol` is not being polyfilled, for instance in a NextJS app with preact.

This PR is adding an additional check to verify that `Symbol` is native to the browser and not polyfilled, allowing for `<input type="range" />` to behave as expected.

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -16,7 +16,9 @@ const CAMEL_PROPS = /^(?:accent|alignment|arabic|baseline|cap|clip(?!PathU)|colo
 // type="file|checkbox|radio", plus "range" in IE11.
 // (IE11 doesn't support Symbol, which we use here to turn `rad` into `ra` which matches "range")
 const ONCHANGE_INPUT_TYPES =
-	typeof Symbol != 'undefined' ? /fil|che|rad/i : /fil|che|ra/i;
+	typeof Symbol != 'undefined' && typeof Symbol() == 'symbol'
+		? /fil|che|rad/i
+		: /fil|che|ra/i;
 
 // Some libraries like `react-virtualized` explicitly check for this.
 Component.prototype.isReactComponent = {};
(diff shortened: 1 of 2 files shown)


marvinhagemeister (commented):
Good catch, this is great! Let's add a quick unit test to make sure we don't regress on this in the future. Our suite runs tests in all browsers including IE11 automatically, so a test where we test if a value was changed should do the job 👍

gcraftyg:
Thanks for taking a look at this @marvinhagemeister.

To clarify on the test coverage, we could amend the existing test: https://github.com/preactjs/preact/blob/9aa7d38c45d7e6782d4ca42c7059a2982aec4c49/compat/test/browser/events.test.js#L73-L85 or add a new test, to include a `Symbol` polyfill when it is ran in IE11. That would replicate the problematic scenario we're guarding against. But I'm not sure if that's the same as what you were suggesting.

I'm happy to add the test, just looking for clarification. Thanks!

marvinhagemeister:
@gcraftyg yup, that sounds perfect. Either way, amending or adding a new test is fine, as long as we have something that checks for it in the CI 👍

gcraftyg:
I pushed up a test that guards against the regression. In order to polyfill `Symbol` for the test, I converted the input `type` check into a function call. It was originally a variable assignment, but that was being assigned before the tests were run. 

My preference would be to keep it as a variable, but I'm not aware of a way to inject a polyfill for a specific test before the module is imported.

marvinhagemeister (approved):
Sweet, the test is perfect! Thank you so much for your PR :raised_hands:

JoviDeCroock (approved):
Awesome, thank you so much!

PR #2819, opened 2020-11-07
(chore) - trial codesandbox-ci

JoviDeCroock:
This will give us a:
- nightly build for a PR
- trial environment to inject into bug repro's


JoviDeCroock:
We can disable comments by adding silent to the config, wdyt?

developit:
17% faster with codesandbox-ci turned on lol

developit (approved):

developit:
Might also be good to update the Codesandbox template (even just this one) to include a bit more stuff so that we can have a nice little test env for each PR.

PR #2820, opened 2020-11-08
Add displayName to context

max-voronov:
#2454 added support for context displayName.

The type definition is lacking the `displayName` property too.


marvinhagemeister (approved):
Good catch! Thank you so much for your PR :tada:

PR #2822, opened 2020-11-11
add animateTransform tag type

HuHguZ:
please add this tag, my company team uses this. thanks! :)


marvinhagemeister (approved):
Great catch, looks like we missed one! Thanks for making a PR 👍

PR #2755, opened 2020-09-15
Hydration suspend tests

andrewiggins:
Start at some tests for hydration suspend and resume

PR #2823, opened 2020-11-11
Fix prototype spies not being reset in tests

marvinhagemeister:
Looks like we never reset previous spies which lead to the failures experienced in #2755 .


JoviDeCroock (approved):

developit:
Nice catch.

PR #2829, opened 2020-11-13
Update csstype from version 2 to version 3

bz2:
Improved typing, and smaller type file with fewer internals exposed.

There are changes to the re-exported types but this feature exposing csstype has only just been released in preact. (https://github.com/preactjs/preact/pull/1797)

Expose TLength and TTime generics, and adapt length only to accept numbers of any kind, to match the 'px' appending logic in preact.

[Version 3 differences](https://github.com/frenic/csstype#version-30).

Now sure how type definitions should be tested ideally?

Have tested locally, including using [tsuquyomi](https://github.com/Quramy/tsuquyomi) for completions (which is the [reason](https://github.com/frenic/csstype#generics) for the `& {}` hack).

style completions

Have also replaced the usage in one of my projects, which was using version 3 before:

```diff
-export type {Properties as CSSProperties} from 'csstype';
-
+export type CSSProperties = preact.JSX.CSSProperties<string | 0>;
```

Which works fine with my (limited) use of the `style` attribute.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3548,9 +3548,9 @@
 			}
 		},
 		"csstype": {
-			"version": "2.6.13",
-			"resolved": "https://registry.npmjs.org/csstype/-/csstype-2.6.13.tgz",
-			"integrity": "sha512-ul26pfSQTZW8dcOnD2iiJssfXw0gdNVX9IJDH/X3K5DGPfj+fUYe3kB+swUY6BF3oZDxaID3AJt+9/ojSAE05A==",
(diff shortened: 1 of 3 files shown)


marvinhagemeister (approved):
Sweet! The only way we've found to test TS types is to add some code using a particular feature to one of our `.ts` files in our test suite and compile them. If they throw, something is off and if they don't we assume that everything is fine. I'm good with merging this PR as is though :+1: 

Thank you again so much for your contribution :raised_hands:

PR #2833, opened 2020-11-16
Search full list for oldDomNodes

### LINKED ISSUE #2783: Unnecessary unmounting/mounting children with keys

# Reproduction

### Steps to reproduce

### Expected Behavior

Good behavior

### Actual Behavior

Bad behavior

JoviDeCroock:
This fixes keying in large lists

Fixes #2783 https://codesandbox.io/s/busy-buck-gyzf1

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -321,7 +321,7 @@ function placeChild(
 			for (
 				let sibDom = oldDom, j = 0;
 				(sibDom = sibDom.nextSibling) && j < oldChildren.length;
-				j += 2
+				j++
 			) {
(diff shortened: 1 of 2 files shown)


pavulon:
I think this fixed one of the isolated repros, but might not be enough for all the issues linked here.

The problem is still present when you use components as children, eg. `h(Item, {key: id})` instead of `h('article', {key: id})`. https://codesandbox.io/s/new-waterfall-dpwu8

JoviDeCroock:
Closing due to functional boundaries not solved yet, this will probably be a v11 thing

PR #2834, opened 2020-11-16
Save some bytes by removing appendChild in favor of insertBefore

JoviDeCroock:


marvinhagemeister:
That's strange. Seems like it has a strong effect on text updates. I would've guessed that there wouldn't be a difference. cc @developit

developit:
insertBefore(x, null) does have a tiny bit of overhead compared to appendChild.

JoviDeCroock:
Implemented in v11

PR #2843, opened 2020-11-23
Add autocapitalize global attribute to JSX types

marcoms:
`autocapitalize` is a global HTML attribute: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize


src/jsx.d.ts
```diff
@@ -622,6 +622,7 @@ export namespace JSXInternal {
 		alt?: string;
 		as?: string;
 		async?: boolean;
+		autoCapitalize?: string;
```
marvinhagemeister, line 625: According to the MDN link the `autocapitalize` value doesn't work with any string, but only with specific values. Something like `autocapitalize="asdasfa"` is invalid. The MDN page lists these values as valid ones: `off`, `none`, `on`, `sentences`, `words` and `characters`. Let's limit the typings to those values instead of allowing any string. We can do that via the union syntax:

```ts
autoCapitalize?: "off" | "none" | "on" | "sentences" | "words" | "characters";
```

marcoms: Thanks for the tip, hopefully this is still good one year later!

rschristian:
Going to close this out, already added in:

https://github.com/preactjs/preact/blob/4f08ae4d607964ae8222e910d3b209def75626b0/src/jsx.d.ts#L786-L799

Sorry!

Issue #2646, opened 2020-07-23
Uncontrolled select support?


thurfir:
Hi, the _defaultValue_ attribute for `<select>` does not work in preact: https://stackblitz.com/edit/preact-bug-defaultvalue-zn7r9a . 

Switching to a selected attribute (in `<option>` ) seems to work but causes some strange behaviour on late renderings (no sandbox yet).

ForsakenHarmony:
Does `defaultValue` even exist for `<select>`?

thurfir:
Hi @ForsakenHarmony , yes it exists; React even asked me (with a console warning) to replace <option default... /> by defaultValue during my tests ;-) .

andrewiggins:
To be clear for anyone looking to work on this, `defaultValue` on `<select>` is not a DOM prop ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement)) but is something React supports, so support for `defaultValue` on `<select>` needs to be in `preact/compat`.

`<option>` does have a `defaultSelected` property ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLOptionElement)), but not sure what React does with this.

Rafi993:
Hello there, If this issue is up for taking I would like to help.

themre:
Hi, i just encountered this bug with defaultvalue on the select and i have a stackblitz example here: https://stackblitz.com/edit/preact-starter-template-yktscr?file=App.jsx. I used `defaultValue` on the select, but no luck. Or am I missing something?

PR #2848, opened 2020-11-25
Adding defaultValue support in select tag

Rafi993:
## Description

`defaultValue` for `select` tag is supported by react even though it is not part of the html spec. This PR adds `defaultValue` attribute support for select tag

## Related Ticket
closes https://github.com/preactjs/preact/issues/2646

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -162,8 +162,13 @@ options.vnode = vnode => {
 		// Adding support for defaultValue in select tag
 		if (type == 'select' && normalizedProps.defaultValue != null) {
 			normalizedProps.value = toChildArray(props.children).forEach(child => {
-				child.props.selected =
-					normalizedProps.defaultValue == child.props.value;
+				if (normalizedProps.multiple) {
+					child.props.selected =
+						normalizedProps.defaultValue.indexOf(child.props.value) != -1;
+				} else {
+					child.props.selected =
+						normalizedProps.defaultValue == child.props.value;
+				}
 			});
 		}
 
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -159,6 +159,14 @@ options.vnode = vnode => {
 			});
 		}
 
+		// Adding support for defaultValue in select tag
+		if (type == 'select' && normalizedProps.defaultValue) {
+			normalizedProps.value = toChildArray(props.children).forEach(child => {
+				child.props.selected =
+					normalizedProps.defaultValue.indexOf(child.props.value) != -1;
+			});
+		}
+
 		vnode.props = normalizedProps;
 	}
 
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -163,7 +163,7 @@ options.vnode = vnode => {
 		if (type == 'select' && normalizedProps.defaultValue != null) {
 			normalizedProps.value = toChildArray(props.children).forEach(child => {
 				child.props.selected =
-					normalizedProps.defaultValue.indexOf(child.props.value) != -1;
+					normalizedProps.defaultValue == child.props.value;
 			});
 		}
 
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -160,7 +160,7 @@ options.vnode = vnode => {
 		}
 
 		// Adding support for defaultValue in select tag
-		if (type == 'select' && normalizedProps.defaultValue) {
+		if (type == 'select' && normalizedProps.defaultValue != null) {
 			normalizedProps.value = toChildArray(props.children).forEach(child => {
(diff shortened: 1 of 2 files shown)


Rafi993:
I might need help to fix those failing github actions.

compat/src/render.js
```diff
@@ -159,6 +159,14 @@ options.vnode = vnode => {
 			});
 		}
 
+		// Adding support for defaultValue in select tag
+		if (type == 'select' && normalizedProps.defaultValue) {
```
marvinhagemeister, line 163: Let's enhance this check to `normalizeProps.defaultValue != null`. This will ensure we still apply this normalization even if users pass in `0` 👍

compat/src/render.js
```diff
@@ -159,6 +159,14 @@ options.vnode = vnode => {
...
+			normalizedProps.value = toChildArray(props.children).forEach(child => {
+				child.props.selected =
+					normalizedProps.defaultValue.indexOf(child.props.value) != -1;
```
marvinhagemeister, line 166: `defaultValue` can be a non-array value too, like in the repro of the original reporter: https://stackblitz.com/edit/preact-bug-defaultvalue-zn7r9a .

Let's add that check here 👍

compat/test/browser/render.test.js
```diff
@@ -110,6 +110,22 @@ describe('compat render', () => {
 		expect(scratch.firstElementChild).to.have.property('value', 'foo');
 	});
 
+	it('should support defaultValue for select tag', () => {
```
marvinhagemeister, line 113: Love this test!

marvinhagemeister (commented):
This is great! Left a few minor comments, let me know what you think 👍

marvinhagemeister (commented):
This is great! Left a few minor comments, let me know what you think 👍

marvinhagemeister:
Disregard the failing GitHub actions. They seem to be always failing when a PR is done from another repo. They are not caused by your PR 👍

Rafi993:
> Disregard the failing GitHub actions. They seem to be always failing when a PR is done from another repo. They are not caused by your PR

Thank you @marvinhagemeister for your time. I'll make those changes.

JoviDeCroock (approved):
Nice work, thank you!

compat/src/render.js
Rafi993: Right now I have converted it to single select but multi select has issues. When I tried the following code

```javascript
		// Adding support for defaultValue in select tag
		if (type == 'select' && normalizedProps.defaultValue != null) {
			normalizedProps.value = toChildArray(props.children).forEach(child => {
				if (normalizedProps.multiple) {
					child.props.selected =
						normalizedProps.defaultValue.indexOf(child.props.value) != -1;
				} else {
					child.props.selected =
						normalizedProps.defaultValue == child.props.value;
				}
			});
		}
```

for multi selection only the last value is selected. any thoughts on why this is happening ?

Rafi993: but I tried logging the value in the loop it was selected during iteration but some got overwritten during the next iteration.

Rafi993: oops, I meant to comment under the other comment https://github.com/preactjs/preact/pull/2848#discussion_r530210401 did it here by mistake.

compat/src/render.js
Rafi993: Right now I have converted it to single select but multi select has issues. When I tried the following code

```javascript
		// Adding support for defaultValue in select tag
		if (type == 'select' && normalizedProps.defaultValue != null) {
			normalizedProps.value = toChildArray(props.children).forEach(child => {
				if (normalizedProps.multiple) {
					child.props.selected =
						normalizedProps.defaultValue.indexOf(child.props.value) != -1;
				} else {
					child.props.selected =
						normalizedProps.defaultValue == child.props.value;
				}
			});
		}
```

for multi selection only the last value is selected. any thoughts on why this is happening ?

Rafi993: but I tried logging the value in the loop it was selected during iteration but some got overwritten during the next iteration.

Rafi993: When I stopped the test from watch mode and ran it again it worked :)

marvinhagemeister (approved):
Awesome, thank you so much for the fantastic PR 🙌 💯

grefrit:
Hello. It doesn't work in situations for example like this. Because it considers that select have one child
image

marvinhagemeister:
@grefrit Can you open an issue for that? That way we can track the progress on this and make sure that we don't forget about it. Comments on closed PRs like this one here don't show up on the project anymore unless one manually searches for closed PRs.

PR #2851, opened 2020-11-27
Add filter benchmark

marvinhagemeister:
This PR adds the codesandbox from #2783 to our benchmark suite. It measures the performance when a long list of items is filtered and items are removed from the middle.

### DIFF
--- a/benches/src/filter_list.html
+++ b/benches/src/filter_list.html
@@ -31,12 +31,16 @@
 			import { measureName, measureMemory } from './util.js';
 			import { render, createElement as h, Component } from 'framework';
 
+			function Row(props) {
+				return h('article', null, props.children);
+			}
+
 			function App(props) {
 				return h('div', {}, [
 					h(
 						'div',
 						{ class: 'items' },
-						props.items.map(id => h('article', { key: id }, id))
+						props.items.map(id => h(Row, { key: id }, id))
 					)
 				]);
 			}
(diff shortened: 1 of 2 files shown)


benches/src/filter_list.html
```diff
@@ -0,0 +1,87 @@
...
+		<script type="module">
+			import { measureName, measureMemory } from './util.js';
+			import { render, createElement as h, Component } from 'framework';
+
+			function App(props) {
+				return h('div', {}, [
+					h(
+						'div',
+						{ class: 'items' },
+						props.items.map(id => h('article', { key: id }, id))
```
JoviDeCroock, line 39: Could we also add one that does `props.items.map(id => h(Row, { key: id }, id))` because this is the harder issue we're facing in diff-children

JoviDeCroock (approved):
Left a nit that we could tackle in another PR, looks good to me :shipit:

benches/src/filter_list.html
```diff
@@ -0,0 +1,87 @@
...
+					await new Promise(r => requestAnimationFrame(r));
+				}
+			}
+
+			warmup().then(async () => {
+				performance.mark('start');
+				runPatch();
+				await new Promise(r => requestAnimationFrame(r));
+				performance.mark('stop');
+				performance.measure('duration', 'start', 'stop');
```
andrewiggins, line 81: You can use `measureName` here so it stays in sync.
```suggestion
				performance.measure(measureName, 'start', 'stop');
```

andrewiggins (approved):
This is great! Let me know if you have any feedback on the benching infra

benches/src/filter_list.html
marvinhagemeister: TIL :+1:

PR #2854, opened 2020-11-29
Catch hydration errors and progressively hydrate array boundaries

### LINKED ISSUE #2539: bug when hydrating with error boundary

JoviDeCroock:
Fix error catching during hydration and progressive hydration of array-boundaries.

Fixes #2539

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -241,11 +241,19 @@ export function diff(
 		newVNode._original = null;
 		// if hydrating or creating initial tree, bailout preserves DOM:
 		if (isHydrating || excessDomChildren != null) {
-			newVNode._dom = oldDom;
-			newVNode._hydrating = !!isHydrating;
-			excessDomChildren[excessDomChildren.indexOf(oldDom)] = null;
-			// ^ could possibly be simplified to:
-			// excessDomChildren.length = 0;
+			if (e.then) {
+				newVNode._dom = oldDom;
+				newVNode._hydrating = !!isHydrating;
+				excessDomChildren.length = 0;
+			} else {
+				// When we are hydrating and have excessDomChildren we don't know the _children this VNode
+				// should receive so it's safer to unmount them. Else on the subsequent error-boundary diff,
+				// we won't know the oldDom and insert an additional node instead of replace the prerendered one. (#2539)
+				for (tmp = excessDomChildren.length; tmp--; ) {
+					if (excessDomChildren[tmp] != null)
+						removeNode(excessDomChildren[tmp]);
+				}
+			}
 		}
 		options._catchError(e, newVNode, oldVNode);
 	}
--- a/compat/test/browser/suspense-hydration.test.js
+++ b/compat/test/browser/suspense-hydration.test.js
@@ -164,6 +164,30 @@ describe('suspense hydration', () => {
 		});
 	});
 
+	it('should properly hydrate when the suspended component is a list', () => {
+		const items = ['Todo 1', 'Todo 2'];
+		scratch.innerHTML = '<p>Todo 1</p><p>Todo 2</p>';
+		clearLog();
+
+		const [Lazy, resolve] = createLazy();
+		hydrate(
+			<Suspense>
+				<Lazy />
+			</Suspense>,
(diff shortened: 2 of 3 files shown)


src/diff/index.js
```diff
@@ -241,11 +241,19 @@ export function diff(
 		newVNode._original = null;
 		// if hydrating or creating initial tree, bailout preserves DOM:
 		if (isHydrating || excessDomChildren != null) {
-			newVNode._dom = oldDom;
-			newVNode._hydrating = !!isHydrating;
-			excessDomChildren[excessDomChildren.indexOf(oldDom)] = null;
-			// ^ could possibly be simplified to:
-			// excessDomChildren.length = 0;
+			if (e.then) {
```
JoviDeCroock, line 244: This logic should only kick in when we are in a boundary, this means that only if we notice a promise is thrown up we should attempt at progressively hydrating this boundary.

src/diff/index.js
```diff
@@ -241,11 +241,19 @@ export function diff(
...
+				newVNode._dom = oldDom;
+				newVNode._hydrating = !!isHydrating;
+				excessDomChildren.length = 0;
```
JoviDeCroock, line 247: Formerly we did `excessDomChildren[excessDomChildren.indexOf(oldDom)] = null;` which didn't catch the case if the direct child of your throwing promise was a component that returned `x.map(Component)`.

src/diff/index.js
```diff
@@ -241,11 +241,19 @@ export function diff(
...
+			} else {
+				// When we are hydrating and have excessDomChildren we don't know the _children this VNode
+				// should receive so it's safer to unmount them. Else on the subsequent error-boundary diff,
+				// we won't know the oldDom and insert an additional node instead of replace the prerendered one. (#2539)
+				for (tmp = excessDomChildren.length; tmp--; ) {
```
JoviDeCroock, line 252: When we have an error being thrown during hydration we should clear the dom else we'll append the oldDom with the new one when we rerender.

src/diff/index.js
developit: Hmm - elsewhere we use component._dirty to trigger this - it's valid to throw a non-Promise and catch it, and that should trigger progressive hydration.

JoviDeCroock: I mean, an error shouldn't hold progressive hydration. An error boundary will then trigger bugs like #2539

Issue #2668, opened 2020-07-30
Missing: `defaultValue` Typescript Types


amiller-gh:
### Reproduction

Preact 10.4.6 does not contain typescript types for React's `defaultValue` despite being supported by `preact/compat`

### Steps to reproduce
Try to use it in a Typescript project 🙂 

I was able to work around the missing types by adding this override declaration to my project:
```ts
declare module 'preact/src/jsx' {
  /* eslint-disable-next-line @typescript-eslint/no-namespace */
  namespace JSXInternal {
    interface HTMLAttributes<RefType extends EventTarget = EventTarget> {
      defaultValue?: string;
    }
  }
}
```

### Expected Behavior
It builds!

### Actual Behavior
It doesn't...
```
Property 'defaultValue' does not exist on type 'HTMLAttributes<HTMLInputElement>'.
```

Rafi993:
Hello there, If this issue is not taken I would like to help

JoviDeCroock:
Resolved by https://github.com/preactjs/preact/commit/3957522a9b5393681ca66ad879e046c2f978a3e9

PR #2857, opened 2020-12-08
Adding support for defaultValue in typings

Rafi993:
## What does it do?
Adds support for **defaultValue** in typescript for HTMLElement

## Related issue

Closes https://github.com/preactjs/preact/issues/2668

## QA Instructions, Screenshots, Recordings

1. Create a typescript preact typescript project with preact cli
2. create input element with attribute defaultValue
3. It should not have typescript error

## Added tests?

- [ ] yes
- [x] no

### DIFF
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -126,3 +126,11 @@ declare namespace React {
 		toArray: (children: preact.ComponentChildren) => preact.VNode<{}>[];
 	};
 }
+
+declare module 'preact/src/jsx' {
+	namespace JSXInternal {
+		interface HTMLAttributes<RefType extends EventTarget = EventTarget> {
+			defaultValue?: string;
+		}
+	}
+}


compat/src/index.d.ts
```diff
@@ -126,3 +126,11 @@ declare namespace React {
 		toArray: (children: preact.ComponentChildren) => preact.VNode<{}>[];
 	};
 }
+
+declare module 'preact/src/jsx' {
```
Rafi993, line 130: I couldn't extend name space JSX directly above so I did this do extend it.

esquevin: I think this could be done directly in `src/jsx.d.ts` https://github.com/preactjs/preact/blob/bd3c6559fcc8a0c07798023fd9927c9bb03a844d/src/jsx.d.ts#L653

It's a global preact issue, not specifically a preact/compat one

Rafi993: But does the preact without compact support ```default``` key?

PR #2863, opened 2020-12-12
Simplify vnode re-use detection (-2 B)

marvinhagemeister:
This PR simplifies our check to detect if a `vnode` is already in use and if we need to clone it. Noticed that while exploring alternative algorithms in `diffChildren`. Saves `2 B`.


andrewiggins (approved):
Oh nice catch!

JoviDeCroock (approved):

PR #2866, opened 2020-12-13
In benches, abstract framework differences with createRoot API

andrewiggins:
This change abstracts the API differences between preact v8 and preact v10 using a new `createRoot` API that the benchmarks will use. This new API allows some new benchmarks (e.g. hydrate1k) to work for preact v8.


marvinhagemeister (approved):

andrewiggins:
Benches are failing cuz I need a proxy for the `preact-master` library. Trying to figure out how to make that work well during development and in CI

andrewiggins:
@marvinhagemeister FYI I just updated the new filter_list bench to use `createRoot` too

marvinhagemeister (approved):

PR #2885, opened 2020-12-23
Fix WTR saucelabs configuration

LarsDenBakker:
This makes the tests run on saucelabs. The tunnel identifier is not needed. If you do want to set it, you also need to set it on the sauce connect proxy (the third parameter). Otherwise they are out of sync.

While debugging I found some other issues which I fixed in the saucelabs plugin and updated the versions.

There are still some failing tests on IE11. There is now a `--manual` flag which lets you easily debug in any browser, locally or on saucelabs. That's probably useful for debugging IE11.

Some of the failures seem to be related to rendering differences. Perhaps something like https://www.npmjs.com/package/@open-wc/semantic-dom-diff can help with that. There are also some errors related to symbols, perhaps a mismatch with the way they are polyfilled? I'm not sure how you guys did that in Karma. Let me know if it might be something in the legacy plugin.

For the other browsers since you're testing on the latest versions, I recommend using the playwright launcher to test on Chromium/Webkit/Firefox, and keep saucelabs for IE11 only. That would be a lot faster. I could prepare a PR for that if you're interested.


marvinhagemeister (approved):
This is awesome! Thank you so much for looking into it :heart: :raised_hands:

PR #2886, opened 2020-12-24
fix portal tests

JoviDeCroock:


src/diff/patch.js
```diff
@@ -138,9 +140,27 @@ function patchComponent(
 			: tmp._defaultValue
 		: globalContext;
 
-	// Get component and set it to `c`
-	c = newVNode._component = oldVNode._component;
-	clearProcessingException = c._processingException = c._pendingError;
+	if (oldVNode._component) {
```
JoviDeCroock, line 143: For `Suspense` we need to check whether or not we are in a mount situation, my first thought was to add this in `children.js` as a seperate branch but this seemed to get into a lot more erroneous scenario's. Will try to look deeper into this.

src/render.js
```diff
@@ -49,7 +49,7 @@ export function render(vnode, parentDom, replaceNode) {
 			: null;
 
 	if (isHydrating) {
-		replaceNode = excessDomChildren[0];
+		replaceNode = excessDomChildren ? excessDomChildren[0] : null;
```
JoviDeCroock, line 52: In the above we set `excessDomChildren` to null potentially so we need to check whether or not this is possible else we have to set it to null

andrewiggins: Oh! good catch!

src/diff/patch.js
andrewiggins: Oh right. Yea, I'm thinking that in `children.js` we need to add some complexity to our check to see if we should go down the mount path or patch path. 3 scenarios come to mind:
1. There is no matching oldVNode
2. The previous oldVNode was suspended while mounting
3. The previous oldVNode was suspended while hydrating

Those last two would require deeper changes to the diff, but something we should tackle eventually

andrewiggins (approved):
Aweseome! Thanks for taking a look!

src/diff/patch.js
JoviDeCroock: We could add a check there with if vnode._component and switch to mount if so, then we could move to using an unmount function in Suspense but that would be a new Suspense implementation entirely so we should maybe defer that to v11?

PR #2889, opened 2020-12-29
Correct `_nextDom` pointer if it is being unmounted

andrewiggins:
While investigating ways to simplify and improve `placeChild` I stumbled upon a bug in how we diff Fragments. If child of a Fragment is unmounted, a Fragment's `_nextDom` pointer could point to this unmounted child. In situations where this happens, the Fragment's parent will correct this by re-appending all the children after the Fragment. While the DOM output is correct, the way we go about achieving it is sub-optimal. This change fixes the `_nextDom` to not point at a DOM that is going to be unmounted.

For example, take the following code:
```html
<div>
  <Fragment> <!-- Fragment 1 -->
    <div>1</div>
    <div>2</div> <!-- this node will get unmounted -->
  </Fragment>
  <Fragment> <!-- Fragment 2 -->
    <div>A</div>
  </Fragment>
</div>
```

If doing a diff from the top where `<div>2` gets unmounted, Fragment 1 should finish it's `diffChildren` call with a `_nextDom` pointer to `<div>A`.  However, before this fix, when Fragment 1 diffs `<div>1`, it sets its `_nextDom` pointer to `<div>1.nextSibling` which is `<div>2` (it hasn't been unmounted yet). This fix detects that the `_nextDom` pointer will be unmounted and resets it to the next DOM sibling, in this case `<div>A`.

This PR fixes a long standing Fragment issue that prevented us from asserting expected DOM operations on Fragment diffing and this PR enables us to them turn on! We can now catch regressions in how we mount & move around fragment children!

Specifically, on master the test "should handle changing node type within a Component that returns a Fragment #1326" in fragments.test.js has more than necessary dom operations. The same is true for the tests I've added. All have reduced number of DOM ops.

### DIFF
--- a/test/browser/fragments.test.js
+++ b/test/browser/fragments.test.js
@@ -8,8 +8,6 @@ import { logCall, clearLog, getLog } from '../_util/logCall';
 /* eslint-disable react/jsx-boolean-value */
 
 describe('Fragment', () => {
-	let expectDomLog = false;
-
 	/** @type {HTMLDivElement} */
 	let scratch;
 
@@ -19,9 +17,7 @@ describe('Fragment', () => {
 	let ops = [];
 
 	function expectDomLogToBe(expectedOperations, message) {
-		if (expectDomLog) {
-			expect(getLog()).to.deep.equal(expectedOperations, message);
-		}
+		expect(getLog()).to.deep.equal(expectedOperations, message);
 	}
 
 	class Stateful extends Component {
@@ -711,7 +707,7 @@ describe('Fragment', () => {
 		);
 	});
 
-	it('should not preserve state when switching to a keyed fragment to an array', () => {
+	it('should not preserve state when switching between a keyed fragment and an array', () => {
 		function Foo({ condition }) {
 			return condition ? (
 				<div>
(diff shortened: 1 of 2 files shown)


test/browser/fragments.test.js
```diff
@@ -739,29 +735,30 @@ describe('Fragment', () => {
 		clearLog();
 		render(<Foo condition={false} />, scratch);
 
-		expect(ops).to.deep.equal([]);
+		expect(ops).to.deep.equal([]); // Component should not have updated (empty op log)
 		expect(scratch.innerHTML).to.equal(html);
 		expectDomLogToBe([
-			'<div>1Hello1.insertBefore(<span>1, <span>1)',
```
andrewiggins, line 745: This original dom ops were just wrong. They don't match what the test was doing.

test/browser/fragments.test.js
```diff
@@ -1150,32 +1147,12 @@ describe('Fragment', () => {
 		clearLog();
 		render(<Foo condition={false} />, scratch);
 		expect(scratch.innerHTML).to.equal(html, 'rendering from true to false');
-		expectDomLogToBe(
```
andrewiggins, line 1153: Since this test was written we converted all nested arrays to Fragments so in this test where DOM ops previously happened (due to nested arrays previously being flattened) there are now none since nested arrays become Fragments

JoviDeCroock (approved):
Nice find!

marvinhagemeister (approved):

PR #2890, opened 2020-12-29
Refactor Portal implementation

marvinhagemeister:
This PR cherry picks the Portal changes we did yesterday and applies them to `master`.

### DIFF
--- a/compat/src/portals.js
+++ b/compat/src/portals.js
@@ -22,6 +22,7 @@ function Portal(props) {
 	_this.componentWillUnmount = function() {
 		render(null, _this._temp);
 		_this._temp = null;
+		_this._container = null;
 	};
 
 	// When we change container we should clear our old container and
@@ -46,7 +47,7 @@ function Portal(props) {
 				},
 				insertBefore(child, before) {
 					this.childNodes.push(child);
-					_this._container.insertBefore(child, before);
+					_this._container.appendChild(child, before);
 				},
 				removeChild(child) {
 					this.childNodes.splice(this.childNodes.indexOf(child) >>> 1, 1);
--- a/compat/src/portals.js
+++ b/compat/src/portals.js
@@ -1,5 +1,8 @@
-import { createElement, hydrate, render, __u as _unmount } from 'preact';
+import { createElement, render } from 'preact';
 
+/**
+ * @param {import('../../src/index').RenderableProps<{ context: any }>} props
+ */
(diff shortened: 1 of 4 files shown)


compat/test/browser/portals.test.js
```diff
@@ -489,7 +489,7 @@ describe('Portal', () => {
 
 		render(<App />, scratch);
 		expect(scratch.innerHTML).to.equal(
-			'<div id="outer">1</div><div id="inner">0</div><div id="app">test</div>'
+			'<div id="inner">0</div><div id="outer">1</div><div id="app">test</div>'
```
JoviDeCroock, line 492: This does mean we slightly deviate from the React behavior but imo that's fine

JoviDeCroock (approved):
This looks great!

compat/src/portals.js
```diff
@@ -1,82 +1,71 @@
...
+			_this._temp = {
+				parentNode: container,
+				childNodes: [],
+				appendChild(child) {
+					this.childNodes.push(child);
+					_this._container.appendChild(child);
+				},
+				insertBefore(child, before) {
+					this.childNodes.push(child);
+					_this._container.insertBefore(child, before);
```
JoviDeCroock, line 49: I guess even this should do `_this._container.appendChild(child);` to match React behavior, this would fix the test you needed to adjust

marvinhagemeister: Tried that out but this section seems to never be called in that test. It seems to always hit the `appendChild` branch instead.

marvinhagemeister: mh seems like the `insertBefore` is never hit at all.

JoviDeCroock: Oh, the other way around then, insertBefore the firstChild of the parentContainer 😅

compat/src/portals.js
```diff
@@ -1,82 +1,71 @@
...
-		{ context: _this.context },
-		props._vnode
-	);
 
 	_this.componentWillUnmount = function() {
-		let parent = _this._temp.parentNode;
-		if (parent) parent.removeChild(_this._temp);
-		_unmount(_this._wrap);
+		render(null, _this._temp);
+		_this._temp = null;
```
andrewiggins, line 24: Should we null out `_this.container` too? So we don't hold on to the DOM reference?

andrewiggins (approved):
Nice!

PR #2891, opened 2020-12-30
Fix Children.toArray not setting index keys

### LINKED ISSUE #2888: Preact toChildArray does not assign keys
### Reproduction

https://fanatical-dot-bumper.glitch.me/

In [React documentation](https://reactjs.org/docs/react-api.html#reactchildrentoarray), they say

> That is, toArray prefixes each key in the returned array so that each element’s key is scoped to the input array containing it.

In other words, it will assign keys, even if they were undefined before. This is not the case for Preact's implementation and causes issues when using `preact/compat`

marvinhagemeister:
Turns out that React always sets keys when calling `Children.toArray` that are based on the array index.

Fixes #2888

### DIFF
--- a/compat/src/Children.js
+++ b/compat/src/Children.js
@@ -17,5 +17,14 @@ export const Children = {
 		if (normalized.length !== 1) throw 'Children.only';
 		return normalized[0];
 	},
-	toArray: toChildArray
+	toArray(children) {
+		return toChildArray(children).map((child, i) => {
+			// Only normalized vnodes will be keyed #2888
+			if (child != null && typeof child === 'object') {
+				child.key = child.key != null ? child.key : i;
+			}
+
+			return child;
+		});
+	}
 };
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):

compat/src/Children.js
```diff
@@ -17,5 +17,14 @@ export const Children = {
 		if (normalized.length !== 1) throw 'Children.only';
 		return normalized[0];
 	},
-	toArray: toChildArray
+	toArray(children) {
+		return toChildArray(children).map((child, i) => {
+			// Only normalized vnodes will be keyed #2888
+			if (child != null && typeof child === 'object') {
+				child.key = child.key != null ? child.key : i;
```
andrewiggins, line 24: I think we may have to be a little fancier here... check out the note in the react docs: https://reactjs.org/docs/react-api.html#reactchildrentoarray 

Would be curious what the use case is

developit: Strong +1 for "what is the use-case for this". I can't think of a situation in which this behavior is actually desirable, even if it's what React currently does.

marvinhagemeister (commented):
After some more investigation, I've come to the same conclusion as @andrewiggins and @developit . Can't find a good use case for this change.

PR #2893, opened 2020-12-31
Add support for pretty printing karma console logs

marvinhagemeister:
Not sure why I didn't think of this earlier. To get pretty console logs out of karma we can sidestep their own (limited) console argument serializer by simply overwriting it. With this in place we now have the same console output as if we'd write native node code. This should make debugging our vnode trees much easier as they are now longer cut off and printed in a single line.

Before:
image

After:
image

Stack traces have also received some attention and now easier to scan for the eyes. Jumping to that line via cmd+click works now too 👍 

Before:
image

After:
image


marvinhagemeister:
Hmrpf looks like it doesn't play well with sinon spies :grimacing:

andrewiggins (approved):
Ah, this is great! Thanks for doing this!! 🎉

PR #2898, opened 2021-01-02
Always instantiate refs array (-6 B)

andrewiggins:
Lazily instantiating the refs array was added before we had benchmarks. So this change is just to evaluate if the bytes are worth it. I suspect that it may not have a meaningful impact on perf so trying this in a PR to see if it matters.


src/diff/children.js
```diff
@@ -36,7 +36,8 @@ export function diffChildren(
 	oldDom,
 	isHydrating
 ) {
-	let i, j, oldVNode, childVNode, newDom, firstChildDom, refs;
+	let i, j, oldVNode, childVNode, newDom, firstChildDom;
+	let refs = [];
```
andrewiggins, line 40: @developit I vaguely remember you mentioning something about some JS engines having an escape analysis feature that makes objects (and arrays?) locally declared that never leave the function essentially free from a GC perspective or something. Do you think we could benefit from that here? Or should we leave this as it is?

JoviDeCroock (approved):

andrewiggins:
Marking it as draft for now since I'm not sure this change is worth it yet. Seems replace1k regresses but that could be a fluke...

src/diff/children.js
developit: Yes - that should apply here.

src/diff/children.js
```diff
@@ -253,10 +253,8 @@ export function diffChildren(
 	}
 
 	// Set refs only after unmount
-	if (refs) {
-		for (i = 0; i < refs.length; i++) {
-			applyRef(refs[i], refs[++i], refs[++i]);
-		}
+	for (i = 0; i < refs.length; i++) {
+		applyRef(refs[i], refs[++i], refs[++i]);
```
developit, line 257: Not sure if Terser is doing it already, but we know that `i=0` going into this loop because of L238:

```suggestion
	while (i < refs.length) {
		applyRef(refs[i++], refs[i++], refs[i++]);
```

andrewiggins: Dang. Had hopes for this one. Increases by 2 bytes on my machine. Here is the terser diff if you are curious:

image

developit: I haven't checked in-situ but I think this one is smaller:

```suggestion
	while (refs[++i]) {
		applyRef(refs[i], refs[++i], refs[++i]);
```

JoviDeCroock:
Closing: micro-perf draft against the v10 ref-array shape; v11 rewrote ref handling (refQueue).

PR #2899, opened 2021-01-02
Improve hydration tests

andrewiggins:
Improve our hydration tests to actually assert what hydration should do: attach event listeners. Did this for normal and suspended hydration as well as fix a bug in our Suspense implementation of suspended hydration so these new tests pass.

### DIFF
--- a/compat/test/browser/suspense-hydration.test.js
+++ b/compat/test/browser/suspense-hydration.test.js
@@ -22,6 +22,11 @@ describe('suspense hydration', () => {
 		rerender,
 		unhandledEvents = [];
 
+	const List = ({ children }) => <ul>{children}</ul>;
+	const ListItem = ({ children, onClick = null }) => (
+		<li onClick={onClick}>{children}</li>
+	);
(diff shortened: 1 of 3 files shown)


compat/test/browser/suspense-hydration.test.js
```diff
@@ -164,6 +216,196 @@ describe('suspense hydration', () => {
...
+			expect(listeners[2]).to.have.been.calledOnce;
+
+			scratch
+				.querySelector('li:last-child')
+				.dispatchEvent(createEvent('click'));
+			expect(listeners[4]).to.have.been.calledTwice;
+		});
+	});
+
+	it.skip('should properly hydrate suspense when resolves to a Fragment', () => {
```
andrewiggins, line 343: Currently not supported but I didn't realize that until after I wrote the tests so since I wrote it I figured I'd add it so that maybe one day we can enable it.

marvinhagemeister (approved):
This is awesome :star:

PR #2901, opened 2021-01-02
Prepare test code to be run in ES2015+ environments

marvinhagemeister:
Today I played around a bit with running our test suite in native ES2015+ environments. That is to say that I skipped the transpilation step to ES5. By doing so I noticed that the prototype of a native class is not iterable. So calling `Object.keys(MyClass.prototype)` will always return an empty array. This breaks some of our tests.

Since we don't iterate over prototypes that often I just inlined our helper functions.


andrewiggins (approved):

andrewiggins:
Oooo running our tests using modern syntax would be really nice

PR #2903, opened 2021-01-03
Fix bug when rerendering a suspended hydration node

andrewiggins:
Added a test where a suspended hydration node is rerendered but still suspends which revealed a bug. This PR fixes it and adds some more tests. This completes the TODOs at the bottom of this test file


marvinhagemeister (approved):
Love those tests :raised_hands: :100:

Issue #2897, opened 2021-01-01
createPortal throws for 10.5.8 in dev mode


commenthol:
### Reproduction

demo: [preact-createportal-demo](https://github.com/pmwcs/preact-createportal-issue)

### Steps to reproduce

Run demo in dev mode `npm run dev`.

### Actual Behavior

If using `createPortal()` in preact@10.5.8 using dev mode `npm run dev` then following exception is thrown on rendering.

```
debug.js?c91e:105 Uncaught (in promise) Error: Expected a valid HTML node as a second argument to render.	Received [object Object] instead: render(<T />, [object Object]);
    at Object.preact__WEBPACK_IMPORTED_MODULE_0__.options.__ (debug.js?c91e:105)
    at O (render.js?50ff:17)
    at d.W [as constructor] (portals.js?463e:41)
    at d.M [as render] (index.js?c988:505)
    at $ (index.js?c988:168)
    at m (children.js?b0ec:147)
    at $ (index.js?c988:168)
    at eval (component.js?dcb2:221)
    at Array.some (<anonymous>)
    at g (component.js?dcb2:216)
```

preact@10.5.7 does not show this behavior.

After bundling with `npm run build` the demo runs without problems while serving with `npm run serve`.

### Expected Behavior

No Error if using `createPortal` in dev mode.

PR #2904, opened 2021-01-03
add nodeType to the render element stub

JoviDeCroock:
Fixes https://github.com/preactjs/preact/issues/2897


marvinhagemeister (approved):
Good catch! :+1:

PR #2906, opened 2021-01-03
Rename "coverage" env to "COVERAGE"

marvinhagemeister:
This fixes invalid code in test runners with more basic string replace based define plugins as they tend to accidentally replace parts of a variable.


JoviDeCroock (approved):
Woop, consistency 🎉

PR #2910, opened 2021-01-04
fix updating suspended component

tanhauhau:
replace the `parentDom` of a suspended component to a detached `<div />`, allowing `diff` to continue updating the component without affecting the actual dom.

when updating a suspended children that will no longer suspending, allow `diff` to insert elements to the detached `<div />`, and moved it over to original `parentDom` if the `<Suspense />` unsuspense.

this change allow to re-enable some of the test cases, but i'm not sure whether this is the right approach on this, probably will add a few more tests, but hope to know whether im approach it in the right direction.

### DIFF
--- a/compat/test/browser/suspense.test.js
+++ b/compat/test/browser/suspense.test.js
@@ -659,7 +659,6 @@ describe('suspense', () => {
 		});
 	});
 
-	// TODO: Fix this test
 	it('should allow children to update state while suspending', () => {
 		/** @type {(state: { s: string }) => void} */
 		let setState;
@@ -1678,7 +1677,7 @@ describe('suspense', () => {
 		expect(scratch.innerHTML).to.eql(`<div>conditional hide</div>`);
 	});
 
-	xit('should support sCU=false when un-suspending', () => {
+	it('should support sCU=false when un-suspending', () => {
 		// See #2176 #2125
 		const [Suspender, suspend] = createSuspender(() => <div>Hello</div>);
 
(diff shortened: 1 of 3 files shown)


tanhauhau:
fyi @kenrick95

marvinhagemeister (approved):
Woa, this is outstanding! Thank you so much for the PR 🙌 🚀

compat/test/browser/suspense.test.js
```diff
@@ -1706,25 +1761,141 @@ describe('suspense', () => {
...
+					this.setState({ promise: null, content });
+				}
+
+				render() {
+					if (this.state.promise) {
+						throw this.state.promise;
+					}
+
+					return this.state.content;
+				}
```
andrewiggins, line 1790: Haven't seen this pattern before. Is this something you do in your own apps or just a quick way to write a test? Specifically referring to throwing a promise that never resolves. Would just like to better understand this use case 😅

andrewiggins (commented):
@tanhauhau Thanks again for your PR :) Sorry for the belated review, but just had one question about better understanding how you use this new functionality

compat/test/browser/suspense.test.js
tanhauhau: Well, it was copied from the test case "should allow suspended children to update", which was written by you 🤣

I believe it's just a quick wat to write the test.

andrewiggins: OH! Ahahahaha 😂😂😊 Got it, thanks :)

I'm thinking we may need to re-evaluate if this use case is really necessary at this moment (unsuspending without resolving the Promise). [React doesn't support it](https://codesandbox.io/s/resume-with-rerender-kim18) and currently it requires that we override the render method of a developer's component which feels icky. I know we do this for `componentWillUnmount` already, but its not a great pattern and one we should move away from (an `options.unmount` plugin would be better). We'll likely revisit this later. Feel free to leave your thoughts on PR #2917 😄

tanhauhau: Hmmm.. I wonder will adding options.commit and options.unmount be in the plan for Preact X, i'd love to help adding them in

PR #2912, opened 2021-01-05
Compat: Add basic shims for some scheduler functions

marvinhagemeister:
Was notified of a library that makes use of some experimental concurrent mode APIs from React. This PR adds the necessary shims for that.

See https://github.com/dai-shi/use-context-selector/pull/36/files

cc @mischnic

### DIFF
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -125,4 +125,16 @@ declare namespace React {
 		only: (children: preact.ComponentChildren) => preact.ComponentChild;
 		toArray: (children: preact.ComponentChildren) => preact.VNode<{}>[];
 	};
+
+	// scheduler
+	export const unstable_ImmediatePriority: number;
+	export const unstable_UserBlockingPriority: number;
+	export const unstable_NormalPriority: number;
+	export const unstable_LowPriority: number;
+	export const unstable_IdlePriority: number;
+	export function unstable_runWithPriority(
+		priority: number,
+		callback: () => void
+	): void;
+	export const unstable_now: () => number;
 }
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -32,6 +32,7 @@ import {
 	REACT_ELEMENT_TYPE,
 	__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED
 } from './render';
+export * from './scheduler';
 
 const version = '16.8.0'; // trick libraries to think we are react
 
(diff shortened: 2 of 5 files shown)


mischnic:
Great! There's also at least one export on `react-dom`: https://github.com/dai-shi/use-context-selector/blob/master/src/batchedUpdates.ts

marvinhagemeister:
@mischnic That's already present in `preact/compat` :)

https://github.com/preactjs/preact/blob/989315cf01c06b000d2d62e35f077a5e25a9b5d8/compat/src/index.js#L101

mischnic:
🙈

yisar:
I don't know why preact must insist on running.
In fact, this kind of implementation is totally incorrect. The update model of preact is completely different from react, which will lead to more and more differences.

marvinhagemeister:
@yisar I'm not sure if you're aware, but our implementation of `__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED` isn't the same as in React either. And yet it keeps working like a charm. The surprising thing of working on the compat layer is that actual libraries are way less reliant on specific implementation details than one might assume. Like the library linked in the original comment. It doesn't rely on specific behavior, but rather just calls these functions and that's it.

> I don't know why preact must insist on running.

Going into other frameworks issue trackers and posting something akin to "your framework sucks" is just rude. You might want to take a look at our [Code of Conduct](https://github.com/preactjs/preact/blob/master/CODE_OF_CONDUCT.md) page again.

yisar:
@marvinhagemeister Once you introduce the shim, a lot of disagreements will arise because the shim is inaccurate.

In essence, this API is not just to execute a callback. The reason why `use-context-selector` uses this API is to solve a tearing problem, which does not exist in preact.

Of course, this is not important. What I need to explain is that once the shim is introduced, there will be more differences. The introduction of shim is not a good behavior, it is full of uncertainty and instability.

It's also worth mentioning that I don't think preact is terrible, I just think it's unwise to introduce an unnecessary shim.

You can give up `use-context-selector`, can't you?

cristianbote:
@yisar preact does not need to have accurate scheduler implementation, as you noted as well, so why not have the APIs from scheduler exported for the libraries that make use of them? Regardless of their accuracy, as long as they do not break the functionality, there's  no harm in having them.

Marvin already added a block of comment where all of the concerns that one might have, are addressed. So, any discussion regarding the accuracy, or validity of them should start from there.

cristianbote (approved):
Great job @marvinhagemeister! This will make the adoption of libraries much easier 🎉

yisar:
@cristianbote 
> so why not have the APIs from scheduler exported for the libraries that make use of them
You're right. If it's just for sharing the third-party library, it can be done without causing damage.

But what if users need to implement their own libraries with the APIs of the `preact/compat` ? At this point, these shims will bring a lot of trouble, because these APIs should not be here. Their semantics and uses should not be here.

I notice that this PR has been merged, and I don't want to argue about this.

PR #2914, opened 2021-01-05
Fix karma benchmarks not compiling

marvinhagemeister:
@andrewiggins Found a bug in `karma-esbuild` and this upgrades it to the version with the fix.


andrewiggins (approved):
Thanks!!

PR #2918, opened 2021-01-09
Fix SauceLabs runner

andrewiggins:
The previous checks in the karma config no longer worked since we don't have a CI-master workflow anymore. I've simplified it to just look for an env variable directly


andrewiggins:
I cancelled the benchmarks since they don't apply to this change

marvinhagemeister (approved):

PR #2919, opened 2021-01-11
use options.unmount instead of overriding component.componentWillUnmount

tanhauhau:
Following up to https://github.com/preactjs/preact/pull/2917, use `options.unmount` instead of overriding `componentWillUnmount`

this will check `component._onResolve` on every unmount though, not sure how much will this add up to slow things down

### DIFF
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -29,6 +29,15 @@ options.unmount = function(vnode) {
 	if (component && component._onResolve) {
 		component._onResolve();
 	}
+
+	// if the component is still hydrating
+	// most likely it is because the component is suspended
+	// we set the vnode.type as `null` so that it is not a typeof function
+	// so the unmount will remove the vnode._dom
+	if (component && vnode._hydrating === true) {
+		vnode.type = null;
+	}
+
 	if (oldUnmount) oldUnmount(vnode);
 };
 
@@ -119,6 +128,7 @@ Suspense.prototype._childDidSuspend = function(promise, suspendingVNode) {
 		if (resolved) return;
 
 		resolved = true;
+		suspendingComponent._onResolve = null;
 
 		if (resolve) {
 			resolve(onSuspensionComplete);
@@ -127,10 +137,7 @@ Suspense.prototype._childDidSuspend = function(promise, suspendingVNode) {
 		}
 	};
 
-	suspendingComponent._onResolve = () => {
-		suspendingComponent._onResolve = null;
-		onResolved();
-	};
+	suspendingComponent._onResolve = onResolved;
 
 	const onSuspensionComplete = () => {
 		if (!--c._pendingSuspensionCount) {
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -22,6 +22,16 @@ options._catchError = function(error, newVNode, oldVNode) {
 	oldCatchError(error, newVNode, oldVNode);
 };
 
+const oldUnmount = options.unmount;
+options.unmount = function(vnode) {
+	/** @type {import('./internal').Component} */
+	const component = vnode._component;
+	if (component && component._onResolve) {
+		component._onResolve();
+	}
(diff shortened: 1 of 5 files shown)


src/internal.d.ts
```diff
@@ -39,6 +39,8 @@ export interface Options extends preact.Options {
 	_skipEffects?: boolean;
 	/** Attach a hook that is invoked after an error is caught in a component but before calling lifecycle hooks */
 	_catchError(error: any, vnode: VNode, oldVNode?: VNode | undefined): void;
+	/** Attach a hook that is invoked after a component is unmounted. */
+	_unmount?(vnode: VNode): void;
```
JoviDeCroock, line 43: Why did we add a type for a hook that isn't used? Afaik we're using the public hook `unmount`

tanhauhau: oh i didnt know about it, i saw it's used by hooks, and other preact/ modules

JoviDeCroock: Should just be `options.unmount` source https://github.com/preactjs/preact/blob/master/src/diff/index.js on line 468

compat/src/suspense.js
```diff
@@ -22,6 +22,15 @@ options._catchError = function(error, newVNode, oldVNode) {
 	oldCatchError(error, newVNode, oldVNode);
 };
 
+const oldUnmount = options.unmount;
```
JoviDeCroock, line 25: You are never calling oldUnmount, this could lead to a plugin chain failing.

tanhauhau: oops good catch 🙈

compat/src/suspense.js
```diff
@@ -22,6 +22,16 @@ options._catchError = function(error, newVNode, oldVNode) {
...
+options.unmount = function(vnode) {
+	/** @type {import('./internal').Component} */
+	const component = vnode._component;
+	if (component && component._onResolve) {
+		component._onResolve();
+	}
+	oldUnmount(vnode);
```
JoviDeCroock, line 32:
```suggestion
	if (oldUnmount) oldUnmount(vnode);
```

src/internal.d.ts
tanhauhau: oh... got it. i've removed this type.

marvinhagemeister: The typing change needs to be reverted too, otherwise the PR looks good to me 👍 Great work!

developit:
From the size bot:

**Size Change:** +29 B (0%) 

**Total Size:** 42 kB

|:--- |:---:|:---:|:---:|
| `compat/dist/compat.js` | 3.35 kB | +10 B (0%) |  |
| `compat/dist/compat.module.js` | 3.34 kB | +9 B (0%) |  |
| `compat/dist/compat.umd.js` | 3.4 kB | +10 B (0%) |  |

compat/src/suspense.js
```diff
@@ -119,14 +127,9 @@ Suspense.prototype._childDidSuspend = function(promise, suspendingVNode) {
...
-		suspendingComponent.componentWillUnmount;
-	suspendingComponent.componentWillUnmount = () => {
+	suspendingComponent._onResolve = () => {
+		suspendingComponent._onResolve = null;
 		onResolved();
-
-		if (suspendingComponent._suspendedComponentWillUnmount) {
-			suspendingComponent._suspendedComponentWillUnmount();
-		}
 	};
```
developit, line 133: I think this could be simplified to:

```js
	suspendingComponent._onResolve = onResolved;
```

... as long as the following was added around L112:

```js
		suspendingComponent._onResolve = null;
```

tanhauhau: got it! updated!

tanhauhau (commented):
I've added some test cases and relevant fixes from #2989

JoviDeCroock (approved):

JoviDeCroock:
Thank you so much for all the efforts @tanhauhau

PR #2921, opened 2021-01-12
Fix broken watcher after build error in tests

marvinhagemeister:
Turns out `karma`'s error reporting isn't built with watch mode in mind. Whenever one forwards an error to `karma` the watcher fails to process new events. So `karma-esbuild` will now rely on `esbuild`'s logging and return an empty file instead.


marvinhagemeister:
Replaced by #2923

PR #2931, opened 2021-01-15
Fix compat/jsx-dev-runtime missing in package

### LINKED ISSUE #2930: jsx-dev-runtime file is missing in preact/compat after upgrading to 10.5.10
### Reproduction

I've been using Preact with Emotion in a Next.js app. We've had to jump through some [hoops](https://github.com/preactjs/next-plugin-preact/issues/16) to get it to work without having to import Emotion's JSX pragma in each file that uses the `css` prop, but that's all right.

With the last release of Preact that hoop jumping should be fixed by the addition of the `jsx-dev-runtime` files, but after upgrading, it seems to be missing from the `preact` folder.

https://github.com/nickrttn/emotion-preact-404

### Steps to reproduce

**The issue on its own**

1. Initialize a new project (I used `yarn init`, all default options)
2. Install `preact@10.5.10`
3. Check the `./node_modules/preact/compat` folder to see that the `jsx-dev-runtime` files that were added in `10.5.10` are missing here.

**The issue in my specific case**

1. Clone https://github.com/nickrttn/emotion-preact-404
2. Run `yarn`
3. Run `yarn dev` or `yarn build` and an error will appear
4. Check the `./node_modules/preact/compat` folder to see that the `jsx-dev-runtime` files that were added in `10.5.10` are missing here.

### Expected Behavior

`jsx-dev-runtime` files to be present.

### Actual Behavior

`jsx-dev-runtime` files are missing.

marvinhagemeister:
Fixes #2930

### DIFF
--- a/package.json
+++ b/package.json
@@ -171,6 +171,8 @@
 		"compat/server.js",
 		"compat/jsx-runtime.js",
 		"compat/jsx-runtime.mjs",
+		"compat/jsx-dev-runtime.js",
+		"compat/jsx-dev-runtime.mjs",
 		"compat/package.json",
 		"debug/dist",
 		"debug/src",


JoviDeCroock (approved):

sventschui:
Thanks a lot @marvinhagemeister !

PR #2932, opened 2021-01-15
Tests: Fix vscode breakpoints not being hit

marvinhagemeister:
Turns out that inline source mapping paths are always resolved relative to the source file which had the `//# sourceMappingURL=` comment. This doesn't play well with breakpoints set from vscode as they expect it to be a straight forward mapping between paths. This is usually the absolute path to the file and vscode has a `pathMapping` setting to swap out prefixes.

The way webpack solves it, is by always resolving paths to the current working directory and appending `webpack:///./` in front of every url. Didn't want to use that, so instead we'll always use the absolute mapping instead. This leads to the launch configuration being shortened to:

```json
{
  "type": "chrome",
  "request": "attach",
  "name": "Attach Karma",
  "address": "localhost",
  "port": 9333
}
```

Screenshot from 2021-01-15 19-06-51

Screenshot from 2021-01-15 19-09-44

To be able to do that I had switch away from inline source maps in `karma-esbuild`. They are now always a separate file. An interesting advantage of this approach is that live-editing in Chrome of JS files is much smoother, and IE11 doesn't lag as much when inspecting tests :tada:


JoviDeCroock (approved):

andrewiggins (approved):
🎉🎉🎉

PR #2934, opened 2021-01-16
Fail build on missing compat entries in files entry

marvinhagemeister:
Add a post build script to verify that all entry files in `compat` are present in the `"files"` entry in `package.json`. Discovered that `compat/test-utils.js` was never shipped in our npm package with that. In the long run the hope is that everyone plays nice and respects the defined package entries in `"exports"`. Until then this should allow us to catch missing files and prevent issues like #2930 from happening.


andrewiggins (approved):
Good call!

PR #2935, opened 2021-01-17
Tests: Fix stack traces not sourcemapped

marvinhagemeister:
Turns out that for source-mapped stack traces to work we need to set an undocumented property on karma's `file` object. The recent update to `karma-esbuild` fixes that.


andrewiggins (approved):
🙌🙌

PR #2936, opened 2021-01-17
Create a separate code path for mounting and hydration

andrewiggins:
This PR is the first PR of our renderer restructure. It creates a separate code path for mounting and hydration from diffing. While this does duplicate some code between the two the code paths, it makes the renderer much easier to reason about since a single function doesn't have to concern itself with all forms of rendering (mounting, hydration, and diffing). Further, the mounting code path can be dramatically simplified since it can assume it is mounting/hydrating. This PR doesn't include all the simplifications we can now make, but is a first start we'd like to build off of.

Further optimizations and simplifications will come in future PRs.


andrewiggins:
Not sure why our compressed size action is failing... Seems to happen with changes to mangle.json. Perhaps the git checkouts aren't properly updating the file?

JoviDeCroock (approved):
Nice work!

marvinhagemeister (approved):
Let's go!! :raised_hands:

PR #2938, opened 2021-01-18
Add test case setting contentEditable to false

marvinhagemeister:
Missed in my last PR that `false` has a special meaning, similar to aria attributes.

Fixes #2926 .


developit:
I think #2329 might support this without the extra 2 custom behaviors (if I can get CI to work).

marvinhagemeister:
@developit Fix was merged with #2939, but let's add the test case from this PR regardless.

JoviDeCroock (approved):

developit:
Good call @marvinhagemeister - I'd meant to copy it over.

PR #2939, opened 2021-01-18
Use try/catch for properties instead of an allow list, remove `xlink:`

developit:
This PR includes two fundamental changes to the way we do prop diffing:

1. Instead of safeguarding against setting read-only reflected DOM properties, it attempts to set the property in a `try/catch` and falls back to setting an attribute

2. The codepath for detecting `xlink:href` / `xlinkHref` props and applying them via `setAttributeNS()` is simply removed. This was only ever used for the `xlink:href` attribute, which was removed from SVG 2. The non-namespaced `href` attribute works exactly the same way. As a bridging solution, since there are lots of folks using `<use xlinkHref="..">`, I've retained the code that strips any `xlink` namespace prefix off of the prop name.

Fixes #2926.

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -126,7 +126,10 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
 
 		if (typeof value === 'function') {
 			// never serialize functions as attribute values
-		} else if (value != null && (value !== false || /^ar/.test(name))) {
+		} else if (
+			value != null &&
+			(value !== false || (name[0] === 'a' && name[1] === 'r'))
+		) {
 			dom.setAttribute(name, value);
 		} else {
 			dom.removeAttribute(name);
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -56,13 +56,7 @@ function setStyle(style, key, value) {
 export function setProperty(dom, name, value, oldValue, isSvg) {
 	let useCapture, nameLower, proxy;
 
-	if (isSvg && name == 'className') name = 'class';
-
-	// if (isSvg) {
-	// 	if (name === 'className') name = 'class';
-	// } else if (name === 'class') name += 'Name';
-
-	if (name === 'style') {
+	o: if (name === 'style') {
 		if (typeof value == 'string') {
 			dom.style.cssText = value;
 		} else {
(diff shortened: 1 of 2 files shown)


src/diff/props.js
```diff
@@ -103,49 +97,39 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
...
-					name.toLowerCase(),
-					value
-				);
-			}
+	} else if (name !== 'dangerouslySetInnerHTML') {
+		if (isSvg) {
+			// Normalize incorrect prop usage for SVG:
+			// - xlink:href / xlinkHref --> href (xlink:href was removed from SVG and isn't needed)
+			// - className --> class
+			name = name.replace(/xlink[H:h]/, 'h').replace(/Name$/, '');
```
andrewiggins, line 105: This might break the [`attributeName` SVG attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/attributeName)

src/diff/props.js
```diff
@@ -103,49 +97,39 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
...
+		// ARIA-attributes have a different notion of boolean values.
+		// The value `false` is different from the attribute not
+		// existing on the DOM, so we can't remove it. For non-boolean
+		// ARIA-attributes we could treat false as a removal, but the
+		// amount of exceptions would cost us too many bytes. On top of
+		// that other VDOM frameworks also always stringify `false`.
+
+		if (typeof value === 'function') {
+			// never serialize functions as attribute values
+		} else if (value != null && (value !== false || /^ar/.test(name))) {
```
andrewiggins, line 129: This saves a couple bytes on my machine.

```suggestion
		} else if (
			value != null &&
			(value !== false || (name[0] === 'a' && name[1] === 'r'))
		) {
```

src/diff/props.js
andrewiggins: I think this works. It adds a couple bytes but on my machine they are offset by the change I suggested in my other comment below.

```suggestion
			name = name.replace(/xlink[H:h]/, 'h').replace(/sName$/, 's');
```

andrewiggins (commented):
This is great!! 🙌🎉

JoviDeCroock (approved):

src/diff/props.js
developit: Ah good catch, I assumed I was safe because this was SVG land, but that's actually where casing goes wild haha.

src/diff/props.js
developit: dang! I played around with this and for some reason never thought of that.

developit:
I will kick CI again in an hour haha

src/diff/props.js
```diff
@@ -90,62 +84,59 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
...
-				!/^ar/.test(name))
+			name !== 'href' &&
+			name !== 'list' &&
+			name !== 'form' &&
+			name !== 'download' &&
+			name in dom
+		) {
+			try {
+				dom[name] = value == null ? '' : value;
+				// labelled break is 1b smaller here than a return statement (sorry)
```
ForsakenHarmony, line 119: is that worth it?

marvinhagemeister: I think it's fine, given that the comment explains what's happening.

marvinhagemeister (approved):
Fantastic stuff 🙌

PR #2940, opened 2021-01-18
Simplify top-level render and remove dead code

andrewiggins:
Simplify top-level render a bit by removing IS_HYDRATE constant and making sure we use the same condition for checking replaceNode everywhere so it gzips better.

Also, remove some code from our suspended hydration catching mechanism I believe is unused. It bubbles up the value of the `_hydrating` flag to the VNode that catches the suspension. However in our Suspense implementation, it looks at the value of the `_hydrating` on the VNode that suspended (not Suspense itself) to determine if a node suspended while hydrating. So we have no need to bubble up the `_hydrating` flag.

### DIFF
--- a/src/diff/catch-error.js
+++ b/src/diff/catch-error.js
@@ -9,8 +9,6 @@ export function _catchError(error, vnode) {
 	/** @type {import('../internal').Component} */
 	let component, ctor, handled;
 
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
Sweet 🙌

PR #2942, opened 2021-01-20
Golf element diffing [-17b]

developit:
This takes advantage of the fact that `Text.prototype.localName === undefined` to skip a few property lookups, savings bytes and hopefully also improving performance a tiny bit.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -315,6 +315,9 @@ function diffElementNodes(
 			// if newVNode matches an element in excessDomChildren or the `dom`
 			// argument matches an element in excessDomChildren, remove it from
 			// excessDomChildren so it isn't later removed in diffChildren
+			//
+			// Note: This takes advantage of Text nodes having `.localName=undefined`,
+			// which is loosely equal to Text VNodes' `.type=null`. Elements use string equality.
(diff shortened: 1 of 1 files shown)


src/diff/index.js
```diff
@@ -314,13 +315,7 @@ function diffElementNodes(
...
 			// argument matches an element in excessDomChildren, remove it from
 			// excessDomChildren so it isn't later removed in diffChildren
-			if (
-				child != null &&
-				((newVNode.type === null
-					? child.nodeType === 3
-					: child.localName === newVNode.type) ||
-					dom == child)
-			) {
+			if (child != null && (dom == child || child.localName == nodeType)) {
```
developit, line 318: I'm leaving this as a suggestion to avoid the comment push re-running benchmarks:

```suggestion
			//
			// Note: This takes advantage of Text nodes having `.localName=undefined`,
			// which is loosely equal to Text VNodes' `.type=null`. Elements use string equality.
			if (child != null && (dom == child || child.localName == nodeType)) {
```

marvinhagemeister (approved):

andrewiggins:
Yea, I really got fix up this reporter thing to not hit the limit so much lol

JoviDeCroock (approved):

PR #2943, opened 2021-01-20
Pass all props to createElement instead of just `is` [-5b]

developit:
The second argument to createElement accepts an object, but only looks at the `is` property. We can just always pass `props` (or we could remove this feature? not sure if it's actually still being used).


developit:
yay, looks like it's performance-neutral.

marvinhagemeister (approved):

PR #2944, opened 2021-01-20
Separate benchmarks into their own workflow

andrewiggins:
This change enables us to only run benchmarks on PRs that change files relevant to the benchmarks.

For example, if I did this correctly, this PR should not have any benchmarks run (though I'll make a silly change to a source file to test them).

I'll mark as ready after I've tested.


andrewiggins:
(I cancelled the benchmarks to not use our API limits since this is just a test that they are invoked)

marvinhagemeister (approved):

PR #2952, opened 2021-01-21
Add addHookName to preact/devtools to relabel native hooks in devtools

marvinhagemeister:
This PR adds a new `addHookName` function to the `preact/devtools` entry. It can be used to replace generic `useState` names with something specific.

image

See the related issue on the devtools repo for more background information: https://github.com/preactjs/preact-devtools/issues/280

### DIFF
--- a/devtools/package.json
+++ b/devtools/package.json
@@ -9,6 +9,7 @@
 	"umd:main": "dist/devtools.umd.js",
 	"source": "src/index.js",
 	"license": "MIT",
+	"types": "src/index.d.ts",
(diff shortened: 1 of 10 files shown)


marvinhagemeister:
@developit Per your suggestion I've moved it out of hooks into `preact/devtools`. Didn't move it to `preact/debug` because there are valid production scenarios where one cares about readable hooks, but doesn't want to include the whole of `preact/debug`. Let me know what you think!

JoviDeCroock (approved):

PR #2953, opened 2021-01-22
Use artifact flow to report benchmarks results

andrewiggins:
Instead of reporting results in each benchmark job (which requires coordinating updating the comment between the concurrent jobs), we'll collect all the results in an artifact and then report the results after the benchmarks have finished.

This change dramatically reduces the number GitHub APIs calls and so should reduce the likelihood of our action getting throttled. The action still does more API calls than I like but figured I get this update out first to reduce the current strain.

Further, this serves as a building block to setting up the flow to enable benchmark reports for PRs from forks!! :D

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -89,12 +89,6 @@ jobs:
         with:
           name: results
           path: benches/results/text_update.json
-      - uses: andrewiggins/tachometer-reporter-action@v2
-        with:
-          path: benches/results/text_update.json
-          report-id: text_update
-          base-bench-name: preact-master
-          pr-bench-name: preact-local
 
   bench_many_updates:
     name: Bench many_updates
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -58,7 +58,7 @@ jobs:
           name: build-output
           path: preact.tgz
       - name: Initialize Tachometer Results Report
-        uses: andrewiggins/tachometer-reporter-action@v2
+        uses: andrewiggins/tachometer-reporter-action@v2.3.0
```
andrewiggins, line 61: Gonna update this to "@v2" once I verify this works and publish it to the v2 tag

andrewiggins: Done!

marvinhagemeister (approved):

Issue #2949, opened 2021-01-21
Strange difference in behaviour between preact and react in combination with @material-ui and conditional operator


masbaehr:
This is the code to test the behavior difference.

Note: I know that it can be fixed by using the "boolswitch &&" / "!boolswitch &&" Syntax. But i noticed this after switching from react to preact. I'm not entirely sure if this might be a material-ui shortcoming but as it happened after switching to Preact i thought it might be better suited to post the issue here. 

```javascript
import React from 'react';
import TextField from '@material-ui/core/TextField';


function Debug(props) {
 
  const [boolswitch, setboolswitch] = React.useState(false);

  return (
    <div>
        
    {boolswitch ? '' : 
        
        <TextField  
          style={{width: "100%"}}
          label={"Username"}
          variant="outlined"
          margin="normal"
          onKeyPress={event => {

          }}
          onChange={event => {}}
        />
    }
      {boolswitch ? '' : 
         <TextField  
          style={{width: "100%"}}
          label={"Password"}
          variant="outlined"
          margin="normal"
          onKeyPress={event => {

          }}
          onChange={event => {}}
        />
        
       
    }
    {boolswitch ?
     
        <TextField  
          style={{width: "100%"}}
          label={"2fa"}
          variant="outlined"
          margin="normal"
          onKeyPress={event => {

          }}
          onChange={event => {}}
        />
        : ''
    }
      <button onClick={event => {
          setboolswitch(true);
      }}>Continue</button>
    </div>
  );
}

export default Debug;


```

First step:

![image](https://user-images.githubusercontent.com/38694044/105323513-83577380-5bca-11eb-9f85-28349d865cab.png)

Second step: 
Problem: The internal value state of the username field is rendered for the now visible 2fa field

![image](https://user-images.githubusercontent.com/38694044/105323546-8b171800-5bca-11eb-8f71-9269ce11004c.png)

I also know that it could be avoided using States for all the values, but this would mean that a state update while typing would be needed which slows down the typing.

Summary: Somehow the   ? '' : ''  Syntax seems to cause a difference while rendering in Preact vs. React

developit:
@masbaehr This can be corrected by adding `key="2fa"` to the last TextField.

masbaehr:
Oh okay, is it a requirement in preact to use "key" for every component? Maybe I missed that? By the way i also noticed some other differences in rendering e.g. the defaultValue is not correctly populated if a variable changes on rerender

[...]

PR #2955, opened 2021-01-23
Add unkeyed swap test for issue #2949

marvinhagemeister:
Don't have a fix for it yet, but figured the test case is good to have in our suite regardless until we find a fix.


JoviDeCroock:
Closing: test-only PR for #2949; children diffing was rewritten in v11 (findMatchingIndex) and the test would need porting.

PR #2501, opened 2020-04-24
support switching context value in useContext

JoviDeCroock:

PR #2958, opened 2021-01-24
implement useContextSelector

JoviDeCroock:
Todo:

- [ ] await final naming decision `useSelectedContext` vs `useContextSelector`
- [x] Allow for changing the selector inside of the `createContext` subscription (https://github.com/preactjs/preact/pull/2501/files#diff-c580d3b3982d2cf5341923532890e75fceb15c1fd1e7993f9b7ba897e3dbe2e1R230)

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -250,8 +250,9 @@ export function useContext(context, selector) {
 	// This is probably not safe to convert to "!"
 	if (state._value == null) {
 		state._value = true;
-		provider.sub(currentComponent, selector);
+		state._subscription = provider.sub(currentComponent, selector);
 	}
+
 	return provider.props.value;
 }
 
@@ -261,6 +262,9 @@ export function useContext(context, selector) {
  */
 export function useContextSelector(context, selector) {
 	const contextValue = useContext(context, selector)
+	const state = getHookState(currentIndex, 11);
+	state._subscription[1] = selector;
+
 	return selector(contextValue);
 }
 
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -248,7 +248,7 @@ export function useContext(context, selector) {
 	state._context = context;
 	if (!provider) return context._defaultValue;
 	// This is probably not safe to convert to "!"
-	if (state._value == null) {
+	if (!state. _subscription) {
 		state._value = true;
 		state._subscription = provider.sub(currentComponent, selector);
 	}
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -235,7 +235,7 @@ export function useCallback(callback, args) {
 /**
  * @param {import('./internal').PreactContext} context
  */
-export function useContext(context) {
+export function useContext(context, selector) {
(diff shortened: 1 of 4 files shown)


src/create-context.js
```diff
@@ -40,17 +40,22 @@ export function createContext(defaultValue, contextId) {
...
 
-				this.sub = c => {
-					subs.push(c);
+				this.sub = (c, selector) => {
+					const entry = [c, selector]
+					const newLength = subs.push(entry);
 					let old = c.componentWillUnmount;
 					c.componentWillUnmount = () => {
-						subs.splice(subs.indexOf(c), 1);
+						subs.splice(newLength - 1, 1);
```
marvinhagemeister, line 55: Not sure if it is save to cache the position in the subscription array. If another subscription is removed the array will change and `newLength` may shift by one, effectively pointing to the wrong index.

src/create-context.js
```diff
@@ -40,17 +40,22 @@ export function createContext(defaultValue, contextId) {
...
+							if (!c[1] || c[1](_props.value) !== c[1](this.props.value))
+								enqueueRender(c[0])
+						});
 					}
 				};
 
-				this.sub = c => {
-					subs.push(c);
+				this.sub = (c, selector) => {
+					const entry = [c, selector]
```
developit, line 51:
```suggestion
				this.sub = function() {
					const entry = arguments;
```

hooks/src/index.js
```diff
@@ -250,11 +250,24 @@ export function useContext(context) {
 	// This is probably not safe to convert to "!"
 	if (state._value == null) {
```
developit, line 251: This might be smaller now:

```suggestion
	if (!state. _subscription) {
```

hooks/src/index.js
```diff
@@ -250,11 +250,24 @@ export function useContext(context) {
...
 }
 
+/**
+ * @param {import('./internal').PreactContext} context
+ * @param {(c: any) => any} selector
+ */
+export function useContextSelector(context, selector) {
+	const contextValue = useContext(context, selector)
+	const state = getHookState(currentIndex, 11);
+	state._subscription[1] = selector;
```
developit, line 266: Since nothing currently assumes `useContext()` will ignore updates to `selector`, maybe we could just have it do this there?

```js
export function useContext(context, selector) {
	const state = getHookState(currentIndex++, selector ? 11 : 9);
	state._context = context;
	if (!provider) return context._defaultValue;
	const sub = state._subscription || (state._subscription = provider.sub(currentComponent));
	sub[1] = selector;
	return selector ? selector(provider.props.value) : provider.props.value;
}

export const useSelectedContext = useContext;
```

marvinhagemeister: Ohh I like that idea :+1:

developit: another option that might be worth checking (really just for size) would be fully duplicating the logic here into `useContext` / `useContextSelector` - duplication tends to be better when compressed, might be smaller than the two ternaries above.

PR #2962, opened 2021-01-25
Update size action to only run when src files change

andrewiggins:


.github/workflows/size.yml
```diff
@@ -1,15 +1,17 @@
 name: Compressed Size
 
-on: [pull_request]
+on:
+  pull_request:
+    branches:
+      - '**'
+    paths:
+      - '**/src/**.js'
```
andrewiggins, line 8: I think this glob should do the right thing... but do keep an eye out for the workflow not running when it should

JoviDeCroock (approved):

.github/workflows/size.yml
```diff
@@ -1,15 +1,17 @@
...
 
 jobs:
   build:
     runs-on: ubuntu-latest
-
     steps:
-      - uses: actions/checkout@v2-beta
-        with:
-          fetch-depth: 1
```
andrewiggins, line 12: `fetch-depth` defaults to `1` in v2

PR #2965, opened 2021-01-25
Merge from master into restructure

andrewiggins:
Merge last week's work on master into restructure


src/diff/mount.js
```diff
@@ -29,20 +29,17 @@ export function mount(
 	startDom,
 	isHydrating
 ) {
-	let tmp,
```
andrewiggins, line 32: This change (and the similar one in `patch.js` are some bonus golf I discovered while working on the merge :D

JoviDeCroock (approved):

marvinhagemeister (approved):

PR #2969, opened 2021-01-28
golf -_pendingEffects

shoonia:
There is reset is pending effects in any case


shoonia:
I'm not sure what that don't affect to parent nodes yet.

PR #2970, opened 2021-01-30
Remove excessDomChildren parameter

andrewiggins:
Remove excessDomChildren parameter and instead use the startDom parameter to pass hydration DOM nodes down the tree.

This PR also introduces the concept of modes on VNodes. This mode property determines whether a VNode is hydrating or suspended, etc.


andrewiggins:
Lovin the movement on that hydrate1k benchmark!

JoviDeCroock (approved):
WOW 😅 that's some nice work right there

marvinhagemeister (approved):

PR #2971, opened 2021-02-01
Deprecate component.base

marvinhagemeister:


JoviDeCroock (approved):

developit (approved):
LGTM. Since it's in follow-ups.md we should be able to remember to include this in the upgrade shim.

PR #2972, opened 2021-02-02
remove _clearProcessingException

JoviDeCroock:


src/diff/catch-error.js
```diff
@@ -7,26 +9,26 @@
  */
 export function _catchError(error, vnode) {
 	/** @type {import('../internal').Component} */
-	let component, ctor, handled;
+	let component, ctor;
+
+	if (vnode._component) vnode._component._mode = MODE_ERRORED;
```
developit, line 14: I don't think we did this at the other mode callsites yet, but this seems like a case where we need to account for the possibility of combined mode flags (maybe?):

```suggestion
	if (vnode._component) vnode._component._mode |= MODE_ERRORED;
```

If there's no combinations here yet we could punt on it for now - I know we'd agreed to only do bitfields via wrapper functions to keep the code clear.

src/diff/catch-error.js
```diff
@@ -7,26 +9,26 @@
...
 
 	for (; (vnode = vnode._parent); ) {
-		if ((component = vnode._component) && !component._processingException) {
+		if ((component = vnode._component) && component._mode !== MODE_ERRORED) {
```
developit, line 17: (as per above)

```suggestion
		if ((component = vnode._component) && component._mode ^ MODE_ERRORED) {
```

developit:
Question: do components have `_mode`? I thought that was only a [VNode property](https://github.com/preactjs/preact/blob/be57a26289283e04b69035884ceb870444b2192b/src/create-element.js#L74).

FWIW it seems like here we would be fine to store this info in `vnode._mode` instead of `vnode._component._mode`, and doing so actually nicely skips some property accesses.

JoviDeCroock:
@developit feels unsafe to me since that'll force us to always carry over mode from oldVNode this is a solution until we can store on a backer

JoviDeCroock:
Current:

```
Build "preact" to dist:
       4208 B: preact.js.gz
       3809 B: preact.js.br
       4220 B: preact.module.js.gz
       3835 B: preact.module.js.br
       4277 B: preact.umd.js.gz
       3867 B: preact.umd.js.br
```

Before:

```
Build "preact" to dist:
       4233 B: preact.js.gz
       3834 B: preact.js.br
       4238 B: preact.module.js.gz
       3846 B: preact.module.js.br
       4297 B: preact.umd.js.gz
       3872 B: preact.umd.js.br
```

marvinhagemeister (approved):
Great work 👍

PR #2977, opened 2021-02-06
restructure-defaultprops

marvinhagemeister:
This PR moves `defaultProps` handling out of core into compat for v11.

Migration:

```jsx
// Preact < 11
function Foo(props) {
  return <div>Hello {props.name}</div>;
}

Foo.defaultProps = {
  name: "world",
};

// vs Preact 11
function Foo({ name = "world" }) {
  return <div>Hello {name}</div>
}
```

One thing I'm unsure about if `defaultProps` is actually tied to class components. Of course, we can use it with functional components per se, but that doesn't make much sense. For class component it kinda does as there is no unified way to change props before component lifecycles are invoked, apart from `defaultProps`.


porfirioribeiro:
For functional components i think removing defaultProps makes much sense, i don't remember the last time i used it, but for class components it makes sense to have them.
Sure we can do it like this:
```javascript
class MyComponent extends Component{
  render({ name = "world" }) {
    return <div>Hello {name}</div>
  }
}
```
But if you need it on any other of the lifecycle events it would not be easy, or you would need to replicate the default value.

Now if we manage to move `Component` out of core, maybe `defaultProps` could be something tied to class components only?

marvinhagemeister:
@porfirioribeiro yup, I came to the same conclusion when opening this PR (see the last paragraph) :+1:

developit:
Yeah, kinda thinking that as well. Maybe the better option here would be to have this move into the extracted class components implementation once that's landed?

One other option (not sure if it's better/worse) would be to apply defaultProps during diffing. We used to do this, but IIRC it ended up being faster+smaller to do it in createElement.

andrewiggins:
Re: what Jason suggested, could we move this into the Component's base constructor?

marvinhagemeister:
Moving it to the Component constructor might not be sufficient when we reuse component instances on updates:

```jsx
function Foo(props) {
  return <div>{props.foo}</div>;
}

Foo.defaultProps = {
  foo: "defaultValue"
};

  const [v, set] = useState(true);
  return (
    <div>
      <button onClick={() => set(!v)}>toggle</button>
      <Foo foo={v ? "bob" : undefined} />
    </div>
  );
}
```

Output:

1. render `bob`
2. render after state update: `defaultValue`

JoviDeCroock (approved):
I've been thinking about this but the types are hard to get right 😅

JoviDeCroock:
Closing: targets the abandoned `restructure` branch, and the idea already shipped — in v11 `defaultProps` lives in compat (`compat/src/render.js`), out of core. Thanks for the work here!

PR #2978, opened 2021-02-06
Improve performance of vnodeId generation

developit:
This PR removes the need for an `options._vnodeId` shared counter by creating separate "keyspaces" within `vnodeId` (`vnode._original`) using signed integers. Despite not sharing the ID counter, `h()` allocates positive monotonic ID's and `jsx()` negative IDs, which prevents collisions. It also doubles the number of available ID values. The performance improvement here is a result of avoiding a property get + set on every VNode, not sure if that will come through in the benchmarks since we already do a loop and set many properties on `props`.

Another option that might be worth testing would be to keep IDs unsigned by having `jsx()` generate IDs starting at 9e9. This would create a collision scenario if `h()` were to create 9 billion VNodes.


marvinhagemeister (approved):
Nice one!

PR #2981, opened 2021-02-06
add a few minor tests

43081j:
its not much but i just wanted to get my head around the source and figured filling a couple of minor gaps in might help.


developit:
Nice! We were actually just talking about removing defaultProps support in #2977, but realized it's still fairly widely used for class-based components. These tests seem handy.

marvinhagemeister (approved):
This is awesome :raised_hands:

PR #2983, opened 2021-02-07
Add support for root nodes

marvinhagemeister:
This PR introduces the concept of a root node which specifies into which dom node its children should be rendered into. This ensures that the vnode tree contains _all_ vnodes and isn't spliced up into random detached branches like before. It's based on the ideas in #2627, minus the `createRoot` API itself.

With these in place we can experiment with removing the `_parentDom` property on component instances, too. Previously we didn't have a way to get the topmost dom node because the vnode tree didn't store any reference to it.

~I skipped two Portals test that we need to get back working for v11. Conceptually Portals always render before the first child of the container node. But existing portals should remain where they are and this leads to awkward outcomes like:~

```html
<div><i>Portal #1</i><div>hello</div><i>Portal #2</i></div>
```

~Personally I feel like this PR is good enough to be merged and unblocks the work on extracting component classes, despite skipping those two Portal tests. Or should I spent more time trying to get those passing?~

### DIFF
--- a/src/diff/component.js
+++ b/src/diff/component.js
@@ -147,6 +147,19 @@ export function renderComponent(
 
 	if ((tmp = options._render)) tmp(internal);
 
+	// Root nodes signal that we attempt to render into a specific DOM node
+	// on the page. Root nodes can occur anywhere in the tree and not just
+	// at the top.
+	if (newProps._parentNode) {
+		parentDom = newProps._parentNode;
+
+		// The `startDom` variable might point to a node from another
+		// tree from a previous render
+		if (startDom != null && startDom.parentNode !== parentDom) {
+			startDom = null;
+		}
+	}
+
 	internal._flags &= ~DIRTY_BIT;
 	c._internal = internal;
 	c._parentDom = parentDom;
--- a/src/diff/component.js
+++ b/src/diff/component.js
@@ -153,8 +153,8 @@ export function renderComponent(
 	if (newProps._parentDom) {
 		parentDom = newProps._parentDom;
 
-		if (oldVNode && oldVNode._dom) {
-			startDom = oldVNode._dom;
+		if (internal && internal._dom) {
+			startDom = internal._dom;
 		}
 
 		// The `startDom` variable might point to a node from another
--- a/src/diff/component.js
+++ b/src/diff/component.js
@@ -153,6 +153,10 @@ export function renderComponent(
 	if (newProps._parentDom) {
 		parentDom = newProps._parentDom;
 
+		if (oldVNode && oldVNode._dom) {
+			startDom = oldVNode._dom;
+		}
+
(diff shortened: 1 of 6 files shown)


src/diff/component.js
```diff
@@ -142,6 +142,19 @@ export function renderComponent(
 
 	if ((tmp = options._render)) tmp(newVNode);
 
+	// Root nodes signal that we attempt to render into a specific DOM node
+	// on the page. Root nodes can occur anywhere in the tree and not just
+	// at the top.
+	if (newProps._parentNode) {
+		parentDom = newProps._parentNode;
+
```
andrewiggins, line 150: Adding the following before checking `startDom` fixes the skipped tests for me. I think what it is saying is "Let's resume the diff of the portal from the previously rendered child of the portal"

```suggestion

		if (oldVNode && oldVNode._dom) {
			startDom = oldVNode._dom;
		}

```

andrewiggins: I think we also need to capture the previous value of `startDom` here and return it at the end of this function so that the diff can resume from where it left off before the Portal. In other words, the diff should act as if the portal isn't there and so we need to manage the `startDom` pointer so that VNodes around the portal see no difference (i.e. startDom doesn't change).

As this is implemented today, this function will return the `nextSibling` of the last child of the Portal's children which won't be in the same DOM tree as the rest of the diff.

I don't think our tests hit this today cuz most of them put portals at the end of a `<div>`:
```
<div>
    <p>Hello</p>
    {createPortal()}
</div>
```
at which point the `nextSibling` value of the Portal child is thrown away since it is the last child of a div.

I'm working on a test case for this bug...

andrewiggins: Okay here is the test case that fails:
```js
it("should not needlessly append siblings to Portals", () => {
  /** @type {() => void} */
  let update;
  function Foo(props) {
    const [state, setState] = useState(false);
    update = () => setState((state) => !state);
    return (
      <div>
        <p>Hello</p>
        {createPortal(props.children, scratch)}
        <p>World!</p>
      </div>
    );
  }

  render(
    <Foo>
      <div>foobar</div>
    </Foo>,
    scratch
  );
  expect(scratch.innerHTML).to.equal(
    "<div>foobar</div><div><p>Hello</p><p>World!</p></div>"
  );

  clearLog();
  update();
  rerender();

  expect(scratch.innerHTML).to.equal(
    "<div>foobar</div><div><p>Hello</p><p>World!</p></div>"
  );
  expect(getLog()).to.deep.equal([]);
});
```

marvinhagemeister: That's amazing! You are right on the money with this. Storing the previous `startDom` pointer and returning it at the end is what makes all tests pass 👍

andrewiggins (approved):
🎉🎉

PR #2984, opened 2021-02-07
[Draft] Export createVNode

andrewiggins:
This PR is just an experiment to see what kind of perf gains are possible if `createElement` didn't have to do any massaging of  inputs (no loops over props, etc.) and just returned the vnode object.


JoviDeCroock:
Closing: `createVNode` stays internal-only in v11 with no public export planned; this perf experiment is superseded.

PR #2988, opened 2021-02-08
fix reordering issue of memoized component when the component initially render null

tanhauhau:
Found a reorder issue for memoised component, as the `oldVNode.vnode._children` still points to the `oldVNode`, causing it to not able to find the correct nextSibling through `getDomSibling`


JoviDeCroock (approved):

marvinhagemeister (approved):
This is fantastic 🙌

PR #2989, opened 2021-02-08
fix updating suspended component after hydration

tanhauhau:
as the Suspense component not rendering a placeholder during hydration, if it rerenders before the promise resolves, the same component get suspended twice, causing an infinite loop during `componentWillUnmount`

not sure this is the best fix, or maybe we should set the `<Suspense />`  shouldComponentUpdate to `false` until all the promise resolves?

### DIFF
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -109,8 +109,7 @@ Suspense.prototype._childDidSuspend = function(promise, suspendingVNode) {
 		if (resolved) return;
 
 		resolved = true;
-		suspendingComponent.componentWillUnmount =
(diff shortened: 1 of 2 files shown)


andrewiggins:
Does your change in #2919 also fix this?

tanhauhau:
haven't tried it yet, but from the looks of it, as the `<Suspense />` may be suspended by the same component multiple times, the `c._pendingSuspensionCount` may not be counting correctly

tanhauhau:
I'm closing this PR and cherry-picked the test cases into #2919

PR #2991, opened 2021-02-08
Update esbuild + karma-esbuild

marvinhagemeister:
The watch feature now uses esbuild's new built in watch capabilities instead of spawning a chokidar instance. This leads to less CPU usage. cc @developit


andrewiggins (approved):
🎉

PR #2995, opened 2021-02-09
Improve fragment unmounting

### LINKED ISSUE #2987: One element may be rendered in place of another in case of conditional rendering
### Reproduction

https://codesandbox.io/s/romantic-galileo-5lrjs?file=/src/index.js:73-1770

### Steps to reproduce

Run the CodeSandbox above.

In the code sandbox I have two elements that are rendered conditionally (the first div and the last div, they may be rendered or not rendered depending on `renderFirstElement` and `renderLastElement` state props), and some divs containing numbers between them (I called the model for them `childrenProps`).

The first render is correct: Preact renders the first div, then 20 divs with numbers from 5 to 24 and then the last div is rendered. The order is correct.

But then, because I change the state in `componentDidMount`, happens the second render. I set `renderFirstElement` to false (so that the first div wouldn't be rendered), and change `childrenProps` (I delete the last 5 elements and add 5 elements at the beginning). And something strange happens: the **last** div is rendered where the **first** div was!

### Expected Behavior

You may see how it should work in React CodeSandbox: https://codesandbox.io/s/divine-haze-pxgf6?file=/src/App.js.
The "last div" always stays at the bottom.

### Actual Behavior

When the "first div" is removed from the DOM, the "last div" takes its place.

andrewiggins:
This PR fixes a bug where we would use `getDomSibling` with `oldChildren` when the oldVNode tree was broken (since we set `oldChildren[i] = undefined` if we are re-using the vnodes). Ideally this kind of bug will become less relevant with "backing tree" since it should always have a complete tree.

Still need to investigate the impact of this change and investigate ways to make it smaller, as well as remove the comments (though I put them here for now so others can read them and see what I tried).

Fixes #2987


JoviDeCroock (approved):

JoviDeCroock:
Closing: fixes a v10 `getDomSibling`/`oldChildren` staleness bug; v11 rewrote this around `_index`/`_parent` and handles the stale-sibling case inline (see #5065).

PR #2998, opened 2021-02-09
Fix stale internal._dom pointers after mount

marvinhagemeister:


src/diff/patch.js
```diff
@@ -162,4 +162,6 @@ function patchDOMElement(
 	) {
 		setProperty(dom, 'checked', tmp, oldProps.checked, false);
 	}
+
+	return dom;
```
developit, line 166: QQ: couldn't we have this set `internal._dom = dom` instead of returning? It's only used in that one spot.

marvinhagemeister: Good point, updated the PR with the suggestion 👍

andrewiggins: Yup we could. I think at the point when this pattern was established the return + set was smaller. I'd say let's leave a golf TODO and revisit this once the branch stabilizes

andrewiggins: Oh nevermind, Marvin fixed it :) All good.

PR #3000, opened 2021-02-09
Fix another inversed bail out check

marvinhagemeister:


marvinhagemeister:
Nevermind, pushed an empty branch 😬

PR #3002, opened 2021-02-10
Fix some tests for restructure backing tree

andrewiggins:
- Renamed `vnode` to `internal` in hooks
- ~~Disabled debug tests cuz they seem to be infinite looping (whoops!)~~ Fixed by Marvin!
- Renamed `internal` to `parentInternal` in `diffChildren`
- Fixed some null placeholder tests by holding onto old DOM node for comparison later
- Moved call to `reorderChildren` to `renderComponent` in `shouldComponentUpdate` code path
- Fixed some suspense hydration by checking value of `internal._children` to determine if mounting or diffing should occur
- Moved copying of parent internal properties to `createInternal`
- Pass internal to all option hooks
- Copy latest VNode._original property after a successful diff
- Fix `getDomSibling` tests

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -103,11 +103,7 @@ export function diffChildren(
 		let oldVNodeRef;
 		let nextDomSibling;
 		if (childInternal == null) {
-			childInternal = createInternal(childVNode);
-			// TODO: Move this into createInternal
-			childInternal._parent = internal;
-			childInternal._depth = internal._depth + 1;
-			childInternal._mode = internal._mode;
+			childInternal = createInternal(childVNode, internal);
 
 			// We are mounting a new VNode
 			nextDomSibling = mount(
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -20,7 +20,7 @@ import { createInternal } from '../tree';
  * @param {import('../internal').ComponentChildren[]} renderResult
  * @param {import('../internal').VNode} newParentVNode The new virtual node
  * whose children should be diff'ed against oldParentVNode
- * @param {import('../internal').Internal} internal The Internal node
+ * @param {import('../internal').Internal} parentInternal The Internal node
  * whose children should be diff'ed against newParentVNode
  * @param {object} globalContext The current context object - modified by
  * getChildContext
@@ -34,7 +34,7 @@ export function diffChildren(
 	parentDom,
 	renderResult,
 	newParentVNode,
-	internal,
+	parentInternal,
 	globalContext,
 	isSvg,
 	commitQueue,
@@ -48,7 +48,7 @@ export function diffChildren(
 	/** @type {import('../internal').VNode | string} */
 	let childVNode;
 
-	let oldChildren = internal._children || EMPTY_ARR;
+	let oldChildren = parentInternal._children || EMPTY_ARR;
 	let oldChildrenLength = oldChildren.length;
 
 	const newChildren = [];
@@ -111,7 +111,7 @@ export function diffChildren(
 		let oldVNodeRef;
 		let nextDomSibling;
 		if (childInternal == null) {
-			childInternal = createInternal(childVNode, internal);
+			childInternal = createInternal(childVNode, parentInternal);
 
 			// We are mounting a new VNode
 			nextDomSibling = mount(
@@ -196,7 +196,7 @@ export function diffChildren(
 			//
 			// To fix it we make sure to reset the inferred value, so that our own
 			// value check in `diff()` won't be skipped.
-			if (internal.type === 'option') {
+			if (parentInternal.type === 'option') {
 				// @ts-ignore We have validated that the type of parentDOM is 'option'
 				// in the above check
 				parentDom.value = '';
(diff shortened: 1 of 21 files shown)


src/diff/children.js
```diff
@@ -142,6 +143,7 @@ export function diffChildren(
 		} else {
 			// this was previously captured after patch(), which seems backwards?
 			oldVNodeRef = childInternal.ref;
+			prevDom = childInternal._dom;
```
andrewiggins, line 146: Argh, really wondering if there is a better way to do this... It's required for a test or two in `placeholders.test.js` I believe

src/diff/children.js
```diff
@@ -175,17 +177,6 @@ export function diffChildren(
...
-				if (
-					childInternal != null &&
-					childVNode.props &&
-					childVNode.props.children != null &&
-					childVNode.props.children === childInternal.props.children
-				) {
-					startDom = reorderChildren(childInternal, startDom, parentDom);
-				} else {
-					startDom = nextDomSibling;
-				}
```
andrewiggins, line 188: This was an area where I don't really understand the implications of this change, but it just worked. At some point I'd like to dig into this more to understand what it was here originally to see if we are missing some edge cases or something....

src/diff/children.js
```diff
@@ -205,41 +196,41 @@ export function diffChildren(
...
 			startDom &&
 			childInternal != null &&
-			childInternal._dom == startDom &&
+			prevDom == startDom &&
 			startDom.parentNode != parentDom
 		) {
 			// The above condition is to handle null placeholders. See test in placeholder.test.js:
 			// `efficiently replace null placeholders in parent rerenders`
-			startDom = getDomSibling(childInternal);
+			startDom = nextDomSibling; // Note: change works in restructure. Not dependent on backing tree
```
andrewiggins, line 212: This change isn't related to backing nodes and works on `restructure` directly. Just including it here since it was an easy change to make

src/diff/component.js
```diff
@@ -163,7 +168,7 @@ export function renderComponent(
 
 	let nextDomSibling;
 
-	if (isNew) {
+	if (internal._children == null) {
```
andrewiggins, line 171: This change is necessary for suspense hydration. Since the suspending node will have a valid internal and component instance after the first render (the one that suspends), `isNew` will be `false` on subsequent renders. However, particularly for suspense hydration we want the children to be mounted/hydrated even though the component isn't new. So using this condition to detect that scenario which works for both `isNew == false` and suspense hydration.

src/diff/patch.js
```diff
@@ -36,8 +36,7 @@ export function patch(
 	// constructor as undefined. This to prevent JSON-injection.
 	if (newVNode.constructor !== undefined) return null;
 
-	// @TODO newVNode is pretty useless here, it contains no tree information
-	if (options._diff) options._diff(newVNode);
+	if (options._diff) options._diff(internal);
```
andrewiggins, line 39: Hooks need internals

src/diff/patch.js
```diff
@@ -70,11 +69,16 @@ export function patch(
 			nextDomSibling = internal._dom.nextSibling;
 		}
 
-		if (options.diffed) options.diffed(newVNode);
+		if (options.diffed) options.diffed(internal);
+
+		// We successfully rendered this VNode, unset any stored hydration/bailout state:
+		internal._mode = MODE_NONE;
+		// Once we have successfully rendered the new VNode, copy it's ID over
+		internal._original = newVNode._original;
```
andrewiggins, line 77: Once a newVNode has successfully rendered, we should copy it's ID over in case it is memoized and that VNode id is used again.

src/index.d.ts
```diff
@@ -259,9 +259,9 @@ declare namespace preact {
 		/** Attach a hook that is invoked whenever a VNode is created. */
 		vnode?(vnode: VNode): void;
 		/** Attach a hook that is invoked immediately before a vnode is unmounted. */
-		unmount?(vnode: VNode): void;
+		unmount?(vnode: VNode): void; // TODO: What is the correct type for this parameter?
```
andrewiggins, line 262: These are public interfaces and so we should decide what interface to expose here at some point...

andrewiggins (commented):
Leaving some inline explanations

src/diff/mount.js
```diff
@@ -240,16 +240,13 @@ export function mountChildren(
 		// Terser removes the `continue` here and wraps the loop body
 		// in a `if (childVNode) { ... } condition
 		if (childVNode == null) {
+			parentInternal._children[i] = null;
```
marvinhagemeister, line 243: +1 Noticed that too that we were leaving a hole here when stepping through our diff.

marvinhagemeister:
"Fix some tests" -> Fixes nearly all the tests 😅

src/diff/component.js
marvinhagemeister: I'm most excited about changing that component instance thing in `Suspense`. Should make things a lot easier 👍

src/diff/children.js
marvinhagemeister: I'm good with the change if all the tests pass. If this piece of code was necessary we should be able to run into a case and turn that into a test. For now I'm good with removing it 👍

marvinhagemeister (approved):

src/diff/children.js
andrewiggins: Added an inline TODO and a follow up to start at this code a bit more to find a more elegant way to do this.

PR #3017, opened 2021-02-15
Convert c._force to a bitfield on internal._mode

marvinhagemeister:
This allows potentially any node to "forcefully" update and gets us one step closer to extracting class-components from core 🎉

### DIFF
--- a/src/diff/component.js
+++ b/src/diff/component.js
@@ -196,8 +196,6 @@ export function renderComponent(
 		commitQueue.push(c);
 	}
 
-	internal._mode &= ~FORCE_UPDATE;
-
 	return nextDomSibling;
 }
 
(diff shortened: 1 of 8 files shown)


src/diff/component.js
```diff
@@ -195,7 +196,7 @@ export function renderComponent(
 		commitQueue.push(c);
 	}
 
-	c._force = false;
+	internal._mode &= ~FORCE_UPDATE;
```
andrewiggins, line 199: Ooo, could we add `FORCE_UPDATE` to the `RESET_MODE` constant and then just reset at the end of `patch` and `mount` when we reset other flags? Could save us a byte or two to not have this additional line of code

andrewiggins (approved):
This is great! Excited by this move to bit flags :D

src/diff/component.js
marvinhagemeister: Neat, you're right that's better 👍

PR #3018, opened 2021-02-15
Convert c._dirty to a bitfield on internal._mode

marvinhagemeister:
Based on #3017 . Marked as draft until that is merged.

### DIFF
--- a/src/diff/component.js
+++ b/src/diff/component.js
@@ -4,7 +4,7 @@ import { assign } from '../util';
 import { Component } from '../component';
 import { mountChildren } from './mount';
 import { diffChildren, reorderChildren } from './children';
-import { FORCE_UPDATE } from '../constants';
+import { DIRTY, FORCE_UPDATE } from '../constants';
 
 /**
  * Diff two virtual nodes and apply proper changes to the DOM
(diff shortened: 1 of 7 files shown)


src/diff/component.js
```diff
@@ -145,7 +147,7 @@ export function renderComponent(
 
 	if ((tmp = options._render)) tmp(internal);
 
-	c._dirty = false;
+	internal._mode &= ~DIRTY;
```
andrewiggins, line 150: Hmm I wonder if we could add `DIRTY` to `RESET_MODE`... technically it is a change in the existing behavior since we clear dirty before diffing children. Moving it to `RESET_MODE` would change it so we only clear dirty after diffing children. Perhaps for this PR we shouldn't change its behavior but perhaps it's something to consider for the future

andrewiggins (approved):
...dirty bit: https://youtu.be/D7K3wFXJFsQ 💃🕺🎊

src/diff/component.js
marvinhagemeister: Tried that, but doing so breaks 42 tests :S

andrewiggins: lolol yea okay good call - let's not do that then

PR #3019, opened 2021-02-15
Add custom branch for function components

marvinhagemeister:
This is a step towards making class and function components separate things, so that we can extract class components out of core. It basically copies the path for class components and removes lifecycles not needed in our other components like `createContext`

Depends on #3017 + #3018


JoviDeCroock:
Closing: targets the abandoned `restructure` branch. The internals-split rewrite it built on was not pursued; v11 shipped a different architecture from `main`.

PR #3031, opened 2021-02-19
Combine _mode & _flags fields

andrewiggins:
Since type and mode flags are shared now, I prefaced all "type" related constants with `TYPE_` and all "mode" related constants with `MODE_`.

Let's bikeshed the names in the comments.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,18 +1,27 @@
-// Internal._mode bitfield constants
-/** Normal mount/patch. Not hydration or side-effecting top-level render. */
-export const MODE_NONE = 0;
+// Internal._flags bitfield constants
+export const TYPE_TEXT = 1 << 0;
+export const TYPE_ELEMENT = 1 << 1;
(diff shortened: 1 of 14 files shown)


src/constants.js
```diff
@@ -1,18 +1,27 @@
...
 /** Signifies this VNode errored on the previous render */
-export const MODE_ERRORED = 8;
+export const MODE_ERRORED = 1 << 8;
+
 /** Signifies that bailout checks will be bypassed */
-export const FORCE_UPDATE = 16;
+export const FORCE_UPDATE = 1 << 9;
 /** Signifies that a node needs to be updated */
-export const DIRTY = 32;
+export const DIRTY_BIT = 1 << 10;
```
andrewiggins, line 24: Are these modes? I left them as is for now, but open to changing them at other's suggestion.

marvinhagemeister (approved):
🙌

PR #3033, opened 2021-02-20
Consistently use flags to determine Internal type

andrewiggins:
Also, skip array children array allocations when using `dangerouslySetInnerHTML`, as well as simplify the `props.children` check in `mountDOMElement`. 

Depends on #3032

### DIFF
--- a/src/diff/mount.js
+++ b/src/diff/mount.js
@@ -185,8 +185,8 @@ function mountDOMElement(dom, internal, globalContext, isSvg, commitQueue) {
 			if (!isHydrating && newHtml.__html) {
 				dom.innerHTML = newHtml.__html;
 			}
-			internal._children = [];
-		} else if ((i = internal.props.children) != internal._children) {
+			internal._children = null;
+		} else if ((i = internal.props.children) != null) {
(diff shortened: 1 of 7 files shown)


src/diff/mount.js
```diff
@@ -185,8 +185,8 @@ function mountDOMElement(dom, internal, globalContext, isSvg, commitQueue) {
 			if (!isHydrating && newHtml.__html) {
 				dom.innerHTML = newHtml.__html;
 			}
-			internal._children = [];
-		} else if ((i = internal.props.children) != internal._children) {
+			internal._children = null;
+		} else if ((i = internal.props.children) != null) {
```
andrewiggins, line 189: When mounting, `internal._children` is always `null` at this point since `mountChildren` hasn't been called on this internal yet. So this condition was only true when `props.children == null`. So I simplified this check to just use `null` directly.

Also, there isn't a reason to have this check in `patch` cuz `props.children` is always an array VNodes and `internal._children` is always an array of Internals so they will never be equal.

marvinhagemeister (approved):

PR #3036, opened 2021-02-21
Replace _parentDom on Component with a lookup function

andrewiggins:
Instead of keeping a `_parentDom` pointer on every Component instance, this PR adds a new helper, `getParentDom`, that climbs up the tree to find the first DOM Internal or root Internal (an internal with a `_parentDom` prop, see #2983) and returns its DOM pointer.

Since the `component._parentDom` field was used to check if a component was still mounted, I added a `MODE_UNMOUNTING` flag to indicate that an internal has been unmounted. This flag is used in a couple of places where we don't want to rerender or run effects of components that have just been unmounted.

This PR also replaces the `TYPE_FRAGMENT` flag (which wasn't used anywhere) with a `TYPE_ROOT` flag. If a function component VNode contains a `_parentDom` prop when it's Internal is being created, it'll get the `TYPE_ROOT` flag. This flag signals that this internal has a `_parentDom` prop that should change the parent DOM node of its children. I wanted a flag for root nodes so that we didn't have to do a megamorphic `props._parentDom` lookup on every component (wondering if that is the slowdowns in text_update and many_updates in #2983...).

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -12,7 +12,7 @@ import {
 	getDisplayName
 } from './component-stack';
 import { assign } from './util';
-import { MODE_UNMOUNTED } from 'preact/src/constants';
+import { MODE_UNMOUNTING } from 'preact/src/constants';
 
 const isWeakMapSupported = typeof WeakMap == 'function';
 
@@ -374,7 +374,7 @@ Component.prototype.setState = function(update, callback) {
 					)}`
 			);
 		}
-	} else if (this._internal._flags & MODE_UNMOUNTED) {
+	} else if (this._internal._flags & MODE_UNMOUNTING) {
 		console.warn(
 			`Can't call "this.setState" on an unmounted component. This is a no-op, ` +
 				`but it indicates a memory leak in your application. To fix, cancel all ` +
@@ -397,7 +397,7 @@ Component.prototype.forceUpdate = function(callback) {
 					getCurrentInternal()
 				)}`
 		);
-	} else if (this._internal._flags & MODE_UNMOUNTED) {
+	} else if (this._internal._flags & MODE_UNMOUNTING) {
 		console.warn(
 			`Can't call "this.forceUpdate" on an unmounted component. This is a no-op, ` +
 				`but it indicates a memory leak in your application. To fix, cancel all ` +
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -12,6 +12,7 @@ import {
 	getDisplayName
 } from './component-stack';
 import { assign } from './util';
+import { MODE_UNMOUNTED } from 'preact/src/constants';
 
 const isWeakMapSupported = typeof WeakMap == 'function';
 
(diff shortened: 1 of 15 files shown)


src/component.js
```diff
@@ -83,52 +84,17 @@ Component.prototype.forceUpdate = function(callback) {
  */
 Component.prototype.render = Fragment;
 
-/**
- * @param {import('./internal').Internal} internal
- * @param {number | null} [childIndex]
- * @returns {import('./internal').PreactNode}
- */
-export function getDomSibling(internal, childIndex) {
```
andrewiggins, line 91: I just moved `getDomSibling` and `updateDomPointers` to `tree.js`

src/component.js
```diff
@@ -228,9 +176,7 @@ function process() {
 		rerenderQueue = [];
 		// Don't update `renderCount` yet. Keep its value non-zero to prevent unnecessary
 		// process() calls from getting scheduled while `queue` is still being consumed.
-		queue.some(c => {
-			if (c._internal._flags & DIRTY_BIT) rerenderComponent(c);
-		});
+		queue.some(rerenderComponent);
```
andrewiggins, line 179: I moved the dirty check into `rerenderComponent` and combined it with the unmounting check

src/internal.d.ts
```diff
@@ -115,13 +113,6 @@ export interface VNode<P = {}> extends preact.VNode<P> {
 	_vnodeId: number;
 }
 
-export type InternalTypeFlags =
-	| CONSTANTS['TYPE_TEXT']
-	| CONSTANTS['TYPE_ELEMENT']
-	| CONSTANTS['TYPE_CLASS']
-	| CONSTANTS['TYPE_FUNCTION']
-	| CONSTANTS['TYPE_COMPONENT'];
-
```
andrewiggins, line 124: I think the change of the TYPE_* flags from number constants (e.g. `4`) to bit shifts (`1 << 2`) made these types just `number` instead of constants :/

I still prefer the `1 << 2` syntax though cuz it is easier to see how many flags we have and what the next number is if we need to add or modify a flag. So I went ahead and removed these types since it doesn't provide us any extra typing info or guarantees.

debug/src/debug.js
```diff
@@ -371,7 +374,7 @@ Component.prototype.setState = function(update, callback) {
 					)}`
 			);
 		}
-	} else if (this._parentDom == null) {
+	} else if (this._internal._flags & MODE_UNMOUNTING) {
```
marvinhagemeister, line 377: Huge fan of using flags here instead of checking the `_parentDom` pointer 🙌

src/tree.js
```diff
@@ -53,11 +53,11 @@ export function createInternal(vnode, parentInternal) {
 
 		// flags = typeof type === 'function' ? COMPONENT_NODE : ELEMENT_NODE;
 		flags =
-			type === Fragment
-				? TYPE_FRAGMENT
-				: typeof type === 'function'
+			typeof type === 'function'
 				? type.prototype && 'render' in type.prototype
 					? TYPE_CLASS
+					: props._parentDom
```
marvinhagemeister, line 59: Makes me wonder if we should add a dedicated root function in the future instead of using a Fragment. Something to try out later

marvinhagemeister (approved):
Really excited about having less pointers to take care of. This is fantastic 🙌

PR #3041, opened 2021-02-23
Properly handle deeply nested repeat errors

andrewiggins:
Fixes a bug where re-throwing an error deeply nested from the error boundary wouldn't properly bubble up.

I had to bring back the dancing PENDING_ERROR & RERENDERING_ERROR dual flags from master to get all tests to pass. I added some comments explaining the two flags and the purpose they serve. Let me know if they makes sense.

### DIFF
--- a/src/diff/catch-error.js
+++ b/src/diff/catch-error.js
@@ -1,4 +1,9 @@
-import { DIRTY_BIT, MODE_ERRORED } from '../constants';
+import {
+	DIRTY_BIT,
+	MODE_RERENDERING_ERROR,
+	MODE_PENDING_ERROR,
+	TYPE_COMPONENT
+} from '../constants';
 
 /**
  * Find the closest error boundary to a thrown error and call it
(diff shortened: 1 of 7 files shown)


src/diff/catch-error.js
```diff
@@ -8,26 +13,27 @@ import { DIRTY_BIT, MODE_ERRORED } from '../constants';
  * is the highest parent that was being unmounted)
  */
 export function _catchError(error, internal) {
-	/** @type {import('../internal').Component} */
-	let component;
```
andrewiggins, line 12: I dropped this variable and just inlined the `internal._component` access. Thought it would simplify this code a bit to help it make more sense. We can golf it later (though removing this variable also reduced bytes lol).

I converted the for loop to a while loop for the same reason. Terser will convert this to a for loop for us. No byte change between the two, fyi.

marvinhagemeister (approved):

PR #3043, opened 2021-02-25
fix: Adding missing SVGFEFunc types

### LINKED ISSUE #3040: Missing JSX.IntrinsicElements typings
It would be helpful to add type definitions for `feFuncA`, `feFuncR`, `feFuncG` and `feFuncB` elements.

### Reproduction

I can't get type checking working in properly codesandbox, but here's what it looks like in VS Code:

image

Looking at [`jsx.d.ts`](https://github.com/preactjs/preact/blob/master/src/jsx.d.ts#L920), I'd expect to find them around L920.

Compare to the [react types file](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v16/index.d.ts#L2799)

### Expected Behaviour

Use these elements without getting a type error.

### Actual Behaviour

Type error: `Property 'feFuncR' does not exist on type 'JSX.IntrinsicElements'.ts(2339)`

rschristian:
Closes #3040

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -918,6 +918,10 @@ export namespace JSXInternal {
 		feDisplacementMap: SVGAttributes<SVGFEDisplacementMapElement>;
 		feDropShadow: SVGAttributes<SVGFEDropShadowElement>;
 		feFlood: SVGAttributes<SVGFEFloodElement>;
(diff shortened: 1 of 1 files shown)


andrewiggins (approved):
Thanks!!

PR #3053, opened 2021-03-08
Use portals to shift suspended DOM out of UI tree

andrewiggins:
## Terminology

<dl>
<dt>root node</dt>
<dd>A Fragment-like node with a special `_parentDOM` prop (mangled in build output) that signals that all DOM children under this node should be mounted in the specified `_parentDOM` node. See #2983</dd>

<dt>to suspend</dt>
<dd>To throw a promise while rendering. The component that does this is called a "suspending component" (see below)</dd>

<dt>Suspense</dt>
<dd>The name of the component that "catches"' or "handles" a thrown promise. A Suspense component does not "suspend" - it does not throw a promise while rendering. It handles components that do suspend. It may be in a "suspended" state meaning it has handled a thrown promise and is awaiting the promise the settle before rerendering</dd>

<dt>suspending component</dt>
<dd>The component that threw a promise when attempting to render. It is a child of a Suspense component</dd>

<dt>suspending DOM</dt>
<dd>The DOM nodes under a Suspense component that will visually be replaced with the Suspense's fallback component when a child of Suspense suspends</dd>
</dl>

## New root node behavior

Instead of manually moving the suspended DOM from the UI tree in Suspense, this PR extends our new "root node" concept in core to support Suspense's use case. Specifically, if a root node's `_parentDOM` property matches the parentDOM of the siblings it is rendering in, then the diff should behave as if this root node was just a Fragment. This logic also handles if a root node transitions from behaving like a Fragment to behaving like a Portal, and will adjust the `nextDomSibling` return value around the DOM nodes that were moved into a different parent tree. Finally, it supports transitioning a root node from behaving like a Portal to behaving like a Fragment, moving all of it's children DOM into it's new parent tree and adjust the `nextDomSibling` return value accordingly.

## How Suspense uses root nodes

Suspense uses the above behavior to move the suspended children into a disconnected DOM parent when rendering the fallback after a VNode suspends. On initial render, the Portal under Suspense will have the same parentDOM as its siblings, so it'll behave like a Fragment. When rendering the fallback, Suspense will re-render and change the Portal's parentDOM to a disconnected DOM node. The suspended children get re-rendered and re-parented to the new disconnected DOM node.

In order to properly re-parent suspended children, we need to ensure that any DOM nodes previously rendered by the suspending component are also moved into the new inert parent DOM. We ensure this by calling `reorderChildren` when a component throws a promise.

More comments and explanations are inline in the code comments and in my PR review.

## Future PRs

* Explore moving some of Suspense's instance properties into component state. `state._suspended` could be replaced with `state._pendingSuspensionCount !== 0`.
* Explore eagerly rerendering components as their promises resolve, removing the need for the `this._suspenders` array
* Could we rewrite Suspense using hooks instead of class state?

### DIFF
--- a/compat/src/portals.js
+++ b/compat/src/portals.js
@@ -17,5 +17,16 @@ function Portal(props) {
  * @param {import('./internal').PreactElement} container The DOM node to continue rendering in to.
  */
 export function createPortal(vnode, container) {
+	// Note: We can't use Fragment here because a component that returned a Portal
+	// (e.g. `const App = () => createPortal(...)`) wouldn't work. Our diff
+	// collapses Fragments without keys that are returned directly from components
+	// into just an array and sets that as the children array of the component.
+	//
+	// We also can't use keyed Fragments here cuz it might lead to weird edge
+	// cases when toggling between two sibling portals if we use a shared keyed or
+	// lead to unnecessary re-mounts if trying to generate a new key on each call.
+	//
+	// So the simplest solution seems to be just to use an unique type for Portal
+	// to skip the Fragment collapsing logic when diffing components
 	return createElement(Portal, { _parentDom: container }, vnode);
 }
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -159,17 +159,17 @@ Suspense.prototype._childDidSuspend = function(promise, suspendingInternal) {
 
 	const onSuspensionComplete = () => {
 		if (!--c._pendingSuspensionCount) {
-			/** @type {import('./internal').Internal} */
-			let suspendedInternal;
-			if (c.state._suspended) {
-				suspendedInternal = c.state._suspended;
-			} else {
-				// _suspended will be unset if we are suspending while hydrating. If so
-				// the suspendedInternal is still the first child of Suspense since we
-				// never rendered the fallback
-				suspendedInternal = c._internal._children[0];
-			}
-
+			// /** @type {import('./internal').Internal} */
+			// let suspendedInternal;
+			// if (c.state._suspended) {
+			// 	suspendedInternal = c.state._suspended;
+			// } else {
+			// 	// _suspended will be unset if we are suspending while hydrating. If so
+			// 	// the suspendedInternal is still the first child of Suspense since we
+			// 	// never rendered the fallback
+			// 	suspendedInternal = c._internal._children[0];
+			// }
+			//
 			// c._internal._children[0] = removeOriginal(
 			// 	suspendedInternal,
 			// 	suspendedInternal._component._parentDom,
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -129,6 +129,9 @@ Suspense.prototype = new Component();
  * @param {import('./internal').Internal} suspendingInternal The suspending component
  */
 Suspense.prototype._childDidSuspend = function(promise, suspendingInternal) {
+	// @TODO Investigate what this logic does when a component throws while
+	// rendering the fallback...
+
 	const suspendingComponent = suspendingInternal._component;
 
 	/** @type {import('./internal').SuspenseComponent} */
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -1,5 +1,6 @@
 import { Component, createElement, options, Fragment } from 'preact';
 import { TYPE_ELEMENT, FORCE_UPDATE, MODE_HYDRATE } from '../../src/constants';
+import { getParentDom } from '../../src/tree';
 import { assign } from './util';
 
 const oldCatchError = options._catchError;
@@ -114,6 +115,7 @@ export function Suspense() {
 	this._pendingSuspensionCount = 0;
 	this._suspenders = null;
 	this._detachOnNextRender = null;
+	this._parentDom = null;
 }
 
 // Things we do here to save some bytes but are not proper JS inheritance:
@@ -168,11 +170,11 @@ Suspense.prototype._childDidSuspend = function(promise, suspendingInternal) {
 				suspendedInternal = c._internal._children[0];
 			}
 
-			c._internal._children[0] = removeOriginal(
-				suspendedInternal,
-				suspendedInternal._component._parentDom,
-				suspendedInternal._component._originalParentDom
-			);
+			// c._internal._children[0] = removeOriginal(
+			// 	suspendedInternal,
+			// 	suspendedInternal._component._parentDom,
+			// 	suspendedInternal._component._originalParentDom
+			// );
 
 			c.setState({ _suspended: (c._detachOnNextRender = null) });
 
@@ -186,6 +188,8 @@ Suspense.prototype._childDidSuspend = function(promise, suspendingInternal) {
 					suspended.forceUpdate();
 				}
 			}
+
+			this._parentDom = null;
 		}
 	};
 
@@ -198,15 +202,18 @@ Suspense.prototype._childDidSuspend = function(promise, suspendingInternal) {
 		(suspendingInternal._flags & MODE_HYDRATE) === MODE_HYDRATE;
 
 	if (!c._pendingSuspensionCount++ && !wasHydrating) {
+		this._parentDom = document.createElement('div');
 		c.setState({
 			_suspended: (c._detachOnNextRender = c._internal._children[0])
 		});
 	}
+
 	promise.then(onResolved, onResolved);
 };
 
 Suspense.prototype.componentWillUnmount = function() {
 	this._suspenders = [];
+	this._parentDom = null;
 };
 
 /**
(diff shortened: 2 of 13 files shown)


compat/src/portals.js
```diff
@@ -17,5 +17,16 @@ function Portal(props) {
...
+	// collapses Fragments without keys that are returned directly from components
+	// into just an array and sets that as the children array of the component.
+	//
+	// We also can't use keyed Fragments here cuz it might lead to weird edge
+	// cases when toggling between two sibling portals if we use a shared keyed or
+	// lead to unnecessary re-mounts if trying to generate a new key on each call.
+	//
+	// So the simplest solution seems to be just to use an unique type for Portal
+	// to skip the Fragment collapsing logic when diffing components
 	return createElement(Portal, { _parentDom: container }, vnode);
```
andrewiggins, line 31: Given how simple the Portal implementation is, I wonder if we should move it core... Though we'd have to export it which would be more bytes we can't crawl back...

compat/src/suspense.js
```diff
@@ -41,79 +42,14 @@ options.unmount = function(internal) {
...
- */
-function detachedClone(internal, detachedParent, parentDom) {
-	if (internal) {
-		if (internal._component && internal._component.__hooks) {
-			internal._component.__hooks._list.forEach(effect => {
-				if (typeof effect._cleanup == 'function') effect._cleanup();
-			});
-
-			internal._component.__hooks = null;
-		}
```
andrewiggins, line 58: Hmm I'm not sure why we are clearing all hook state here. I don't think it makes sense for a `<Counter />`  component to have it's state cleared when a sibling suspends. And we only clear hook state and not class component state which seems inconsistent.

For context here is the related issue (#2795) and PR (#2796). Let me know if I'm missing a use case or something! :)

FWIW, React does not clear hook state or run effect cleanups when suspending (I copied our relevant tests into these code sandboxes:
* https://codesandbox.io/s/does-react-reset-hooks-when-suspending-mm1xn
* https://codesandbox.io/s/does-react-clean-effects-when-suspending-7pmri

compat/src/suspense.js
```diff
@@ -41,79 +42,14 @@ options.unmount = function(internal) {
...
-
-	return internal;
-}
-
-/**
- * @param {import('./internal').Internal} internal
- * @param {import('./internal').PreactElement} detachedParent
- * @param {import('./internal').PreactElement} originalParent
- */
-function removeOriginal(internal, detachedParent, originalParent) {
```
andrewiggins, line 83: We no longer reset the `_vnodeId` property or add the `FORCE_UPDATE` flag to all suspended children so any vnode equality or `sCU` lifecycles under the Suspense node are still respected. We do still call `.forceUpdate` on the component that suspended (i.e. threw the promise) to ensure it gets rendered, but any intermediate children between it and Suspense won't necessarily be re-rendered in this new implementation.

I think we gone back and forth on this. If anyone knows of any use cases where we must re-render intermediate children, let me know!

compat/src/suspense.js
```diff
@@ -128,15 +64,20 @@ Suspense.prototype = new Component();
  */
 Suspense.prototype._childDidSuspend = function(promise, suspendingInternal) {
 	const suspendingComponent = suspendingInternal._component;
+	if (suspendingComponent._onResolve != null) {
+		// This component has already been handled by a Suspense component. Do
+		// nothing
+		return;
+	}
```
andrewiggins, line 71: Since we re-render the suspending component component again when rendering the fallback, it'll throw again. This check ensures that additional throw doesn't duplicate the resolution logic. We already have logic to prevent re-entry (line 85: `if (resolved) return;`) but I like this additional check to prevent the number of promise callbacks from exploding.

compat/test/browser/suspense.test.js
```diff
@@ -586,7 +509,52 @@ describe('suspense', () => {
 		});
 	});
 
-	it('should allow children to update state while suspending', () => {
+	it('should keep hook state of siblings when suspending', () => {
```
andrewiggins, line 512: This test (and the other added test below it) are here just to ensure that hook state and class state have the same behavior in various Suspense scenarios, which wasn't true before. These tests fail on master.

compat/test/browser/suspense.test.js
```diff
@@ -772,8 +788,8 @@ describe('suspense', () => {
 				expect(scratch.innerHTML).to.eql(
 					`<div>Hello first 2</div><div>Hello second 2</div>`
 				);
-				expect(Suspender1.prototype.render).to.have.been.calledThrice;
-				expect(Suspender2.prototype.render).to.have.been.calledThrice;
+				expect(Suspender1.prototype.render).to.have.been.calledTwice;
+				expect(Suspender2.prototype.render).to.have.been.calledTwice;
```
andrewiggins, line 792: This behavior change is an example of my previous comment where we no longer rerender intermediate components. Here, Suspender doesn't throw. It renders a lazy child which throws. So Suspender's child is the component that re-renders an additional time here. Will try to fix this so we can assert the right thing here.

debug/test/browser/debug-suspense.test.js
```diff
@@ -106,8 +106,8 @@ describe('debug with suspense', () => {
 
 				return loader.then(() => {
 					rerender();
-					expect(console.warn).to.be.calledTwice;
-					expect(warnings[1].includes('MyLazyLoaded')).to.equal(true);
+					expect(console.warn).to.be.calledThrice;
+					expect(warnings[2].includes('MyLazyLoaded')).to.equal(true);
```
andrewiggins, line 110: Since lazy is re-rendered when mounting the fallback, these checks get triggered an additional time.

src/diff/component.js
```diff
@@ -190,8 +189,17 @@ export function renderComponent(
...
+		if (c.getSnapshotBeforeUpdate != null) {
+			snapshot = c.getSnapshotBeforeUpdate(oldProps, oldState);
+		}
+
+		// Only schedule componentDidUpdate if the component successfully rendered
+		if (c.componentDidUpdate != null) {
+			c._renderCallbacks.push(() => {
+				c.componentDidUpdate(oldProps, oldState, snapshot);
+			});
+		}
```
andrewiggins, line 202: This change is to prevent queueing up extra calls to `cDU` for each aborted render of a suspending component. Since a suspending component is guaranteed to be rendered at least 3 times (1. initial render, 2. re-render of fallback, 3. successfuly re-render when promise resolves), I figured we should not call cDU so many times unnecessarily.

src/diff/patch.js
```diff
@@ -82,6 +82,12 @@ export function patch(
 		// Once we have successfully rendered the new VNode, copy it's ID over
 		internal._vnodeId = newVNode._vnodeId;
 	} catch (e) {
+		if (e.then) {
+			// If a promise was thrown, reorderChildren in case this component is
+			// being hidden or moved
+			reorderChildren(internal, startDom, parentDom);
+		}
+
```
andrewiggins, line 90: This is the code that ensures a suspending component's DOM children are re-parented to the new disconnected DOM parent when mounting the fallback

compat/src/suspense.js
```diff
@@ -215,35 +141,15 @@ Suspense.prototype.componentWillUnmount = function() {
...
-				this._detachOnNextRender,
-				detachedParent,
-				(detachedComponent._originalParentDom = detachedComponent._parentDom)
-			);
-		}
-
-		this._detachOnNextRender = null;
+	if (this._parentDom == null) {
+		this._parentDom = getParentDom(this._internal);
 	}
```
andrewiggins, line 146: `this._parentDom` is null on renders that aren't suspended. To ensure that Suspense behaves like a Fragment, we need to set the Portal's parentDOM prop to match Suspense's parentDOM. So we manually search for it here.

andrewiggins: Note, this is necessary because root nodes must have a non-null `_parentDOM` prop on initial render so they get the `TYPE_ROOT` flag. If instead we were to use another signal `e.g. vnode.type == Portal` or something then we could say "for root nodes with a nullish _parentDOM, default to the current parentDOM" in diff/components.js

src/diff/mount.js
```diff
@@ -126,6 +123,8 @@ function mountDOMElement(dom, internal, globalContext, isSvg, commitQueue) {
 		} else if (dom.data !== newProps) {
 			dom.data = newProps;
 		}
+
+		internal._dom = dom;
```
andrewiggins, line 127: When a Suspense component initially renders, it should behave like a Fragment and not move its children. To do this, it needs to pass its current parentDOM to the `createPortal` function. To get its current parentDOM, we need to ensure that its parent Internals have their `_dom` pointer set before recursing down the tree. These `internal._dom = dom;` lines do that.

src/diff/mount.js
```diff
@@ -199,6 +200,9 @@ function mountDOMElement(dom, internal, globalContext, isSvg, commitQueue) {
 			);
 		}
 
+		// To override mountChildren setting _dom
+		internal._dom = dom;
+
```
andrewiggins, line 205: Why do we set `internal._dom = dom` so many times in this function you ask? Well currently `diffChildren` sets `parentInternal._dom = firstDomChild` after all children are diffed so this line overrides that change. Once we remove the `_dom` property on Components, all that extra dom tracking (including the `parentInternal._dom = firstDomChild` line) will go away.

andrewiggins: Actually, as I type this I'm wondering if instead we should just wrap the `parentInternal._dom = firstChildDom` in an if statement that only applies it for TYPE_COMPONENT... I may do that instead.

andrewiggins: Done

compat/src/suspense.js
andrewiggins: FYI, this is PR #2661 added the `removeOriginal` function, for additional context

marvinhagemeister (approved):
Love these changes a lot! I always had trouble understanding our Suspense implementation before and this PR makes the code much easier to follow🙌

andrewiggins:
FYI, I found an edge case that this breaks in ([the edge case I mention in the TODO comments](https://github.com/preactjs/preact/pull/3053/files#diff-5cf05a25c60d8fcfa19e5dcac743666edfcd9d402384438406b7244b42de9e2aR267)). I'm investigating at the moment

compat/test/browser/suspense.test.js
andrewiggins: Rewrote this test to directly spy Lazy which actually throws.

marvinhagemeister:
Oh no I didn't saw your comment in the mobile view 🙈 Let's PR any fixes directly to restructure

andrewiggins:
@JoviDeCroock & @developit Would still appreciate any feedback or thoughts you have on this PR :)

debug/test/browser/debug-suspense.test.js
developit: This one seems curious to me: we invoke the component that threw in order to un-render it?

PR #3054, opened 2021-03-10
Add decoding TypeScript attribute

sumanthratna:
currently, when I try using: 
```jsx
<img src="..." decoding="async" />
```

I get this error:
```
TS2322: Type '{ alt: string; src: string; style: { width: string; height: string; }; className: string; decoding: string; }' is not assignable to type 'HTMLAttributes<HTMLImageElement>'.
  Property 'decoding' does not exist on type 'HTMLAttributes<HTMLImageElement>'.
```

this patch fixes this issue

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -656,6 +656,7 @@ export namespace JSXInternal {
 		disabled?: boolean;
 		disableRemotePlayback?: boolean;
 		download?: any;
+		decoding?: 'sync' | 'async' | 'auto';
 		draggable?: boolean;
 		encType?: string;
 		form?: string;
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -811,6 +811,7 @@ export namespace JSXInternal {
 		dd: HTMLAttributes<HTMLElement>;
 		del: HTMLAttributes<HTMLModElement>;
 		details: HTMLAttributes<HTMLDetailsElement>;
+		decoding: HTMLAttributes<HTMLImageElement>;
 		dfn: HTMLAttributes<HTMLElement>;
 		dialog: HTMLAttributes<HTMLDialogElement>;
 		div: HTMLAttributes<HTMLDivElement>;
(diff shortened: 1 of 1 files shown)


rschristian (changes requested):
If you look at where you added that type, those are all HTML elements. What you did was make `<decoding />` a valid HTML element, and it's the type of a `HTMLImageElement`.

What you'll actually want to do is create a `decoding` entry [under the `HTMLAttributes`](https://github.com/preactjs/preact/blob/3cb470ee83919a5f6f1b6897980ec524e7d07aa9/src/jsx.d.ts#L614) and set it so it matches [the attribute's options](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-decoding).

Like this:

```
decoding?: 'sync' | 'async' | 'auto';
```

sumanthratna:
> If you look at where you added that type, those are all HTML elements. What you did was make `<decoding />` a valid HTML element, and it's the type of a `HTMLImageElement`.
> 
> What you'll actually want to do is create a `decoding` entry [under the `HTMLAttributes`](https://github.com/preactjs/preact/blob/3cb470ee83919a5f6f1b6897980ec524e7d07aa9/src/jsx.d.ts#L614) and set it so it matches [the attribute's options](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-decoding).
> 
> Like this:
> 
> ```
> decoding?: 'sync' | 'async' | 'auto';
> ```

whoops, too many types to keep track of :) should be fixed now @rschristian. thanks for taking a look at this PR!

rschristian (approved):

rschristian:
Thanks!

sumanthratna:
@rschristian I know this _just_ got merged but do you have an ETA on the next release?

rschristian:
I do not, but what you can do is manually add the types until then: 

declaration.d.ts
```
interface HotPatch {
	decoding?: 'sync' | 'async' | 'auto';
}

declare module 'preact' {
	namespace JSX {
		interface IntrinsicAttributes extends HotPatch {}
	}
	interface Attributes extends HotPatch {}
}
```

Should silence any errors you'd get

PR #3056, opened 2021-03-10
Fix the useRef typing to include undefined when called without initial value

cmlenz:
Fix the `PropRef` typing in hooks to include the `undefined` type for the `current` property, as it will be undefined (at least) on initial render.

This was changed in https://github.com/preactjs/preact/pull/2803 when the `current` property was made non-optional, but it still needs to include the `undefined` type. The `current` property will always be `undefined` on initial render, for example:

```tsx
import {useRef} from 'preact/hooks';
​
function Test() {
  const ref = useRef<HTMLButtonElement>()
  console.log(ref.current.textContent) // oops
  return <button ref={ref}>Click</button>
}
```

This change made in this PR mirrors the corresponding override in the React typings:

```ts
function useRef<T = undefined>(): MutableRefObject<T | undefined>;
```

See https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e47aff8a3a1331d0c540550db79dd5065ab735e5/types/react/index.d.ts#L1045

### DIFF
--- a/compat/test/ts/forward-ref.tsx
+++ b/compat/test/ts/forward-ref.tsx
@@ -1,7 +1,7 @@
 import React from '../../src';
 
 const MyInput: React.ForwardFn<{ id: string }, { focus(): void }> = (props, ref) => {
-  const inputRef = React.useRef<HTMLInputElement>()
+  const inputRef = React.useRef<HTMLInputElement>(null)
 
   React.useImperativeHandle(ref, () => ({
     focus: () => {
--- a/hooks/src/index.d.ts
+++ b/hooks/src/index.d.ts
@@ -1,4 +1,4 @@
-import { PreactContext, Ref as PreactRef } from '../..';
+import { PreactContext, Ref as PreactRef, RefObject } from '../..';
 
 type Inputs = ReadonlyArray<unknown>;
 
@@ -40,6 +40,7 @@ export function useReducer<S, A, I>(
 	init: (arg: I) => S
 ): [S, (action: A) => void];
 
+/** @deprecated Use the `Ref` type instead. */
 type PropRef<T> = { current: T };
 type Ref<T> = { current: T };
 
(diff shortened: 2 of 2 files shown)


marvinhagemeister (commented):
This is awesome! Just checked it out but I didn't get a compile error. Presumably, we need to remove the optional sign (=`?`) at this line:

https://github.com/preactjs/preact/blob/cd8d65882d08af9f3b7d143d441d128ff068705c/hooks/src/index.d.ts#L53

When that's removed accessing `myRef.current.foo` errors as expected.

cmlenz:
You are right, this change does not do what it was intended to do. Sorry for not checking this thoroughly myself.

In my example snippet from above, the `current` property will always be `undefined` on first render, but the typings in both React and Preact choose the pragmatic approach of not including `null` or `undefined` in its type. While this may mask bugs that TypeScript should normally warn against, the alternative would add a lot of null-checking that is not what library users would expect, apparently. Even the version of `useRef` that expects an argument excludes `null` and `undefined` from the return type, even though they are clearly possible.

I do wonder why this is different between the `Ref` type in hooks and the `Ref` type in core, which *does* include `null`. Seems to me that the simplest thing would be to use that type for both `createRef` and `useRef`.

Anyway, after playing around with this some more, I have come away more confused than before, but have learned this:

1. The `useRef` typings in Preact no longer make sense since #2803, as making the parameter optional means the compiler will not choose the second argument-less override. Makes me wonder why #2803 was submitted in the first place.
2. Also, that override has its own return type `PropRef<T>`, which is (outside of this PR) exactly the same as the `Ref<T>` type in hooks.

My proposal would be to simplify the typings by keeping the argument optional, removing the override, and removing (or deprecating?) the `PropRef<T>` type alias. I can update the PR with this change if that sounds like a good idea to you :)

cmlenz:
I went ahead and updated the PR.

marvinhagemeister (approved):
Love it, this works and is much better 🙌
+1 on deprecating the `PropRef` type

PR #3064, opened 2021-03-14
Fix tabIndex not getting reset in IE11

marvinhagemeister:
Follow up to #3060. Verified via browerstack.


developit (commented):
LGTM. Let's just make sure to remove this in 11.

developit (approved):

PR #3072, opened 2021-03-19
Only reposition Suspense children when suspending & improve suspending around Fragments

andrewiggins:
In #3053, upon suspending, all the children under the Suspense's portal were rerendered. This PR changes that behavior such that the children under Suspense's Portal no longer rerender when they are suspended. Instead, the Portal's previous VNode ID is reused just causing this children to be re-positioned under the Portal's parentDOM. This change requires moving the root node `startDom` handling outside of `renderComponent` so that it still happens even when a Portal's VNode ID is the same.

This change also improves handling Suspense in and around Fragments. The biggest fix is just making sure that a Portal's (or root node's) `_dom` pointer doesn't bubble up to its parent Internals when it has a different `parentDom` pointer. Those little special `if` conditionals will hopefully go away when we stop bubbling `_dom` pointer up the tree through component Internals.

Also, re-enabled the suspense hydration around `sCU` components since backing tree enables this to just work.

### DIFF
--- a/follow-ups.md
+++ b/follow-ups.md
@@ -61,6 +61,8 @@
   - Ensure all tree traversal functions handle root nodes with different
     parentDOM. For example, getChildDom & getDomSibling should skip over root
     nodes iif they have different parentDOMs.
+  - Remove TYPE_ROOT special handling in mountChildren/diffChildren related to
+    skipping set \_dom on parent Internals through Portal/root nodes
 - Fix Suspense tests:
   - "should correctly render nested Suspense components without intermediate DOM #2747"
 - Fix Suspense hydration tests:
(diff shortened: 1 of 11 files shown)


JoviDeCroock (approved):

follow-ups.md
```diff
@@ -13,6 +28,22 @@
 - Rewrite commit loop to operate on internals not components
 - rewrite hooks to operate on internals?
 - Always assign a number to `_vnodeId` (not null, use 0 to clear?)
+- Consider converting Fragment (and other special nodes, e.g. root, Portal,
+  context?) to be numbers instead of functions with special diff handling
+  - One benefit of using numbers for Roots/Portals is that a portal could be
```
marvinhagemeister, line 33: +10000

marvinhagemeister (approved):
This is soo good 👍👍

PR #3073, opened 2021-03-20
Remove DOM pointers from Component nodes

andrewiggins:
This PR removes the setting and bubbling up of DOM pointers through Component backing nodes. This change allows me to remove some tracking code in diff/mountChildren, as well as the `updateDomPointers()` helper! Though `updateDomPointers()` has basically been replaced with `getChildDom()`.  Note, `getChildDom()` takes an optional index parameter that is used to specify which child to begin searching for a child dom element. By default, it starts with 0 so behaves like a `getFirstChildDom()` function, but `getDomSibling` has use for specifying which child it should begin searching at, hence the index param.

In order for `getDomSibling()` and `getChildDom()` to work, I had to teach them about portals/root nodes. Both functions should only search from DOM nodes through root nodes if the portal/root nodes `_parentDom` property matches its parent's `parentDom`. When a root node's parent DOM is the same as its parent, the root node behaves like a Fragment. However if the parent DOMs are different, then both functions should skip over searching through the portal/root node since any DOM under it aren't actually siblings/children of the tree we are searching in.

### DIFF
--- a/follow-ups.md
+++ b/follow-ups.md
@@ -30,10 +30,14 @@
 - Always assign a number to `_vnodeId` (not null, use 0 to clear?)
 - Consider converting Fragment (and other special nodes, e.g. root, Portal,
   context?) to be numbers instead of functions with special diff handling
-  - One benefit of using numbers for Roots/Portals is that a portal could be
-    created with a `null` \_parentDOM prop and still be marked as a root type.
-    Particularly, this is useful for Suspense and would remove the need for the
-    call to `getParentDom` in `Suspense.render`
+
+  One benefit of using numbers for Roots/Portals is that a portal could be
+  created with a `null` \_parentDOM prop and still be marked as a root type.
+  Particularly, this is useful for Suspense and would remove the need for the
+  call to `getParentDom` in `Suspense.render`
+  
+  With the above suggestion, remember to update getParentDom, getDomSibling,
+  etc. to handle a `null` `_parentDom`
 
 ## Suspense follow ups
 
--- a/follow-ups.md
+++ b/follow-ups.md
@@ -6,6 +6,7 @@
 
 ## Root node follow ups
 
+- Investigate if the return value of `createRoot()` can be re-used as a root Node...
 - Add specific tests for root nodes
   - initial mount with same/different parentDOM as parent
   - mounting on rerender with same/different parentDOM as parent
@@ -35,7 +36,7 @@
   created with a `null` \_parentDOM prop and still be marked as a root type.
   Particularly, this is useful for Suspense and would remove the need for the
   call to `getParentDom` in `Suspense.render`
-  
+
   With the above suggestion, remember to update getParentDom, getDomSibling,
   etc. to handle a `null` `_parentDom`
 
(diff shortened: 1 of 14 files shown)


src/diff/children.js
```diff
@@ -119,7 +120,10 @@ export function diffChildren(
 		}
 		// If this node suspended during hydration, and no other flags are set:
 		// @TODO: might be better to explicitly check for MODE_ERRORED here.
-		else if ((childInternal._flags ^ (MODE_HYDRATE | MODE_SUSPENDED)) === 0) {
```
andrewiggins, line 122: @developit This change didn't work for me cause the result is never `0`. When both `MODE_HYDRATE` and `MODE_SUSPENDED` flags are set, the result is all other flags set in `_flags`.

Here is a simplified example. Imagine the left most bit is a flag that represents the type of the internal:

| \_flags | Hydrate \| Suspended | Result with ^ | What we want |
| ------- | -------------------- | ------------- | ------------ |
| 100     | 011                  | 111           | false        |
| 101     | 011                  | 110           | false        |
| 110     | 011                  | 101           | false        |
| 111     | 011                  | 100           | true         |

src/tree.js
```diff
@@ -83,6 +84,11 @@ export function createInternal(vnode, parentInternal) {
 	return internal;
 }
 
+const shouldSearchComponent = internal =>
+	internal._flags & TYPE_COMPONENT &&
+	(!(internal._flags & TYPE_ROOT) ||
+		internal.props._parentDom == getParentDom(internal._parent));
```
andrewiggins, line 90: This is the logic that says only search through a Component if it is not a portal/root node or if it is a root node, search through it only if it's parentDom matches its parent's parentDom

src/render.js
```diff
@@ -37,7 +37,7 @@ export function render(vnode, parentDom, replaceNode) {
 			{},
 			parentDom.ownerSVGElement !== undefined,
 			commitQueue,
-			rootInternal._dom
+			getChildDom(rootInternal, 0) // TODO: could this just be parentDom.firstChild? Probably once we drop replaceNode
```
marvinhagemeister, line 40: I'm so ready to drop `replaceNode` 🙌

test/browser/components.test.js
```diff
@@ -1771,63 +1771,6 @@ describe('Components', () => {
 		});
 	});
 
-	it('should set component._internal._dom when sCU returns false', () => {
```
marvinhagemeister, line 1774: +1 for removing this test. I think it was one of the last remaining ones where we tested implementation details directly 👍

test/browser/getDomSibling.test.js
```diff
@@ -379,4 +382,182 @@ describe('getDomSibling', () => {
 		expect(divAInternal.key).to.equal('A');
 		expect(getDomSibling(divAInternal)).to.equal(null);
 	});
+
+	it("should return null if it's the only child", () => {
```
marvinhagemeister, line 386: Man these new tests are amazing!! 💯

JoviDeCroock (approved):
This is amazing 🙌

follow-ups.md
```diff
@@ -57,18 +62,18 @@
...
-- Consider further removing `_dom` pointers from non-dom VNodes
-  - Ensure all tree traversal functions handle root nodes with different
-    parentDOM. For example, getChildDom & getDomSibling should skip over root
-    nodes iif they have different parentDOMs.
-  - Remove TYPE_ROOT special handling in mountChildren/diffChildren related to
-    skipping set \_dom on parent Internals through Portal/root nodes
+- Logic to begin (re)rendering an internal is sorta scattered around (top level
+  render/hydrate, rerenderComponent, diffChildren), particularly the logic to
+  resume suspended hydration. Could we consolidate this logic into a single
+  function?
```
marvinhagemeister, line 68: Agree, we should definitely consolidate that logic

follow-ups.md
```diff
@@ -30,10 +31,14 @@
...
   context?) to be numbers instead of functions with special diff handling
-  - One benefit of using numbers for Roots/Portals is that a portal could be
-    created with a `null` \_parentDOM prop and still be marked as a root type.
-    Particularly, this is useful for Suspense and would remove the need for the
-    call to `getParentDom` in `Suspense.render`
+
+  One benefit of using numbers for Roots/Portals is that a portal could be
+  created with a `null` \_parentDOM prop and still be marked as a root type.
+  Particularly, this is useful for Suspense and would remove the need for the
+  call to `getParentDom` in `Suspense.render`
```
marvinhagemeister, line 38: This makes me extremely excited for devtools, because in 10.x we need to pass function references around to detect if we're dealing with `Fragment`s or other special components. Would be much easier if it's just a serializable value like a number 👍

marvinhagemeister (approved):
I'm exremely excited about this!! Those DOM pointers were a frequent wall I ran into when attempting 2-phase experiments. You're comments about consolidating update logic is spot on too. This is a huge milestone!! Opens a lot of nice improvements for us 👍 💯

src/diff/children.js
developit: What you ended up with here makes more sense to me anyway and I believe will also be fastest (x & 6 === 6).

PR #3079, opened 2021-03-23
fix: make typings deno compatible

### LINKED ISSUE #3075: Type definitions, Deno support.

**Describe the bug**

Because of the way the type definitions in Preact are defined, they are unable to be used within Deno. See: https://github.com/denoland/deno/issues/9852

**To Reproduce**

The issue linked above has a reproduction steps, but I have also created a slim Github repo that reproduces this issue: https://github.com/tbremer/preact-deno-types-error

1. Ensure Deno is available on your system
    - Installation: https://deno.land/#installation
1. Clone the repository
1. Optionally pre-cache the modules with: deno cache --config tsconfig.json deps.ts
1. run the mod.tsx file: deno run --config tsconfig.json mod.tsx
1. Error is presented

**Expected behavior**

JSX IntrinsicElements should be consumable in the Deno runtime. 

**Additional Info**

I don't know if I have all the right context to open an immediately mergable pull request, but I would be happy to open an initial take on the problem and collaborate from there until it's resolved.

lucacasonato:
This commit changes the types to use ESM d.ts files intead of UMD/global
files. Added an additional test to validate that the global `preact`
Typescript namespace is still present.

Fixes #3075

### DIFF
--- a/compat/src/suspense-list.d.ts
+++ b/compat/src/suspense-list.d.ts
@@ -1,11 +1,11 @@
-import { Component, ComponentChild } from '../../src';
+import { Component, ComponentChild, ComponentChildren } from '../../src';
 
 //
 // SuspenseList
 // -----------------------------------
 
 export interface SuspenseListProps {
-	children?: preact.ComponentChildren;
+	children?: ComponentChildren;
 	revealOrder?: 'forwards' | 'backwards' | 'together';
 }
 
--- a/compat/src/suspense.d.ts
+++ b/compat/src/suspense.d.ts
@@ -1,4 +1,4 @@
-import { Component, ComponentChild } from '../../src';
+import { Component, ComponentChild, ComponentChildren } from '../../src';
 
 //
 // Suspense/lazy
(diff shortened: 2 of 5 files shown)


tbremer:
I am not too familiar with Preact's release process, but is there some way I can help move this PR along?

JoviDeCroock (approved):
Looks good to me, thank you so much!

developit:
Just so I'm following correctly, this removes the ambient `preact` namespace?

lucacasonato:
No the ambient namespace is still there (the `export as namespace preact`) for legact compat when using preact in UMD environments. This is tested by the new test I added.

The previous syntax used (`declare namespace preact`, and then `export = preact`) was really UMD centric, because it creates a global namespace and assigns that namespace as the export for this module. This is all very legacy TS syntax that breaks down quickly when you enable too many modern options, and disable CommonJS support (Deno). It has since been replaced with something that just looks like ESM, where types and functions are exported using the `export ...` syntax. This is what TS recommends nowadays: https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html#reference-example. Because preact ships as ESM JS files, the typings should match that as closely as possible.

I'll rebase the PR real quick too - seems to have gotten quite out of sync with `main` since I opened it.

PR #3080, opened 2021-03-24
Move render callbacks to Internals

andrewiggins:
... and rename to `commitCallbacks`.

This change is a step to further separate the core diff from a specific Component implementation. Though I suspect that `_commitCallbacks` won't forever stay on Internal (depends on how we implement an extensible component API), I figured I'd move it to Internals for now so we can start to peel away other parts of the codebase.

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -56,7 +56,7 @@ Component.prototype.setState = function(update, callback) {
 	if (update == null) return;
 
 	if (this._internal) {
-		if (callback) this._renderCallbacks.push(callback);
+		if (callback) this._commitCallbacks.push(callback);
 		enqueueRender(this);
 	}
 };
@@ -73,7 +73,7 @@ Component.prototype.forceUpdate = function(callback) {
 		// is coming from. We need this because forceUpdate should never call
 		// shouldComponentUpdate
 		this._internal._flags |= FORCE_UPDATE;
-		if (callback) this._renderCallbacks.push(callback);
+		if (callback) this._commitCallbacks.push(callback);
 		enqueueRender(this);
 	}
 };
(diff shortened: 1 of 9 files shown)


src/component.js
```diff
@@ -56,7 +56,7 @@ Component.prototype.setState = function(update, callback) {
 	if (update == null) return;
 
 	if (this._internal) {
-		if (callback) this._renderCallbacks.push(callback);
+		if (callback) addCommitCallback(this._internal, () => callback.call(this));
```
JoviDeCroock, line 59:
```suggestion
		if (callback) addCommitCallback(this._internal, callback.bind(this));
```

src/component.js
```diff
@@ -73,7 +73,7 @@ Component.prototype.forceUpdate = function(callback) {
 		// is coming from. We need this because forceUpdate should never call
 		// shouldComponentUpdate
 		this._internal._flags |= FORCE_UPDATE;
-		if (callback) this._renderCallbacks.push(callback);
+		if (callback) addCommitCallback(this._internal, () => callback.call(this));
```
JoviDeCroock, line 76:
```suggestion
		if (callback) addCommitCallback(this._internal, callback.bind(this));
```

JoviDeCroock (approved):

src/component.js
marvinhagemeister: @developit here is something to benchmark for you! I recall `bind()` being slower in <= Chrome 59 and being made faster in recent versions. We should check how `bind()` vs a closure behaves in other browsers.

marvinhagemeister:
Let's merge this, we can check the closure vs bind later 👍

marvinhagemeister (approved):

PR #3082, opened 2021-03-24
(restructure) - move hooks to internal._data

JoviDeCroock:


hooks/src/index.js
```diff
@@ -138,20 +137,20 @@ export function useReducer(reducer, initialState, init) {
...
+	if (!hookState._internal) {
 		hookState._value = [
 			!init ? invokeOrReturn(undefined, initialState) : init(initialState),
 
 			action => {
 				const nextValue = hookState._reducer(hookState._value[0], action);
 				if (hookState._value[0] !== nextValue) {
 					hookState._value = [nextValue, hookState._value[1]];
-					hookState._component.setState({});
+					hookState._internal._component.setState({});
```
JoviDeCroock, line 148: A good way forward would be to have an update mechanism on an internal I guess 😅

hooks/src/index.js
```diff
@@ -240,7 +239,7 @@ export function useCallback(callback, args) {
  * @param {import('./internal').PreactContext} context
  */
 export function useContext(context) {
-	const provider = currentComponent.context[context._id];
+	const provider = currentInternal._component.context[context._id];
```
JoviDeCroock, line 242: A future PR could attach `.context` on an internal (and also call provider.sub with an internal)

hooks/src/index.js
```diff
@@ -30,17 +30,17 @@ const RAF_TIMEOUT = 100;
...
 
 options._render = internal => {
 	if (oldBeforeRender) oldBeforeRender(internal);
 
-	currentComponent = internal._component;
+	currentInternal = internal;
 	currentIndex = 0;
 
-	const hooks = currentComponent.__hooks;
+	const hooks = currentInternal.__hooks;
```
andrewiggins, line 43: I'm thinking that `__hooks` should still live off of the `_component` property. Perhaps a better name for that property would be `_componentData` or `_instance`.

The idea is that Internal's shape should be pre-defined up front and not change to keep our core algo fast and avoid re-compiles on shape changes. Since at this point class components are still in core, `_component` will already exist and be an object but in the future if that's not true, I'd imagine in `options.render`, hooks would do something like `if (internal._component == null) { internal._component = { __hooks: [], ... /* etc. */ }`

andrewiggins (commented):
This is looking good! Thanks for doing this!!

Though curious what the heck caused the memory benches to freak out... I don't think any of them use hooks so no clue why this change would affect them...

hooks/src/index.js
andrewiggins: Yea, was thinking we need a way to expose something like `rerenderInternal`. Will come later for sure

hooks/src/index.js
JoviDeCroock: So it was good before? Figured we'd want tobget rid of the property all together 😅

JoviDeCroock:
I think the benchmarks are against the root branch as I initially targetted the wrong branch

hooks/src/index.js
developit: Yeah I like it either being on _component, or on that new _data "grab bag" object.

hooks/src/index.js
developit: what about the component._renderer API from Keyed?

hooks/src/index.js
JoviDeCroock: personally would prefer moving away from `_component`, is `_data` already started if so would it be a good keep-sake for all `hooks`/`components`/`composition`?

JoviDeCroock:
Creating new PR for PERF tests

PR #3087, opened 2021-03-26
Invoke commitQueue callbacks with their component as the context

developit:
This is an optimization of #3080. It turns out all of the commit queue callbacks we currently use either don't care about `this`, or expect it to be the Component instance associated with an internal, so we can just call the callbacks on `internal._component`.


JoviDeCroock (approved):

andrewiggins (approved):
LGTM!

marvinhagemeister (approved):

PR #3089, opened 2021-03-27
Move `isSVG` param to a flag on Internal

andrewiggins:
Remove the `isSvg` param from mount/patch/diff and instead set it as a `MODE_` flag on the relevant internals.

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -7,7 +7,7 @@ import options from '../options';
  * changes to
  * @param {object} newProps The new props
  * @param {object} oldProps The old props
- * @param {boolean} isSvg Whether or not this node is an SVG node
+ * @param {number} isSvg 0 if not an SVG element, else it is an SVG element
  */
 export function diffProps(dom, newProps, oldProps, isSvg) {
 	let i;
@@ -49,7 +49,7 @@ function setStyle(style, key, value) {
  * @param {string} name The name of the property to set
  * @param {*} value The value to set the property to
  * @param {*} oldValue The old value the property had
- * @param {boolean} isSvg Whether or not this DOM node is an SVG node or not
+ * @param {number} isSvg 0 if not an SVG element, else it is an SVG element
  */
 export function setProperty(dom, name, value, oldValue, isSvg) {
 	let useCapture;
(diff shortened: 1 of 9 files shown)


src/diff/props.js
```diff
@@ -49,7 +49,7 @@ function setStyle(style, key, value) {
  * @param {string} name The name of the property to set
  * @param {*} value The value to set the property to
  * @param {*} oldValue The old value the property had
- * @param {boolean} isSvg Whether or not this DOM node is an SVG node or not
+ * @param {number} isSvg 0 if not an SVG element, else it is an SVG element
```
andrewiggins, line 52: This parameter is only needed to convert `xlink;href`/`xlinkHref` to `href` and `className` to `class`. I wonder if there is a smaller way to determine this behavior without needing this param. Perhaps drop support for `xlink:href` and `xlinkHref` by providing a codemod and figure something else out for `className` -> `class`...

Something to think about for a future PR

JoviDeCroock (approved):

src/tree.js
```diff
@@ -79,6 +80,15 @@ export function createInternal(vnode, parentInternal) {
 		_depth: parentInternal ? parentInternal._depth + 1 : 0
 	};
 
+	if (internal._flags & TYPE_ELEMENT && type === 'svg') {
```
marvinhagemeister, line 83: I wonder if the slight slowdown in the `replace1k` benchmark is caused by these checks being called for every internal, including text nodes. My assumption is that we don't need that information for text nodes and can move these checks to the branch above where we do the flag detection logic.

marvinhagemeister (approved):
LGTM, left one small comment of an idea to try out, but I'm good with merging either way. I expect us to do more performance passes once we are close to a proper v11 release anyway in the future.

src/tree.js
andrewiggins: Updated so these only happen in the non-text node branch. Had to move the application of inherited modes higher up in this method though since these SVG checks (specifically the `foreignObject` one) need to happen after inherited modes have been applied (the `foreignObject` mode effectively turns off the inherited MODE_SVG flag).

In my local testing this didn't increase the size much. And I think I like it this way better cuz it explicitly enables us to modify MODE inheritance in special cases

marvinhagemeister (approved):

PR #3091, opened 2021-03-30
fix: should override children if null is provided as an argument

clyfish:
https://github.com/facebook/react/blob/v17.0.2/packages/react/src/__tests__/ReactElement-test.js#L264-L273

React overrides children if null is provided as an argument, but preact doesn't.

### DIFF
--- a/src/clone-element.js
+++ b/src/clone-element.js
@@ -25,7 +25,7 @@ export function cloneElement(vnode, props, children) {
 			children.push(arguments[i]);
 		}
 	}
-	if (children != null) {
+	if (arguments.length > 2) {
 		normalizedProps.children = children;
 	}
 
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -26,7 +26,7 @@ export function createElement(type, props, children) {
 			children.push(arguments[i]);
 		}
 	}
-	if (children != null) {
+	if (arguments.length > 2) {
 		normalizedProps.children = children;
 	}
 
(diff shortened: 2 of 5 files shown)


clyfish:
BTW, the new `jsx-runtime` also overrides children if null is provided as an argument.

clyfish:
@developit  @marvinhagemeister  Please review this PR.

We are trying to switch from React to Preact, but this is a blocking issue.

React renders `<div children="A">{null}</div>` to `<div></div>`, but preact without jsx-runtime renders to `<div>A</div>`.

src/clone-element.js
```diff
@@ -25,7 +25,7 @@ export function cloneElement(vnode, props, children) {
 			children.push(arguments[i]);
 		}
 	}
-	if (children != null) {
+	if (arguments.length > 2) {
```
marvinhagemeister, line 28: Let's add a test case for the `cloneElement` case and than it's good to go 👍

marvinhagemeister (commented):
Sorry for the late response. The PR is awesome! Left one minor comment and then it's ready to be merged 👍

JoviDeCroock (approved):

marvinhagemeister:
Awesome, thank you so much for making a PR 🙌

developit:
I think this causes `createElement('div', { children: 'foo' }, undefined)` to produce `{ children: undefined }`.

PR #3103, opened 2021-04-03
Separate internal component handling from component implementation

andrewiggins:
My goal with this change is to better understand what parts of our component implementation are generic and which are specific to React-style components. So I've added a separate `renderReactComponent` which is intended to only contain code for rendering react-style  components and `renderComponent` contains generic code for rendering components agnostically. There is a lot of cleanup I'd like to do and I'm not satisfied with some of the existing hacks here, but I thought I'd share to get some early feedback.

This PR is probably better understood by looking at the commits individually. I'll try to summarize the changes here:

* Add `renderReactComponent` function (481fad5)
* Add `SKIP` sentinel to communicate children should be skipped (481fad5) (bleh, currently not a fan of this)
* Add `setGlobalContext` function to allow a component implementation to change the globalContext passed to its children (481fad5) (bleh, currently not a fan of this)
* Always pass newVNode to renderComponent (c4c2fe5)
* Move VNode equality check into patch so we handle both component and DOM node VNode equality in the same place (3666939) (Note: b598683 first moves it outside of `renderReactComponent`)
* Move DIRTY_BIT handling out of component implementation (5e93a1b)
* Copy props from VNode to internal in `renderComponent` so component implementation doesn't have to (b687b49)

The remaining pieces to figure out are to determine how we can do react component things without importing from Preact internals, assuming that's something we are interested in pursuing 😅


follow-ups.md
```diff
@@ -6,6 +6,18 @@
 - `[MAJOR]` Remove `replaceNode`, [use this technique instead](https://gist.github.com/developit/f321a9ef092ad39f54f8d7c8f99eb29a))
 - `[MAJOR]` Removed select `<option>` fix for IE11, using select in IE11 will always require you to specify a `value` attribute
 
+## Component plugin prep follow ups
+
+- Consider exposing flags to replace various bools in `rendererState`
+	- Move setting COMMIT_FLAG to `setState`/`forceUpdate`
+- Consider adding TYPE_ERROR_BOUNDARY
```
andrewiggins, line 13: These are already addressed in #3107

marvinhagemeister (approved):
I like these changes a lot! Not having to drag a `_globalContext` around and instead switching the one from the component is much nicer and requires less pointers to keep around. This is a fantastic refactor 🙌 💯

developit:
This all looks good (and like a boatload of work, nice job). One thing that was gnawing at me as I read through though: what was the reason for moving `globalContext` to be global instead of being an object passed down through the tree as a function parameter? I recall we talked about this, but I don't remember the reasoning.

JoviDeCroock (approved):

src/diff/reactComponents.js
```diff
@@ -0,0 +1,195 @@
...
+/**
+ * Diff two virtual nodes and apply proper changes to the DOM
+ * @param {import('../internal').VNode} newVNode The new virtual node
+ * @param {import('../internal').Internal} internal The component's backing Internal node
+ * @param {import('../internal').RendererState} rendererState An object
+ * containing the current state of the renderer. See comments in the
+ * RendererState interface.
+ * @returns {import('../internal').PreactNode} pointer to the next DOM node (in order) to be rendered (or null)
+ */
+export function renderReactComponent(newVNode, internal, rendererState) {
```
JoviDeCroock, line 13: Personally would name this `renderComponent` rather than a specific name

src/diff/component.js
```diff
@@ -27,173 +30,41 @@ export function renderComponent(
...
+	rendererState.skip = false;
+	rendererState.commit = false;
+	rendererState.force = (internal._flags & FORCE_UPDATE) == FORCE_UPDATE;
 
-			// TODO: Returning undefined here (i.e. return;) passes all tests. That seems
-			// like a bug. Should validate that we have test coverage for sCU that
-			// returns Fragments with multiple DOM children
-			return reorderChildren(internal, startDom, parentDom);
-		}
+	const renderResult = renderReactComponent(newVNode, internal, rendererState);
```
JoviDeCroock, line 49: I wonder if it would be worth differentiating and making a hot path: if class `renderComponent` if func `type.render()`

JoviDeCroock:
Closing: targets the abandoned `restructure` branch (component-internals split that was not pursued in the shipped v11).

PR #3106, opened 2021-04-10
Avoid leaking internal vnodes

developit:
We started calling `options.vnode()` for internal "text" VNodes and copied VNodes at some point, as best as I can tell it was unintentional. We don't officially support the Text VNode shape, and copied VNodes shouldn't be passed through `options.vnode` again because they've already been processed.

This should be good for performance, though it likely won't show up in benchmarks because it primarily affects `preact/compat`.

### DIFF
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -78,7 +78,8 @@ export function createVNode(type, props, key, ref, original) {
 		_original: original == null ? ++options._vnodeId : original
 	};
 
-	if (options.vnode != null) options.vnode(vnode);
+	// Only invoke the vnode hook if this was *not* a direct copy:
+	if (original == null && options.vnode != null) options.vnode(vnode);
 
 	return vnode;
 }


andrewiggins (approved):

src/create-element.js
```diff
@@ -78,7 +78,8 @@ export function createVNode(type, props, key, ref, original) {
 		_original: original == null ? ++options._vnodeId : original
 	};
 
-	if (options.vnode != null) options.vnode(vnode);
+	// Only invoke the vnode hook if this was *not* a direct copy:
+	if (original == null && options.vnode != null) options.vnode(vnode);
```
JoviDeCroock, line 82: My concern here becomes a bit prefresh-related, we'd break prefresh for ESM-runtimes like wmr, vite & snowpack. Consider the following VDOM-tree:

```
const Parent = () => <Child />
```

When we upate Child internally we assign a new type and call forceUpdate, now if we update Parent the reference will still contain a stale version of `.type` in _children[0] which means that it'll render the previous version of said functional node.

We'd have to change from [this](https://github.com/JoviDeCroock/prefresh/blob/main/packages/core/src/runtime/vnode.js#L14) to something that gets all parent vnodes and updates their own child.type

developit: When Parent is invoked, it will call h(Child) and run options.vnode on that VNode - wouldn't that keep prefresh working?

PR #3107, opened 2021-04-12
Expose flags to component plugins

andrewiggins:
This PR builds on #3103 but uses flags to replace many of the booleans in in `rendererState`. It also adds a `TYPE_ERROR_BOUNDARY` flag that component plugins can set if a component can handle errors. Our `catchError` upward search uses this flag now instead of inspecting each `TYPE_COMPONENT` for lifecycle methods.

### DIFF
--- a/test/browser/lifecycles/componentDidCatch.test.js
+++ b/test/browser/lifecycles/componentDidCatch.test.js
@@ -655,49 +655,6 @@ describe('Lifecycle methods', () => {
 			expect(scratch).to.have.property('textContent', 'Error: Error!');
 		});
 
-		it('should bubble through already-dirty components', () => {
-			let middle;
-			class Middle extends Component {
-				render({ children }) {
-					middle = this;
-					// This component marks itself as dirty during rendering.
-					// The fact that it is marked as dirty should not prevent propagation of the error through this component.
-					if (this.state.a !== 1) {
-						this.setState({ a: 1 });
-					}
-					const markDirty = () => {
-						if (this.state.b !== 1) {
-							this.setState({ b: 1 });
-						}
-					};
-					// NOTE: the above avoids setting the DIRTY flag during rendering, circumventing the propagation issue.
-					// However, manually marking a component as dirty during rendering *will* prevent error propagation:
-					// this._internal.flags |= DIRTY
-					return [<Fragment ref={markDirty} />, children];
-				}
-			}
-			sinon.spy(Middle.prototype, 'render');
-
-			ThrowErr.prototype.render = throwExpectedError;
-
-			render(
-				<Receiver>
-					<Middle>
-						<ThrowErr />
-					</Middle>
-				</Receiver>,
-				scratch
-			);
-			rerender();
-
-			expect(middle)
-				.to.haveOwnProperty('state')
-				.that.deep.equals({ a: 1, b: 1 });
-			expect(Receiver.prototype.componentDidCatch).to.have.been.calledWith(
-				expectedError
-			);
-		});
-
 		it('should be called through non-component parent elements', () => {
 			ThrowErr.prototype.render = throwExpectedError;
 			render(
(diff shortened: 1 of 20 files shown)


test/browser/lifecycles/componentDidCatch.test.js
```diff
@@ -655,49 +655,6 @@ describe('Lifecycle methods', () => {
 			expect(scratch).to.have.property('textContent', 'Error: Error!');
 		});
 
-		it('should bubble through already-dirty components', () => {
```
marvinhagemeister, line 658: Curious, did this test not work anymore with dirty flagging and the bit masks?

marvinhagemeister (approved):
Another great PR 💯

test/browser/lifecycles/componentDidCatch.test.js
andrewiggins: ... that's a good question... why did I remove this test? Lemme get back to you on this

andrewiggins: Ah! I think this is what I was thinking:

So previously when handling errors, we would check the "dirty" flag on every component between the thrower and error boundary. So presumably a component in the middle of the thrower and the ErrorBoundary could interfere with the error handling.

However, with the new `TYPE_ERROR_BOUNDARY` flag, we now only check for the dirty flag on components that have specified they are error boundaries, so it is impossible for a middle component to interfere.

But, not sure why I removed the test... Lemme try to add it back

andrewiggins: This test fails in this branch cuz the state is wrong (`{ a: 1 }` vs `{ a: 1, b: 1}`). Here is what happens.

tl;dr: On `restructure`, Middle gets an additional rerender cuz `_catchError` thinks it is trying to handle the error and so it's next state (`{ a: 1, b: 1 }`) is applied. In this branch, we are a bit smarter and recognize that Middle is not an error boundary and so it does not get an extra rerender and its next state is never applied.

On `restructure` (and presumably `master`), the diff behaves likes this:

1. Middle.render calls `setState({ a: 1 })`. `DIRTY_BIT` is set and Middle is added to rerender queue
2. After calling `render`, the next state (`{ a: 1 }`) is applied to `this.state`.
3. The Fragment ref is called, so `setState({ b: 1 })` called and the next state for Middle is now `{ a: 1, b: 1 }`
4. ThrowError is rendered and throws
5. The `_catchError` climbs up the tree and sees that the `DIRTY_BIT` flag is set on Middle and so it assumes Middle is trying to handle the error (narrator: it is not).
6. Now we begin our rerender. The sorted rerender queue is `[Middle]`.
7. Middle rerenders, the next state is applied so `this.state` is now `{ a: 1, b: 1 }`.
8. ThrowError is rendered again and still throws
9. This time, `_catchError` sees that Middle tried to rerender but still threw and so skips it and keeps searching up for a component to handle the error. Receiver has `componentDidCatch` called which calls `this.setState({ error })`. Receiver's `DIRTY_BIT` is set. `_catchError` notices this and assumes Receiver is trying to handle the error (narrator: it is).
10. Our rerender loop sees that another component was added to the queue and so it rerenders it.
11. When Receiver rerenders, it mounts an error view and unmounts Middle
12. The final state of Middle is `{ a: 1, b: 1 }`

On this branch, the diff instead behaves likes this:

1. Middle.render calls `setState({ a: 1 })`. `DIRTY_BIT` is set and Middle is added to rerender queue
2. After calling `render`, the next state (`{ a: 1 }`) is applied to `this.state`.
3. The Fragment ref is called, so `setState({ b: 1 })` is called and the next state for Middle is now `{ a: 1, b: 1 }`
4. ThrowError is rendered and throws
5. Receiver handles the error and calls (`this.setState({ error })`). Receiver is added to the renrender queue. Note, Middle does not handle the error and is skipped here even though its `DIRTY_BIT` is set because it is not `TYPE_ERROR_BOUNDARY`
6. Now we begin our rerender. The sorted rerender queue is `[Receiver, Middle]`.
7. When Receiver rerenders, it mounts an error view and unmounts Middle
8. When Middle rerenders, we see it has unmounted and skip rerendering it.
9. The final state of Middle is `{ a: 1 }` with a never applied next state of `{ a: 1, b: 1 }`

So in summary, I think this new behavior is an improvement on what we do in restructure. Perhaps this test should instead assert the final HTML result instead of the asserting `Middle.state` to ensure that the diff settles to the correct outcome?

PR #3110, opened 2021-04-15
scout the full list off dom-nodes for reusal, else we could needlessl…

JoviDeCroock:
…y recreate nodes


JoviDeCroock:
Hmm, I thought this would be more of a perf improvement 😅

PR #3111, opened 2021-04-15
(restructure) - hooks currentComponent --> currentInternal

JoviDeCroock:
execute hooks on internals, this is one step towards needing to backRef ._internal on _component


hooks/src/internal.d.ts
```diff
@@ -1,11 +1,16 @@
...
-	PreactContext
+	PreactContext,
+	Internal as PreactInternal
 } from '../../src/internal';
 import { Reducer } from '.';
 
 export { PreactContext };
 
+export type Internal = PreactInternal & {
+	_component: Component;
```
andrewiggins, line 11: Doesn't the `PreactInternal` interface already have this field? Was TS complaining about the `| null` on the core Internal type?

andrewiggins (commented):
Thanks for doing this!

hooks/src/internal.d.ts
JoviDeCroock: mainly because we extend `Component` to have `__hooks`

PR #3112, opened 2021-04-15
Ignore undefined children in createElement 3rd arg

developit:
I believe #3091 may have introduced a bug where `h('div', { children: 'foo' }, undefined)` would produce `{ children: undefined }`. This fixes that bug, if it is a bug.

I've left this as a draft because I actually don't know which is the more desirable behavior here. From an implementation standpoint, I actually think the `arguments.length > 2` approach is best, because it allows us to seamlessly transition to rest parameters since the semantics are then the same:

```
function createElement(type, props, ...children) {
  if (children.length !== 0) props.children = children;
}
```

### DIFF
--- a/src/clone-element.js
+++ b/src/clone-element.js
@@ -25,7 +25,7 @@ export function cloneElement(vnode, props, children) {
 			children.push(arguments[i]);
 		}
 	}
-	if (arguments.length > 2) {
+	if (children !== undefined) {
 		normalizedProps.children = children;
 	}
 
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -26,7 +26,7 @@ export function createElement(type, props, children) {
 			children.push(arguments[i]);
 		}
 	}
-	if (arguments.length > 2) {
+	if (children !== undefined) {
 		normalizedProps.children = children;
 	}
 


clyfish:
IMO, we should match the behavior of `react` and `babel-plugin-transform-react-jsx` with the new jsx runtime.

The new jsx runtime (which preact also supports with `preact/jsx-runtime`) transforms

```
<div children="foo">{undefined}</div>
```

to

```
import { jsx as _jsx } from "preact/jsx-runtime";

/*#__PURE__*/
_jsx("div", {
  children: "foo",
  children: undefined
});
```

developit:
That is a bug in Babel and doesn't affect createElement(), it only affects jsxs()

clyfish:
The bug affects both React.createElement and Babel, if it is a bug.
I think Preact needs to be consistent with React and Babel, whether it is a bug or not.
https://github.com/facebook/react/blob/v17.0.2/packages/react/src/ReactElement.js#L386-L400

JoviDeCroock:
Closing: v11 intentionally keeps the `arguments.length > 2` children semantics (`src/create-element.js`), and undefined children are normalized in the diff.

PR #3116, opened 2021-04-18
Inline lifecycle handling into `internal._component` check

andrewiggins:
This change removes the `isNew` variable

Dependent on #3107 for now

Great catch by @developit


JoviDeCroock:
Closing: targets the abandoned `restructure-component-prep` branch; the lifecycle-inlining experiment does not apply to the shipped v11.

PR #3117, opened 2021-04-18
Simplify `skipRemove` flag handling in unmount

andrewiggins:


src/diff/unmount.js
```diff
@@ -42,9 +33,19 @@ export function unmount(internal, parentInternal, skipRemove) {
 
 	if ((r = internal._children)) {
 		for (let i = 0; i < r.length; i++) {
-			if (r[i]) unmount(r[i], parentInternal, skipRemove);
+			if (r[i]) {
+				unmount(
+					r[i],
+					parentInternal,
+					~internal._flags & TYPE_ROOT && internal._flags & TYPE_DOM
```
andrewiggins, line 40: I'm assuming that these bitwise operations are fast enough to do on every child/call to unmount (it saves some bytes!). If they are not, I think it'll show up on 02_replace1k which unmounts and mounts a tree of DOM every iteration.

developit: @andrewiggins I think we can pass a constant `1` here, if TYPE_ROOT handling is moved to line 46:

```js
// Line 46:
if (!skipRemove && internal._dom || ~internal._flags & TYPE_ROOT) {
  // either this is skipRemove=0, or a root Internal which we always remove:
  removeNode(internal._dom);
}
```

developit: or something like:

```js
	if (internal._flags & TYPE_DOM) {
		if (skipRemove) internal._dom = null;
		else skipRemove = 1;  // only used for children after this point
	}

	if ((r = internal._children)) {
		for (let i = 0; i < r.length; i++) {
			if (r[i]) unmount(r[i], parentInternal, skipRemove);
		}
	}

	// this will only be present for TYPE_DOM that wasn't set to null above due to skipRemove
	if (internal._dom) removeNode(internal._dom);
	internal._dom = null;
```

JoviDeCroock (approved):

src/diff/unmount.js
developit: I pushed a corrected version of my suggestions in 69993fd.

marvinhagemeister (approved):

developit (approved):

developit:
(from testing)

```
function gen(skipRemove, flags) {
  return skipRemove ? ~flags & TYPE_ROOT : flags & TYPE_DOM;
}
for (const skipRemove of [0,1,2,4])
  for (const flags of [1, 2, 4, 16])
    console.log(`skipRemove:${skipRemove}, flags:${flags}, ${gen(skipRemove, flags)}`);
```

PR #3115, opened 2021-04-18
[WIP] Use shift to empty queues

andrewiggins:
PR to check out the perf impact of using shift for our queues. Since these code paths aren't as hot as the core diff I'm not expecting much change.

Feel free to push to this branch if you want to experiment with other implementations

Thanks to @developit for the suggestion :)

PR #3120, opened 2021-04-20
Simplify unmount logic

andrewiggins:
Simplify unmounting logic to save some bytes. Should offset the bytes from #3115.


marvinhagemeister (approved):
Neat, that looks way cleaner 👍

PR #3124, opened 2021-04-20
(Suspense) - keep hookState around on subsequent suspensions

JoviDeCroock:
Currently React does not seem to reset state/refs, only reruns effects/memo's not entirely sure if that's only on first-mount or also subsequent suspension, need to do some more research on that.

Hmm given [this sandbox](https://codesandbox.io/s/sharp-matsumoto-1zj2u?file=/src/index.js) it does look like we're doing the right thing in resetting stuff


compat/src/suspense.js
```diff
@@ -43,12 +43,11 @@ options.unmount = function(vnode) {
 
 function detachedClone(vnode, detachedParent, parentDom) {
 	if (vnode) {
-		if (vnode._component && vnode._component.__hooks) {
-			vnode._component.__hooks._list.forEach(effect => {
-				if (typeof effect._cleanup == 'function') effect._cleanup();
-			});
-
+		if (vnode._component && vnode._component.__hooks && !vnode._dom) {
```
JoviDeCroock, line 46: This means that we reset hooks-state only on first-mount, not really sure if this is the most ideal way to detect this but for now we do it like this.

So during the traversal up during initial mount we reset all hooks

compat/src/suspense.js
```diff
@@ -43,12 +43,11 @@ options.unmount = function(vnode) {
...
 			vnode._component.__hooks = null;
+		} else if (vnode._component && vnode._component.__hooks) {
```
JoviDeCroock, line 48: If this is a previously mounted component suspending due to an update we just clear potentially scheduled effects

hooks/src/index.js
```diff
@@ -366,6 +366,8 @@ function invokeEffect(hook) {
 	// and move the currentComponent away.
 	const comp = currentComponent;
 	hook._cleanup = hook._value();
+	hook._args = hook._pendingArgs;
```
JoviDeCroock, line 369: This measure is taken so that when a hook is scheduled but cancelled due to Suspense or unmount we aren't eagerly changing the _args, we are now introducing an _pending state that will only be applied when the cb executes.

compat/src/suspense.js
andrewiggins: I think only the component that suspended has effects cleared, so this logic shouldn't run as part of the traversal. At least that is my observation here: https://codesandbox.io/s/parent-effects-when-child-suspends-1lsex

compat/src/suspense.js
andrewiggins: Ah dang yea. In restructure we could some of our flags to better detect this :/

hooks/src/index.js
andrewiggins: Oh nice! Good call. Does this have any observable change we could test?

andrewiggins: Thinking more about this change, does it mean we would only run effects if they changed from the original deps? So if I initiate two rerenders quickly that both suspend and the final rerender matches the original deps, then we wouldn't rerun effects here?

I'm thinking something like this:
```
Initial render deps: [1]
1st rerender deps: [2] // Suspends
2nd rerender deps: [1] // Happens before 1st rerender unsuspends. With this change would effects run with these args?
// Should they? Maybe they shouldn't?
```
```

File: compat/test/browser/suspense.test.js [hunk 4]
```diff
@@ -104,7 +102,7 @@ describe('suspense', () => {
 		});
 	});
 
-	it('should reset hooks of components', () => {
+	it('should set state correctly', () => {
```
andrewiggins, line 105: Do we have tests that cover the new behavior?

andrewiggins (commented):
Looking good! Left a couple question to clarify my understanding of Suspense

compat/src/suspense.js
JoviDeCroock: Good point

PR #3125, opened 2021-04-21
(restructure) - deprecate ie11 support

JoviDeCroock:


developit:
Interesting this didn't improve benchmarks more - I wonder if we should be jumping straight to restructuring?

I was thinking it might be worth switching the transpiler config to use targets.esmodules rather than last 2 versions.

babel.config.js
```diff
@@ -25,7 +25,7 @@ module.exports = function(api) {
 					modules: noModules ? false : 'auto',
 					exclude: ['@babel/plugin-transform-typeof-symbol'],
 					targets: {
-						browsers: ['last 2 versions', 'IE >= 9']
+						browsers: ['last 2 versions', 'not IE 11']
```
marvinhagemeister, line 28: Apparently `not IE 11` includes IE 10, blackberry 10 and other relics from the past: https://browserslist.dev/?q=bGFzdCAyIHZlcnNpb25zLCBub3QgSUUgMTE%3D

Not sure what a good replacement value would be that only includes somewhat modern browsers. Maybe @developit knows more?

developit: Need to include 'not dead'. But this was the main reason I am thinking we might be better off using targets.esmodules.

JoviDeCroock: esbuild seems to fail when I do that 😅

developit: Probably a simplified browserslist implementation. This should work and is equivalent:

```
"Firefox>=60, chrome>=61, and_chr>=61, Safari>=10.1, iOS>=10.3, edge>=16, opera>=48, op_mob>=48, Samsung>=8.2, not dead"
```

JoviDeCroock: I don't think microbundle actually looks at this babel.config 😅

developit: correct - it uses `"browserslist"` from the package.json.

developit:
Weird, coverage never ran on the merge commit.

JoviDeCroock:
Yes coverage seems to be acting up on `main` and my other PR as well

JoviDeCroock:
I think there's still a few suites not running

developit:
@danielweck heh - after I took that screenshot, someone from the V8 team pointed out that the `for..of` test was incorrectly referencing a global variable instead of the loop variable. Now things look a bit more like what you posted (notably with Safari suffering from quite poor `for..of` performance).

fzzle:
Chrome 90:
image

gustavpursche:
Firefox 90.0a1 (2021-05-20) on Ubuntu:

Screenshot from 2021-05-24 15-59-25

developit:
Nice! Thanks for the cross-platform checks @fzzle & @gustavpursche! I think they confirm things nicely - we likely can't make generalized recommendation between "unrolled" vs "unrolled X2", but the fact that those two are consistently either nearly-fastest (Chrome/Edge ~92) or significantly faster (Firefox & Safari) means we should be leaning _slightly_ towards unrolling where it can be done without sacrificing other metrics (size, simplicity, etc). FWIW I think the Edge result is a bit of an anomaly, and m93 will likely exhibit performance slightly more in line with m90 (a very small advantage for unrolling).

src/create-context.js
```diff
@@ -40,7 +40,9 @@ export function createContext(defaultValue, contextId) {
 						// 	c.context[contextId] = _props.value;
 						// 	enqueueRender(c);
 						// });
-						subs.some(enqueueRender);
+						for (let sub of subs) {
```
developit, line 43: lingering for..of here

src/diff/component.js
```diff
@@ -62,7 +62,7 @@ export function renderComponent(
 		c = internal._component;
 	} else {
 		// Instantiate the new component
-		if ('prototype' in type && type.prototype.render) {
+		if (internal._flags & TYPE_CLASS) {
```
andrewiggins, line 65: Could we leave this line as-is for now? I have another PR where the `TYPE_CLASS` flag gets removed as part of the "make our internals more agnostic to component implementation" work

developit: Definitely - I couldn't generate any meaningful metrics to support this change when I tried, best to just leave it as-is.

PR #3130, opened 2021-04-26
Rely directly on performance.now

developit:
React doesn't export the scheduler methods or constants from `react` or `react-dom`.

### DIFF
--- a/compat/src/scheduler.js
+++ b/compat/src/scheduler.js
@@ -18,7 +18,4 @@ export function unstable_runWithPriority(priority, callback) {
 	return callback();
 }
 
-export const unstable_now =
-	typeof performance === 'object' && typeof performance.now === 'function'
-		? performance.now.bind(performance)
-		: () => Date.now();
+export const unstable_now = () => performance.now();
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
 		"build:devtools": "microbundle build --raw --cwd devtools",
 		"build:hooks": "microbundle build --raw --cwd hooks",
 		"build:test-utils": "microbundle build --raw --cwd test-utils",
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -62,6 +62,10 @@
 			"require": "./compat/server.js",
 			"import": "./compat/server.mjs"
 		},
+		"./compat/scheduler": {
+			"require": "./compat/dist/scheduler.js",
```
marvinhagemeister, line 66: The CI fails because this file is not present there. Maybe we need to build first?

marvinhagemeister (approved):
Let's get the CI green and then this is ready to be merged 👍

package.json
developit: that's an option, yeah. We don't currently provide the `scheduler` package, and it's only ever CJS. So I was thinking maybe we move it from `compat/src/scheduler.js` to `compat/scheduler.{mjs,js}` just like we do for `compat/server`?

marvinhagemeister: sounds good to me 👍

compat/src/scheduler.js
```diff
@@ -18,7 +18,4 @@ export function unstable_runWithPriority(priority, callback) {
 	return callback();
 }
 
-export const unstable_now =
-	typeof performance === 'object' && typeof performance.now === 'function'
-		? performance.now.bind(performance)
-		: () => Date.now();
+export const unstable_now = () => performance.now();
```
JoviDeCroock, line 21:
```suggestion
export const unstable_now = performance.now;
```

developit: heh - I checked because I'd hoped to do the same, but this throws an error due to the context being set to `globalThis`. I've changed it back to using bind.

PR #3132, opened 2021-04-27
Move VNode matching logic into its own function

andrewiggins:
While doing some child diffing investigations, I noticed locally that moving our VNode matching logic into its own function and improve some of the run times on 02_replace1k benchmark. The final duration doesn't change too much but locally it appears to remove some of the main-thread optimizations that occured in the early runs of 02_replace1k

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -35,7 +35,7 @@ export function diffChildren(
 	commitQueue,
 	startDom
 ) {
-	let i, j, newDom, refs;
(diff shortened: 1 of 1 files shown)


andrewiggins:
If you expand the results for `02_replace1k` you'll see these tables. These are the speedups I mentioned in my PR description that are motivating this change:

image

JoviDeCroock (approved):

PR #3136, opened 2021-05-02
remove prevDom as there is no need for it

JoviDeCroock:


andrewiggins (approved):
Oh wonderful! Glad to see this go.

PR #3139, opened 2021-05-06
Update signature of `lazy` to reflect behavior

JonasKruckenberg:
The `lazy` function of `preach/compat` accepts the promise to either return a `{ default: T }` (as `import()` would) or just `T`. This was missing in the typescript declaration however and this PR dresses that.

### DIFF
--- a/compat/src/suspense.d.ts
+++ b/compat/src/suspense.d.ts
@@ -3,7 +3,7 @@ import { Component, ComponentChild } from '../../src';
 //
 // Suspense/lazy
 // -----------------------------------
-export function lazy<T>(loader: () => Promise<{ default: T }>): T;
+export function lazy<T>(loader: () => Promise<{ default: T } | T>): T;
 
 export interface SuspenseProps {
 	children?: preact.ComponentChildren;


developit (approved):

ForsakenHarmony:
> The `lazy` function of `preach/compat`

`preach` for preact 🙏

birtles:
For anyone else arriving here because things broke after updating to 10.5.14, it looks like if you have:

```ts
const MyComponent = lazy(() =>
  import('./my-component').then((m) => {
    return {
      default: m.MyComponent,
    };
  })
);
```

then as of this change it seems like TypeScript will pick the RHS of the union `{ default: T } | T` and resolve `T` as `{ default: React.FunctionComponent<Props> }` and hence the type of `MyComponent` will be `{ default: React.FunctionComponent<Props> }` giving you errors like `TS2604: JSX element type 'MyComponent' does not have any construct or call signatures` when you try to use it.

For me the solution was to supply the generic parameter manually:

```ts
import type { Props as MyComponentProps } from './my-component';

const MyComponent = lazy<React.FunctionComponent<MyComponentProps>>(() =>
  import('./my-component').then((m) => {
    return {
      default: m.MyComponent,
    };
  })
);
```

marvinhagemeister:
@birtles Can you file an issue for that? It sounds like something worth fixing so that user's don't have to pass the generic manually.

birtles:
> @birtles Can you file an issue for that? It sounds like something worth fixing so that user's don't have to pass the generic manually.

Sure. I filed #3219 for now. I hope you don't mind that I didn't provide a CodeSandbox to repro the issue -- I'm not too familiar with CodeSandbox and the template doesn't appear to include TypeScript so I'd rather not spend time trying to get that set up.

marvinhagemeister:
All good, it's much easier for us to track issues via issues in GitHub compared to comments on closed PRs. That makes sure it doesn't get lost somewhere 👍

PR #3143, opened 2021-05-09
Use Array.prototype.slice.call for children arguments

fzzle:


developit:
Solid performance win!

src/clone-element.js
```diff
@@ -19,14 +19,11 @@ export function cloneElement(vnode, props, children) {
...
-		children = [children];
-		for (i = 3; i < arguments.length; i++) {
-			children.push(arguments[i]);
-		}
-	}
 	if (arguments.length > 2) {
-		normalizedProps.children = children;
+		normalizedProps.children =
+			arguments.length > 3
+				? Array.prototype.slice.call(arguments, 2)
```
developit, line 25: Hoisting the .slice access out of this function should improve perf a tiny bit more:

```js
const slice = Array.prototype.slice;
// etc

		normalizedProps.children =
			arguments.length > 3 ? slice.call(arguments, 2) : children;
```

fzzle: Jep, and perhaps move it to util so it can be shared (less bytes)?

JoviDeCroock (approved):
Awesome find

src/clone-element.js
developit: Usually when something is only used in 2 places we duplicate - gzip compression makes the duplicates smaller than an intermediary variable.

developit: 100% agreed on the size thing. It actually does still run for fork-PRs, but it can't post a comment. I need to use the workaround Andre came up with that lets benchmarks run.

PR #3144, opened 2021-05-10
Improve typing of forwardRef

cmlenz:
The specified ref type is now included in the typing of the functional component returned by `forwardRef`.

This is similar to the corresponding React typings, which currently look like [this](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e293a5bb95e4c77482da1d91dde77d4a1d6e397c/types/react/index.d.ts#L800):

```ts
function forwardRef<T, P = {}>(
  render: ForwardRefRenderFunction<T, P>
): ForwardRefExoticComponent<PropsWithoutRef<P> & RefAttributes<T>>;
```

### DIFF
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -105,7 +105,7 @@ declare namespace React {
 
 	export function forwardRef<R, P = {}>(
 		fn: ForwardFn<P, R>
-	): preact.FunctionalComponent<P>;
+	): preact.FunctionalComponent<Omit<P, 'ref'> & { ref?: preact.RefObject<R> }>;
 
 	export function unstable_batchedUpdates(
 		callback: (arg?: any) => void,


brookback:
Heads up that this change was a breaking change for me. I had to override this locally with

```diff
-): preact.FunctionalComponent<Omit<P, 'ref'> & { ref?: preact.RefObject<R> }>;
+): preact.FunctionalComponent<Omit<P, 'ref'> & { ref?: preact.Ref<R> }>;
```

cmlenz:
Yeah, sorry, there's PR #3214 with a fix.

brookback:
Cool, thanks!

Also note that you don't need `Omit<P, 'ref'>` in there. The `&` operator is enough.

```ts
const fn = <P = {}>(p: P & {ref?: number}) => console.log(p.ref);

// Won't compile
// fn({
//    ref: 'foo',
// });

fn({
    ref: 1,
});

fn({});
```
[Source](https://www.typescriptlang.org/play?jsx=0&strict=true#code/MYewdgzgLgBAZmGBeGAeACsmBvAvgPgAoAHALhkwDIcAnAUzgH5ywBXAWwCM6bcBKZPhihIIADZ0AdGJABzEpPpw+AbgBQagPSaYAdXAByWKHbEAlhK06EhbFZgOYS8gbggQBgDT3+6tTbtHJwZyAEZvXw0A3yA)

This caused some problems in my codebase, so I replaced it with:

```diff
+): preact.FunctionalComponent<Omit<P, 'ref'> & { ref?: preact.Ref<R> }>;
-): preact.FunctionalComponent<P & { ref?: preact.Ref<R> }>;
```

PR #3152, opened 2021-05-13
save to set and unsubscribe in options.unmount

JoviDeCroock:


src/create-context.js
```diff
@@ -18,10 +19,19 @@ export function createContext(defaultValue, contextId) {
...
-				let subs = [];
+				let subs = new Set();
 				let ctx = {};
 				ctx[contextId] = this;
 
+				const oldUnmount = options.unmount;
+				options.unmount = internal => {
+					if (subs.has(internal._component)) {
+						subs.delete(internal._component);
+					}
```
developit, line 30:
```suggestion
					subs.delete(internal._component);
```

developit:
Interesting that the size is net larger. I think maybe the way this ends up being smaller is if we use a WeakMap for the provider-to-subscriber mappings, which could allow using a single `options.unmount` hook and replacing the `shouldComponentUpdate` injection with a direct call to update subscribers from within the body of the render function (just a guess).

src/create-context.js
```diff
@@ -1,9 +1,18 @@
...
-export let i = 0;
+export let nextContextId = 0;
 
-export function createContext(defaultValue, contextId) {
-	contextId = '__cC' + i++;
+const providers = new Set();
+
+export const unsubscribeFromContext = component => {
+	// if this was a context provider, delete() returns true and we exit:
+	if (providers.delete(component)) return;
```
andrewiggins, line 9: Huh, would've thought calling `Set.delete` on every unmount would've regressed 02_replace1k unmounting perf but it seems fine. Neat!

developit: I wonder if we should check this with the synchronous test runner.

PR #3157, opened 2021-05-16
Use Brotli instead of gzip

fzzle:
Let's optimize for Brotli size:
- It's supported by most modern browsers (and not IE 11).
- It performs better than gzip.

### DIFF
--- a/.github/workflows/size.yml
+++ b/.github/workflows/size.yml
@@ -15,3 +15,4 @@ jobs:
       - uses: preactjs/compressed-size-action@v1
         with:
           repo-token: '${{ secrets.GITHUB_TOKEN }}'
+          compression: 'brotli'


rschristian:
The key statistic there is not how many browsers support Brotli, but how many sites are distributing it. Browsers only consume what they're given, so we'd need to take a look at what's actually being given.

According to this [W3Techs report](https://w3techs.com/technologies/details/ce-brotlicompression), as of the 17th of May, 2021, only 23% of the world's top 10 million sites use Brotli. I'm not sure how reliable that report is, so take it with a pretty big grain of salt, but that sounds about right with my expectations.

Not that I imagine optimizing for Brotli will hurt Gzip size in any great way, but ignoring Gzip for Brotli, which is a small-ish minority at the moment, seems odd. Switching from Gzip here would mean size regressions for 77% of the world might go unnoticed.

fzzle:
> The key statistic there is not how many browsers support Brotli, but how many sites are distributing it. Browsers only consume what they're given, so we'd need to take a look at what's actually being given.
> 
> According to this [W3Techs report](https://w3techs.com/technologies/details/ce-brotlicompression), as of the 17th of May, 2021, only 23% of the world's top 10 million sites use Brotli. I'm not sure how reliable that report is, so take it with a pretty big grain of salt, but that sounds about right with my expectations.
> 
> Not that I imagine optimizing for Brotli will hurt Gzip size in any great way, but ignoring Gzip for Brotli, which is a small-ish minority at the moment, seems odd. Switching from Gzip here would mean size regressions for 77% of the world might go unnoticed.

I'm not sure that I would consider that a key statistic. Brotli is new - most websites aren't. A key statistic would be the % of websites created in 2021 that uses Brotli or the % of Preact websites that uses Brotli. I think that'd be a very different number.

A key selling point of Preact is its Brotli size ("Fast 3kB alternative..."). It'd be easier to change that to 2kB with Brotli than gzip.

developit:
This is a tricky one: while Brotli is certainly the more logical choice from a purely technical standpoint, changing the compression algorithm we use to track size invalidates all prior measurements.

I think the only way we can do this would be to report both brotli and gzipped sizes for a period of time, eventually moving to Brotli-only when there are fewer long-term PRs (such as when the current rewrite lands and has had a minute to settle).

PR #3158, opened 2021-05-17
(restructure) - use node.remove rather than removeNode helper

JoviDeCroock:


developit:
Likely test strategy needs to normalize the two removal types. Not sure what's up with the fragments thing though...

andrewiggins (approved):
Oh neat! Good catch with the Text.prototype. I didn't realize Text.prototype had its own .remove method! 🎉

PR #3163, opened 2021-05-25
[restructure] Simplify normalizeToVNode() so all new types are treated as text by default

developit:
This just inverts the type checks in `normalizeToVNode` so that the default case (for types we don't track or use) is to convert things to a String.


JoviDeCroock (approved):

developit:
Update: it seems like the switch variant with string passthrough would be best:
https://esbench.com/bench/60ad2eca6c89f600a57008a4
Screen Shot 2021-05-25 at 3 26 55 PM

developit:
I have determined that it is impossible to improve on Andre's implementation (lol).

PR #3169, opened 2021-06-02
(restructure) - Remove is-non-dimensional

JoviDeCroock:
Copied from https://github.com/preactjs/preact/pull/2629

Breaking:
- Removes automatic postfixing of dimensional

New:
- Warning for when people rely on this for easy migration
- Automatic postfixing for non-dimensional in compat

### DIFF
--- a/compat/src/util.js
+++ b/compat/src/util.js
@@ -10,6 +10,8 @@ export function assign(obj, props) {
 	return /** @type {O & P} */ (obj);
 }
 
+export const IS_NON_DIMENSIONAL = /^(-|f[lo].*[^se]$|g.{5,}[^ps]$|z|o[pr]|(W.{5})?[lL]i.*(t|mp)$|an|(bo|s).{4}Im|sca|m.{6}[ds]|ta|c.*[st]$|wido|ini)/;
+
 /**
  * Check if two objects have a different shape
  * @param {object} a
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -232,7 +232,7 @@ export function initDebug() {
 					for (let i in style) {
 						if (typeof style[i] === 'number' && !IS_NON_DIMENSIONAL.test(i)) {
 							console.warn(
-								`Inline CSS value is missing a unit like "rem" or "px: "${i}: ${style[i]}"\n` +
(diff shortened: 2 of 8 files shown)


debug/src/debug.js
```diff
@@ -221,6 +222,22 @@ export function initDebug() {
...
+					!isCompatNode &&
+					key === 'style' &&
+					vnode.props[key] !== null &&
+					typeof vnode.props[key] === 'object'
+				) {
+					const style = vnode.props[key];
+					for (let i in style) {
+						if (typeof style[i] === 'number' && !IS_NON_DIMENSIONAL.test(i)) {
+							console.warn(
+								`Inline CSS value is missing a unit like "rem" or "px: "${i}: ${style[i]}"\n` +
```
developit, line 235:
```suggestion
								`Inline CSS value is missing a unit like "rem" or "px": ${i}: ${style[i]}"\n` +
```

developit: Though tbh suggesting `rem` here might be misleading - unitless values were being converted to px, never rem.

```suggestion
								`Numeric CSS property value is missing a "px" unit: ${i}: ${style[i]}"\n` +
```

compat/src/util.js
```diff
@@ -10,6 +10,8 @@ export function assign(obj, props) {
 	return /** @type {O & P} */ (obj);
 }
 
+export const IS_NON_DIMENSIONAL = /^(-|f[lo].*[^se]$|g.{5,}[^ps]$|z|o[pr]|(W.{5})?[lL]i.*(t|mp)$|an|(bo|s).{4}Im|sca|m.{6}[ds]|ta|c.*[st]$|wido|ini)/;
```
developit, line 13: Is this the regex that was suggested a while back? I am wondering if the increased repetition will impact performance.

marvinhagemeister (approved):

compat/src/util.js
JoviDeCroock: It's the new one, while it is a bit slower it's more correct;

image

marvinhagemeister: imo it's perfectly fine as the `style` attribute isn't used that much in my experience. Most React apps I've worked on, only make use of that occasionally for certain areas.

PR #3170, opened 2021-06-03
(restructure) Upgrade Terser to prevent function inlining

developit:
We are heavily reliant on `reduce_funcs`, which was removed in Terser 5, but [added back in 5.2.2 or 5.3](https://github.com/terser/terser/issues/350).
In the meantime, we have to disable `reduce_vars` to prevent function inlining, although this has a large size cost since it also prevents variable inlining.


devtools/package.json
```diff
@@ -4,8 +4,9 @@
 	"version": "1.0.0",
 	"private": true,
 	"description": "Preact bridge for Preact devtools",
+  "exports": "./dist/devtools.mjs",
```
ForsakenHarmony, line 7: 2 spaces here 🤔

Seems like the indent of `package.json` files is different, does npm no longer force some specific indent on these?

developit: npm now detects and uses the current indentation. I was avoiding running Prettier on these files since it converts the indentation over to tabs - we should do that in a dedicated PR probably.

PR #3173, opened 2021-06-04
(restrcuture) run benches when they're changed

developit:
This runs benchmarks in CI not just for `src/**.js` changes, but also when the benchmarks themselves are modified, as well as for any `mangle.json` changes, which play an important role in performance.


developit:
(I'm being bad and merging this eagerly)

marvinhagemeister:
> (I'm being bad and merging this eagerly)

All good, that's what the `restructure` branch is for

PR #3183, opened 2021-06-09
(restructure) - move undefined into constant

JoviDeCroock:


src/constants.js
```diff
@@ -60,3 +60,4 @@ export const RESET_MODE = ~(
 export const INHERITED_MODES = MODE_HYDRATE | MODE_MUTATIVE_HYDRATE | MODE_SVG;
 
 export const EMPTY_ARR = [];
+export const UNDEFINED = undefined;
```
developit, line 63:
```suggestion
export let UNDEFINED;
```

JoviDeCroock: Screenshot 2021-06-10 at 10 00 01

Had a pretty bad impact on both size and perf

marvinhagemeister (approved):

PR #3185, opened 2021-06-10
fix: Fixing undefined initializer case for useState type

### LINKED ISSUE #3181: useState type differences between React and Preact

**Describe the bug**
The type definitions for useState differ slightly.

React: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L913
Preact: https://github.com/preactjs/preact/blob/master/hooks/src/index.d.ts#L10

The React definitions allow for calling useState with no initialState, and it will infer a state of type T | undefined.  In Preact I must pass in an undefined as the initialState, and create a union with undefined for the type.

Preact valid code:
```
const [editingObject, setEditingObject] = useState<{id: number} | undefined>(
        undefined
    );
```

**To Reproduce**

`const [editingObject, setEditingObject] = useState<{ id: number }>();`

**Expected behavior**
This should compile and infer a state of type { id: number } | undefined

rschristian:
Should fix #3181

### DIFF
--- a/hooks/src/index.d.ts
+++ b/hooks/src/index.d.ts
@@ -9,6 +9,11 @@ export type StateUpdater<S> = (value: S | ((prevState: S) => S)) => void;
  */
 export function useState<S>(initialState: S | (() => S)): [S, StateUpdater<S>];
 
+export function useState<S = undefined>(): [
+	S | undefined,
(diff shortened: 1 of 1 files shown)


developit (approved):
Yes! This one has been driving me nuts haha.

PR #3191, opened 2021-06-15
Fix type definitions

craftedsystems:
This PR adds attribute typings for HTMLAttributes

* adds camel case version of `spellCheck`
* adds non-standard attribute `autoCapitalize` and `autocapitalize`

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -755,8 +755,8 @@ export namespace JSXInternal {
 		wrap?: string;
 
 		// Non-standard Attributes
-		autocapitalize?: string;
-		autoCapitalize?: string;
+		autocapitalize?: 'off' | 'none' | 'on' | 'sentences' | 'words' | 'characters';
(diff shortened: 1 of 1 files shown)


src/jsx.d.ts
```diff
@@ -753,6 +754,10 @@ export namespace JSXInternal {
 		wmode?: string;
 		wrap?: string;
 
+		// Non-standard Attributes
+		autocapitalize?: string;
+		autoCapitalize?: string;
```
rschristian, line 759:
```suggestion
		autocapitalize?: 'off' | 'none' | 'on' | 'sentences' | 'words' | 'characters';
		autoCapitalize?: 'off' | 'none' | 'on' | 'sentences' | 'words' | 'characters';
```

Types could be improved by being more specific. As the value has to be one of those values, `string` is very vague by comparison. [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize)

craftedsystems: very good point

rschristian (approved):
Awesome, thanks for adding these!

PR #3192, opened 2021-06-16
refactor(catch-error): use `while` will reduce the volume of output MAP files (-16b)

liuarui:
- I found it here replace `for` for `while`, will reduce the volume of MAP file volumes
- The output `dist` folder will be reduced by 16b
84561623829598_ pic

I am not sure if this is helpful😅


JoviDeCroock:
Hey, this seems to leave the size of Preact [unchanged](https://github.com/preactjs/preact/pull/3192/checks?check_run_id=2837142573)

liuarui:
this seems to just change source map size 🤣

PR #3194, opened 2021-06-18
refactor(diff-children): reduce typeof usage to reduce size

liuarui:
It seems to improve the speed of operation?


liuarui:
reopen

PR #3199, opened 2021-06-19
Remove unecessary assignment

sinnerJoe:
Checked in this way in Firefox console:
```
var a = 5, b = 6;
a = b = a;
a // shows 5
b // shows 5
```
If the leftmost assignment doesn't change the value of the variable then it can be safely removed
image

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -87,7 +87,7 @@ export function diffChildren(
 				childVNode._original
 			);
 		} else {
-			childVNode = newParentVNode._children[i] = childVNode;
+			newParentVNode._children[i] = childVNode;
 		}
 
 		// Terser removes the `continue` here and wraps the loop body


developit:
Hmm - this wouldn't have originally been written, I'm thinking maybe it was a broken merge?

developit:
Ah, mystery solved! @andrewiggins did this because having the extra assignment there improves the compressed size (we do this a lot - lines that are more similar compress better). Removing the extra assignment increases Preact's size by 10b:

|:--- |:---:|:---:|:---:|
| `dist/preact.js` | 3.98 kB | +8 B (0%) |  |
| `dist/preact.min.js` | 4 kB | +10 B (0%) |  |
| `dist/preact.module.js` | 4 kB | +11 B (0%) |  |
| `dist/preact.umd.js` | 4.05 kB | +6 B (0%) |  |

sinnerJoe:
Wow, I would never have thought about that. That's really clever. Thanks for the explanation.

PR #3202, opened 2021-06-21
refactor(diff-children): reduce typeof usage to improve performance 

liuarui:
[esbench](https://esbench.com/bench/60d031f56c89f600a5700e32)


fzzle:
Your benchmark implementation is wrong. You don't use `typeof` in the includes case.
image
`typeof x === 'number'`, etc. are some of the fastest expressions in JS (even faster than equality checks).

liuarui:
> Your benchmark implementation is wrong. You don't use `typeof` in the includes case.
> image
> `typeof x === 'number'`, etc. are some of the fastest expressions in JS (even faster than equality checks).

get😱😱 sorry

PR #3204, opened 2021-06-24
docs: Removing suggestion to install Preact CLI

rschristian:
Just another reference to installing Preact CLI.

While not as explicit as the other instructions to install globally, probably don't want to encourage direct installation, at least without pointing people at the docs first.


marvinhagemeister (approved):
Good catch 👍

Issue #3045, opened 2021-02-26
Preact version 10.5.11 (& 10.5.12) breaks a specific rendering scenario


ludvigeriksson:
As of Preact v 10.5.11 I noticed a scenario in my app in which rendering of an overlay in OpenLayers breaks. 

### Reproduction

I've distilled it down to a minimal example here: https://codesandbox.io/s/gifted-shaw-xvoyg. The background map is only there for demonstration purposes & is not needed for the bug. The file responsible for the overlay rendering is `oloverlay.js`.

### Steps to reproduce

1. Open the sandbox
2. Click somewhere on the map
3. See a red rectangle appear where you clicked (expected behaviour)
4. Change the preact version in the sidebar from 10.5.10 to 10.5.11 or 10.5.12.
5. See the red rectangle is now pinned above the map, & nothing happens when you click the map.

### Expected Behavior

A red rectangle should appear where you clicked on the map.

### Actual Behavior

The red rectangle is pinned above the map.

Aswathprabhu:
I'm too facing a similar issue recently (version **10.5.11** onwards) and I've managed to repro this in an example.
In my example, I'm rendering a custom element which is a modal dialog. Everything works fine in the initial render until I mutate any state in a context wrapping the app. I mutate the state in the modal and the **deep child elements** get **pulled out to the top level** within the modal.

**DOM during Initial Render:**

<img width="1298" alt="image" src="https://user-images.githubusercontent.com/25875367/109934031-0dc6e300-7cf2-11eb-9684-c374ad9a4a30.png">

**DOM on Re-render:**

<img width="1293" alt="image" src="https://user-images.githubusercontent.com/25875367/109934140-2fc06580-7cf2-11eb-9813-8fae0f824067.png">

[Example using Preact 10.5.10](https://torpid-sedate-bakery.glitch.me/) - _Works fine_
[Source Code](https://glitch.com/edit/#!/accessible-silken-jade)

[Example using Preact 10.5.11](https://accessible-silken-jade.glitch.me/) - _Facing issue_
[Source Code](https://glitch.com/edit/#!/accessible-silken-jade)

MaximePlancke:
We are also facing the same problem. Moving back to 10.5.10 solved it. We haven't tested on 10.5.13 yet

ludvigeriksson:
Tested & broken on 10.5.13 as well

Aswathprabhu:
[...]

PR #3210, opened 2021-07-01
(fix) - prevent eager child removal

JoviDeCroock:
fixes: https://github.com/preactjs/preact/issues/3045


src/render.js
```diff
@@ -48,7 +48,7 @@ export function render(vnode, parentDom, replaceNode) {
 			? [replaceNode]
 			: oldVNode
 			? null
-			: parentDom.firstChild
+			: parentDom.childNodes.length
```
developit, line 51: this seems equivalent

developit (approved):

andrewiggins:
Nice! Can we also add a test for this case?

JoviDeCroock:
@andrewiggins good point

PR #3217, opened 2021-07-04
make 'null' default initial value in useRef hook

### LINKED ISSUE #3216: `null` should be the default of useRef() hook

I don't know what version of `preact` and `typescript` I had before but since I upgraded to `preact` 10.5.14 and `typescript` 4.3.5 suddenly I was getting...

```
ERROR in /Users/peterbe/dev/GROCER/groce/src/routes/list/list-item.tsx(134,15):
TS2322: Type 'Ref<HTMLInputElement | undefined>' is not assignable to type 'Ref<HTMLInputElement> | undefined'.
  Type 'Ref<HTMLInputElement | undefined>' is not assignable to type 'RefObject<HTMLInputElement>'.
    Types of property 'current' are incompatible.
      Type 'HTMLInputElement | undefined' is not assignable to type 'HTMLInputElement | null'.
        Type 'undefined' is not assignable to type 'HTMLInputElement | null'.
```

The solution was simple to type but not obvious when I started :)
```diff
-  const textInputRef = useRef<HTMLInputElement>();
-  const descriptionInputRef = useRef<HTMLTextAreaElement>();
-  const groupInputRef = useRef<HTMLInputElement>();
+  const textInputRef = useRef<HTMLInputElement>(null);
+  const descriptionInputRef = useRef<HTMLTextAreaElement>(null);
+  const groupInputRef = useRef<HTMLInputElement>(null);
```

I got it from: https://github.com/preactjs/preact/issues/2268

What I would expect is to be able to omit the `null` and make that the default initial value. 

[My own comment here](https://github.com/preactjs/preact/issues/2268#issuecomment-873616349)

[Relevant code today](https://github.com/preactjs/preact/blob/ba8353cf3a105e45958168525c496902ab8524b6/hooks/src/index.js#L189-L192)

I think

```js
const textInputRef = useRef<HTMLInputElement>();
```
is a nicer and more intuitive API because, "of course", the ref hasn't been assigned yet.

peterbe:
Fixes #3216

I would *loove* some guidance on how to write a unit test for this. Hence the Draft state.

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -186,7 +186,7 @@ export function useLayoutEffect(callback, args) {
 	}
 }
 
-export function useRef(initialValue) {
+export function useRef(initialValue = null) {
 	currentHook = 5;
 	return useMemo(() => ({ current: initialValue }), []);
 }


JoviDeCroock:
This would deviate from the expected behavior and break React compatability https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v16/index.d.ts#L1022

developit:
We could address this via Types instead - have the default value be `null | undefined` (which are considered equivalent internally).

peterbe:
> This would deviate from the expected behavior and break React compatability https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v16/index.d.ts#L1022

[You're darn right!](https://codesandbox.io/s/eager-pasteur-1o0bl?file=/src/App.tsx)
Screen Shot 2021-07-04 at 2 25 37 PM

And now, on second look, I notice my React projects have things like `const inputRef = React.useRef<null | HTMLInputElement>(null);` but even there I think you can get away with spelling out the type as `<HTMLInputElement>` if you still use the `null` as an initial value to the constructor. 

So, close this (and the issue I filed first)?



> We could address this via Types instead - have the default value be `null | undefined` (which are considered equivalent internally).

Not sure what you're referring to here.

JoviDeCroock:
> We could address this via Types instead - have the default value be `null | undefined` (which are considered equivalent internally).

This would also deviate from the stable types that we share with React now

PR #3222, opened 2021-07-08
Suggest fix for useRef-types in hooks

JoviDeCroock:


robertknight:
What is the rationale for these changes? Is there a related issue?

tomasklaen:
This originated from slack. The issue is that `Ref` types are broken since 10.5.14:

image

This is because components expect `Ref<T | null>`, but the `useRef` hook overload with no params is typed as:

```ts
declare function useRef<T>(): Ref<T | undefined>;
```

The quickest fix for this would be to just switch the `undefined` for `null`, which is how I'm currently patching the preact module types in my project.

But lets make this an opportunity to fix the whole preact Ref typing situation. As developit said it on slack, there are currently 4 or 6 different Ref types spread over `preact` and `preact/hooks` type definitions.

The ideal situation would be to have just one set in stone `Ref` type with no side effects or magic:

```ts
type Ref<T> = {current: T};
```

And since it's implemented in core, put it into `preact` type definitions, and then just import it inside `preact/hooks`.

No special Refs with `null` or `readonly`. That just makes things more complicated, confusing, and harder to implement Ref consuming interfaces.

Also, although `readonly` would be useful for Ref objects intended to be passed into component's `ref` attribute, you just can't make this assumption beforehand, because we don't know where the Ref object that is being created is going to end up or how it's going to be used. This will lead to people confused by why they can't edit their Ref object, which should be mutable.

JoviDeCroock:
@tomasklaen these are similar types to other frameworks implementing these paradigms (can be seen in React as well) I would stick to the deterministic nature of `readonly` for dom-nodes atleast as the assumption can be struck that we'll start out `undefined` and let an external actor take care of it.

tomasklaen:
I see that this is pretty much 1:1 copy of how react defines their types, and compatibility with react is definitely a good thing.

I've also realized you can make non initialized refs mutable by adding `| null` to `T`, which addresses my issues. So no more objections from me :)

But if we care about compatibilty with react, the interfaces should be `RefObject` and `MutableRefObject`. If we don't, they should be `Ref` and `MutableRef`. Right now they are a mix of both.

Oh and what about consolidating all refs from both preact's main definition file and hooks definitions into these 2, defined in core, and imported into hooks? Or do we not care that core has its own different Ref interfaces?

hooks/src/index.d.ts
```diff
@@ -46,8 +46,14 @@ export function useReducer<S, A, I>(
 ): [S, (action: A) => void];
 
 /** @deprecated Use the `Ref` type instead. */
-type PropRef<T> = { current: T };
-type Ref<T> = { current: T };
+type PropRef<T> = MutableRefObject<T>;
+interface Ref<T> {
+	readonly current: T | null;
+}
```
rschristian, line 52: Are we able to instead (slightly) alter `RefObject` from core?

https://github.com/preactjs/preact/blob/4aaecc18d33a9f3139ddd5f30f6d7c745076795f/src/index.d.ts#L40

If not, super nit: `RefObject` is now unused, doesn't need to be imported

https://github.com/preactjs/preact/blob/4aaecc18d33a9f3139ddd5f30f6d7c745076795f/hooks/src/index.d.ts#L1

`Ref` & `MutableRef` or `RefObject` & `MutableRefObject` would be nice for consistency's sake too, as previously mentioned.

rschristian (approved):

Issue #3171, opened 2021-06-03
Preact 10.x unnecessarily updates DOM for <select> options


albv:
**Describe the bug**
Preact 10.x unnecessarily updates DOM (options's values specifically) for `<select>` element even if it's not changed.

**To Reproduce**
https://codesandbox.io/s/objective-jepsen-g55oi

1. Open Chrome Dev Tools and inspect select element options
2. Click on 'Render' button
3. See that every option's value is updated (highlighted with pink)

**Expected behavior**
DOM is not updated at all

developit:
Yep, this is on the radar. That whole behavior is being removed in Preact 11, as it is a workaround for Internet Explorer 11.

RReverser:
I'm running into this as well - I'm frequently updating a large form from an external source and I see all the `select>option[value]` constantly rerendering:

![image](https://user-images.githubusercontent.com/557590/125535479-ef765df8-817a-426a-978b-a033f5f8a184.png)

Another issue I've noticed - not sure whether related or not - is that if I omit `value` attribute on `option`s (which is perfectly valid per HTML and should default to text content of the `option`), Preact adds it and sets to an empty string.

RReverser:
> I'm running into this as well - I'm frequently updating a large form from an external source and I see all the select>option[value] constantly rerendering:

Also: in Chrome Canary it's actually even worse - opening a dropdown immediately closes, I assume due to `option` changes. On Chrome stable it still works fine-ish.

albv:
@RReverser FWIW I ended up using something like this https://codesandbox.io/s/focused-lichterman-m68bb

JoviDeCroock:
Hmm I guess [this IE11 workaround](https://github.com/preactjs/preact/blob/master/src/diff/children.js#L182-L196) is the biggest issue here where we have to reset value and have it set later through the `diff`. This is not present in 11 but if this is causing issues we should look at an alternative workaround.

PR #3226, opened 2021-07-14
fix bug in chrome where select values continuously rerender

JoviDeCroock:
Resolves https://github.com/preactjs/preact/issues/3171

If this poses issues in IE11:

```js
options.diffed = vnode => {
  if (vnode.type === 'option') {
    const props = vnode.props;
    if (!('value' in props)) {
      vnode.dom.value = vnode.dom.textContent;
    }
  }
}
```

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -180,21 +180,7 @@ export function diffChildren(
 				);
 			}
 
-			// Browsers will infer an option's `value` from `textContent` when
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):

developit (approved):

developit:
Let's make sure to release this as a minor and include the patch/workaround in release notes.

PR #3228, opened 2021-07-16
compat internal types: import and re-export PreactElement

henryqdineen:
Since `PreactElement` is used in this module we should also `import` it instead of only doing the `export {} from`.

https://github.com/henryqdineen/preact/blob/a1309b8e1a51ee51ca93848d69dc30a5d3ad0356/compat/src/internal.d.ts#L25

I understand that these are just internal types that are used for `ts-check` but it was just something that I noticed. It seems like it has been innocuous so far. Thanks!


rschristian (approved):
Sorry this was missed and sat for so long

Issue #2797, opened 2020-10-13
preact rendering <style> tag inside <noscript>, whereas react does not


jaredh159:
I'm not sure if this matters to the library maintainers, or how much compatibility/interop with React is a goal, but I noticed a subtle difference between React and Preact that caused a fairly big blocker trying to switch a Gatsby site over to preact.

Basically, this jsx:

```jsx
<noscript>
  <style dangerouslySetInnerHTML={{ __html: "body { background-color: pink }" }} />
</noscript>
```

Renders differently between React and preact.  In react, you get an **empty** `<noscript>` tag.  Preact renders out the `<style>` tag inside the noscript.  Codepens below.

The use-case here is that I'm trying to use `gatsby-background-image` which adds styles to `noscript` like the example above. With react, nothing renders, but with preact, I get extra style tags which load in un-optimized images, destroying any performance benefit gained from switching to preact.

### Reproduction

Codepen with React:
https://codepen.io/jaredh159/pen/NWrGyPq?editors=1010

Codepan with Preact:
https://codepen.io/jaredh159/pen/mdEeXJg?editors=1010

_Note: you'll have to use the dev tools to inspect to see the `<noscript>` contents, as it's not visible, of course._

### Steps to reproduce

See codepens above ^^^

### Expected Behavior

I would have hoped that preact matched the behavior of react.

### Actual Behavior

preact rendered the contents of the `<noscript>` tag, unlike React.

marvinhagemeister:
That sounds like React isn't following the [HTML-Spec](https://html.spec.whatwg.org/multipage/scripting.html#the-noscript-element) correctly. It states:

> When scripting is disabled, not in a `<head>` element: transparent, but there must be no `<noscript>` element descendants.

Which translates as any of the usual elements is a valid descendant of a `<noscript>` element, even `<style>` tags. The rules are different when the `<noscript>` is inside the `<head>` section of the document. In the linked codesandbox that's not the case.

Looks like React has a bug there. I can't come up for valid reasons to disallow style-tags in that context as it is often used to customize the fallback experience on sites I've seen.

jaredh159:
@marvinhagemeister yeah, i understand your reasoning here.  digging into it a bit more, it seems react made a conscious decision to ignore the contents of the `<noscript>` when they are client-rendering:

[...]

PR #3238, opened 2021-07-30
[compat] Skip rendering <noscript> contents on the client

developit:
This fixes the issue from #2797, and [this issue](https://zenn.dev/d_suke/scraps/f7013c4554ac43) using Next.js's `<Image>` component with Preact.


JoviDeCroock (approved):

marvinhagemeister:
Let's add a test for this 👍

developit:
I also want to find a way to remove the IS_DOM boolean. It should be possible to get that information from which Preact hook is called.

marvinhagemeister:
> I also want to find a way to remove the IS_DOM boolean. It should be possible to get that information from which Preact hook is called.

Random idea: Maybe we could use an options property for that? Something like `!options._server` that we could minify on the client and only set in `preact-render-to-string`

PR #3251, opened 2021-08-23
(restructure) - controlled components

JoviDeCroock:
This is a bit of a new feature a big nuisance in Preact X was that when people had components like this:

```js
import { useState } from 'preact/hooks';

const Input = () => {
  const [value, setValue} = useState('');
  const onInput = e => {
    if (e.target.value.length > 3) return;
    setValue(e.target.value);
  }
  return <input value={value} onInput={onInput} />
}
```

This would transition to an uncontrolled component as the underlying dom-element would update to the new character but there was no trigger to the render cycle for this meaning that the control expressed in this event-handler wasn't respected.


developit:
Question as I peek through here: does this require that a controlled input have an onInput handler in order to run patch() on every keystroke/change? Wondering if that gets blocked by shouldComponentUpdate and reverts back to uncontrolled.

JoviDeCroock:
@developit I'm not entirely sure I understand the question, controlled components as it stands require both a `value` and a notion of a `changeHandler` here being `onInput` if it bails out of an update due to `return;` it restores the previous value as that means a bailout

developit:
Heh - sorry, I typed that out and didn't realize how vague it ended up being 😅

I was under the impression the following would be considered "controlled", and would not permit any user modification of the input:

```js
function X() {
  return <input value="foo" />
}
```

You're saying the change handler is actually required in order to flip the input into controlled mode?

src/diff/mount.js
```diff
@@ -188,6 +188,15 @@ function mountDOMElement(dom, internal, globalContext, commitQueue) {
 		}
 
 		let newHtml, newValue, newChildren;
+		if (internal.type === 'input') {
```
JoviDeCroock, line 191: When we create the new component we see whether it has a `checked` or `value` property, this will make it controlled

src/diff/patch.js
```diff
@@ -205,4 +205,17 @@ function patchDOMElement(dom, newVNode, internal, globalContext, commitQueue) {
 			dom.firstChild
 		);
 	}
+
+	if ('value' in newProps && dom._isControlled) {
```
JoviDeCroock, line 209: After diffing the children we check whether the value got out of sync, if it did we update it. We also update the `_prevValue` to prepare for the next event hitting our controlled component

src/diff/props.js
```diff
@@ -116,6 +116,14 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
  */
 function eventProxy(e) {
 	this._listeners[e.type + false](options.event ? options.event(e) : e);
+	if (this._isControlled && this._prevValue) {
```
JoviDeCroock, line 119: After something potentially altered the state we check if this is a controlled component with a previous value if so set the previous value

marvinhagemeister:
@JoviDeCroock This looks great! 🙌 

> I was under the impression the following would be considered "controlled", and would not permit any user modification of the input:
> 
> ```js
> function X() {
>   return <input value="foo" />
> }
> ```

That's a good case we should cover in our tests 👍

JoviDeCroock:
@marvinhagemeister Yeah the test covers that case (in how far we can do that in JSDom) 😅

marvinhagemeister (approved):
LGMT 👍

EmmanuelOga:
Note: Reverted on https://github.com/preactjs/preact/issues/4008

Issue #1611, opened 2019-05-07
DOM event props capitalization: onFocusIn v. onFocusin v. onfocusin


maranomynet:
I was trying to bind a [`focusin`](https://developer.mozilla.org/en-US/docs/Web/API/Element/focusin_event) event handler to an element, and found out that the only way to do it was to use an all-lower case prop `onfocusin`.

Neither `onFocusin` nor `onFocusIn` work.

I'm not sure if this is a Preact 10 beta related issue, or simply a documentation issue.

Are, perhaps, camelCased event prop-names strictly a React-compatibility feature, and all-lower case event props encouraged by Preact?

If so, that fact should be clearly mentioned under [What's missing?](https://preactjs.com/guide/differences-to-react#what-s-missing-) or [What's Different?](https://preactjs.com/guide/differences-to-react#what-s-different-) in the Preact Guide.

(Using `preact@10.0.0-beta.1` – Firefox, Chrome)

cristianbote:
Hey @maranomynet, indeed, `onfocusin/out` is not _supported_[1]. Also, looking over in `react` it seems as well that it's not supported https://github.com/facebook/react/issues/6410.

As an alternative perhaps you could rely on `focus/blur` events?

[1] The supported part is actually driven by the fact that the event exists on the Global Event Interface. MDN gives a 404 on these ones, even though they are linked on their website.

maranomynet:
Ah, I see now that onFocus does _indeed_ "bubble" like in React! Wonderful!

My initial interpretation of "no synthetic events, just plain browser events" and what must have been a simple coding error with `onFocus` led me to hastily assume I needed to resort to `focusin`/`focusout`.

(...similar to the difference in Preact's `onChange` and React's event of same name)

Out of curiosity: Is there anything technically wrong with useing lower case event props instead of the camelCased ones?
I assume the hyperscript (`h()`) function simply tacks them on the DOM Element like any other JavaScript property, but does it really hurt outside being non-idiomatic?

marvinhagemeister:
@maranomynet no, there is nothing wrong with using lowercased events. In fact we map those 1:1 to native DOM events👍

PR #3257, opened 2021-09-07
Add focusin/focusout to type definitions

### LINKED ISSUE #3186: Add onfocusin/onfocusout to jsx.d.ts
Based on the comments in https://github.com/preactjs/preact/issues/1611, it sounds like the recommendation for using the focusin + focusout DOM events is to use the standard all-lowercase forms (onfocusin/onfocusout) which will result in Preact calling addEventListener for the corresponding events.  This works fine at runtime, but we're running into issues in our TypeScript-only project as these properties are not defined in the list of IntrinsicElement properties.  As a result, we're forced to use workarounds like patching IntrinsicElements ourselves or using the spread operator on the element to sidestep type-checking.

Would it make sense to add these two props to jsx.d.ts to make it easier to use these in TypeScript?  Thanks!

boarwell:
Resolves #3186.
Added onfocusin/onfocusout attributes, all lowercased according to the comments on #1611.

---
Before:
image

After:
image

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -445,6 +445,10 @@ export namespace JSXInternal {
 		// Focus Events
 		onFocus?: FocusEventHandler<Target>;
 		onFocusCapture?: FocusEventHandler<Target>;
+		onfocusin?: FocusEventHandler<Target>;
+		onfocusinCapture?: FocusEventHandler<Target>;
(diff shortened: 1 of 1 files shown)


rschristian (approved):
Awesome, thanks for adding these!

PR #3259, opened 2021-09-08
restrict camelCase prop renaming to non-custom elements

jramanat-oracle:


developit (approved):
This looks good to me @jramanat-oracle! Want to un-mark it as "draft"? That this behavior was being applied to Custom Elements was an oversight, so we should be good to merge.

jramanat-oracle:
Done, thanks for the quick review!

PR #3264, opened 2021-09-15
Fix tests in IE11

developit:


marvinhagemeister:
@developit Executing tests locally prints the following warnings:

```bash
marvinhagemeister ~/dev/github/preact (tests-fix-ie11*) $ yarn test:karma:watch
yarn run v1.22.11
$ cross-env BABEL_NO_MODULES=true karma start karma.conf.js --no-single-run
15 09 2021 22:14:10.573:INFO [esbuild]: Compiling...
 > node_modules/chai/index.mjs:3:7: error: Transforming const to the configured target environment ("es5") is not supported yet
    3 │ export const expect = chai.expect;
      ╵        ~~~~~

 > node_modules/chai/index.mjs:4:7: error: Transforming const to the configured target environment ("es5") is not supported yet
    4 │ export const version = chai.version;
      ╵        ~~~~~
```

Maybe we need to downtranspile `chai` manually before passing it to `esbuild`?

JoviDeCroock (approved):

karma.conf.js
```diff
@@ -326,7 +342,7 @@ module.exports = function(config) {
 			singleBundle: false,
 
 			// esbuild options
-			target: 'es2015',
+			target: 'es5',
```
developit, line 345: This no longer complains about `const` because it is being fed `var` (due to Babel running on all JS files). Setting it to `es5` here just prevents ESbuild from generating ES2015 glue code (for module bindings etc).

src/diff/index.js
```diff
@@ -433,7 +433,9 @@ function diffElementNodes(
 				// despite the attribute not being present. When the attribute
 				// is missing the progress bar is treated as indeterminate.
 				// To fix that we'll always update it when it is 0 for progress elements
-				(i !== dom.value || (nodeType === 'progress' && !i))
+				(i !== oldProps.value ||
```
developit, line 436: This fixes a bug in IE11 where we never apply the value prop for `<option value="X">X<option>`. In that browser, `option.value` is `X` as inferred from the option's text content, however inferred option values can't be selected via `select.value`, so we need to assign to it manually.

This also fixes a potential issue where the text content of an `<option>` is initially some value, but then later changes - if we skip assigning `option.value`, the value-to-text binding remains live instead of being correctly locked to the `value` prop from VDOM.

JoviDeCroock (approved):

PR #3267, opened 2021-09-24
types: use FunctionComponent instead of FunctionalComponent

jung-han:
A type error occurred when using FunctionComponent using `forwardRef`, so this part was changed to use FunctionComponent.
Please let me know if you have any additional corrections to make!


rschristian:
I can't reproduce any type error, might've been fixed at some point in an alternative fashion.

Thanks for the PR though.

PR #3274, opened 2021-09-30
refactor(types): Allow Promise<void> for useEffect callback

### LINKED ISSUE #3239: Type `EffectCallback` – allow async functions?
When asynchronously downloading data in a `useEffect()` callback, it would be nice if we could use an async function (whose return value would be ignored).

```ts
// Currently:
const [data, setData] = useState<null | Data>(null);
useEffect(() => {
  downloadData()
  .then((newData: Data) => {
    setData(newData);
  });
});

// Wish:
const [data, setData] = useState<null | Data>(null);
useEffect(async () => {
  const newData = await downloadData();
  setData(newData);
});
```

Current type:

```ts
type EffectCallback = () => void | (() => void);
```

Suggestion:

```ts
type EffectCallback = () => (void | Promise<void> | (() => void));
```

rschristian:
Closes #3239

Allows for behavior like the following, without TS throwing an error:

```
const [data, setData] = useState<null | Data>(null);

useEffect(async () => {
  const newData = await downloadData();
  setData(newData);
}, []);
```

### DIFF
--- a/hooks/src/index.d.ts
+++ b/hooks/src/index.d.ts
@@ -62,7 +62,7 @@ export function useRef<T>(initialValue: null): RefObject<T>;
 export function useRef<T>(initialValue: T): Ref<T>;
 export function useRef<T>(): Ref<T | undefined>;
 
-type EffectCallback = () => void | (() => void);
+type EffectCallback = () => void | Promise<void> | (() => void);
 /**
  * Accepts a function that contains imperative, possibly effectful code.
  * The effects run after browser paint, without blocking it.


marshtomp:
This PR is ridiculous.
Effect callback shouldn't be asynchronous. In your case, if the component unmount during your async `downloadData`, it will lead to an error.
Async function is also NOT-ALLOWED in React EffectCallback. You may wrap your dangerous, async call in a new function to avoid the TS error.

hooks/src/index.d.ts
```diff
@@ -62,7 +62,7 @@ export function useRef<T>(initialValue: null): RefObject<T>;
 export function useRef<T>(initialValue: T): Ref<T>;
 export function useRef<T>(): Ref<T | undefined>;
 
-type EffectCallback = () => void | (() => void);
+type EffectCallback = () => void | Promise<void> | (() => void);
```
JoviDeCroock, line 65: It looks like React doesn't allow for this https://www.runpkg.com/?@types/react@17.0.21/index.d.ts#901

rschristian: I didn't think it did, though would more lax types from Preact be an issue?

rschristian:
@marshtomp Easy there chief.

Lord knows I don't contribute much here, and I'm a far cry away from familiar with the internals, but this seemed to have a go-ahead on the linked issue.

Definitely not a perfect setup by any means, and it is limited, but can be used effectively. If not for TS types getting in the way, that is.

marshtomp:
It is common to do some async tasks in useEffect, but it is necessary to return a cleanup callback synchronously in most cases.
To my understanding, `type EffectCallback = () => Promise<void>` means do async task without cleanup?
If so, I don't think it's a good change...

hooks/src/index.d.ts
JoviDeCroock: I see what we're aiming to do, the thing that scares me about this is that we "allow" people to invoke race conditions, what if this effect invokes twice due to two renders with two different sets of arguments making this downloadData return two distinct values. What's stopping the user from going into erroneous states? It's implied that the user uses an `AbortController`, ... but it's hard to allow this and then not extensively document what they need to check for.

rschristian: Definitely.

Certainly happy to close this out if you're of the opinion that it's likely to do more harm than good.

I have no arguments for or against, just was scrolling through open issues and came across Jason's comment asking for a PR.

JoviDeCroock: I mean, that's only my opinion, open to anyone else chiming in, just thought I'd leave my reasoning 😄

robertknight:
The "rules of hooks" ESLint package maintained by the React team has specific checks in it that warn you not to do this. See [this lint check](https://github.com/facebook/react/blob/f2c381131fb58c107e6153255bc8d1d6340db506/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js#L103) [1]. As such I think we should avoid making this change.

[1] The warning proposes a workaround which appears to have the same result, but my understanding is that the purpose is to make the developer at least think about the possibility of effects running concurrently and whether they should provide code to eg. cancel existing data fetches when a new one starts.

rschristian:
Sure, and I think that's a great linter check. However, along the same lines of the 'exhaustive-deps' rule, there are valid circumstances in which you could use the above. Linters should raise an eyebrow at this of course, but seeing as how many build tools tie in TS type checking, this just seems like a more strict enforcement of this lint rule.

Should TS types enforce "best practices", rather what's actually possible? Not sure, and I'm not much of a proponent for either side. Just some thoughts.

robertknight:
> Should TS types enforce "best practices", rather what's actually possible?

I view types as generally stronger assertions than lint checks, because they specify the API contract that the library provides to its users. Preact needs to know how to handle the return value of the callback. If the types declare that developers are allowed to return a Promise, then we're effectively promising to keep supporting this in future, or declaring a breaking change otherwise. This may require extra code and complicate future changes.

For library authors, there is also the issue of portability between React/Preact to consider. Keeping the APIs aligned encourages code to be portable, which is helpful for the ecosystem as a whole.

hooks/src/index.d.ts
marvinhagemeister: Agree with @JoviDeCroock . In most client projects people tend to want to add async/await and introduce the mentioned race conditions unknowingly.

Issue #3159, opened 2021-05-17
Preact compat: class is forwarded without className


dios-david:
Hey guys 👋 

I know you already put lot of effort into fixing class/className issues in Preact compat, but I just found another weird thing which doesn't happen when rendering with React.

I prepared a unit test to [demonstrate the issue here](https://github.com/preactjs/preact/commit/57779bb68354fa1083f2b674fcf44b7bb1c7e87d):
```jsx
const Foo = ({ className, ...props }) => {
	return (
		<section className={className}>
			<div {...props} />
		</section>
	);
};

render(<Foo className="bar" />, scratch);
expect(scratch.firstChild.className).to.equal('bar'); // the <section> element
expect(scratch.firstChild.firstChild.className).to.equal(''); // the <div> element
```

This test fails with:

```
FAILED TESTS:
  compat render
    className normalization
      ✖ shout not forward class when className is not forwarded
        Chrome Headless 90.0.4430.212 (Mac OS 10.15.7)
      AssertionError: expected 'bar' to equal ''

      + expected - actual

      -bar
      
    at Context.<anonymous> (compat/test/browser/render.test.js:24144:60)
```

I think this happens because after doing `{ className, ...props }`, `props` becomes something like: `{ class: 'bar' }` which then gets passed to `<div>`.
Then when normalising the props [here](https://github.com/preactjs/preact/blob/master/compat/src/render.js#L189) the `classNameDescriptor` is added to the normalised props because `class != className`, making `{ class: 'bar' }` to be used when rendering that inner `<div>` element.

If I do `<div className="" {...props} />` instead of `<div {...props} />` then it works fine, as `className` takes precedence.
Alternatively, I could do `{ className, class, ...props }` when destructuring the props, but this is a bit hacky.

Do you think there is any better way of dealing with this? Is this something that compat should handle better to match how React behaves?

developit:
Thanks for the detailed repro! I think in order to fix this we will need to make `class` a non-enumerable descriptor when props.className is provided.

saadtazi:
Note: I got the issue while using `createPortal` (which is imoprted from `preact/compat`): here is a [codeSandbox](https://codesandbox.io/s/bug-preact-class-96iyv) to repro: `props` received by the component contains `className` and `class` when `preact/compat` is imported (it doesn't happen if it is not imported). I hope it helps.

PR #3279, opened 2021-10-08
fix className leak

JoviDeCroock:
Fixes https://github.com/preactjs/preact/issues/3159
Fixes https://github.com/preactjs/preact/issues/2582

I honestly think it's highly indeterministic to alter the props passed into a component, generally in React (compat) people will pas in either `class` or `className` and won't expect this to be changed. Introducing an additional prop is only going to confuse people. Note that we did not do this before

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -184,7 +184,7 @@ options.vnode = vnode => {
 		vnode.props = normalizedProps;
 	}
 
-	if (type && props.class != props.className) {
+	if (type && typeof type === 'string' && props.class != props.className) {
 		classNameDescriptor.enumerable = 'className' in props;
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
Awesome 🙌

nullpaper:
As a heads up this breaks Router5 `<Link>` component styles and presumably any React components that are expecting a `className` prop but getting a preact `class`. In short:

```jsx
// Preact `10.5.15`
<Link class='green' />
// -> <a class='green' /> 
// -> WORKS! 🎉

// Preact `10.6.0` (ths PR)
<Link class='green' />
// -> <a />
// -> No class propagation 💥
```

This is a breaking change for a whole class of components.

PR #3298, opened 2021-10-28
Compat: statically re-export hooks

developit:
This only matters for the generated CommonJS bundle, which was previously doing a loop that caused Node's CJS named export inference to fail.


JoviDeCroock:
Closing: v11 removed CJS/UMD builds and already statically re-exports hooks in compat, so the named-export-inference issue is gone.

PR #3315, opened 2021-11-10
(restructure) - remove _component from our hooks implementation

JoviDeCroock:
This is a run-up to removing `Component` from our core, this PR also splits up functional from class components

### DIFF
--- a/compat/src/memo.js
+++ b/compat/src/memo.js
@@ -1,4 +1,4 @@
-import { createElement } from 'preact';
+import { createElement, Component } from 'preact';
 import { shallowDiffers } from './util';
 
 /**
@@ -9,7 +9,16 @@ import { shallowDiffers } from './util';
  * @returns {import('./internal').FunctionComponent}
  */
 export function memo(c, comparer) {
-	function shouldUpdate(nextProps) {
+	function Memoed() {}
+
+	Memoed.prototype = new Component();
+	Memoed.displayName = 'Memo(' + (c.displayName || c.name) + ')';
+	// eslint-disable-next-line
+	Memoed.prototype.render = function(props) {
+		return createElement(c, props);
+	};
+
+	Memoed.prototype.shouldComponentUpdate = function(nextProps) {
 		let ref = this.props.ref;
 		let updateRef = ref == nextProps.ref;
 		if (!updateRef && ref) {
@@ -21,13 +30,8 @@ export function memo(c, comparer) {
 		}
 
 		return !comparer(this.props, nextProps) || !updateRef;
-	}
+	};
 
-	function Memoed(props) {
-		this.shouldComponentUpdate = shouldUpdate;
-		return createElement(c, props);
-	}
-	Memoed.displayName = 'Memo(' + (c.displayName || c.name) + ')';
 	Memoed.prototype.isReactComponent = true;
 	Memoed._forwarded = true;
 	return Memoed;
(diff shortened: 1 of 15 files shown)


src/diff/component.js
```diff
@@ -7,12 +7,132 @@ import {
...
+
+		c.props = newProps;
+		c.context = componentContext;
+		c.constructor = type;
+		c.render = doRender;
+		internal.flags |= DIRTY_BIT;
+		isNew = true;
+	}
+
+	if (
```
JoviDeCroock, line 66: Currently needed because of `compat.memo`

src/diff/component.js
```diff
@@ -7,12 +7,132 @@ import {
...
+
+		if (provider) provider._subs.add(internal);
+
+		c.props = newProps;
+		c.context = componentContext;
+		c.constructor = type;
+		c.render = doRender;
+		internal.flags |= DIRTY_BIT;
+		isNew = true;
+	}
```
JoviDeCroock, line 64: In an ideal world we get rid of `c` entirely but there's a few moving parts still relying on this like context

compat/src/memo.js
```diff
@@ -9,7 +9,16 @@ import { shallowDiffers } from './util';
...
+	function Memoed() {}
+
+	Memoed.prototype = new Component();
+	Memoed.displayName = 'Memo(' + (c.displayName || c.name) + ')';
+	// eslint-disable-next-line
+	Memoed.prototype.render = function(props) {
+		return createElement(c, props);
+	};
+
+	Memoed.prototype.shouldComponentUpdate = function(nextProps) {
```
developit, line 21: Might be worth hoisting this out to a function in module scope so that all `memo()` calls can use the same function reference. Would need to store `comparer` on the instance though I guess, which probably costs too many bytes.

src/tree.js
```diff
@@ -83,16 +84,23 @@ export function createInternal(vnode, parentInternal) {
...
 		_vnodeId: vnodeId,
 		_dom: null,
 		_component: null,
-		flags,
 		_context: null,
 		_depth: parentInternal ? parentInternal._depth + 1 : 0
 	};
 
+	if (typeof type == 'function') {
+		internal.rerender = enqueueRender.bind(null, internal);
```
developit, line 100: I think we could get a bit of a performance boost by skipping this bind:

```js
internal.render = enqueueRender;

// then to enqueue a render for an internal:
x._internal.render(x._internal);
```

(using `render` here should compress better than `rerender`)

src/tree.js
```diff
@@ -83,16 +84,23 @@ export function createInternal(vnode, parentInternal) {
...
+		internal.data = {};
```
developit, line 101: Could compute this the same way the other properties are computed differently for function vs element Internals (on L70):

```diff
function createInternal() {
+  let data;
	// ...
+		if (flags & TYPE_ELEMENT === 0) {
+			data = {};
+		} else if (type === 'svg') {
-		if (flags & TYPE_ELEMENT && type === 'svg') {
	// ...  

	const internal = {
		// ...
+		data,
+		render: enqueueRender,
	}
}
```

developit:
I'm thinking about ideas for how to avoid the `internal.render()` bound thing:

1. `import { render } from 'preact'; render(internal)` to enqueue any internal for re-rendering?
2. `internal.render = enqueueRender`, then `internal.render(internal)` (cheaper because no bind)

compat/src/memo.js
JoviDeCroock: Bytes increase is indeed quite on the higher side 😅

PR #3319, opened 2021-11-13
(restructure) - tree-shaking compat

JoviDeCroock:
This would aim at better supporting our users only using certain parts of compat

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -23,6 +23,10 @@ let oldBeforeUnmount = options.unmount;
 const RAF_TIMEOUT = 100;
 let prevRaf;
 
+// TODO: these options hooks are also still side-effects and we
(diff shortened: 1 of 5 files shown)


hooks/src/index.js
```diff
@@ -23,6 +23,10 @@ let oldBeforeUnmount = options.unmount;
 const RAF_TIMEOUT = 100;
 let prevRaf;
 
+// TODO: these options hooks are also still side-effects and we
```
JoviDeCroock, line 26: I don't think it's possible to do this for hooks 😅

JoviDeCroock:
Last commit causes some discrepancies due to the assumption that we always use preact/compat.<render|hydrate> for a react-like root. I'll alter this 😅

PR #3328, opened 2021-11-16
fix hydrate when we are passing document as the parent

JoviDeCroock:


package-lock.json
```diff
@@ -10562,6 +10562,8 @@
 		},
 		"node_modules/npm-merge-driver/node_modules/ansi-regex": {
 			"version": "2.1.1",
+			"resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
```
JoviDeCroock, line 10565: latest npm seems to alter out lockfile

jfsiii:
@JoviDeCroock I'm interested in this re: https://github.com/remix-run/remix/issues/425. It seems to be the only remaining task. Is this change no longer needed/wanted? Will it be handled in another ticket? Thanks for the work and any information.

JoviDeCroock:
Looking for a more elegant solution to this in all fairness

PR #3330, opened 2021-11-17
(restructure) - avoid potential cold path

JoviDeCroock:


JoviDeCroock:
Perf does not seem impacted 🎉

PR #3331, opened 2021-11-17
Optimization: use replaceChild where possible

developit:
This uses `ParentNode.replaceChild()` in the case where insertion is happening at the last index in a list, since it is not possible for an old DOM node to be left in-position after the inserted node. It does not avoid the `.removeChild()` call for the old DOM node, since that is a cheap noop for already-removed nodes.

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -175,7 +175,8 @@ export function diffChildren(
 					oldVNode,
 					oldChildren,
 					newDom,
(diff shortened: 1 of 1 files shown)


JoviDeCroock:
Closing: micro-opt draft; the rewritten v11 children diff uses insertBefore-based mechanics, so this would need full re-derivation.

PR #3334, opened 2021-11-21
(restructure) - optimize normalization effort

JoviDeCroock:


src/create-element.js
```diff
@@ -88,20 +88,17 @@ export function normalizeToVNode(childVNode) {
...
-		// eslint-disable-next-line valid-typeof
-		typeof childVNode == 'bigint'
-	) {
-		return childVNode + '';
-	} else if (Array.isArray(childVNode)) {
-		return createVNode(Fragment, { children: childVNode }, null, null, 0);
-	}
-
-	return childVNode;
+	return typeof childVNode === 'function' ? childVNode : childVNode + '';
```
developit, line 101: I doubt it will show up much in benchmarks, but we should check if it's worth making the string cast conditional here so that we don't cast `childVNode` when its already a string.

PR #3337, opened 2021-11-22
Hydrate HTML Comments: part 3

developit:
This is an addendum to #3327 that changes how comment/cdata/doctype/etc nodes are handled during hydration.
Instead of skipping over those nodes when performing matching, they are excluded from `exccessDomChildren` entirely.
Also, any insertions performed during diffing will now use the next Element/Text node instead of `.nextSibling`, which could be a comment.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -342,7 +342,7 @@ module.exports = function(config) {
 			singleBundle: false,
 
 			// esbuild options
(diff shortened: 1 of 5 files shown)


JoviDeCroock:
Closing: addendum to the closed #3327; v11 rewrote hydration/comment handling around suspense markers.

Issue #2715, opened 2020-08-25
Infinite loop caused by unstable context value


mischnic:
### Reproduction

https://codesandbox.io/s/preact-batch-state-context-z71k8

```js
// **** works
import { render } from "react-dom";
import * as React from "react";

// **** doesn't work:
// import { render } from "preact/compat";
// import * as React from "preact/compat";

// ----------------------------------------------------------

const Context = React.createContext(null);

function ModalProvider(props) {
	let [modalCount, setModalCount] = React.useState(0);
	let context = {
		modalCount,
		addModal() {
			setModalCount((count) => count + 1);
		},
		removeModal() {
			setModalCount((count) => count - 1);
		}
	};

	return <Context.Provider value={context}>{props.children}</Context.Provider>;
}

export function useModal() {
	let context = React.useContext(Context);
	React.useEffect(() => {
		console.log("useModal mount");

		context.addModal();
		return () => {
			console.log("useModal unmount");
			context.removeModal();
		};
	}, [context]);
}

// ----------------------------------------------------------

function Popover() {
	useModal();
	return <div>Popover</div>;
}

function App() {
	return (
		<ModalProvider>
			<Popover />
		</ModalProvider>
	);
}

render(<App />, document.getElementById("root"));
```

### Steps to reproduce

- Open the codesandbox
- Currently, it's using React which works as expected
- Change the imports to use Preact instead
- An infinite loop caused by `useEffect` and `setState` via the context occurs

### Expected Behavior

No infinite loop (as React)

### Actual Behavior

I think Preact batches updates differently, so the useEffect and useEffect cleanup callback don't cancel out.

marvinhagemeister:
Just had a quick look at this and wrapping the context value with a `useMemo` resolves the infinite loop. This leads me to assume that React either has some infinite loop detection and bails out somewhere, or that they have special handling for objects passed into a context provider.

LXSMNSYC:
@marvinhagemeister, if I remember correctly, React's render phase tries to detect a re-render request and throws an error after a defined number of cycles has been reached before it can go to the commit phase

joeythomaschaske:
@marvinhagemeister Can you provide code to explain what you mean?

mischnic:
(This was the fix that solved it for me: https://github.com/adobe/react-spectrum/commit/78ac497400786ba2572d3ee076a158038e4a409e)

```diff
 export function ModalProvider(props: ModalProviderProps) {
   let {children} = props;
[...]

Issue #3297, opened 2021-10-28
[preact/compat] causes page to stuck when rendering radix-ui Dialog


jackyef:
- [x] Check if updating to the latest Preact version resolves the issue (tested, it doesn't resolve the issue)

**Describe the bug**
When using preact/compat with radix-ui, rendering `Dialog.Content` would cause the page to stuck indefinitely.

**To Reproduce**

[Open this stackblitz project](https://stackblitz.com/edit/nextjs-preact-webpack5-odz79y?file=components/Dialog.js), and click on the "Trigger" button to open the `Dialog`. The page should now be unresponsive.

1. Go to https://stackblitz.com/edit/nextjs-preact-webpack5-odz79y?file=components/Dialog.js
2. Click on the "Trigger" button to open the `Dialog` component
3. The page should now be unresponsive.

**Expected behavior**
The same code works with React. The expectation is that it also works with preact/compat.

nikuda:
Related issue: https://github.com/radix-ui/primitives/issues/863

Though it's not clear which React specific features Radix UI uses.

jas-per:
I think there is a similar issue with the latest version of react-bootstrap (see https://github.com/react-bootstrap/react-bootstrap/issues/6131)

Preact/compat gets into an endless loop when using tooltips in react-bootstrap - from what I see this could be related to the use of portals because the issues with radix-ui dialog and chakra-ui popover seem to be similar to me

would be great to get some feedback on this - having to choose bootstrap5 or preact would be a pity in the long run

JoviDeCroock:
I mean, I don't think we'd implement synthetic events just for portals 😅 to me it feels kinda odd having to leverage DOM-events outside of a VDOM portal

maon-fp:
We ran into this problem when using react-bootstrap, too. @JoviDeCroock If synthetic events are not to be implemented: is there an alternative?

JoviDeCroock:
Fixed in 10.10.1 https://stackblitz.com/edit/nextjs-preact-webpack5-dmpk3a?file=package.json

PR #3348, opened 2021-11-28
Bail on batched useState setter ending in an equal result

JoviDeCroock:
When we batch state updates from `useState` it can end up in an equal result, this is a micro-optimization that a lot of UI-libraries seem to rely on.

In the case of https://github.com/preactjs/preact/issues/2715 we can see that we have an effect that does +1 --> unmount -1 --> remounts +1 which ends up in the same Number value for our `useState` meaning we can bailout of this state update which also relieves us from the infinite loop that follows

Fixes: https://github.com/preactjs/preact/issues/2715

To verify:

- https://github.com/preactjs/preact/issues/3297 (EDIT: does not get fixed)
- https://github.com/chakra-ui/chakra-ui/issues/4397

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -21,7 +21,7 @@ let oldBeforeUnmount = options.unmount;
 const RAF_TIMEOUT = 100;
 let prevRaf;
 
-options._diff = vnode => {
+options._diff = (vnode, oldVNode) => {
 	currentComponent = null;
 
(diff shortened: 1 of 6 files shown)


JoviDeCroock:
Could be good for 11 later on

JoviDeCroock:
After landing #3553 there must be a better way to see equality across batches and cancel the render.

PR #3366, opened 2021-12-09
Move default props to compat

JoviDeCroock:
This moves the implementation of default-props to `preact/compat`, this saves some bytes as generally people resort to doing `({ x= 'y' })` anyway in their functional components.

The TS-types for compat are now slightly altered to stitch `defaultProps` on top


compat/src/render.js
```diff
@@ -210,6 +210,12 @@ options.vnode = vnode => {
 			if (props.className != null) normalizedProps.class = props.className;
 			Object.defineProperty(normalizedProps, 'className', classNameDescriptor);
 		}
+	} else if (type.defaultProps != null) {
```
developit, line 213: I'm wondering if we should guard against type being undefined here? (`createElement(X.MissingProp)`)

developit (approved):

PR #3371, opened 2021-12-09
10.6.4

JoviDeCroock:
This commit will spawn a codesandbox package which we can use to verify whether we fix https://github.com/withastro/astro/issues/2169


natemoo-re:
Wow, thanks so much for digging into this! 🎉

marvinhagemeister (approved):

PR #3376, opened 2021-12-15
[wip] Remove distance limit from keyed matching

### LINKED ISSUE #2783: Unnecessary unmounting/mounting children with keys

developit:
This may fix #2783.


developit:
This confirms that we won't be fixing this in the 10.x codebase, as a proper fix requires two-pass child reconciliation, which is in 11.

PR #3377, opened 2021-12-15
Optimizations 1

developit:


mangle.json
```diff
@@ -34,6 +34,7 @@
   "props": {
     "cname": 6,
     "props": {
+      "$flags": "f",
```
JoviDeCroock, line 37: I would opt to keep this to `flags` to encourage use of library authors for additional extensions that we don't want to support first-party

developit: I've been going back and forth on it. It pains me that this is 23b just for `lags` 👁

src/component.js
```diff
@@ -90,36 +94,37 @@ Component.prototype.forceUpdate = function(callback) {
...
-			internal.ref, // since the mode flag should bypass key/ref handling
-			0
-		);
-
-		const commitQueue = [];
-		patch(parentDom, vnode, internal, commitQueue, startDom);
-		commitRoot(commitQueue, internal);
-	}
-}
+const renderDirtyInternal = internal => {
```
JoviDeCroock, line 100: Let's maybe rename this to renderInternal :p

developit: renderEnqueuedInternal?

src/create-context.js
```diff
@@ -34,11 +29,11 @@ export const createContext = (defaultValue, contextId) => {
 				ctx = {};
 				ctx[contextId] = this;
 				this.getChildContext = () => ctx;
-				providers.add(this._internal);
+				providers.add(this);
 			}
 			// re-render subscribers in response to value change
 			else if (props.value !== this._prev) {
-				this._subs.forEach(enqueueRender);
+				this._subs.forEach(i => i.render());
```
JoviDeCroock, line 36:
```suggestion
				this._subs.forEach(i => { i.render() });
```

Saves some bytes

src/component.js
JoviDeCroock: Yes something like that, it's a nit, the dirty part feels a bit odd 😅

mangle.json
JoviDeCroock: Hmm, the community being able to use this without looking at this file kind off boats well for me but yeah, one opinion 😅 also the types won't ever reflect .f unless we start doing two different type files

src/create-context.js
developit: ha! can't believe I missed that.

developit: @JoviDeCroock would you believe that actually adds 3b?! gzip

mangle.json
developit: true about types too (though that makes me think we should probably already be doing internal VS external files given we have other minified properties)

JoviDeCroock (approved):

src/create-element.js
```diff
@@ -1,4 +1,4 @@
-import { UNDEFINED } from './constants';
+import { UNDEFINED, EMPTY_ARR } from './constants';
```
developit, line 1:
```suggestion
import { UNDEFINED } from './constants';
```

src/create-context.js
cristianbote: Maybe it comes from the `return` keyword added at build time 🤷

Thompson1985 (approved):

PR #3378, opened 2021-12-16
Automate building npm package for release

andrewiggins:
This PR adds a new Release workflow that runs whenever a new tag is created. Upon new tag creation, it'll run a build which produces as an artifact the built npm package for the commit the tag was created on. It then creates a draft release and uploads the built npm package as an asset to the npm package.

This PR also adds a script that will download the npm package from a release and publish it to NPM.

See the changes to our CONTRIBUTING guide to understand how our publish process will change.

NOTE: 2FA is still required for npm publishes. This PR does not automate the publishing a new package, just the building of the repo to create the npm package that still must be manually published.

Unfortunately, I can only test this after this PR is completed so there might be a couple of fast follows after this one to fix bugs 😅

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,7 @@ on:
     branches:
       - master
       - restructure
+  workflow_call:
 
 jobs:
   build_test:
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
 			"require": "./jsx-runtime/dist/jsxRuntime.js",
 			"import": "./jsx-runtime/dist/jsxRuntime.mjs"
 		},
-    "./compat/server": {
+		"./compat/server": {
 			"require": "./compat/server.js",
 			"import": "./compat/server.mjs"
 		},
@@ -66,7 +66,7 @@
 			"require": "./compat/jsx-runtime.js",
 			"import": "./compat/jsx-runtime.mjs"
 		},
-    "./compat/jsx-dev-runtime": {
+		"./compat/jsx-dev-runtime": {
 			"require": "./compat/jsx-dev-runtime.js",
 			"import": "./compat/jsx-dev-runtime.mjs"
 		},
@@ -235,6 +235,8 @@
 	"bugs": "https://github.com/preactjs/preact/issues",
 	"homepage": "https://preactjs.com",
 	"devDependencies": {
+		"@actions/github": "^5.0.0",
+		"@actions/glob": "^0.2.0",
 		"@babel/core": "^7.7.0",
 		"@babel/plugin-proposal-object-rest-spread": "^7.6.2",
 		"@babel/plugin-transform-react-jsx": "^7.7.0",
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -235,6 +235,8 @@
 	"bugs": "https://github.com/preactjs/preact/issues",
 	"homepage": "https://preactjs.com",
 	"devDependencies": {
+		"@actions/github": "^5.0.0",
+		"@actions/glob": "^0.2.0",
```
andrewiggins, line 239: These are only used for typings. Happy to remove them if people prefer to not have the additional devDeps

.github/workflows/ci.yml
```diff
@@ -9,6 +9,7 @@ on:
     branches:
       - master
       - restructure
+  workflow_call:
```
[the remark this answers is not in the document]
developit: TIL there are still whole parts of the YAML syntax I've never seen

package.json
developit: all good.

scripts/release/create-gh-release.js
```diff
@@ -0,0 +1,59 @@
...
+		console.log('No existing release found. Creating a new draft...');
+
+		// No existing release for this tag found so let's create a release
+		// API Documentation: https://docs.github.com/en/rest/reference/repos#create-a-release
+		// Octokit Documentation: https://octokit.github.io/rest.js/v18#repos-create-release
+		const createReleaseRes = await github.rest.repos.createRelease({
+			...context.repo,
+			tag_name,
+			name: tag_name,
+			body: '', // TODO: Maybe run changelogged and prefill the body?
```
developit, line 43: this would be super slick.

scripts/release/publish.mjs
```diff
@@ -0,0 +1,140 @@
...
+	DEBUG = opts.debug;
+
+	log.debug('Git tag:', tag);
+	log.debug('Options:', opts);
+
+	// 1. Find a release with the matching tag
+	/** @type {Release} */
+	let release;
+
+	for await (let releasePage of getReleases()) {
```
developit, line 69: QQ: would it work to use the "get release by tag name" API?
`GET https://api.github.com/repos/preactjs/preact/releases/tags/${tag}`

https://docs.github.com/en/rest/reference/releases#get-a-release-by-tag-name

scripts/release/publish.mjs
```diff
@@ -0,0 +1,140 @@
...
+		packageAsset.browser_download_url,
+		{
+			method: 'GET',
+			maxRedirections: 30
+		},
+		() => fs.createWriteStream(packageAsset.name)
+	);
+
+	// 3. Run npm publish
+	const args = ['publish', packageAsset.name];
```
developit, line 119: I didn't know this was supported! that's super awesome, I love having the separate build/download/publish steps.

developit (approved):

scripts/release/publish.mjs
andrewiggins: Oh! I didn't use this at first cuz it doesn't return draft releases (and I thought we'd need them here), but since we are running this after publishing the release, it'd be much simpler. Good call!

PR #3205, opened 2021-06-24
[test] Restructure skew diff

developit:

PR #3388, opened 2021-12-28
Child diffing with skew based diff

JoviDeCroock:
Updates #3205 to the newest version of restructure

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -2,7 +2,6 @@ import { applyRef } from './refs';
 import { normalizeToVNode } from '../create-element';
 import {
 	TYPE_COMPONENT,
-	TYPE_TEXT,
 	MODE_HYDRATE,
 	MODE_SUSPENDED,
 	EMPTY_ARR,
@@ -12,7 +11,7 @@ import {
 import { mount } from './mount';
 import { patch } from './patch';
 import { unmount } from './unmount';
-import { createInternal, getDomSibling, getChildDom } from '../tree';
+import { createInternal, getDomSibling } from '../tree';
 
 /**
  * Diff the children of a virtual node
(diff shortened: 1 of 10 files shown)


src/diff/children.js
```diff
@@ -56,25 +56,36 @@ export function diffChildren(
...
 
 			// We are mounting a new VNode
-			nextDomSibling = mount(
+			mount(
 				parentDom,
 				childVNode,
 				childInternal,
 				commitQueue,
-				startDom
+				getDomSibling(parentInternal, skewedIndex)
```
JoviDeCroock, line 88: This currently has room for improvement

src/diff/patch.js
```diff
@@ -144,20 +130,24 @@ export function patch(parentDom, newVNode, internal, commitQueue, startDom) {
...
-				nextDomSibling = mountChildren(
+				mountChildren(
 					parentDom,
 					Array.isArray(renderResult) ? renderResult : [renderResult],
 					internal,
 					commitQueue,
-					startDom
+					(internal._flags & (MODE_HYDRATE | MODE_SUSPENDED)) ===
+						(MODE_HYDRATE | MODE_SUSPENDED)
+						? internal._dom
```
JoviDeCroock, line 140: I wonder if the issue lays here with the failed resumed hydration

src/diff/patch.js
```diff
@@ -28,19 +29,14 @@ import { Fragment } from '../create-element';
...
-		if (newVNode !== internal.props) {
-			dom.data = newVNode;
-			internal.props = newVNode;
-		}
-
-		return dom.nextSibling;
+	if (flags & TYPE_TEXT && newVNode !== internal.props) {
+		dom.data = newVNode;
+		internal.props = newVNode;
 	}
```
developit, line 40: This needs to always bail out of patch() for Text nodes, even when they're not updated:


```suggestion
	if (flags & TYPE_TEXT) {
		if (newVNode !== internal.props) {
			dom.data = newVNode;
			internal.props = newVNode;
		}

		return;
	}
```

PR #3396, opened 2022-01-07
Update skipped tests

JoviDeCroock:
We shouldn't be supporting this test as conceptually when swapping unkeyed elements we unmount/remount else there could be some dangerous side-effects in carrying over state

Refs are expected to throw https://codesandbox.io/s/gracious-brook-pi8us?file=/src/App.js


developit (approved):

developit:
LGTM. I could see us altering hydration for out-of-order trees at some point in the future as a way to simplify (in 11 we droped backward seek, but maybe we look at dropping seek entirely). We can update or remove those tests when we make those changes though, better to have coverage of how it works today.

Issue #3373, opened 2021-12-10
useId hook


dios-david:
Usually for accessibility reasons we need to generate IDs for DOM elements and reference those IDs via e.g. `aria-labelledby` attributes on other DOM elements.

Implementing a custom deterministic ID generator is not straightforward:
- If you use random ID generators then most probably IDs will mismatch during hydration, as the IDs generated during hydration will be different than the IDs which were generated on SSR.
- Using a global counter for ID generation might not be SSR-safe, as the counter needs to be unique for each render. Creating a context to store the counter might help tho.

React 18 introduces a [useId hook](https://blog.saeloun.com/2021/12/09/react-18-useid-api#useid-api) which solves the problem of mismatching IDs on SSR/hydration, as the generated IDs are deterministic and based on the node location in the tree which will match on SSR and on hydration (if I get it right).

I think this would be really useful for most of the Preact users as well. What do you think?

kmcaloon:
Same with [Reach UI](https://reach.tech/). Running into errors when trying to switch to using preact/compat.

JoviDeCroock:
I have a working version in https://github.com/preactjs/preact/pull/3583 just need to update some of the server-side libraries to add support for the `_children` and `_parent` properties 😅

JoviDeCroock:
An issue that seems very apparent now coming to think about it is that a large amount of users use React on the server and Preact on the client 😅

mjgerace:
@JoviDeCroock  will you cut a new release soon with these changes?

JoviDeCroock:
@mjgerace https://github.com/preactjs/preact/pull/3583#issuecomment-1236406632 sorry but this is really getting on my nerves

mjgerace:
@JoviDeCroock my apologies! I was just wondering about timeline - as you know, this will enable many folks to use React 18 component libraries safely. Please take the comment more as eagerness to be able to use new features rather than an attempt to annoy or beg. Take your time :)

PR #3402, opened 2022-01-09
useId hook

JoviDeCroock:
Implements https://github.com/preactjs/preact/issues/3373

Currently this hook leverages a combination of the component `_vnodeId` (ensures uniqueness across switching subtrees and roots), the vnode depth (ensures uniqueness within a subtree) and the order of the hook. 

EDIT1:
Replacing the usage of `_vnodeId` and `_depth` with a root-identifier and the dom-depth because in SSR currently there are a lot of cases where a subtree isn't rendered until we're on the client due to it being personal information this could skew `_vnodeId` in our current implementation. Same goes for `_depth` if SSR has an additional or missing FunctionalComponent the `_depth` will be erroneous, we avoid missmatches during hydration so `_domDepth` is far more reliable.

This adds tracking for a rootId and _domDepth

TODO: uniqueness across siblings, this could prove troublesome if siblings are fenced with Fragments

```
Parent
  Fragment -- Child -- useId();
  Fragment -- Child -- useId();
```

In this example both Child's will see a _domDepth of 0 and a parentPosition of 0 so this could produce the same id which is erroneous.

EDIT2: replaced the counter that counted the amount of `useId` invocations scoped to a vnode to be a global counter which will bypass the issue outlined above. 

EDIT3: the uniqueness across siblings has been caught by looking at the index the child has inside of the parentInternal._children, we could now scope the counter to be by vnode to avoid big numbers if we incorporate something like a `_domDepth` unless we want to assume `_depth` to be consistent

This is currently blocked by changes needed to RenderToString


hooks/src/index.js
```diff
@@ -307,6 +311,16 @@ export function useErrorBoundary(cb) {
 	];
 }
 
+
+export function useId() {
+	const state = getHookState(currentIndex++, 11);
+	if (!state._id) {
+		currentIdCounter++;
+		state._id = '' + currentInternal._vnodeId + currentInternal._depth + currentIdCounter
```
JoviDeCroock, line 319: Personally I'm a bit hesitant on leveraging `_vnodeId` as it's less deterministic but I feel we need some root-identifier for `render`. `_depth` scares me a bit as well if components return different component trees (I.e. add some functional components) in server vs client. I think separating into `_domDepth` would resolve this, WDYT?

src/create-root.js
```diff
@@ -11,6 +11,8 @@ import { mount } from './diff/mount';
 import { patch } from './diff/patch';
 import { createInternal } from './tree';
 
+let rootId = 0;
```
JoviDeCroock, line 14: We might want to introduce some randomness here in case there are multiple completely separate Preact installations running, i.e. 2 widgets as those will be two roots with a resetting `_domDepth`

marvinhagemeister: FYI: In devtools I'm using this to sort of namespace the ids of different preact installations

```js
const namespace = Math.floor(Math.random() * 2 ** 32);
```

https://github.com/preactjs/preact-devtools/blob/15762ba29ea1cfeca6964bb09dadf4dc625d2bde/src/adapter/hook.ts#L211

hooks/src/index.js
```diff
@@ -307,6 +310,16 @@ export function useErrorBoundary(cb) {
 	];
 }
 
+
+export function useId() {
+	const state = getHookState(currentIndex++, 11);
+	if (!state._id) {
+		currentIdCounter++;
+		state._id = (currentInternal._rootId + currentInternal._domDepth + currentIdCounter).toString(32)
```
ForsakenHarmony, line 318: Do you know how collision resistant this is?

Also, do we need to worry about preact users using a simple ID (single digit) themselves?

src/create-root.js
ForsakenHarmony: Randomness would break hydration with SSR, though, no?

marvinhagemeister: Right, good point about the randomness!

hooks/src/index.js
JoviDeCroock: Prefixed  with `_P` now

hooks/src/index.js
```diff
@@ -348,6 +351,21 @@ export function useErrorBoundary(cb) {
...
+export function useId() {
+	const state = getHookState(currentIndex++, 11);
+	if (!state._id) {
+		currentIdCounter++;
+
+		state._id =
+			'_P' +
+			currentInternal._rootId +
+			(currentInternal._parent._children.indexOf(currentInternal) +
+				currentInternal._depth +
```
JoviDeCroock, line 363: _domDepth might be more reliable here as sometimes people wrap their ssr with a wrapper for the routing, ...

dvoytenko:
Does this approach support `Suspense`?

JoviDeCroock:
@dvoytenko it does yes as it does not take any global counters, it will rely on the state of the subtree which in suspense is a stable thing to rely on.

A combination of a rootId, the position of the node in the parents children and the amount of useId invocations for a node. In React this implementation is a lot more complex as they bitshift forks of the tree I assume for concurrency issues

dvoytenko:
Is it closed because there's an alternative PR?

PR #3405, opened 2022-01-10
fix (un)mounting of refs

### LINKED ISSUE #3404: Issue 3397 is not fully fixed: childInternal is not updated when ref transitions from non-null to null
Applies to: main branch (11.x) 

To reproduce, use test case included with issue 3397
1) Press 'Render with ref'. <span>Hello World with ref!</span> gets logged as expected
2) Press 'Render with no ref'. Nothing gets logged as expected
3) Press 'Render with ref'. Two messages logged: 'null' and '<span>Hello World with ref!</span>'

In (3), the expected behavior is that only one message should be logged (<span>Hello World with ref!</span>)

I think the issue is that childInternal.ref only gets updated when childVNode.ref is non-null. We should be updating it every time IMO, i.e. the assignment should happen before the 'if' block:

			childInternal.ref = childVNode.ref;
			if (childVNode.ref) {
				if (!refs) refs = [];
				// childInternal.ref = childVNode.ref;
				refs.push(
					oldVNodeRef,
					childVNode.ref,
					childInternal.flags & TYPE_COMPONENT
						? childInternal._component || newDom
						: newDom,
					childInternal
				);
			}

JoviDeCroock:
Fixes #3404 

https://codesandbox.io/s/hopeful-leaf-mci3t?file=/src/App.js

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -113,7 +113,7 @@ export function diffChildren(
 
 		newDom = childInternal._dom;
 
-		if (childVNode.ref || childInternal.ref) {
+		if (childVNode.ref || oldVNodeRef) {
 			if (!refs) refs = [];
 			childInternal.ref = childVNode.ref;
 			refs.push(
@@ -124,8 +124,6 @@ export function diffChildren(
 					: newDom,
 				childInternal
 			);
-		} else if (childInternal.ref) {
-			childInternal.ref = childVNode.ref;
 		}
 
 		go: if (mountingChild) {
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -113,9 +113,9 @@ export function diffChildren(
 
 		newDom = childInternal._dom;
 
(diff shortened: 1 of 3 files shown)


developit (approved):

src/diff/children.js
```diff
@@ -113,9 +113,9 @@ export function diffChildren(
 
 		newDom = childInternal._dom;
 
-		if (childVNode.ref) {
+		if (childVNode.ref || oldVNodeRef) {
```
developit, line 116: Wondering if we even need the conditional here, or could just always assign.

developit (approved):

src/diff/children.js
JoviDeCroock: To me this felt like a performance bottleneck as we keep on piling into the array then for every element, now it's only for elements that once had a ref in the current or prev cycle

PR #3408, opened 2022-01-11
Move `createPortal` to core

JoviDeCroock:
The implementation has grown to a very minimal amount of code so it seems worth moving this to Preact so importing `createPortal` doesn't require all the compat side-effects


developit (approved):
Awesome!

PR #3410, opened 2022-01-14
Update Dependencies

developit:
I didn't update Karma yet, we're 2 versions back and it'll likely break some tests.


JoviDeCroock (approved):

developit:
Waiting for #3425 to be merged before this one so we can check size impact.

PR #3411, opened 2022-01-14
Node 16 for CodeSandbox CI

developit:
Hoping this fixes the jest-worker failures we're getting (from Microbundle + Terser).


JoviDeCroock (dismissed):

JoviDeCroock:
Wrong button 😅 maybe microbundle has an old jest-worker that isn't compatible anymore

PR #3416, opened 2022-01-17
fix effect ordering

### LINKED ISSUE #3415: Hooks order in portals are wrong after 10.6.3
First of all, thank you for such a great library!

**Describe the bug**

In 10.6.3 (and 10.6.4) update hooks order was changed. Now it differs from React (inverted). But the main issue is hooks order with `createPortal`. `createPortal`-level hooks fire before portal's content hooks, which is wrong and a breaking change.

**To Reproduce**

The main concept is to have 2 useEffect hooks in a portal parent and child. Child hook is expected to be fired before the parent.

```jsx
const Modal = ({ children }) => {
  return createPortal(<div className="modal">{children}</div>, document.getElementById('modals'))
}

const ModalButton = () => {
  useEffect(() => {
    // set an event listener
  }, [])

  return <button>Action</button>
}

const ExampleModal = () => {
  useEffect(() => {
    // fire an event
  }, [])

  return (
    <Modal>
      <ModalButton />
    </Modal>
}
```

10.6.2 behaviour: https://codesandbox.io/s/hooks-order-preact-10-6-2-m90mp

Correct order as expected:
```
Child 1 
Child 2 
Child 3 
Parent 
Portal Child 1 
Portal Child 2 
Portal Child 3 
Portal Parent 
Portal
```

10.6.4 behaviour: https://codesandbox.io/s/hooks-order-preact-10-6-4-n4ny3

Unexpected order: 
```
Portal Child 3 
Portal Child 2 
Portal Child 1 
Child 3 
Child 2 
Child 1 
Portal 
Portal Parent 
Parent 
```

As you can see Portal useEffect fires earlier than direct child hook.

**Expected behavior**

Hooks order should be as expected from child to parent in all cases.

JoviDeCroock:
Hooks should run by subtree so consider a tree structured like the following

```
       Parent
     C1      C2
  GC1  GC2     GC3  
```

If all these dispatch effects then we should run: `GC1 GC2 C1 GC3 C2 Parent`, currently the depth sorting is making it so that we execute `GC3 GC2 GC1 C2 C1 Parent` which seems a bit out of band.

fixes #3415 

https://codesandbox.io/s/sweet-zeh-cxn1c?file=/src/App.js
https://codesandbox.io/s/recursing-moore-8bylr?file=/src/App.js

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -45,7 +45,7 @@ options.diffed = vnode => {
 
 	const c = vnode._component;
 	if (c && c.__hooks && c.__hooks._pendingEffects.length) {
-		afterPaint(afterPaintEffects.unshift(c));
+		afterPaint(afterPaintEffects.push(c));
 	}
 	currentComponent = null;
 };
@@ -289,7 +289,7 @@ export function useErrorBoundary(cb) {
  */
 function flushAfterPaintEffects() {
 	let component;
-	while ((component = afterPaintEffects.pop())) {
+	while ((component = afterPaintEffects.shift())) {
 		if (!component._parentDom) continue;
 		try {
 			component.__hooks._pendingEffects.forEach(invokeCleanup);
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -290,10 +290,11 @@ export function useErrorBoundary(cb) {
 function flushAfterPaintEffects() {
 	let component;
 	// sort the queue by depth (outermost to innermost)
-	afterPaintEffects.sort((a, b) => (a._vnode._depth - b._vnode._depth) || (
+	afterPaintEffects.sort(
+		(a, b) =>
+			a._vnode._depth - b._vnode._depth ||
 			a._vnode._parent._children.indexOf(a._vnode) -
-			b._vnode._parent._children.indexOf(b._vnode)
-		);
+				b._vnode._parent._children.indexOf(b._vnode)
 	);
 	while ((component = afterPaintEffects.pop())) {
 		if (!component._parentDom) continue;
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -291,11 +291,11 @@ function flushAfterPaintEffects() {
 	let component;
 	// sort the queue by depth (outermost to innermost)
 	afterPaintEffects.sort((a, b) => {
-		if (a._vnode._depth !== b._vnode._depth)
-			return a._vnode._depth - b._vnode._depth;
-		return (
-			b._vnode._parent._children.indexOf(b._vnode) -
-			a._vnode._parent._children.indexOf(a._vnode)
+		let vnodeA = a._vnode;
+		let vnodeB = b._vnode;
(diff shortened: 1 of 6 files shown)


hooks/src/index.js
```diff
@@ -290,8 +290,15 @@ export function useErrorBoundary(cb) {
...
 	// sort the queue by depth (outermost to innermost)
-	afterPaintEffects.sort((a, b) => a._vnode._depth - b._vnode._depth);
-	while (component = afterPaintEffects.pop()) {
+	afterPaintEffects.sort((a, b) => {
+		if (a._vnode._depth !== b._vnode._depth)
+			return a._vnode._depth - b._vnode._depth;
+		return (
+			b._vnode._parent._children.indexOf(b._vnode) -
+			a._vnode._parent._children.indexOf(a._vnode)
+		);
```
developit, line 299: Possibly smaller/faster:

```suggestion
		let vnodeA = a._vnode;
		let vnodeB = b._vnode;
		return (vnodeA._depth - vnodeB._depth) || (
			vnodeB._parent._children.indexOf(vnodeB) -
			vnodeA._parent._children.indexOf(vnodeA)
		);
```

developit:
my suggestion is ~6b larger in hooks, I'm thinking the intermediary variables for `vnodeA` / `vnodeB` weren't worthwhile.

hooks/test/browser/combinations.test.js
```diff
@@ -300,7 +300,9 @@ describe('combinations', () => {
 		]);
 	});
 
-	it('should run effects child-first even for children separated by memoization', () => {
+	// TODO: I actually think this is an acceptable failure, because we update child first and then parent
+	// the effects are out of order
```
developit, line 304: Agreed - this is updating via a top-level render, which is super uncommon. Normally the child and parent would update via setState or forceUpdate(), which would mean they'd go into the update queue and that queue would flush them inner-to-outer. I wonder if we could keep this test and make it pass by switching to invoking `setState()` on both components rather than `rerender()` of the outer?

JoviDeCroock: Well this is impossible to bypass in tests at least, we render `Parent` but `Child` is memoized meaning we won't enter that. We exit diffing and push this one on the `afterPaintEffects` queue, then we do a new setState and push that one on the queue

marvinhagemeister (approved):
Fantastic work as usual! 🙌

hooks/test/browser/combinations.test.js
andrewiggins: @developit @JoviDeCroock (playing catch up over the last month 😅) This test was added to regress this PR and the related issue: #3354

Are we saying that this test doesn't accurately represent the issue, that we longer think the issue is a bug, or something else?

JoviDeCroock: I would be inclined to label this as an edge case; it's hard to differentiate between these updates because it are not trickle down renders. Fixing this would imply us inverting the render queue completely. This behavior seems inline with React too

PR #3424, opened 2022-01-20
implement currentOwner for emotion/mui

### LINKED ISSUE #3395: Error when trying to use preact X with MUI via UMD
The example is using the latest version (10.6.4)

**Describe the bug**
We are using preact X for our frontend (browser-side) via UMD (and requireJS), and are trying to get MUI running.
It seems to me that `preact/compat` does not add compatibility for `ReactCurrentOwner` from the secret internals (`__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED`), which MUI tries to use and then runs into the following error:
`material-ui.development.js:4510 Uncaught TypeError: Cannot read properties of undefined (reading 'current')`

**To Reproduce**

1. Go to 'https://codepen.io/Lucas-HMS/pen/vYeajrr?editors=1010' with the Dev Tools open
2. See the error in the browser console/debugger
The error happens at the following location: `return ReactElement(type, key, ref, self, source, ReactCurrentOwner.current, props);` (line 4510 in material-ui.development.js for example)

We are using one of the latest chrome versions "96.0.4664.110".

I tried to create a hack to define the missing internal. The error goes away, but no MUI components are rendered afterwards. Preact components work still fine.
You can enable the hack on line 73 (`//"ReactCompatHack",`)

**Expected behavior**
MUI should work with Preact X via UMD

JoviDeCroock:
Emotion uses their own implementation of `jsxDev` that uses currentOwner to wrap the component.

```js
return ReactElement(type, key, ref, self, source, ReactCurrentOwner.current, props);
```

Fixes #3395

Investigating this further this looks like it is because React is inlined https://github.com/facebook/react/blob/main/packages/react/src/jsx/ReactJSXElement.js#L354

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -123,8 +123,7 @@ options.vnode = vnode => {
 			if (IS_DOM && i === 'children' && type === 'noscript') {
 				// Emulate React's behavior of not rendering the contents of noscript tags on the client.
 				continue;
-			}
-			else if (i === 'value' && 'defaultValue' in props && value == null) {
+			} else if (i === 'value' && 'defaultValue' in props && value == null) {
 				// Skip applying value if it is null/undefined and we already set
 				// a default value
 				continue;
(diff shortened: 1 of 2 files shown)


Ninerian:
What happened to these changes? We are now experiencing the some problems with current owner

JoviDeCroock:
@Ninerian check the linked issue, it's an issue with mui vendoring react not a preact concern basically

Aloento:
it's helped

Issue #2789, opened 2020-10-06
Exception in Slate (RTE) when pressing enter or backspace


shhlkien:
I use slate (https://docs.slatejs.org/) with preact, all things in the latest version. The problem begins when I used preact hooks.
Slate provides rich text editor and comes from react. When used it with react, everything is ok, but when used preact, my app was crashed.
I think there is a problem with preact hooks
This problem only occurs in Firefox

### Reproduction
Error screen:
![image](https://user-images.githubusercontent.com/23404165/95208199-fc48e000-0812-11eb-90db-ceaec9795d76.png)

Sandbox: https://codesandbox.io/s/youthful-boyd-z3l36

### Steps to reproduce
Run the app, type something then press `enter` or `backspace`

### Expected Behavior
The Slate editor adds new line or deletes characters

### Actual Behavior
App is crashed

shhlkien:
For the title, I've tested [draft-js](https://draftjs.org/) and the result is same. This is a hooks problem.

sventschui:
This is not related to hooks but rather FF not implementing onBeforeInput but slate relying on it. React ships a rather incomplete polyfill that breaks with spellcheck/autocorrect.

I'll close this as a duplicate of https://github.com/preactjs/preact/issues/1422, have a look at the comment I just left there.

PR #3430, opened 2022-01-22
Normalize CompositionEvent listeners in preact/compat

hpneo:
This PR normalizes `compositionstart`, `compositionupdate` and `compositionend` event listeners in the same way Preact already does for `beforeinput`.

This solves an issue with Slate, where trying to type a character with an accent (like `ú`) causes an exception.

[A similar issue](https://github.com/preactjs/preact/issues/2789) is fixed by this PR (see  https://codesandbox.io/s/pensive-cookies-wpdej). See also https://github.com/ianstormtaylor/slate/issues/3903#issuecomment-887576458


JoviDeCroock (approved):
Nice, this looks good, double checked the event reference and there should be no collisions

PR #3440, opened 2022-01-31
repeat renders that result in a pending state

JoviDeCroock:
When our render-function has a condition that results in us having to repeat this render (can be seen with the `DIRTY_BIT`) then we can safely just reexecute that render. To do this successfully however we need to also dispose pending effects, memo, ... so these are executed correctly with the updated state and have no duplicate effects for our state.

This PR adds reexecution of render function when we have pending state and also adds a detection for this, when a render is repeated we retain `state` but clear memoization/effects

### DIFF
--- a/hooks/src/internal.d.ts
+++ b/hooks/src/internal.d.ts
@@ -7,9 +7,11 @@ import { Reducer } from '.';
 
 export { PreactContext };
 
-export type Internal = PreactInternal & {
-	_component: Component;
-};
+export interface Internal extends PreactInternal {
+	data: PreactInternal.data & {
+		__hooks: ComponentHooks;
+	};
+}
 
 /**
  * The type of arguments passed to a Hook function. While this type is not
--- a/src/diff/component.js
+++ b/src/diff/component.js
@@ -46,13 +46,14 @@ export function renderFunctionComponent(
 	internal.props = c.props = newProps;
 
 	c._internal = internal;
-	// eslint-disable-next-line
-	while (1) {
+	let counter = 0;
+	while (counter < 25) {
+		counter++;
 		internal.flags &= ~DIRTY_BIT;
 		if ((tmp = options._render)) tmp(internal);
 		tmp = type.call(c, c.props, c.context);
 		if (!(internal.flags & DIRTY_BIT)) {
-			break;
+			counter = 25;
 		}
 	}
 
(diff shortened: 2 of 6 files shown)


hooks/src/internal.d.ts
```diff
@@ -7,9 +7,11 @@ import { Reducer } from '.';
 
 export { PreactContext };
 
-export type Internal = PreactInternal & {
-	_component: Component;
-};
+export interface Internal extends PreactInternal {
+	data: PreactInternal.data & {
+		__hooks: ComponentHooks;
```
developit, line 12: We could probably have this live in the base Internal interface.

src/diff/component.js
```diff
@@ -45,12 +45,16 @@ export function renderFunctionComponent(
...
 	internal.props = c.props = newProps;
 
-	if ((tmp = options._render)) tmp(internal);
-
-	internal.flags &= ~DIRTY_BIT;
 	c._internal = internal;
-
-	tmp = type.call(c, c.props, c.context);
+	// eslint-disable-next-line
+	while (1) {
```
developit, line 50: Could L54 be merged into this loop condition?


```suggestion
	while (internal.flags & DIRTY_BIT) {
```

The dirty bit will be set on the first iteration, then the first line will unset it, and the loop will exit unless state is updated.

JoviDeCroock: Well components subject from trickle down updates won't be dirty as their state isn't set, they are just children from a state-updated parent which means they would never enter this clause

jridgewell: This also needs a count check (react uses 25) to break an infinite loop.

hooks/src/internal.d.ts
marvinhagemeister: Don't mind really having it here. If we're super strict about our addons it's the right place to put it here.

src/diff/component.js
```diff
@@ -45,12 +45,17 @@ export function renderFunctionComponent(
...
-
-	tmp = type.call(c, c.props, c.context);
+	let counter = 0;
+	while (counter < 25) {
+		counter++;
+		internal.flags &= ~DIRTY_BIT;
+		if ((tmp = options._render)) tmp(internal);
+		tmp = type.call(c, c.props, c.context);
+		if (!(internal.flags & DIRTY_BIT)) {
+			counter = 25;
```
JoviDeCroock, line 56: We could opt to also unset the dirty bit here as we then completely break out of the loop

marvinhagemeister (approved):
Let's gooooooooooo 💯

PR #3442, opened 2022-02-01
verify perf of creating vnodes as classes

JoviDeCroock:


src/create-element.js
```diff
@@ -37,6 +37,19 @@ export function createElement(type, props, children) {
...
 }
 
+function VNode(type, props, key, ref, original) {
+	// V8 seems to be better at detecting type shapes if the object is allocated from the same call site
+	// Do not inline into createElement and coerceToVNode!
+	this.type = type;
+	this.props = props;
+	this.key = key;
+	this.ref = ref;
+	this.constructor = undefined;
```
developit, line 47: this doesn't work - need to change the `vnode.constructor == null` check elsewhere to look for `vnode.constructor === VNode` (likely what's causing the regression)

PR #3443, opened 2022-02-04
Avoid rerunning effects/memo's from uncommited renders

JoviDeCroock:
We reset all the effects/memo's rather than just the select scheduled ones, we also eagerly upgraded memo-values


marvinhagemeister (approved):
Fantastic work as always 🎉

PR #3445, opened 2022-02-10
make sure compat is triggered in new jsx runtime

Austaras:
`preact` relies on `compat/render` to be required/imported to generate compatible jsx object to react. This is generally enough, however as the current esm/cjs transistion, sometimes two copy of preact will exist in runtime for each kind of module. If `compat` is only imported once, everthing compat related would be broken.

This pr is not enough for compeletely fix this issue, but would make preact working when someone only import the new jsx function is imported. This is part of fix for gatsbyjs/gatsby#34783

### DIFF
--- a/compat/jsx-dev-runtime.js
+++ b/compat/jsx-dev-runtime.js
@@ -1,1 +1,3 @@
+require('preact/compat');
+
 module.exports = require('preact/jsx-runtime');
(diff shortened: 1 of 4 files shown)


developit (approved):

developit:
It might also be possible to ship this:

```js
import { createElement, Fragment } from 'preact/compat';

function jsx(type, props, key, __source, __self) {
  const vnode = createElement(type, props);
  vnode.key = key;
  vnode.__source = __source;
  vnode.__self = __self;
  return vnode;
}

export { jsx, jsx as jsxs, jsx as jsxDEV, Fragment };
```

marvinhagemeister (approved):

marvinhagemeister:
@Austaras Thank you so much for making a PR for this 🎉

PR #3448, opened 2022-02-13
Fix cursor reset in Safari in input field

### LINKED ISSUE #1502: [10.0.0-alpha.2] Safari resets cursor in controlled form inputs
Haven't dived in to see what's going on at all yet, but Safari is not happy with controlled form field components under Preact X. When typing, the cursor is constantly reset to the end of the form field:

1. Place your cursor in the middle of a controlled `<input>`
2. Type several characters

## Expected results

All of the characters appear at the point where cursor was placed. This works correctly in Chrome and Firefox.

## Actual results

In Safari, the first character is inserted where the cursor started, but [presumably when the element is re-rendered] then the browser moves the cursor to the end of the field and the remaining characters end up there instead.

## Workaround

None found yet (sticking with controlled components anyway). For now we are directing the small audience of this code to use Chrome/Firefox.

## Details / sample code

This manifests with a basic controlled input, e.g.:

```
const Form = () => {
  let [value, setValue] = useState("Type between →← those.");
  return html`<form>
    <input size="100"
      value=${value}
      onInput=${evt => {
        setValue(evt.target.value);
      }}
    />
  </form>`;
};
```

Full demo available at https://codesandbox.io/s/jj24wqyn0y

marvinhagemeister:
This PR partially reverts 09c30063cb18d0a1f6e2135f5c45381566dfa22b which lead to a regression in Safari. Locally verified in IE11, Safari, Firefox and Chrome.

Fixes #1502 (a second time) .

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -433,9 +433,9 @@ function diffElementNodes(
 				// despite the attribute not being present. When the attribute
 				// is missing the progress bar is treated as indeterminate.
 				// To fix that we'll always update it when it is 0 for progress elements
-				(i !== oldProps.value ||
-					i !== dom.value ||
-					(nodeType === 'progress' && !i))
+				// - Don't compare with `oldProps` value, otherwise Safari
+				//   resets input cursor position #1502
+				(i !== dom.value || (nodeType === 'progress' && !i))
 			) {
 				setProperty(dom, 'value', i, oldProps.value, false);
 			}


JoviDeCroock (approved):
Nice one!

developit (approved):
LGTM. I don't recall the reason I changed this, since I didn't apply the same change to the `checked` property handling that immediately follows it.

PR #3450, opened 2022-02-13
Add back IE11 select fix

marvinhagemeister:
@developit Turns out that line in https://github.com/preactjs/preact/commit/09c30063cb18d0a1f6e2135f5c45381566dfa22b was a necessary fix for IE11 after all! It's necessary to get `<select>`-elements to behave properly in IE11 🙈


JoviDeCroock (approved):

developit:
At least we can drop this in 11!

PR #3451, opened 2022-02-13
Add support for <template>-tag

### LINKED ISSUE #3444: The `<template>` tag does not have the correct `content` property

**Describe the bug**
When using Preact to render a `<template>` tag, the template's `content` property remains empty.

The native HTML parser appends all children to the template's `content` property (and the template's `children` remains empty).
But Preact creates DOM nodes by appending them to the template's `children`, so the `content` remains empty.

This also happens in React: https://github.com/facebook/react/issues/19932

**To Reproduce**

Codepen reproducing the issue:
https://codesandbox.io/s/nifty-kowalevski-5syzm?file=/src/index.js

1. Render a `<template>` tag using Preact; eg:
```
render(<template><h1>HELLO</h1></template>, document.getElementById("root"));
```
2. Inspect the rendered `<template>` element in dev tools
3. Using dev tools, check the template's "content" property: `$0.content.children.length` will be 0 (it should be 1).
4. Check the template's "children" property: `$0.children.length` will be 1 (it should be 0).
5. Check the template's innerHTML: `$0.innerHTML` will be an empty string (it should be `"<h1>HELLO</h1>"`)

marvinhagemeister:
Fixes #3444 .

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -303,8 +303,7 @@ function placeChild(
 		newDom != oldDom ||
 		newDom.parentNode == null
 	) {
-		let target =
-			parentDom.nodeName === 'TEMPLATE' ? parentDom.content : parentDom;
+		let target = 'content' in parentDom ? parentDom.content : parentDom;
 
 		outer: if (oldDom == null || oldDom.parentNode !== parentDom) {
 			target.appendChild(newDom);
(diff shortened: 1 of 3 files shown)


JoviDeCroock (approved):

src/diff/children.js
```diff
@@ -303,8 +303,10 @@ function placeChild(
 		newDom != oldDom ||
 		newDom.parentNode == null
 	) {
+		let target = 'content' in parentDom ? parentDom.content : parentDom;
```
marvinhagemeister, line 306: I'm worried that this breaks web components. Need to check that.

Issue #3118, opened 2021-04-19
componentDidCatch should pass the errorInfo parameter.


MasterEric:
**Describe the feature you'd love to see**
Preact's version of componentDidCatch is not compatible with React's variant. The React variant includes a parameter called `errorInfo`; in the latest version of React, this is an object that contains only one string, which contains the component stack trace for the error that was caught.

This is important and useful as it allows improved error reporting for the application, but Preact does not include this functionality by default.

**Additional context (optional)**
See the documentation for Preact: https://preactjs.com/guide/v10/whats-new/#componentdidcatch

Compare with the corresponding documentation for React: https://reactjs.org/docs/error-boundaries.html

el-gringo:
This is required for Sentry JS

https://github.com/getsentry/sentry-javascript/blob/cabb53c67ab202bb12c517f8a4aa65641cac606e/packages/react/src/errorboundary.tsx#L67

adit-hotstar:
When will this fix be released?

PR #3452, opened 2022-02-14
Add support for `errorInfo`

### LINKED ISSUE #2835: componentDidCatch and errorInfo
Hi. I have a question regarding componentDidCatch lifecycle hook. React's componentDidCatch receives two parameters, the first one is an error and the second is errorInfo which contains additional details about the error (like stack trace). I wanted to replace React with preact/compact in my app, however, I didn't manage to do it because I use @sentry/react package and its ErrorBoundry component relies on errorInfo parameter https://github.com/getsentry/sentry-javascript/blob/master/packages/react/src/errorboundary.tsx#L62 so missing `errorInfo`  in preact simply crashes my app.

What's the reason for errorInfo not being present in Preact's componentDidCatch hook? Shouldn't preact/compat be 100% compatible with the React? 

`...preact/compat, which is a thin layer over Preact that attempts to achieve 100% compatibility with React.`

marvinhagemeister:
This PR adds support for the `errorInfo` argument which is the second to `componentDidCatch`. In React it's used to pass the owner tree along. This is often used in combination with [error logging](https://github.com/getsentry/sentry-javascript/blob/d5c3929abc51f05bde1e342cd37274d9aa28c645/packages/react/src/errorboundary.tsx?rgh-link-date=2020-11-18T10%3A51%3A49Z#L70). Note that we'll only collect component traces if `preact/debug` is loaded.

Fixes #2835, #3118  .
Replaces #2836 .

### DIFF
--- a/src/diff/catch-error.js
+++ b/src/diff/catch-error.js
@@ -4,8 +4,10 @@
  * @param {import('../internal').VNode} vnode The vnode that threw
  * the error that was caught (except for unmounting when this parameter
  * is the highest parent that was being unmounted)
+ * @param {import('../internal').VNode} [oldVNode]
+ * @param {import('../internal').ErrorInfo} [errorInfo]
  */
-export function _catchError(error, vnode) {
+export function _catchError(error, vnode, oldVNode, errorInfo) {
 	/** @type {import('../internal').Component} */
 	let component, ctor, handled;
 
@@ -20,7 +22,7 @@ export function _catchError(error, vnode) {
 				}
 
 				if (component.componentDidCatch != null) {
-					component.componentDidCatch(error);
+					component.componentDidCatch(error, errorInfo || {});
 					handled = component._dirty;
 				}
 
--- a/src/internal.d.ts
+++ b/src/internal.d.ts
@@ -20,6 +20,10 @@ export interface DevSource {
 	lineNumber: number;
 }
 
+export interface ErrorInfo {
+	componentStack?: string;
+}
+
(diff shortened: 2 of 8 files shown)


src/internal.d.ts
```diff
@@ -38,6 +38,7 @@ export interface Options extends preact.Options {
 	_skipEffects?: boolean;
 	/** Attach a hook that is invoked after an error is caught in a component but before calling lifecycle hooks */
 	_catchError(error: any, vnode: VNode, oldVNode?: VNode | undefined): void;
```
marvinhagemeister, line 40: TODO: Update this

src/diff/catch-error.js
```diff
@@ -4,8 +4,10 @@
  * @param {import('../internal').VNode} vnode The vnode that threw
  * the error that was caught (except for unmounting when this parameter
  * is the highest parent that was being unmounted)
+ * @param {import('../internal').VNode} oldVNode
+ * @param {{ componentStack?: string }} errorInfo
  */
-export function _catchError(error, vnode) {
+export function _catchError(error, vnode, oldVNode, errorInfo) {
```
marvinhagemeister, line 10: TODO: Update other instances of `options._catchError`

developit (approved):

adit-hotstar:
When will this fix be released?

marvinhagemeister:
@adit-hotstar just published `10.7.0` which includes this PR 👍

PR #3462, opened 2022-02-20
add missing values to mangle

JoviDeCroock:


mangle.json
```diff
@@ -40,6 +40,8 @@
       "$__hooks": "__H",
       "$_list": "__",
       "$_pendingEffects": "__h",
+      "$_pendingArgs": "__a",
```
marvinhagemeister, line 43: These two properties seem to be only used in hooks. Did you encounter them somewhere else? Asking because maybe it was only used to debug the next issue we talked about today or if it fixes an actual issue.

JoviDeCroock: only used in hooks, just noticed we were missing them

developit: IIRC these are only used as properties of internal hookState objects, right?
If so, they don't need `__` prefixes - could just be `a` and `v` (and `_pendingEffects` could be `p`).

PR #3463, opened 2022-02-22
Revise typedefs to support integration

DanielHeath:
I've added these typedefs to get a fair-sized project switched over. Perhaps they're useful to you too?


compat/src/index.d.ts
```diff
@@ -21,7 +21,7 @@ declare namespace React {
 	export import StateUpdater = _hooks.StateUpdater;
 	export import useCallback = _hooks.useCallback;
 	export import useContext = _hooks.useContext;
-	export import useDebugValue = _hooks.useDebugValue;
```
JoviDeCroock, line 24: We probably don't wanna remove this

DanielHeath: Ooh that's not meant to be deleted. For that matter, I don't expect that this PR is necessarily "production grade"; it merely meets my needs.

developit:
This actually looks like its worth landing pretty much as-is, minus the `useDebugValue` removal.

JoviDeCroock:
Closing due to #3534

PR #3464, opened 2022-02-23
Add missing defaultValue and defaultChecked typings

ilogico:
fixes #2857 

These attributes are necessary for working with uncontrolled forms.
Preact already works correctly with these, but the JSX typings are missing.


JoviDeCroock (approved):
Awesome, thank you!

PR #3466, opened 2022-02-25
Bugfix: nested hook state updates return the first state

developit:
This mostly affects useReducer, where dispatching from within the reducer will drop the resulting state update: https://codepen.io/developit/pen/YzEOWGx?editors=0010

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -141,7 +141,7 @@ export function useReducer(reducer, initialState, init) {
 			action => {
 				const prevValue = hookState._value[0];
 				const nextValue = hookState._reducer(prevValue, action);
-				hookState._value[0] = nextValue;
+				hookState._value = [nextValue, hookState._value[1]];
 				if (nextValue !== prevValue) {
 					hookState._component.setState({});
 				}
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -139,9 +139,10 @@ export function useReducer(reducer, initialState, init) {
 			!init ? invokeOrReturn(undefined, initialState) : init(initialState),
 
 			action => {
-				const nextValue = hookState._reducer(hookState._value[0], action);
-				if (hookState._value[0] !== nextValue) {
-					hookState._value = [nextValue, hookState._value[1]];
+				const prevValue = hookState._value[0];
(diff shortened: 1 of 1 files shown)


JoviDeCroock (approved):

hooks/src/index.js
```diff
@@ -139,9 +139,10 @@ export function useReducer(reducer, initialState, init) {
 			!init ? invokeOrReturn(undefined, initialState) : init(initialState),
 
 			action => {
-				const nextValue = hookState._reducer(hookState._value[0], action);
-				if (hookState._value[0] !== nextValue) {
-					hookState._value = [nextValue, hookState._value[1]];
+				const prevValue = hookState._value[0];
+				const nextValue = hookState._reducer(prevValue, action);
+				hookState._value[0] = nextValue;
```
JoviDeCroock, line 144:
```suggestion
				hookState._value = [nextValue, hookState._value[1]];;
```

developit:
I'm no longer convinced this change fixes the linked reproduction. Here's a change that should fix it:

```js

hookState._reducer = reducer;
if (!hookState._component) {
	let actions = [];
	let dispatch = action => {
		if (actions.push(action) !== 1) return;
		let value = hookState._value[0];
		let original = value;
		while (actions.length) {
			value = hookState._reducer(value, actions[0]);
			actions.shift();
		}
		hookState._value = [value, dispatch];
		if (value !== original) hookState._component.setState({});
	};
	hookState._value = [
		!init ? invokeOrReturn(undefined, initialState) : init(initialState),
		dispatch
	];
	hookState._component = currentComponent;
}
```

andrewiggins:
If I'm reading the codepen correctly, I thought dispatching from within a reducer was explicitly disallowed. Academically, reducers are suppose to be pure without side effects I believe. More practically I prefer the simplicity of being able to reason about reducers without having to worry about nested dispatches. I think it keeps our code lean and clean and pushes consumers to similar patterns that are easier to reason about.

Typically to share code between action handlers in reducers I've just created helper functions that can used in both places. Or if I need to do some sort of side effect in two actions I'd put that in a helper and invoke it the same place I call dispatch. Wrapping useReducer and dispatch is another way to share that code.

JoviDeCroock:
Closing due to inactivity

Issue #3218, opened 2021-07-05
[Need help]Cannot read property 'animationstartfalse' of undefined


Stupidism:
- [ ] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
[Our sentry reported this issue](https://sentry.io/share/issue/809a4ca8632d4d7a959809a8082d3367/). And when I search the error message I got 

![image](https://user-images.githubusercontent.com/5801015/124416617-f4aa6d80-dd89-11eb-8d13-a72c67503604.png)

**To Reproduce**

I have no idea how to reproduce at all. I might need help optimizing the sentry config before I can find how to reproduce it.

**Expected behavior**

marvinhagemeister:
Can you follow the stack trace and share the snippet that caused this issue? Hard to tell from the sentry screenshot what is going on as it just show _that_ an error happened, but knowing that an error occured is not enough information to debug it.

Useful information that can help debug this issue is. Which component triggered the error? How does it look like? Where are you using `animationstartfalse`?

JoviDeCroock:
My hunch would be that it happens [here](https://github.com/preactjs/preact/blob/master/src/diff/props.js#L153) but `DOM._listeners` being `undefined` sounds quite.... exotic

Stupidism:
> My hunch would be that it happens [here](https://github.com/preactjs/preact/blob/master/src/diff/props.js#L153) but `DOM._listeners` being `undefined` sounds quite.... exotic. Are you switching the event-handler somehow dynamically, the only way to get into `eventProxy` is [here](https://github.com/preactjs/preact/blob/master/src/diff/props.js#L92-L103) where we set the `dom._listeners` unless we somehow remove & replace that element properties in-place

Thanks for your reply. Your guess looks very close and I understand I must provide more info. For now, I can only get two more info:

1. The error happens only on two machines `35.229.126.86` and `35.203.150.252`. By looking up their IPs, I found they belong to Google Cloud, so I think they must be google spiders.

![image](https://user-images.githubusercontent.com/5801015/126449903-33416f4e-b161-4f7b-b4ba-24a2f47a845e.png)


![image](https://user-images.githubusercontent.com/5801015/126449135-b6a185b9-5dc3-4caf-bf4c-63ee50d6bcee.png)


2. The error only occurs on two pages:

![image](https://user-images.githubusercontent.com/5801015/126450181-ec73d8c0-eac6-4251-b2a2-e2b5df3a86ad.png)


[...]

PR #3473, opened 2022-03-08
Fix event proxy exception

michael-graves-dojo:
Prevent event proxy functions from throwing exceptions in the rare case _listeners is not set

Fixes issue #3218


rasapetter:
Any updates here? The fix seems sane.

Thompson1985 (approved):

JoviDeCroock:
Closing: duplicate of the fix already shipped for #3218 — the `if (this._listeners)` guard is present in `src/diff/props.js`.

PR #3475, opened 2022-03-09
Improve karma test console logging

marvinhagemeister:
- Fix `Error` objects not being serialized for kamra cli logs
- Don't access getters during serialization and just print the getter name instead (causes an infinite loop in #3474)


JoviDeCroock (approved):

developit:
Woot! This was bugging me last week.

PR #3478, opened 2022-03-14
Track context on rerenderState

JoviDeCroock:
This introduces a global called `rendererState` which we can use to track metadata for a root/ongoing render/.... We could expand this to track `_parentDom`, `commitQueue`, ...

Ideally we scope `_commitQueue` by render-root as well as context/... in case we ever decide to go on to async rendering


marvinhagemeister (approved):
I like where this is a lot! A shared object for keeping the current root state around is the right way forward 💯

PR #3480, opened 2022-03-15
Move `preact-render-to-string` to `preact/server`

marvinhagemeister:
This PR merges the https://github.com/preactjs/preact-render-to-string repo into core under `preact/server`. All commits from that repo are preserved and all tags have been renamed to `server-x.x.x`.

On top of that there are a few commits that integrate the tooling into our main one. It has not been updated for internals yet, so I've skipped a few tests. The tests can be run via `npm run test:mocha:server`


server/README.md
```diff
@@ -0,0 +1,102 @@
+# preact-render-to-string
```
JoviDeCroock, line 1: We should probably redo this `preact/server` and maybe something with the logos

server/package.json
```diff
@@ -0,0 +1,30 @@
+{
+	"name": "preact-server",
+	"amdName": "preactServer",
+	"version": "5.1.20",
```
JoviDeCroock, line 4: are we publishing this separately or are we going to keep this at its own version

marvinhagemeister: I think we can safely ignore this as the plan is to publish this as part of core and not as a separate package. The other addons have bogus version numbers to which don't match the same package. So far we haven't gotten an issue for that so I'm assuming that no tool reads it.

server/README.md
marvinhagemeister: Good point, I've updated the title and removed the badges. In the longterm we might want to think about removing it or adding readme's for the other addons too. But that's something for later.

JoviDeCroock (approved):

PR #3482, opened 2022-03-17
[ts] components can return ComponentChildren

developit:
Previously `FunctionComponent` was only typed as returning `VNode | null`, but Preact supports returning Arrays as well as any VDOM text primitive (boolean, etc). Class component render methods were typed as returning `ComponentChild`, which doesn't reflect the fact that this method can also return Arrays.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -87,7 +87,7 @@ export type ComponentProps<
 	: never;
 
 export interface FunctionComponent<P = {}> {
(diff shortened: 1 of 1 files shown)


JoviDeCroock (approved):

marvinhagemeister (approved):

marvinhagemeister:
Looks like this breaks JSX types.

Thompson1985 (approved):

PR #3487, opened 2022-03-22
reset useImperativeHandle ref to null when the component get unmounted

deadem:
When the component get unmounted useImperativeHandle hook should reset ref to null state.

Example:
```js
const Component = forwardRef((props, ref) => {
  useImperativeHandle(ref, () => ({}));
  return <div>Test</div>;
});

  const [mode, setMode] = useState(true);
  const ref = useRef();

  function onClick() {
    setMode(!mode);
  }

  return (
    <Fragment>
      <h1 onClick={onClick}>
        click me. value should change {+!!ref.current + 0}
      </h1>
      {mode ? <Component ref={ref} /> : <div>empty</div>}
    </Fragment>
  );
}
```

[Sandbox React](https://codesandbox.io/s/sleepy-dream-snkpyu)
[Sandbox Preact](https://codesandbox.io/s/frosty-cloud-z9crui)

[React reset ref.current to null on unmount event](https://github.com/facebook/react/blob/d5f1b067c8bbb826b823d0354a28ba31078b70c0/packages/react-reconciler/src/ReactFiberHooks.new.js#L1781) but Preact does not.

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -197,8 +197,13 @@ export function useImperativeHandle(ref, createHandle, args) {
 	currentHook = 6;
 	useLayoutEffect(
 		() => {
-			if (typeof ref == 'function') ref(createHandle());
-			else if (ref) ref.current = createHandle();
+			if (typeof ref == 'function') {
+				ref(createHandle());
+				return () => ref(null);
+			} else if (ref) {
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):
This looks great, thank you!

Issue #3483, opened 2022-03-17
forwardRef((props, { refA, refB }) => { ... }) not working (preact + preact-compat)


baptistecs:
- [x] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
Passing an object of refs to forwardRef is working with React but not on Preact + Preact/Compat
Which is the only bug that prevent us to migrate to Preact :'(

**To Reproduce**
- Using refs in prop is working: https://codesandbox.io/s/refs-in-prop-working-opqe05?file=/components/app.js
- Using forwardRef for one ref is working: https://codesandbox.io/s/forwardref-ref-working-j1mvzh?file=/components/app.js
- Using forwardRef for refs is not working: https://codesandbox.io/s/forwardref-refs-not-working-i9mumy?file=/components/app.js

Check CodeSandbox logs

**Expected behavior**
CodeSandbox logs should return `{A: "everyone", B: "one"}` but it returns `{A: undefined, B: undefined}`

baptistecs:
🤩 Thank you @JoviDeCroock for the fix!

PR #3494, opened 2022-03-28
allow for passing in multiple refs through forwardRef

JoviDeCroock:
Fixes https://github.com/preactjs/preact/issues/3483

We check for the absence of keys on the object as that is a better heuristic, when no ref is passed in we'll have an empty object (componentContext/state) which we checked for by doing `current in ref`

### DIFF
--- a/compat/src/forwardRef.js
+++ b/compat/src/forwardRef.js
@@ -33,7 +33,9 @@ export function forwardRef(fn) {
 		ref = props.ref || ref;
 		return fn(
 			clone,
-			!ref || (typeof ref === 'object' && !('current' in ref)) ? null : ref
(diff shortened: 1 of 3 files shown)


marvinhagemeister (approved):
Good catch!

compat/src/forwardRef.js
```diff
@@ -33,7 +33,9 @@ export function forwardRef(fn) {
 		ref = props.ref || ref;
 		return fn(
 			clone,
-			!ref || (typeof ref === 'object' && !('current' in ref)) ? null : ref
+			!ref || (typeof ref === 'object' && Object.keys(ref).length === 0)
```
developit, line 36: do we even need to guard this at all? we could just check that it's an object and non-null.

PR #3497, opened 2022-03-29
Simplify forwardRef

developit:
The (re)use of the second argument here seems like it may have been to accommodate `forwardRef(forwardRef())`, but that is something that would be considered invalid usage of fowardRef - forwardRef creates a component that calls the given function (technically not function component) with `(props, ref)`, whereas a function component expects `(props, context)`.

### DIFF
--- a/compat/src/forwardRef.js
+++ b/compat/src/forwardRef.js
@@ -24,9 +24,6 @@ export const REACT_FORWARD_SYMBOL =
  * @returns {import('./internal').FunctionComponent}
  */
 export function forwardRef(fn) {
-	// We always have ref in props.ref, except for
-	// mobx-react. It will call this function directly
-	// and always pass ref as the second argument.
 	function Forwarded(props) {
 		let clone = assign({}, props);
 		delete clone.ref;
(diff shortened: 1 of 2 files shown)


compat/src/forwardRef.js
```diff
@@ -27,16 +27,10 @@ export function forwardRef(fn) {
 	// We always have ref in props.ref, except for
 	// mobx-react. It will call this function directly
 	// and always pass ref as the second argument.
-	function Forwarded(props, ref) {
+	function Forwarded(props) {
```
marvinhagemeister, line 30: We might need to double check with `mobx-react` as outlined in the comment for this function if the second argument is still needed.

marvinhagemeister:
Let's add a testcase for this:

```jsx
it('should not leak context into refs', () => {
	class Provider extends Component {
		getChildContext() {
			return { foo: 2 };
		}
		render() {
			return this.props.children;
		}
	}

	let actual;
	const Forwarded = forwardRef((_, ref) => {
		actual = ref;
		return <div />;
	});

	render(
		<Provider>
			<Forwarded />
		</Provider>,
		scratch
	);

	expect(actual).to.equal(null);
});
```

compat/src/forwardRef.js
marvinhagemeister: Just checked, looks like this is not needed anymore 🎉  So the comment can be dropped too 👍

marvinhagemeister (approved):

PR #3498, opened 2022-03-30
add missing stuff for react 18

JoviDeCroock:
These are some preliminary implementations for the v18 methods.

We are still missing useId which has a pending PR to main, useSyncExternalStore can essentially be a state and effect combination and useInsertionEffect can be a useLayoutEffect

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -117,8 +117,29 @@ const flushSync = (callback, arg) => callback(arg);
  */
 const StrictMode = Fragment;
 
+const startTransition = callback => callback();
+const useTransition = () => {
+	const [isPending, setIsPending] = useState(false);
+
+	const start = cb => {
+		setIsPending(true);
+		cb();
+		setIsPending(false);
+	};
+	return [isPending, start];
+};
+
+const useDeferredValue = val => {
+	const value = useRef(val);
+	return value.current;
+};
+
 export * from 'preact/hooks';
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
I think we can add these, I'd just add a few test cases 👍

JoviDeCroock:
@marvinhagemeister thinking about how we could make all of this smaller 😅

compat/src/index.js
```diff
@@ -117,8 +117,58 @@ const flushSync = (callback, arg) => callback(arg);
...
 const StrictMode = Fragment;
 
+const startTransition = callback => callback();
+const useTransition = () => {
+	const [isPending, setIsPending] = useState(false);
+
+	const start = cb => {
+		setIsPending(true);
+		cb();
+		setIsPending(false);
```
gaearon, line 127: I don’t understand what this is meant to do. Won’t all updates be batched here anyway?

compat/src/index.js
```diff
@@ -117,8 +117,58 @@ const flushSync = (callback, arg) => callback(arg);
...
+	};
+	return [isPending, start];
+};
+
+const useDeferredValue = val => {
+	const value = useRef(val);
+	return value.current;
```
gaearon, line 134: This should return `val` instead. The implementation in this PR would forever return the initial value which is wrong.

PR #3508, opened 2022-04-07
add missing containerInfo to portals

### LINKED ISSUE #3507: Portals [React-Compat] missing field and rendering problem
[+] Check if updating to the latest Preact version resolves the issue

React Portals has a property called "containerInfo" which can return the container itself. This is documented somewhere for sure, but I only found out from the Dart docs. https://pub.dev/documentation/react/latest/react_client.react_interop/ReactPortal-class.html
This problem occurred while integrating with this proprietary library: 

https://bryntum.com/docs/scheduler/#  
https://bryntum.com/docs/scheduler/guide/Scheduler/integration/react/guide#rendering-react-components-in-column-cells

Where there is a React wrapper for the components, which uses Portals to move cells around. Relevant snippet:

```js 
if (portal) {
    // Move portal container back to the cell if we have one
    renderElement.appendChild(portal.containerInfo);
    renderElement.portalContainer = portal.containerInfo;
} else {
    // Create new portal container
    ...
    renderElement.portalContainer = _portalContainer; // Create a new portal in the portal container
    portal = ReactDOM.createPortal(reactNode, _portalContainer, portalId); // Add the new portal to Map
```

Here is what React produces:

kép

A portal, with a component inside, with its children.

What Preact produces:

kép

The container is okay, it but the children are missing from it:

kép

Which is also part of the component:

kép

The props children never shows up on the output.

**Expected behavior**

The compat library to incorporate this functions.

** Steps to reproduces ** 

1. Get the 45 day trial, and based of the doc, just override the cell renderer method with JSX

** Partial solution **

During debugging I already managed to link up the missing field with this code snippet, but this is sub optimal.

compat/src/portals.js
```js
export function createPortal(vnode, container) {
	return createElement(Portal, { _vnode: vnode, _container: container, _expose: "containerInfo" });
}
```

compat/src/render.js
```js
let oldVNodeHook = options.vnode;
options.vnode = vnode => {
	let type = vnode.type;
	let props = vnode.props;
	let normalizedProps = props;
	if (vnode.props._expose === "containerInfo") {
		vnode.containerInfo = vnode.props._container;
	}
```

I can do more debugging, just the output of the preact bundle is minified, and thats not ideal tracking whats not working as expected.

Any help would be useful, and would prevent a React switch. :)

JoviDeCroock:
Attempt to fix #3507

### DIFF
--- a/compat/src/portals.js
+++ b/compat/src/portals.js
@@ -76,5 +76,7 @@ function Portal(props) {
  * @param {import('./internal').PreactElement} container The DOM node to continue rendering in to.
  */
 export function createPortal(vnode, container) {
-	return createElement(Portal, { _vnode: vnode, _container: container });
+	const el = createElement(Portal, { _vnode: vnode, _container: container });
+	el.containerInfo = container;
+	return el;
 }
--- a/compat/test/browser/portals.test.js
+++ b/compat/test/browser/portals.test.js
@@ -668,6 +668,29 @@ describe('Portal', () => {
 		expect(calls).to.deep.equal(['Button 1', 'Button 2', 'Modal', 'App']);
 	});
 
+	it('should include containerInfo', () => {
+		let root = document.createElement('div');
+		document.body.appendChild(root);
+
+		const A = () => <span>A</span>;
+
+		let portal;
+		function Foo(props) {
+			portal = createPortal(props.children, root);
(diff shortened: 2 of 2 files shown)


marvinhagemeister (approved):
Awesome 👍

PR #3510, opened 2022-04-08
add missing compat hooks for React 18

JoviDeCroock:
This has a bit more considerations compared to #3498 and tries to account for

- a transition essentially happening after a microtick, we could use other heuristics if someone knows better ones 😅  but yes React hooks into the task queue
- useDeferredValue as rightfully pointed out by @gaearon we do not have to do anything special here (thanks for the feedback again)
- useInsertionEffect from the React implementation looks like a `useEffect` correct me if I am wrong please
- useSyncExternalStore is actually close to the React implementation but skipped some hoops they use for ensuring value consistency during a render (with regards to interrupts, ..)

Currently we are missing useId #3402  does a first attempt at that in our new structure but I noticed we miss `_depth` and `_parent` references in `render-to-string` which could complicate this a bit.

A bit of a personal opinion here, I do think we should add `useId` to `preact/hooks` as it seems like a really useful one to me

> note that this bundle size will most likely be tree-shaken out for _most_ people

### DIFF
--- a/compat/src/react-18-hooks.js
+++ b/compat/src/react-18-hooks.js
@@ -34,24 +34,18 @@ export const useInsertionEffect = useEffect;
 export function useSyncExternalStore(subscribe, getSnapshot) {
 	const [state, setState] = useState(getSnapshot);
 	const value = getSnapshot();
-	if (state !== value) {
-		setState(value);
-	}
 
 	useLayoutEffect(() => {
 		if (value !== state) {
-			setState(state);
+			setState(value);
 		}
 	}, [value]);
 
 	useEffect(() => {
 		return subscribe(() => {
-			const newValue = getSnapshot();
-			if (newValue !== state) {
-				setState(newValue);
-			}
+			setState(getSnapshot());
 		});
-	}, [subscribe, getSnapshot, state]);
+	}, [subscribe, getSnapshot]);
 
 	return state;
 }
--- a/compat/src/react-18-hooks.js
+++ b/compat/src/react-18-hooks.js
@@ -1,32 +1,11 @@
-import { useState, useLayoutEffect, useEffect } from 'preact/hooks';
-
-/**
- * Asynchronously schedule a callback
- * @type {(cb: () => void) => void}
- */
-/* istanbul ignore next */
-// Note the following line isn't tree-shaken by rollup cuz of rollup/rollup#2566
-const defer =
-	typeof Promise == 'function'
-		? Promise.prototype.then.bind(Promise.resolve())
-		: setTimeout;
+import { useState, useLayoutEffect, useEffect, useCallback } from 'preact/hooks';
 
 export function useDeferredValue(val) {
 	return val;
 }
 
 export function useTransition() {
-	const [isPending, setIsPending] = useState(false);
-
-	const start = cb => {
-		setIsPending(true);
-		cb();
-		defer(() => {
-			setIsPending(false);
-		});
-	};
-
-	return [isPending, start];
+  return [false, useCallback(cb => { cb(); }, [])];
 }
 
 export const useInsertionEffect = useEffect;
--- a/compat/src/react-18-hooks.js
+++ b/compat/src/react-18-hooks.js
@@ -1,0 +1,60 @@
+import { useState, useLayoutEffect, useEffect } from 'preact/hooks';
+
+/**
+ * Asynchronously schedule a callback
+ * @type {(cb: () => void) => void}
+ */
+/* istanbul ignore next */
+// Note the following line isn't tree-shaken by rollup cuz of rollup/rollup#2566
+const defer =
+	typeof Promise == 'function'
+		? Promise.prototype.then.bind(Promise.resolve())
+		: setTimeout;
+
+export function useDeferredValue(val) {
+	return val;
+}
+
+export function useTransition() {
+	const [isPending, setIsPending] = useState(false);
+
+	const start = cb => {
+		setIsPending(true);
+		cb();
+		defer(() => {
+			setIsPending(false);
+		});
+	};
+
+	return [isPending, start];
+}
+
+export const useInsertionEffect = useEffect;
+
+export function useSyncExternalStore(subscribe, getSnapshot) {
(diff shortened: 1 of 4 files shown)


compat/src/react-18-hooks.js
```diff
@@ -0,0 +1,54 @@
...
+		cb();
+		defer(() => {
+			setIsPending(false);
+		});
+	};
+
+	return [isPending, start];
+}
+
+export const useInsertionEffect = useEffect;
```
gaearon, line 32: This is not right.

The point of insertion effects is that they run _before_ the DOM is updated. They’re where CSS-in-JS libraries should insert their style tags. If you run them as useEffect (after paint), then the user will see unstyled content.

Running them as a layout effect is also wrong. Because if the user reads layout in a layout effect then they’ll potentially get it before the style is inserted (eg by a sibling component). So the layout will be wrong.

This is why it’s a separate Hook that runs together with DOM mutations and before layout effects or regular effects. We wouldn’t be adding it if it was okay to run in the existing slots.

compat/src/react-18-hooks.js
```diff
@@ -0,0 +1,54 @@
...
+ */
+/* istanbul ignore next */
+// Note the following line isn't tree-shaken by rollup cuz of rollup/rollup#2566
+const defer =
+	typeof Promise == 'function'
+		? Promise.prototype.then.bind(Promise.resolve())
+		: setTimeout;
+
+export function useDeferredValue(val) {
+	return val;
```
gaearon, line 15: Fwiw you could implement a similar mechanism to what you do with startTransition.

compat/src/react-18-hooks.js
JoviDeCroock: Oh, when I took a look at React it seemed to be also a regular effect 😅 I must have completely misunderstood, will give it some thought

gaearon: No prob! See here:

https://reactjs.org/blog/2022/03/29/react-v18.html#useinsertioneffect

gaearon: And here https://github.com/reactwg/react-18/discussions/110

JoviDeCroock: Thank you so much for that ❤️

compat/src/react-18-hooks.js
```diff
@@ -0,0 +1,54 @@
+import { useState, useLayoutEffect, useEffect } from 'preact/hooks';
+
+/**
+ * Asynchronously schedule a callback
+ * @type {(cb: () => void) => void}
+ */
+/* istanbul ignore next */
+// Note the following line isn't tree-shaken by rollup cuz of rollup/rollup#2566
```
ilogico, line 8: Using `/* @__PURE__ */`is always a possibility.

compat/src/react-18-hooks.js
```diff
@@ -0,0 +1,54 @@
...
+const defer =
```
ilogico, line 9: I think `requestAnimationFrame` is more correct than using `Promise.resolve().then` (or `queueMicrotask` which should be more efficient).
From my understanding, the idea is to defer some updates so that more urgent ones can take place. But microtasks have a very high priority.

compat/src/react-18-hooks.js
```diff
@@ -0,0 +1,54 @@
...
+	return val;
+}
+
+export function useTransition() {
+	const [isPending, setIsPending] = useState(false);
+
+	const start = cb => {
+		setIsPending(true);
+		cb();
+		defer(() => {
```
ilogico, line 24: I think this will behave very differently from React.
In React, all state updates inside the callback will be deferred and `isPending` signals that the current state is not up to date.

If the idea is just to stub, the hook without providing the functionality, it might be better to simply do something like:
```js
```export function useTransition() {
  return [false, useCallback(cb => { cb(); }, [])];
}
```

compat/src/react-18-hooks.js
```diff
@@ -0,0 +1,54 @@
...
+		: setTimeout;
+
+export function useDeferredValue(val) {
+	return val;
+}
+
+export function useTransition() {
+	const [isPending, setIsPending] = useState(false);
+
+	const start = cb => {
```
ilogico, line 21: I believe this function should be the same on very render, so `useCallback` should be used.

compat/src/react-18-hooks.js
gaearon: In particular, it doesn’t make sense to set isPending to true and immediately perform the state update. The isPending flag being true is supposed to mean the state update has _not_ been applied.

gaearon: I recommend to take some existing useTransition demo and test the behavior. For example https://github.com/reactwg/react-18/discussions/65.

I think a best-effort attempt could be to do let it paint once with isPending = true and old state, and then after paint, let it update to the final state. This will have some downsides though:

1. The transition update itself would still be non-interruptible. So even if it’s delayed, you’re still going to lock the screen right after. So it doesn’t really improve responsiveness except maybe communicating there’ll be a slowdown right after.

2. Delaying the callback itself seems like significantly different behavior. React doesn’t do that.

3. This still doesn’t handle the (arguably) more important use case in longer term — preventing Suspense fallback from showing. startTransition is an important part of implementing common UI patterns with Suspense.

So all in all, maybe providing an empty naïve shim would be better than doing something half-baked.

compat/src/react-18-hooks.js
```diff
@@ -0,0 +1,33 @@
...
+	useEffect(() => {
+		return subscribe(() => {
+			setState(getSnapshot());
+		});
+	}, [subscribe, getSnapshot]);
+
+	return state;
+}
+
+// TODO: useId... This could be a bit more challenging as RTS does not have
```
eolme, line 32: Can we re-use vnodeId here? For example (pseudo code):
```js
export const useId = () => '' + vnodeId;
```

developit: Not sure if that would work - every time JSX is reevaluated it generates new VNodes with new IDs, and the IDs are monotonic with no reset options.

PR #3513, opened 2022-04-09
support nodenext in TypeScript 4.7

ilogico:
TypeScript now supports node12's module resolution, which looks at the exports entry in package.json to resolve files.

Because preact publishes type definitions in a different folder, TS fails to resolve them.

https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-beta/#package-json-exports-imports-and-self-referencing

Adding the `"types"` entries in the subpath exports seem to fix it.

This PR adds these entries to the subpath exports that seem to have appropriate type definitions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,12 +11,14 @@
 	"source": "src/index.js",
 	"exports": {
 		".": {
+			"types": "./src/index.d.ts",
 			"browser": "./dist/preact.module.js",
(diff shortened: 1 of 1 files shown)


JoviDeCroock (approved):

misode:
@JoviDeCroock When could I see this in a release?

PR #3518, opened 2022-04-22
Testing out renderer state + alternatives

developit:
I'll be opening a few PRs for variants of this to get numbers from the perf bot. This one is an experiment to check the cost/benefit of arguments vs rendererState.

I'd originally intended to simply replace properties with accessor functions (`rendererState.context = x` becomes `setCurrentContext(x)`) in 8ea55ce, however Terser's inlining has degraded considerably since we last relied on this technique. In particular, it fails to inline any function that is used twice from within the same caller:

```js
function patch() {
  let ctx = getCurrentContext(); // gets inlined
  if (child) {
    setCurrentContext(childCtx); // can be forced inline via @__INLINE__, but that's not great
    patch(child);
    setCurrentContext(ctx); // will never be inlined. forcing via @__INLINE__ inlines as an IIFE 😶‍🌫️
  }
}
```

So, I decided to try checking whether `context` or `parentDom` (the main two properties on RendererState) were driving performance. When I started looking into this though, I realized that `context` hasn't been an argument for quite some time - it's a backwards walk up the tree. I'd like to do another experimental PR to check if this is a performance tradeoff or not.


Thompson1985 (approved):

JoviDeCroock:
Closing: targets the old long-lived `v11` branch (last updated 2024-04); v11 now lives on `main`. This renderer-state experiment was inconclusive and is superseded by the shipped architecture.

PR #3519, opened 2022-04-22
Experiment: split component lifecycle merge types

developit:
This is an alternative to #3518 that re-unifies component rendering (function and class components taking the same codepath) in order to split apart component mounting vs patching. It may also be possible to keep function/class components split up while also keeping mount/patch separate, but some quick experiments locally seemed to show a fairly major size increase (4 functions with limited similarity, vs 2 similar functions in both `main` and this PR).

The code here is primarily a refactoring of the v11 code from `main`, with a few optimizations from `vm` incorporated. In particular, `renderFunctionComponent()` and `renderClassComponent()` were merged into a new `patchComponent()` method (to match `patchElement`). It also removes short-circuiting of Text Internals, since we hadn't actually intended to bypass `options._diff()` and `options.diffed()` for those.


Thompson1985 (approved):

JoviDeCroock:
Closing: targets the old `v11` branch (now inactive; v11 is on `main`). The lifecycle-merge split is superseded — `main` already merged the function/class render paths.

PR #3520, opened 2022-04-24
Test diff-related changes in isolation

JoviDeCroock:
@developit


developit:
So TLDR here seems to be:  when there's an early bailout path, prefer conditional invocation over early return.

developit (approved):

PR #3526, opened 2022-04-28
Previous start dom optimization

JoviDeCroock:


src/diff/mount.js
```diff
@@ -136,7 +124,7 @@ export function mount(internal, newVNode, startDom) {
 		options._catchError(e, internal);
 	}
 
-	return nextDomSibling;
+	return prevStartDom || nextDomSibling;
```
JoviDeCroock, line 127: @developit I am a bit suspicious here because even if I remove this everything works

developit: We likely are missing a test then:

```js
function Root() {
  return [
    <h1>A</h1>
    createPortal(<h2>B</h2>, document.body),
    <h3>C</h3>
  ];
}

render(<Root />, scratch);
```

developit: Added a test for hydration that fails if the line is removed:
https://github.com/preactjs/preact/pull/3531/files#diff-2e304b2302598aee764c82b1325c8e850f5e37a6f08b1c20fcf1adbd12a08ddeR101

developit (approved):

PR #3531, opened 2022-04-28
Improve Portal hydration test coverage

developit:
As @JoviDeCroock pointed out [here](https://github.com/preactjs/preact/pull/3526/files#r860582571), we didn't have tests covering the case where `startDom` needs to be reset after rendering a portal. As far as I can tell, this is only needed during hydration.


developit:
Merging since it's a simple test add.

PR #3559, opened 2022-06-09
drop IE11 support

JoviDeCroock:
IE11 is coming to an [end on JUNE 15](https://www.lansweeper.com/eol/internet-explorer-11-end-of-life/)

- Removes Promise fallback
- Removes Symbol fallback
- Removes some DOM-specific IE impedements
- Removes the `removeNode` helper in favor of `Element.prototype.remove()`
- Removes the `assign` helper in favor of `Object.assign`


marvinhagemeister (approved):
Let's gooo 🙌🙌

src/component.js
```diff
@@ -176,10 +175,7 @@ let rerenderQueue = [];
  */
 /* istanbul ignore next */
 // Note the following line isn't tree-shaken by rollup cuz of rollup/rollup#2566
-const defer =
-	typeof Promise == 'function'
-		? Promise.prototype.then.bind(Promise.resolve())
-		: setTimeout;
+const defer = Promise.prototype.then.bind(Promise.resolve());
```
porfirioribeiro, line 178: Wouldn't `queueMicrotask` also be a possibility here?

porfirioribeiro (approved):

src/component.js
marvinhagemeister: Good catch. Browser support is looking really good for that too. What do you think @developit ?

developit: I had thought it was worse, but yeah you're right, it's decent. Main concern is Edge 16-18 and Safari 12, but those are looking pretty old these days...

JoviDeCroock:
These changes are applied in https://github.com/preactjs/preact/pull/3589 and will hit main

PR #3560, opened 2022-06-10
make createRoot / hydrateRoot compatible with React spec

3846masa:
* hydrateRoot returns Root instance
    * https://github.com/facebook/react/blob/v18.0.0/packages/react-dom/src/client/ReactDOMRoot.js#L255-L259
* Root instance has `unmount()` method
* `Root.prototype.render()` and `Root.prototype.unmount()` returns void
    * ref. https://github.com/facebook/react/blob/v17.0.2/packages/react-dom/src/client/ReactDOMRoot.js#L15-L20

### DIFF
--- a/compat/client.js
+++ b/compat/client.js
@@ -1,13 +1,19 @@
-const { render, hydrate } = require('preact/compat');
+const { render, hydrate, unmountComponentAtNode } = require('preact/compat');
 
-exports.createRoot = function(container) {
+function createRoot(container) {
 	return {
 		render(children) {
-			return render(children, container);
+			render(children, container);
+		},
+		unmount() {
+			unmountComponentAtNode(container);
 		}
 	};
-};
+}
+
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):

compat/client.js
```diff
@@ -1,13 +1,17 @@
-const { render, hydrate } = require('preact/compat');
+const { render, hydrate, unmountComponentAtNode } = require('preact/compat');
 
 exports.createRoot = function(container) {
```
JoviDeCroock, line 3: if you name this function `createRoot` you won't have to call `exports.createRoot` on line 16

3846masa: I got it. I will change the code.

3846masa:
@JoviDeCroock
Thanks for the review. I changed the code and rebased it.

PR #3563, opened 2022-06-13
Fix size CI failing on node version missmatch

JoviDeCroock:


marvinhagemeister (approved):
I'd be totally fine going straight to 16.x as that's the latest LTS 👍

PR #3553, opened 2022-06-03
implement state settling in X

JoviDeCroock:
Supersedes https://github.com/preactjs/preact/pull/3522
Mirrors https://github.com/preactjs/preact/pull/3440 from Preact 11

This introduces state settling like React in Preact, this is something some libraries rely on to have states settle after a given amount of renders.

Codesandbox as illustration https://codesandbox.io/s/young-cdn-jof35q?file=/src/index.js

PR #3566, opened 2022-06-14
10.8.0

JoviDeCroock:
- add export maps to the subpackages (#3565)
- Ensure both onchange and oninput callbacks are executes when typing (#3562)
- make createRoot / hydrateRoot compatible with React spec (#3560)
- implement state settling in X (#3553)


cristianbote (approved):

marvinhagemeister (approved):

marvinhagemeister:
That one byte improvement!

Issue #3572, opened 2022-06-17
Regex used to convert oninput to oninputcapture is too broad


jramanat-oracle:
- [ ] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
The change in https://github.com/preactjs/preact/pull/3562 was intended to convert oninput to oninputCapture.  However, because it is using /^oninput/, it actually ends up converting oninput* to oninputCapture.  This breaks functionality for custom elements that might be exposing CustomEvents that happen to start with "input".  i.e.

`<my-custom-element oninputAction={...}/> `

used to call addEventListener('inputAction', ...), but will now add a capture listener for the 'input' event.

**To Reproduce**

See above.

**Expected behavior**

See above.

jramanat-oracle:
Think https://github.com/preactjs/preact/pull/3573 is all that's needed.

PR #3573, opened 2022-06-17
restrict "oninputCapture" conversion to just "oninput" vs "oninput*"

jramanat-oracle:
This should fix https://github.com/preactjs/preact/issues/3572


JoviDeCroock:
This looks like the right solution, is it intended to be in draft?

jramanat-oracle:
Oh, I might have misunderstood the contributing guide.  I thought it implied that draft PRs were preferred.  I think the code change should be ready to go, thanks!

developit (approved):
Whoops! Thanks for the fix @jramanat-oracle

PR #3575, opened 2022-06-20
Add `_pendingValue` to `mangle.json`

marvinhagemeister:
PR #3567 added a new property to hooks for pending state. Devtools needs to be able to access these to be able to show them in the sidebar. With this PR we ensure that devtools can rely on a consistent name to access it.

### DIFF
--- a/mangle.json
+++ b/mangle.json
@@ -31,6 +31,7 @@
       "$_list": "__",
       "$_pendingEffects": "__h",
       "$_value": "__",
+      "$_pendingValue": "__V",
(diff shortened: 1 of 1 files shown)


JoviDeCroock (approved):

JoviDeCroock:
Do you need _pendingArgs too?

marvinhagemeister:
@JoviDeCroock good catch! Just checked and devtools doesn't need to read `_args` or `_pendingArgs`. It just grabs the values directly.

PR #3576, opened 2022-06-21
Remove setState on unmounted component warning

### LINKED ISSUE #3470: Can't call "this.setState" on an unmounted component error.

**Describe the bug**
This bug looks to be the same as #2056 but on a version `10.6.6` of Preact.

I am creating a modal (using `@headlessui/react` and `preact/compat`) and when you press the modals close button the following warning appears in the console.

```
debug.js:379 Can't call "this.setState" on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
```

It should be noted that the component works as expected but this warning is definitely something I would like to get fixed. Given there is a similar issue mentioned above I am assuming this is a bug in Preact but it could definitely be an issue on my side so let me know if that is the case.

**To Reproduce**

I had some issues getting CodeSandbox working so I created a [Stackblitz](https://stackblitz.com).

1. Go to [demo application](https://stackblitz.com/edit/vitejs-vite-qkdrfy?file=src/App.jsx)
2. Right-click anywhere in the application preview and press "Inspect" and open the "Console" tab. You will probably want to clear it.
3. Click on 'Open Dialog' to show the modal
4. Observe no error in the console
5. Click on 'Got it, Thanks!' or press `Esc` on your keyboard to close the modal
6. Observe the error above in the console.

**Expected behavior**
No error should appear in the console.

ekwoka:
As discussed in https://github.com/preactjs/preact/issues/3470 this warning can cause unwarranted concern in users.

Following the path of React, where this warning was removed https://github.com/facebook/react/pull/22114#issuecomment-929191048 it can likely be similarly removed from Preact.

Basically, Preact already checks if the component is unmounted, hence its ability to provide the warning, and most users that run into this implement anti-patterns to suppress the warning instead of addressing the actual issue being warned about.

The actual issue is less likely to be an indication of a real memory leak, and more likely to be just a single late update.

With all this in mind, it makes sense to also remove the warning here.

What this does:
- Removes the warning for setting state after a component has unmounted
- Removes test that checked for warning

Resolves #3470

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -377,13 +377,6 @@ Component.prototype.setState = function(update, callback) {
 					`"this.state = {}" directly.\n\n${getOwnerStack(getCurrentVNode())}`
 			);
 		}
-	} else if (this._parentDom == null) {
-		console.warn(
-			`Can't call "this.setState" on an unmounted component. This is a no-op, ` +
-				`but it indicates a memory leak in your application. To fix, cancel all ` +
-				`subscriptions and asynchronous tasks in the componentWillUnmount method.` +
-				`\n\n${getOwnerStack(this._vnode)}`
-		);
 	}
 
 	return setState.call(this, update, callback);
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):
Awesome, thank you!

PR #3577, opened 2022-06-21
Add support for svg property shape-rendering

DannyvanderJagt:
Adding support for the conversion from `shapeRendering` to `shape-rendering` as a SVG property as it seems to be missing.

[Reproducible demo](https://codesandbox.io/s/infallible-pond-o53o79?file=/src/index.js)
[SVG docs - shape-rendering](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/shape-rendering)

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -10,7 +10,7 @@ export const REACT_ELEMENT_TYPE =
 	(typeof Symbol != 'undefined' && Symbol.for && Symbol.for('react.element')) ||
 	0xeac7;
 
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):
Awesome work, thank you!

You disabled "let maintainers update branch" so we'll just need you to update the branch 🙌

marvinhagemeister (approved):
Nice work! Like @JoviDeCroock said we can't update the branch against master ourselves to merge it.

DannyvanderJagt:
Thanks for the library, happy to return a tiny bit 👍

PR #3578, opened 2022-06-21
fix: commit hooks in options.diffed

JoviDeCroock:
Commit hooks in diffed to bypass the issue of us dealing with a component that isn't in the commitQueue, components that are missing effects aren't admitted to the commitQueue and are essentially missing the commit of their pending arguments and values. This wasn't an issue in user-land as it would correct itself on the next invocation of `_render` because we would hit that `else` branch indicating that we are dealing with a new component.

This PR also means we can remove the updating of pending values in the else branch


marvinhagemeister (approved):
Like this approach a lot more conceptually! Thank you so much for the PR ❤️

PR #3583, opened 2022-06-23
Preact X useId

JoviDeCroock:
Currently to the thought of a [prefix tree](https://en.wikipedia.org/wiki/Trie) this would be an alternative to `useId`. This tries to follow our component tree and has a `mask` variable.

Resolves https://github.com/preactjs/preact/issues/3373

## Mount

Every time we encounter a component a long the way we will set `vnode._mask`  to the global counter we are following, this means that every node that _can_ generate an id will have a unique mask. When we invoke `useId` we will look at the `localIdCounter` (reset for every component we see) this with the combination of our prefix will ensure that we generate a consistent `id` even if we are invoking multiple `useId` calls.

## Mounting a new subtree

When we see a new subtree we will look at the parentMask and count from there, this however means that we could cause conflicts, ideally we would need to keep incrementing `mask`. If we can find a way to reset the `mask` counter for SSR environments we would bypass this.

## Notes

- We can optimize subsequent client-side renders by using a counter that increments after hydration but atm I don't think we have a good way to track this

DEMO:  https://codesandbox.io/s/crazy-platform-omyl5l?file=/src/hooks.js

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -40,7 +40,7 @@ options._diff = vnode => {
 				? vnode._parent._children.indexOf(vnode)
 				: 0;
 		vnode._mask = parentMask + position;
-	} else {
+	} else if (!vnode._mask) {
 		vnode._mask = vnode._parent._mask;
 	}
 	currentComponent = null;
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -27,19 +27,16 @@ const RAF_TIMEOUT = 100;
 let prevRaf;
 
 options._diff = vnode => {
-	if (vnode.type === Fragment) {
-		// Skip so we don't have to ensure wrapping fragments in RTS and prepass
-		vnode._mask = '';
-	} else if (typeof vnode.type === 'function' && !vnode._mask) {
+	if (typeof vnode.type === "function" && !vnode._mask && vnode.type !== Fragment) {
 		const parentMask =
-			vnode._parent && vnode._parent._mask ? vnode._parent._mask : '';
+		vnode[PARENT] && vnode[PARENT]._mask ? vnode[PARENT]._mask : "";
 		const position =
-			vnode._parent && vnode._parent._children
-				? vnode._parent._children.indexOf(vnode)
-				: 0;
+		vnode[PARENT] && vnode[PARENT][CHILDREN]
+			? vnode[PARENT][CHILDREN].indexOf(vnode)
+			: 0;
 		vnode._mask = parentMask + position;
 	} else if (!vnode._mask) {
-		vnode._mask = vnode._parent._mask;
+		vnode._mask = vnode[PARENT] ? vnode[PARENT]._mask : "";
 	}
 	currentComponent = null;
 	if (oldBeforeDiff) oldBeforeDiff(vnode);
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -27,16 +27,20 @@ const RAF_TIMEOUT = 100;
 let prevRaf;
 
 options._diff = vnode => {
-	if (typeof vnode.type === "function" && !vnode._mask && vnode.type !== Fragment) {
+	if (
+		typeof vnode.type === 'function' &&
+		!vnode._mask &&
+		vnode.type !== Fragment
+	) {
 		const parentMask =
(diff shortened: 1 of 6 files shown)


hooks/src/index.js
```diff
@@ -55,27 +85,8 @@ options._render = vnode => {
...
-			}
-			hookItem._pendingArgs = undefined;
-			hookItem._pendingValue = EMPTY;
-		});
-	}
-	previousComponent = currentComponent = null;
-};
-
 options._commit = (vnode, commitQueue) => {
+	mask = 0;
```
JoviDeCroock, line 89: Currently this is the reason why we store `_mask` on vnodes, ideally we introduce a better way to reset this on the server-side. We could offer this as an `options` attribute, hwoever that would mean we need to cut a new `renderToString` and `preact-ssr-prepass` minor

hooks/src/index.js
```diff
@@ -26,15 +26,45 @@ let oldBeforeUnmount = options.unmount;
 const RAF_TIMEOUT = 100;
 let prevRaf;
 
+let mask = 0;
+let localIdCounter = 0;
+
 options._diff = vnode => {
```
JoviDeCroock, line 32: We can make this assumption because RTS calls this https://github.com/preactjs/preact-render-to-string/blob/master/src/index.js#L137 and follows a consistent diffing compared to our core algorithm

hooks/src/index.js
```diff
@@ -26,15 +26,45 @@ let oldBeforeUnmount = options.unmount;
...
+		vnode._mask =
+			vnode._parent && vnode._parent._mask
+				? vnode._parent && vnode._parent._mask + 1
+				: mask++;
+	}
 	currentComponent = null;
 	if (oldBeforeDiff) oldBeforeDiff(vnode);
 };
 
+options.diffed = vnode => {
```
JoviDeCroock, line 43: Also called in RTS https://github.com/preactjs/preact-render-to-string/blob/master/src/index.js#L222

BFriedrichs:
Hi just to let you know I checked out this and the PR on render-to-string (https://github.com/preactjs/preact-render-to-string/pull/226)

In my setup this line breaks:
https://github.com/preactjs/preact/blob/186f5da5c8965fccd2760bacf76d7d75e8769d53/hooks/src/index.js#L36-L39

This is because indexOf is not defined on `vnode._parent._children`, I investigated this and it seems like `_children` is a rendered preact object rather than a list. In my case I fixed it by just making the line:
```js
const position = vnode._parent &&
    vnode._parent._children &&
    vnode._parent._children !== vnode &&
    vnode._parent._children.indexOf
```

I realise this is quite hacky (and probably doesn't do what I _actually_ want it to) but at least it doesn't crash anymore. I have basically no knowledge of this project and just needed stuff to work.

JoviDeCroock:
You mean it breaks on the client or during SSR, if it breaks on the client that might be expected I haven't gone through all cases yet.

BFriedrichs:
It breaks during the `render` call of render-to-string.

JoviDeCroock:
Well that might mean you are not using the version of RTS that sets it to an array I think 😅

BFriedrichs:
The more I think about it, given the scuffed setup I'm using it is very possible that the fault is actually somewhere on my end ...

marvinhagemeister (approved):
This is great! 💯

JoviDeCroock:
@developit had an alternative that had a better byte-size impact

mjgerace:
Is this okay to merge?

Wallacy:
Any news here?

marvinhagemeister (approved):

marvinhagemeister (approved):

marvinhagemeister:
Let's merge this 👍

PodaruDragos:
will there be a new release soon that integrates this ?

marvinhagemeister:
@PodaruDragos I definitely share your excitement about `useId`! We merged this PR just two days ago and it was the weekend. We maintainers need some time to recharge and relax too :)

PodaruDragos:
@marvinhagemeister my apologies for sounding demanding, yeah I am really waiting for this for a long time.
Thanks for this feature and thanks for all the good work you guys are doing on this project.

Aloento:
When will the new version be released?

PR #3584, opened 2022-06-23
Preact X controlled components

JoviDeCroock:
In this PR we introduce controlled components, however it runs into the selection issue shown [here](https://jovidecroock.com/blog/controlled-inputs) basically given the following code

```js
const Issue = () => {
  const [value, setValue] = useState('')

  const onInput = (e) => {
    if (e.currentTarget.value.length <= 3) {
      setValue(e.currentTarget.value)
    }
  }

  return <input value={value} onInput={onInput} />
}
```

When we type "aaaa" the fourth "a" won't appear, however if we go to the front of the input (start of selection) and type another a the selection will move our cursor to the end of the input.

You can try for yourself [here](https://codesandbox.io/s/inspiring-mayer-vmev4o?file=/src/index.js)

This can be fixed however with the following code: https://github.com/preactjs/preact/pull/3584/commits/b66ab9d19196b4a05236cb8dfbc5c0fcbc51c95d

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -364,6 +364,21 @@ function diffElementNodes(
 			);
 		}
 
+		if (
+			(nodeType === 'input' ||
+				nodeType === 'textarea' ||
+				nodeType === 'select') &&
+			(newProps.onInput || newProps.onChange)
+		) {
+			if (newProps.value != null) {
+				dom._isControlled = true;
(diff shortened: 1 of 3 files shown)


JoviDeCroock:
Closing as I can't seem to resolve the scroll-restoration issue 😅 also reverting in 11 for now until we can figure this out more

PR #3588, opened 2022-06-24
Try the rendererState in Preact X

JoviDeCroock:


JoviDeCroock:
Impact on perf is negative and it is on byte-size too

PR #3590, opened 2022-06-24
Remove controlled components

JoviDeCroock:
Due to the selection-range issue outlined in #3584


developit:
The size and perf wins here are so compelling.

PR #3592, opened 2022-06-28
Handle capture and normal events in an array on the element

JoviDeCroock:


JoviDeCroock:
Updating this branch to see if 
image
is a fluke

PR #3568, opened 2022-06-15
React 18 hooks

JoviDeCroock:
This implements stub-like behaviour for most hooks and implements a minimal functional version for others, currently we are missing `useId`, this is a bigger piece of work as it will need changes in RenderToString/... to work #3402 implements the CSR version of it.

Resolves #3557
Resolves #3555

PR #3598, opened 2022-07-02
add react 18 hooks

JoviDeCroock:
Just like #3568 misses `useId`

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -122,6 +122,37 @@ const StrictMode = Fragment;
  */
 const flushSync = (callback, arg) => callback(arg);
 
+export function startTransition(cb) {
+	cb();
+}
+
+export function useDeferredValue(val) {
+	return val;
+}
+
+export function useTransition() {
+	return [false, startTransition];
+}
+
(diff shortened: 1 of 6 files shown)


marvinhagemeister (approved):

compat/src/index.js
```diff
@@ -122,6 +122,37 @@ const StrictMode = Fragment;
...
+	const [state, setState] = useState(getSnapshot);
+
+	// TODO: in suspense for data we could have a discrepancy here because Preact won't re-init the "useState"
+	// when this unsuspends which could lead to stale state as the subscription is torn down.
+
+	useEffect(() => {
+		return subscribe(() => {
+			setState(getSnapshot());
+		});
+	}, [subscribe, getSnapshot]);
```
mayank99, line 151: I came across this PR when I was looking into React 18 support in compat. Just wondering, shouldn't the state be set before subscribing? 🤔

```suggestion
	useEffect(() => {
	  const callback = () => setState(getSnapshot());
	  callback();
		return subscribe(callback);
	}, [subscribe, getSnapshot]);
```

PR #3600, opened 2022-07-02
try event optimization in x

JoviDeCroock:
This PR investigates the positive results of #3592 but with some variation

- commit 1: array
- commit 2: object, both in dict mode
- commit 3: what we had before in dict mode

Result 1:
image

Restult 2:
image

Result 3:
image


JoviDeCroock:
Looks like a net negative in X

PR #3603, opened 2022-07-03
propose removal of createRoot from core 

JoviDeCroock:
as it is only available in react-dom/client

let's look at the byte-size impact to make judgement


developit:
30b is compelling, but I'm waffling on this for some reason. I liked the separation we were getting with createRoot. That said, I think our API was already different from React in a way that would confuse folks (`createRoot().hydrate()` vs `hydrateRoot().render()`?).

I need to think more about it though.

JoviDeCroock:
@developit yes it was slightly different already

marvinhagemeister (approved):

JoviDeCroock:
Closing due to inactivity

PR #3606, opened 2022-07-06
Improve tree shaking in compat

JoviDeCroock:
This would improve the tree-shakeability of compat, most of these can be done on-demand apart from the `options.vnode` one, that one should be part of our `jsx` hooks and `createElement`


JoviDeCroock:
Closing this for now but keeping the branch open as I think this is a good experiment to continue later on

PR #3617, opened 2022-07-14
fix: make demo compatible with node 16 and 18

gengjiawen:
mainly webpack and related deps upgrade.
Tested on node 16 and 18.

<del>
https://github.com/sass/node-sass/pull/3257 still not merged, so I make a patch to it.
</del>

replaced with sass.

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -2,7 +2,7 @@
   "name": "demo",
   "main": "index.js",
   "scripts": {
-    "start": "webpack-dev-server --inline"
+    "start": "webpack-dev-server"
   },
   "devDependencies": {
     "@babel/core": "^7.0.0-beta.55",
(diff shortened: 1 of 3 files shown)


demo/package.json
```diff
@@ -16,14 +17,15 @@
...
-    "node-sass": "^4.12.0",
-    "sass-loader": "7.1.0",
-    "style-loader": "0.23.1",
-    "webpack": "4.33.0",
-    "webpack-cli": "^3.3.4",
-    "webpack-dev-server": "^3.7.1"
+    "patch-package": "^6.4.7",
+    "css-loader": "6.7.1",
+    "html-webpack-plugin": "5.5.0",
+    "node-sass": "7.0.1",
```
marvinhagemeister, line 23: What if switched to `sass` instead of `node-sass`? It looks like that's the [recommended way](https://github.com/sass/node-sass#node-sass) to use sass these days. With that I assume we won't need to patch packages anymore and the `sass` package will work out of the box with newer node versions.

marvinhagemeister (commented):
Awesome, thank you so much for the PR. Left a suggestion for the sass handling. Let me know what you think!

demo/package.json
gengjiawen: sass works good. And also eliminate the cpp toolchain. Should be more friendly for windows user now.

JoviDeCroock (approved):
Thank you so much 🙌

PR #3627, opened 2022-07-21
insert comments for adjacent text-nodes

JoviDeCroock:
This PR will insert comments for adjacent text-nodes, when the babel transform runs for `<p>x{' '}y</p>` it will generate 3 text-nodes, when we enter hydration we see 1 dom-node but 3 virtual nodes so we want to correctly hydrate these as 3 dom-nodes hence we insert comments so the DOM is forced to see these as 3 dom-nodes as well

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,7 +4,7 @@ root = true
 indent_style = tab
 end_of_line = lf
 charset = utf-8
-trim_trailing_whitespace = true
+trim_trailing_whitespace = false
 insert_final_newline = true
 
 [{*.json,.*rc,*.yml}]
--- a/package.json
+++ b/package.json
@@ -21,23 +21,23 @@
 		"Samsung>=8.2",
 		"not dead"
 	],
-  "sideEffects": [
-    "./compat/**/*",
-    "./debug/**/*",
-    "./devtools/**/*",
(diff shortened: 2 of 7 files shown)


.editorconfig
```diff
@@ -4,7 +4,7 @@ root = true
 indent_style = tab
 end_of_line = lf
 charset = utf-8
-trim_trailing_whitespace = true
+trim_trailing_whitespace = false
```
JoviDeCroock, line 7: This was tripping me up hard when fixing the tests in `pretty.test.js`

package.json
```diff
@@ -126,7 +126,7 @@
 		"build:test-utils": "npm run -s _bundle -- --cwd test-utils",
 		"build:compat": "npm run -s _bundle -- --cwd compat --globals 'preact/hooks=preactHooks'",
 		"build:jsx": "npm run -s _bundle -- --cwd jsx-runtime",
-		"build:server": "npm run -s _bundle -- --cwd server",
+		"build:server": "cd server && npm run build",
```
JoviDeCroock, line 129: We weren't using the original build resulting in missing bundles

JoviDeCroock:
Closing due to inactivity

PR #3643, opened 2022-08-02
check for undefined __hooks in afterpaint

### LINKED ISSUE #3642: Cannot set properties of null (setting '__h') error when suspending the entire app
Preact version: `10.10.0` (happens in earlier versions aswell).

**Bug Description**

When wrapping the entire `<App>` with Suspense, I get the following error:

```
Uncaught TypeError: Cannot set properties of null (setting '__h')
    at M
```
And the app breaks after navigating to another page (it becomes unresponsive).

This **does not** happen if I only wrap the `<Routes>` component with Suspense.

**To Reproduce**

1. Clone my repo: https://github.com/theninthsky/client-side-rendering/tree/preact.
2. Execute `npm i` and `npm start`.
3. Refresh the page a few times until this error shows (it's inconsistent, as if it depends on script loading times or something).

You can even visit the preview deploy here: https://70f42741.client-side-rendering.pages.dev

**Expected behavior**
No error should be thrown and the app should be responsive.

I'd like to add that I really love Preact and its great compatibility with React codebase, and that the ability to suspend the entire app is very important to me (in order to prevent when I call _[Sequenced Rendering](https://github.com/theninthsky/client-side-rendering/tree/preact#preventing-sequenced-rendering))_.

JoviDeCroock:
Fixes #3642 

During suspensions we reset the hooks state https://github.com/preactjs/preact/blob/master/compat/src/suspense.js#L51 which could turn out problematic if an afterpaint is queued 😅

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -173,10 +173,7 @@ export function useReducer(reducer, initialState, init) {
 				const nextValue = hookState._reducer(currentValue, action);
 
 				if (currentValue !== nextValue) {
-					hookState._nextValue = [
-						nextValue,
-						hookState._value[1]
-					];
+					hookState._nextValue = [nextValue, hookState._value[1]];
 					hookState._component.setState({});
 				}
 			}
@@ -341,7 +338,7 @@ export function useErrorBoundary(cb) {
 function flushAfterPaintEffects() {
 	let component;
 	while ((component = afterPaintEffects.shift())) {
-		if (!component._parentDom) continue;
+		if (!component._parentDom || !component.__hooks) continue;
 		try {
 			component.__hooks._pendingEffects.forEach(invokeCleanup);
 			component.__hooks._pendingEffects.forEach(invokeEffect);


marvinhagemeister (approved):
LGTM 👍

hooks/src/index.js
```diff
@@ -341,7 +338,7 @@ export function useErrorBoundary(cb) {
 function flushAfterPaintEffects() {
 	let component;
 	while ((component = afterPaintEffects.shift())) {
-		if (!component._parentDom) continue;
+		if (!component._parentDom || !component.__hooks) continue;
```
marvinhagemeister, line 341: Is it possible to add a test for this?

JoviDeCroock: I tried but we consume our queue before 😅

marvinhagemeister: Ohh right makes sense. Forgot about our synchronous test logic.

PR #3645, opened 2022-08-04
Multiple state hooks in a component

JoviDeCroock:
Currently the last sCU wins which basically means we are losing a bit of granularity, as well as filtering out React.memo when wrapped. The new flow looks like the following:

- hold a flag whether we have added SCU
- when we traverse if no state-hooks have a pending next state --> update because this means no state setters have been invoked for this component (call former SCU)
- when there are invocations look at whether all of the ones with a nextState have equal states
  - if they all have equal states between next and original bail out
  - if not continue updating (call former SCU)

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -189,10 +189,15 @@ export function useReducer(reducer, initialState, init) {
 					x => x._component
 				);
 				const allHooksEmpty = stateHooks.every(x => !x._nextValue);
+				// When we have no updated hooks in the component we invoke the previous SCU or
+				// traverse the VDOM tree further.
 				if (allHooksEmpty) {
 					return prevScu ? prevScu(p, s, c) : true;
 				}
 
+				// We check whether we have components with a nextValue set that
+				// have values that aren't equal to one another this pushes
+				// us to update further down the tree
 				const shouldSkipUpdating = stateHooks.every(hookItem => {
 					if (!hookItem._nextValue) return true;
 					const currentValue = hookItem._value[0];
@@ -205,6 +210,11 @@ export function useReducer(reducer, initialState, init) {
 					return prevScu ? prevScu(p, s, c) : true;
 				}
 
+				// When all set nextValues are equal to their original value
+				// we bail out of updating.
+				// Thinking: would this be dangerous with a batch of updates where
+				// Comp1 updates --> Comp2 updated in same batch twice but has same eventual state --> this leads to us
+				// not diving into Comp3
 				return false;
 			};
 		}
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -68,7 +68,7 @@ options.diffed = vnode => {
 			if (hookItem._pendingArgs) {
 				hookItem._args = hookItem._pendingArgs;
 			}
-			if (hookItem._pendingValue !== EMPTY) {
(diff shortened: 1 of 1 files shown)


hooks/src/index.js
```diff
@@ -181,15 +181,43 @@ export function useReducer(reducer, initialState, init) {
...
-		hookState._component.shouldComponentUpdate = () => {
-			if (!hookState._nextValue) return true;
+		if (!hookState._component._hasScuFromHooks) {
+			hookState._component.__hooks._hasScuFromHooks = true;
+			const prevScu = hookState._component.shouldComponentUpdate;
+			hookState._component.shouldComponentUpdate = (p, s, c) => {
+				const stateHooks = hookState._component.__hooks._list.filter(
+					x => x._component
+				);
+				const allHooksEmpty = stateHooks.every(x => !x._nextValue);
```
marvinhagemeister, line 191: For state updates the hook value could be `0` or `""` which would be interpreted as empty here. Let's check for `undefined` instead here to avoid casting errors.

```suggestion
				const allHooksEmpty = stateHooks.every(x => x._nextValue === undefined);
```

hooks/src/index.js
```diff
@@ -181,15 +181,43 @@ export function useReducer(reducer, initialState, init) {
...
-		};
+				if (!shouldSkipUpdating) {
+					return prevScu ? prevScu(p, s, c) : true;
+				}
+
+				// When all set nextValues are equal to their original value
+				// we bail out of updating.
+				// Thinking: would this be dangerous with a batch of updates where
+				// Comp1 updates --> Comp2 updated in same batch twice but has same eventual state --> this leads to us
+				// not diving into Comp3
```
marvinhagemeister, line 217: Right now I can't come up with a scenario where that would break. We're already somewhat in edge case territory here with components being re-scheduled as part of the same commit. So I'd say let's keep on eye on that, but I don't expect there to be issues.

marvinhagemeister (approved):
As already mentioned in slack: Really impressed with your debugging skills on this one! This is great 💯

hooks/src/index.js
JoviDeCroock: _nextValue in this case is an array like the normal state hook containing [value, updater] so using ! is the least bytes and correct as it's array or undefined

marvinhagemeister: Makes sense, I was confusing it with the actual value. Thanks for correcting me 👍

hooks/src/index.js
```diff
@@ -181,15 +181,43 @@ export function useReducer(reducer, initialState, init) {
...
+				}
 
-			const currentValue = hookState._value[0];
-			hookState._value = hookState._nextValue;
-			hookState._nextValue = undefined;
+				// We check whether we have components with a nextValue set that
+				// have values that aren't equal to one another this pushes
+				// us to update further down the tree
+				const shouldSkipUpdating = stateHooks.every(hookItem => {
+					if (!hookItem._nextValue) return true;
```
robertknight, line 202: Per Marvin's comment above, shouldn't this check for `_nextValue` being undefined?

JoviDeCroock: > _nextValue in this case is an array like the normal state hook containing [value, updater] so using ! is the least bytes and correct as it's array or undefined

PR #3651, opened 2022-08-07
avoid crashing due to __hooks being null

### LINKED ISSUE #3647: "Cannot read properties of null (reading '__')" error when suspending the entire app
Preact version: `10.10.1`.

**Bug Description**

When wrapping the entire `<App>` with Suspense, I get the following error:

```
Uncaught TypeError: Cannot read properties of null (reading '__')
    at i.__c.shouldComponentUpdate
```
And almost the entire app unmounts (due to the lack of error boundary).

This **does not** happen if I only wrap the `<Routes>` component with Suspense.

**To Reproduce**

1. Clone my repo: https://github.com/theninthsky/client-side-rendering/tree/preact.
2. Execute `npm i` and `npm start`.
3. Refresh the page a few times until this error shows.

You can even visit the preview deploy here: https://04cb527b.client-side-rendering.pages.dev

**Expected behavior**
No error should be thrown and the app should be visible.

JoviDeCroock:
fixes #3647

When suspending we reset `__hooks` to null, when this happens we should update rather than crashing 😅

### DIFF
--- a/compat/test/browser/suspense.test.js
+++ b/compat/test/browser/suspense.test.js
@@ -2088,4 +2088,29 @@ describe('suspense', () => {
 			expect(scratch.innerHTML).to.eql(`<div>123</div>`);
 		});
 	});
+
+	it('should not crash due to suspending a ', () => {
+		const [Suspender, suspend] = createSuspender(() => {
+			const [state] = useState('hello world');
+			return <p>{state}</p>;
+		});
+
+		const suspense = (
+			<Suspense fallback={<div>Suspended...</div>}>
+				<Suspender />
+			</Suspense>
+		);
+
+		render(suspense, scratch);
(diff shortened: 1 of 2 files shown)


compat/test/browser/suspense.test.js
```diff
@@ -2088,4 +2088,29 @@ describe('suspense', () => {
 			expect(scratch.innerHTML).to.eql(`<div>123</div>`);
 		});
 	});
+
+	it('should not crash due to suspending a ', () => {
```
JoviDeCroock, line 2092:
```suggestion
	it('should not crash due to suspending a component with useState', () => {
```

marvinhagemeister (approved):
Lgtm 😎

developit (approved):

PR #3655, opened 2022-08-08
fix `useSyncExternalStore` relying on changed render values

### LINKED ISSUE #3654: @apollo/client + useSyncExternalStore - data updates not working since 10.9.0+
**Describe the bug**

:bow: *my knowledge about anything react 18 is baaad. apologies if the issue is useless. feel free to close if that's the case.*

`useSyncExternalStore` calls in `@apollo/client` don't update any state. It uses a memoized function (`useCallback`) as "subscribe" param.

Issues like this one seem to be already known judging by comments in the codebase. But since `@apollo/client` is a popular package I thought it would be good to open an issue for people like myself...

**To Reproduce**

I created a minimal test that works when using the method shipped in `@apollo/client` ([link](https://github.com/apollographql/apollo-client/blob/6f710004f3c61dad8b0c3b362827cf449d068cd5/src/react/hooks/useSyncExternalStore.ts#L25)).

In `hooks.test.js` under the `useSyncExternalStore` section:
```
it('works with useCallback', () => {
  let toggle;
  const App = () => {
    const [state, setState] = useState(true);
    toggle = setState.bind(this, () => false);

    const value = useSyncExternalStore(
      useCallback(() => {
        return () => {};
      }, [state]),
      () => state ? 'yep' : 'nope'
    );

    return <p>{value}</p>;
  };

  act(() => {
    render(<App />, scratch);
  });
  expect(scratch.innerHTML).to.equal('<p>yep</p>');

  toggle();
  rerender();

  expect(scratch.innerHTML).to.equal('<p>nope</p>');
});
```

When replacing `useSyncExternalStore` with the function that's shipped with `@apollo/client`, the value updates properly.

**Workarounds**
Using `patch-package` (😄) to null out the `useSyncExternalStore` import from `React` works, leading to the shipped method in the library to be used.

JoviDeCroock:
fixes #3654

It looks like Apollo and other libs can have their snapshot changed during render time, i.e. not part of the subscribe, this means that we need to check in a layout effect whether we are due to re-render, luckily this is similar how React does it as well 😅

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -140,7 +140,7 @@ export function useSyncExternalStore(subscribe, getSnapshot) {
 
 	useLayoutEffect(() => {
 		if (value !== state) {
-			setState(value);
+			setState(() => value);
 		}
 	}, [subscribe, value, getSnapshot]);
 
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -136,8 +136,13 @@ export const useInsertionEffect = useLayoutEffect;
 export function useSyncExternalStore(subscribe, getSnapshot) {
 	const [state, setState] = useState(getSnapshot);
 
-	// TODO: in suspense for data we could have a discrepancy here because Preact won't re-init the "useState"
-	// when this unsuspends which could lead to stale state as the subscription is torn down.
+	const value = getSnapshot();
+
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
🎉

Issue #3662, opened 2022-08-11
Infinite loop when using zustand


theninthsky:
Preact version: `10.10.2`.

**Bug Description**

When using [zustand](https://www.npmjs.com/package/zustand), the app freezes and enters an infinite loop.

**To Reproduce**

Clone my repo: https://github.com/theninthsky/client-side-rendering/tree/preact
Run `npm i` and `npm start`

You can visit the preview deploy here: https://e578fdf5.client-side-rendering.pages.dev

Investigation so far: https://github.com/preactjs/preact/issues/3661#issuecomment-1211641512

**Expected behavior**
Integration with zustand should work.

theninthsky:
@JoviDeCroock
Everything works properly now. Thanks!

PR #3663, opened 2022-08-11
fix sync-external-store with zustand

JoviDeCroock:
fixes https://github.com/preactjs/preact/issues/3662

When zustand is used they use the [withSelector](https://github.com/facebook/react/blob/main/packages/use-sync-external-store/src/useSyncExternalStoreWithSelector.js#L39) variant of sync-external-store. This relies on `getSnapshot` not being incorporated in the `useEffect` dependency array.

Creating a test for it atm


marvinhagemeister (approved):
🎉

PR #3670, opened 2022-08-15
fix case where we set SCU multiple times

### LINKED ISSUE #3669: Component not re-rendered after sequence of state updates which leave some states unchanged
When an event handler combines several updates to two states, such that one of the states is unchanged at the end and the other is changed, the component does not re-render. The code below works as expected in Preact 10.10.0 but not in Preact 10.10.1 and 10.10.2.

**Test case:**

```js
import { h, render } from "preact";
import { useState } from "preact/hooks";

function TestWidget() {
  const [saved, setSaved] = useState(false);
  const [saving, setSaving] = useState(false);

  const onSave = (event) => {
    event.preventDefault();

    // Clicking the button does not show the text "Saved!" as expected in Preact 10.10.2.
    //
    // If the `setSaving` lines are commented out, it works.

    setSaving(true);
    setSaved(true);
    setSaving(false);
  };
  return (
    <div>
      <button onClick={onSave}>Save</button>
      {saved && "Saved!"}
    </div>
  );
}

const appEl = document.querySelector("#app");
render(<TestWidget />, appEl);
```

**Steps to reproduce**

Repro: https://codesandbox.io/s/compassionate-jepsen-lp5xyqq4vz

Click on the "Save" button. The text "Saved!" should appear after it.

In Preact 10.10.0 this works. In Preact 10.10.1 and 10.10.2 it does not.

JoviDeCroock:
Fixes #3669

We were setting `hookState._component.__hooks._hasScuFromHooks = true;` rather than `hookState._component._hasScuFromHooks` which becomes problematic as we start setting as many SCU's as there are stateful hooks in a component.

On the bright side found a golfing opportunity

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -43,8 +43,11 @@ options._render = vnode => {
 			hooks._pendingEffects = [];
 			currentComponent._renderCallbacks = [];
 			hooks._list.forEach(hookItem => {
+				if (hookItem._nextValue) {
+					hookItem._value = hookItem._nextValue;
+				}
 				hookItem._pendingValue = EMPTY;
-				hookItem._pendingArgs = undefined;
+				hookItem._nextValue = hookItem._pendingArgs = undefined;
 			});
 		} else {
 			hooks._pendingEffects.forEach(invokeCleanup);
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -43,11 +43,8 @@ options._render = vnode => {
 			hooks._pendingEffects = [];
 			currentComponent._renderCallbacks = [];
 			hooks._list.forEach(hookItem => {
-				if (hookItem._nextValue) {
-					hookItem._value = hookItem._nextValue;
-				}
 				hookItem._pendingValue = EMPTY;
-				hookItem._nextValue = hookItem._pendingArgs = undefined;
+				hookItem._pendingArgs = undefined;
 			});
 		} else {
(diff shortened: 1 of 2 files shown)


hooks/src/index.js
```diff
@@ -43,11 +43,8 @@ options._render = vnode => {
 			hooks._pendingEffects = [];
 			currentComponent._renderCallbacks = [];
 			hooks._list.forEach(hookItem => {
-				if (hookItem._nextValue) {
```
JoviDeCroock, line 46: This can be removed as we are returning `return hookState._nextValue || hookState._value;` from useState either way

marvinhagemeister (approved):
Good catch 👍

hooks/test/browser/useState.test.js
```diff
@@ -234,6 +234,31 @@ describe('useState', () => {
 		expect(scratch.textContent).to.equal('hi');
 	});
 
+	it('correctly updates with multiple state updates', () => {
```
robertknight, line 237: I'd suggest to add a comment mentioning the original issue, since the sequence of updates needed to trigger the problem is quite specific, rather than this being just a general test for multiple states in one component.

robertknight (approved):
The code change looks straightforward. I confirm this fixes the test failures encountered in our app when going from Preact 10.10.0 => 10.10.1: https://github.com/hypothesis/lms/pull/4335.

PR #3671, opened 2022-08-15
Fix hooks calling shouldComponentUpdate without context

developit:
This was breaking code that relies on `this` within shouldComponentUpdate.

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -184,7 +184,7 @@ export function useReducer(reducer, initialState, init) {
 		if (!currentComponent._hasScuFromHooks) {
 			currentComponent._hasScuFromHooks = true;
 			const prevScu = currentComponent.shouldComponentUpdate;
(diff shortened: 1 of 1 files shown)


JoviDeCroock (approved):

hooks/src/index.js
```diff
@@ -193,7 +193,7 @@ export function useReducer(reducer, initialState, init) {
 				// When we have no updated hooks in the component we invoke the previous SCU or
 				// traverse the VDOM tree further.
 				if (allHooksEmpty) {
-					return prevScu ? prevScu(p, s, c) : true;
+					return prevScu ? prevScu.call(this, p, s, c) : true;
```
jridgewell, line 196: Which `this` are you trying to propagate? This one is saved when the `useReducer()` call happened, not when `component.shouldComponentUpdate()` was called.

developit: whoooops, good catch.

PR #3675, opened 2022-08-17
ensure we iterate over all hooks

### LINKED ISSUE #3674: State update issue in Preact 10.10.3
**Describe the bug**

I encountered a scenario in Preact 10.10.3 where a state update does not work after the initial render. The test case below works in earlier Preact releases (I tested 10.10.0-10.10.2).

**Test case:**

```js
import { h, render } from "preact";
import { useState } from "preact/hooks";

function Widget() {
  const [, setUnusedState] = useState(0);
  const [suggestionsListOpen, setSuggestionsListOpen] = useState(false);

  const open = () => {
    // Test works if line below is commented out
    setUnusedState(42);
    setSuggestionsListOpen(true);
  };

  const close = () => {
    setSuggestionsListOpen(false);
  };

  return (
    <div>
      <button onClick={open}>Open</button>
      <button onClick={close}>Close</button>
      <div data-testid="state">{suggestionsListOpen ? "open" : "closed"}</div>
    </div>
  );
}

const root = document.querySelector("#app");
render(<Widget />, root);
```

**Steps to reproduce**

1. Open the sandbox at https://codesandbox.io/s/preact-10-10-3-state-updates-7xs51b, which uses the test case above
2. Click on "Open".
3. Click on "Close"

**Expected behavior:** The text below the button should change to "open" after clicking the "Open" button and change to "closed" after clicking the "Close" button.

**Actual behavior:** The text changes to "Open" after clicking "Open" but does not update after clicking "Close".

After clicking the "Open" button a second time, the Open and Close buttons then behave as expected. The issue has something to do with the sequence of states that are updated. See the note above the code which calls `setUnusedState`.

In Preact 10.10.2 and earlier, the example works as expected.

JoviDeCroock:
Fixes #3674 

Due to `.every` we bailed out of the loop on the first time we find a `false` which in this case means we update from 0 --> 42 for the unused hook and never set the `_nextValue` to `_value` for the second update

CC @robertknight

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -184,7 +184,7 @@ export function useReducer(reducer, initialState, init) {
 		if (!currentComponent._hasScuFromHooks) {
 			currentComponent._hasScuFromHooks = true;
 			const prevScu = currentComponent.shouldComponentUpdate;
-			currentComponent.shouldComponentUpdate = function (p, s, c) {
+			currentComponent.shouldComponentUpdate = function(p, s, c) {
 				if (!hookState._component.__hooks) return true;
 				const stateHooks = hookState._component.__hooks._list.filter(
 					x => x._component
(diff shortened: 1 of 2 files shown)


robertknight (approved):
I confirm this fixes the downstream test failures that were distilled into #3674. The diff looks sensible to me, although I don't have a clear understanding what the high-level intention of the `shouldComponentUpdate` handler generated by this code is. A comment above the block of code explaining the high-level picture would be useful.

Issue #3672, opened 2022-08-16
Types for FunctionalComponent doesn't allow returning just children


oliverfindl:
- [ ] Check if updating to the latest Preact version resolves the issue.

**Describe the bug:**
`FunctionalComponent` doesn't allow returning just `children`, although code compiles successfully (with `// @ts-ignore`) and works as expected.

**To Reproduce:**
```tsx
// DummyComponent.tsx

import { FunctionalComponent, h } from "preact";

const DummyComponent: FunctionalComponent = ({ children }) => {
    return children;
});

export default DummyComponent;
```

Error in VS Code:
`  Type 'ComponentChildren' is not assignable to type 'VNode<any> | null'. `

**Expected behavior:**
No errors.

developit:
I've run into this as well.

Issue #3611, opened 2022-07-09
Types for JSX.Element don't work for components that have children in their interface


lifeiscontent:
- [ ] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
if I have a component that returns children directly, it results in a type error.

**To Reproduce**

https://codesandbox.io/s/ecstatic-montalcini-ter683?file=/src/index.tsx

1. Go to 'index.tsx'
2. See type errors

**Expected behavior**
Not to have type error when returning children

ConcernedHobbit:
`ComponentChildren` is not necessarily one value. It can be an array, and you cannot return multiple siblings without a parent node.
Wrap the `Element`'s return value in a `Fragment`.

You could also do some custom validation to check that what has been passed to the function is just one renderable element.

lifeiscontent:
@ConcernedHobbit what you've described has nothing to do with the issue I've outlined, its a well known fact that the entire JSX ecosystem cannot return strongly typed children, however things like Arrays are returnable by React and other frameworks its still considered a single value but with a fragment you're adding a `DocumentFragment` to the element which might introduce a runtime difference in certain code.

ConcernedHobbit:
@lifeiscontent How do you suppose your element would work in JSX typing? Is the intended behaviour that when it is a child component it would understand that and would not error in typing?
Consider 
`render(<Example><p>Hello</p><p>World</p></Example>, ...)`
which is essentially
`render(<p>Hello</p><p>World</p>, ...)`
or doing the equivelant with an abstraction into the App component. 
I don't think React supports this with `FC` or `ReactNode` typings, is there a semantic difference with preact here?

JoviDeCroock:
My fix introduced a bug, undid for now but will keep looking at it

lifeiscontent:
@ConcernedHobbit Yeah, basically any value that react accepts as children I'd expect preact to do the same, and as just raw JS, preact appears to have the same behavior as React here.

e.g: `null`, `array` `boolean` `number` `string` etc

PR #3676, opened 2022-08-17
allow function component with children

JoviDeCroock:
fixes https://github.com/preactjs/preact/issues/3672
fixes https://github.com/preactjs/preact/issues/3611


marvinhagemeister (approved):
🎉

PR #3680, opened 2022-08-19
undo jsx-element removal

### LINKED ISSUE #3679: Type for `Element` changed to `ElementType`: intended?
Before we update our types, just checking if it was intentional to remove the `Element` type and replace it with `ElementType`? Getting a TS error with `10.10.4` for `h.JSX.Element`

https://github.com/preactjs/preact/commit/7d54883b6e45b3d0cc670254c0078cee3a2aaa87#diff-814abe9ce1d0e89a7fa1ef941142487296d1423ffad605dc1188b544056c2bcfL29

JoviDeCroock:
Resolves #3679 
Re-opens #3611

While working on this I did not realise that we have an override in our tests to make `JSX.Element` work https://github.com/preactjs/preact/blob/master/test/ts/jsx-namespacce-test.tsx#L7 which lead me to believe that TS constructs its own given the `ElementType` you export. This was wrong, this re-introduces the bug of `children` which is a more subtile issue with how we typed `FunctionComponent` and `ClassComponent`

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -87,7 +87,7 @@ export type ComponentProps<
 	: never;
 
 export interface FunctionComponent<P = {}> {
-	(props: RenderableProps<P>, context?: any): ComponentChild;
+	(props: RenderableProps<P>, context?: any): VNode<any>;
 	displayName?: string;
 	defaultProps?: Partial<P>;
 }
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -36,6 +36,7 @@ export namespace JSXInternal {
 					: never;
 		  }[keyof IntrinsicElements]
 		| ComponentType<P>;
+	export interface Element extends VNode<any> {}
 	export type ElementClass = Component<any, any> | FunctionComponent<any>;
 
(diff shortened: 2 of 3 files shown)


marvinhagemeister (approved):

src/index.d.ts
```diff
@@ -87,7 +87,7 @@ export type ComponentProps<
 	: never;
 
 export interface FunctionComponent<P = {}> {
-	(props: RenderableProps<P>, context?: any): ComponentChild;
+	(props: RenderableProps<P>, context?: any): VNode<any>;
```
piotr-cz, line 90: This didn't revert return type to `VNode<any> | null`, as it was in v10.10.3: https://github.com/preactjs/preact/pull/3676/files#diff-619cc2239a7eb4ef05041767e6ab0e4fb069e98ee9d2c83c2d8e1865c475e141L90

Returning `null` from components is valid, as described here:
React Docs > [Preventing Component from Rendering](https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering)

PR #3688, opened 2022-08-20
fix grid properties

JoviDeCroock:
Fix grid and other style properties as defined in https://github.com/preactjs/preact/issues/2607

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,3 +1,3 @@
 export const EMPTY_OBJ = {};
 export const EMPTY_ARR = [];
-export const IS_NON_DIMENSIONAL = /acit|ex(?:s|g|n|p|$)|rph|grid|ows|mnc|ntw|ine[ch]|zoo|^ord|itera/i;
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):

src/constants.js
```diff
@@ -1,3 +1,3 @@
 export const EMPTY_OBJ = {};
 export const EMPTY_ARR = [];
-export const IS_NON_DIMENSIONAL = /acit|ex(?:s|g|n|p|$)|rph|grid|ows|mnc|ntw|ine[ch]|zoo|^ord|itera/i;
+export const IS_NON_DIMENSIONAL = /^(-|f[lo].*[^se]$|g.{5,}[^ps]$|z|o[pr]|(W.{5})?[lL]i.*(t|mp)$|an|(bo|s).{4}Im|sca|m.{6}[ds]|ta|c.*[st]$|wido|ini)/;
```
developit, line 3: We already special-case leading `-` and don't add px units.

```suggestion
export const IS_NON_DIMENSIONAL = /^(f[lo].*[^se]$|g.{5,}[^ps]$|z|o[pr]|(W.{5})?[lL]i.*(t|mp)$|an|(bo|s).{4}Im|sca|m.{6}[ds]|ta|c.*[st]$|wido|ini)/;
```

developit (commented):
This regex is scary.

Issue #2612, opened 2020-07-09
TypeScript types on input/button/etc without JSX


zachasme:
### Reproduction

https://codesandbox.io/s/throbbing-wildflower-jr1t4?file=/index.tsx

### Steps to reproduce

I am trying to use Preact *with* TypeScript but *without* JSX. However, without JSX any regular components (div/input/etc) apparently are typed as generic components.

### Expected Behavior

```ts
// Assuming this line typechecks
const doesWork = <input onInput={onInput} />;
// I would expect the following line to be equivalent
const doesNotWork = h("input", { onInput: onInput });
```

### Actual Behavior

The second line (without JSX) complains that

```
No overload matches this call.
  Overload 1 of 2, '(type: string, props: HTMLAttributes<EventTarget> & SVGAttributes<SVGElement> & Record<string, any>, ...children: ComponentChildren[]): VNode<...>', gave the following error.
    Type '({ currentTarget }: TargetedEvent<HTMLInputElement, Event>) => void' is not assignable to type 'GenericEventHandler<EventTarget> & GenericEventHandler<SVGElement>'.
      Type '({ currentTarget }: TargetedEvent<HTMLInputElement, Event>) => void' is not assignable to type 'GenericEventHandler<SVGElement>'.
        Types of parameters '__0' and 'event' are incompatible.
          Type 'TargetedEvent<SVGElement, Event>' is not assignable to type 'TargetedEvent<HTMLInputElement, Event>'.
            Type 'TargetedEvent<SVGElement, Event>' is not assignable to type '{ readonly currentTarget: HTMLInputElement; }'.
              Types of property 'currentTarget' are incompatible.
                Type 'SVGElement' is missing the following properties from type 'HTMLInputElement': accept, align, alt, autocomplete, and 69 more.
  Overload 2 of 2, '(type: ComponentType<{ onInput: ({ currentTarget }: TargetedEvent<HTMLInputElement, Event>) => void; }>, props: Attributes & { ...; }, ...children: ComponentChildren[]): VNode<...>', gave the following error.
    Argument of type '"input"' is not assignable to parameter of type 'ComponentType<{ onInput: ({ currentTarget }: TargetedEvent<HTMLInputElement, Event>) => void; }>'.ts(2769)
```

PR #3690, opened 2022-08-24
improve types for bare createElement and h calls

### LINKED ISSUE #3618: Types for props in h / createElement function doesn't allow ref
- [ ] Check if updating to the latest Preact version resolves the issue.

**Describe the bug:**
`h` / `createElement` function doesn't allow `ref` in `props`, although code compiles successfully.

**To Reproduce:**
```jsx
// DummyComponent.tsx

import { FunctionalComponent, h } from "preact";
import { useRef } from "preact/hooks";

const DummyComponent: FunctionalComponent = ({ children }) => {
    const node = useRef<HTMLDivElement>();
    return h("div", { ref: node }, children);
});

export default DummyComponent;
```

Error in VS Code:
`Type 'MutableRef<HTMLDivElement | undefined>' is not assignable to type '(Ref<EventTarget> & Ref<SVGElement>) | undefined'. `

**Expected behavior:**
No errors.

JoviDeCroock:
Resolves #3618 
Resolves https://github.com/preactjs/preact/issues/2612

This fine-tunes the element creation calls to derive more state from the context in which they are called. This also makes a special case for input so we can correctly leverage onInput

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -219,7 +219,7 @@ export function createElement<T extends HTMLElement>(
 ): VNode<any>;
 export function createElement<P>(
 	type: ComponentType<P>,
-	props: (ClassAttributes & P) | null,
+	props: (Attributes & P) | null,
 	...children: ComponentChildren[]
 ): VNode<any>;
 export namespace createElement {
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
This is great!

PR #3697, opened 2022-08-28
fix: mac arm build

gengjiawen:
This internally bump `brotli-size`, the old version contains cpp module which not compatible on m1 mac.

The `d.ts` looks auto updated with microbundler update. Not sure should i revert it or leave it be.


JoviDeCroock (commented):
This seems to change a bit more than just the brotli-size version, would it be worth just bumping brotli in our package json?

rschristian:
Newer `microbundle` versions will generate `.d.ts` files if `"types"` is specified in the package file. Will need to run it with `--no-generateTypes` to ensure hand-written types aren't overwritten if brotli for some reason can't be bumped on its own.

developit:
Good point @rschristian - @gengjiawen can you re-do this PR adding `--no-generateTypes` to each microbundle script?  The result should (hopefully) be a PR that only modifies `package.json` and `package-lock.json`.

gengjiawen:
> Good point @rschristian - @gengjiawen can you re-do this PR adding `--no-generateTypes` to each microbundle script? The result should (hopefully) be a PR that only modifies `package.json` and `package-lock.json`.

done.

rschristian:
Has the build size increased? We (unfortunately) don't seem to run compressed size action if the source hasn't been touched.

Last time I checked, updating Microbundle increased the output size by a non-negligible amount.

gengjiawen:
Size not changed much 
on master
```
-rw-r--r-- 1 gitpod gitpod  10K Aug 30 02:05 preact.js
-rw-r--r-- 1 gitpod gitpod 9.9K Aug 30 02:04 preact.min.js
-rw-r--r-- 1 gitpod gitpod 9.9K Aug 30 02:04 preact.mjs
-rw-r--r-- 1 gitpod gitpod 9.9K Aug 30 02:04 preact.module.js
-rw-r--r-- 1 gitpod gitpod  11K Aug 30 02:04 preact.umd.js
```

on this patch

```
-rw-r--r-- 1 gitpod gitpod 9.9K Aug 30 01:44 preact.js
-rw-r--r-- 1 gitpod gitpod 9.9K Aug 30 01:44 preact.min.js
-rw-r--r-- 1 gitpod gitpod 9.9K Aug 30 01:44 preact.mjs
-rw-r--r-- 1 gitpod gitpod 9.9K Aug 30 01:44 preact.module.js
-rw-r--r-- 1 gitpod gitpod  11K Aug 30 01:44 preact.umd.js
```

rschristian:
Rounding (or truncating) to the nearest 100b doesn't exactly give the clearest view. Probably doesn't make much sense to increase output size to solve a build-time platform compat issue.

Can `brotli-size` be upgraded independently as Jovi originally suggested? Microbundle will likely need to be updated at some point but, as it seems there are implications for doing so and it involves more moving pieces (pkg.exports, post-build scripts, etc), I'm not sure if it's worth touching to fix this specific issue if it can be avoided.

JoviDeCroock:
The thing that scares me here is that this seems to be generating additional files 😅

marvinhagemeister (approved):
This is great! Thank you so much for taking the time to address feedback and hanging with us 👍 

@JoviDeCroock @rschristian The current iteration works fine for me and seems to generate the same output as before. I don't see any additional files being generated on my end.

JoviDeCroock:
@marvinhagemeister generates an extra .mjs file for all packages, if that isn't an issue I am okay with merging

```
       1381 B: hooks.mjs.gz
       1277 B: hooks.mjs.br
```

JoviDeCroock:
Want to do that in a follow-up or do we do that now? 😅

marvinhagemeister:
I'm fine with a follow up. I think we dragged this PR long enough.

rschristian:
The resulting `.mjs` files are Microbundle's modern builds, not the desired copy of `.module.js`. Will follow up.

marvinhagemeister:
@rschristian oh, well in that case we should probably keep our script around. Currently it copies the `.module.js` file and overwrites `*.mjs` anyway.

rschristian:
> @rschristian oh, well in that case we should probably keep our script around. Currently it copies the `.module.js` file and overwrites `*.mjs` anyway.

Made a PR but elected to disable the modern output. Figured relying on the overwrite would eventually trip someone up or create issues.

Issue #3542, opened 2022-05-12
v11 ability to specify render() order in parentNode's children


Colisan:
Hi, in [the current documentation](https://preactjs.com/guide/v10/api-reference/#render) it's written that

> The replaceNode-argument will be removed with Preact v11

But no replacement is specified. What is the new way to tell, for example "render this vnode between the 4th and the 5th child of this parent dom element"? I don't see any parameter to render() that currently allow such a configuration.

If already determined, can we add this information into the documentation?

danielweck:
Try this :)

https://gist.github.com/developit/f4c67a2ede71dc2fab7f357f39cff28c

developit:
Heh, thanks @danielweck as always :)

We'll include a nice guide and shim library based on that gist with the 11 launch.

MicahZoltu:
I would like to make sure this doesn't get dropped, but I'm not sure how.  As I understand it, this is still an issue 2 years later (at least, v10 still says deprecated and still points people at that gist I believe), but the code doesn't actually pass type checking (it is passing incompatible objects into into functions) and, worse, it directly references preact internal properties which means if those property names ever change anyone using this code will break.

Is the idea that v11 will provide a properly implemented function with a signature of
```ts
createRootFragment(parent: ContainerNode, replaceNode: ContainerNode | ContainerNode[]): ContainerNode
```
and users who are using the code in the linked gist are basically just using a shim that only needs to work until v11 is released, at which point they'll just delete the shim code entirely and import the function with the same name from v11 library?

The code was recently published to https://github.com/preactjs/preact-root-fragment, suggesting this is in fact a final solution and it says it is compatible with Preact 11 suggesting that people will be expected to use this code long term.  However, this published version contains all of the same problems as the gist which is concerning.

Perhaps the issue is that Preact's types are wrong, and this library just lies about the types to work around it?  Any chance of fixing preact's types if that is the case?

Is there any way to avoid referencing internal preact properties (which I think may be post-mangled names even)?

rschristian:
> v10 still says deprecated and still points people at that gist I believe

Yep, only just made the repo public. Will update that comment shortly though, it's on my to-dos.

> but the code doesn't actually pass type checking (it is passing incompatible objects into into functions) and, worse, it directly references preact internal properties which means if those property names ever change anyone using this code will break.

[...]

PR #3700, opened 2022-08-30
chore: Add deprecation JSDoc for render()'s `replaceNode`

rschristian:
I noticed we have not marked `replaceNode` as deprecated despite it [already being removed in v11](https://github.com/preactjs/preact/blob/56ae63920c65f31e7e10f7df31c5024a71590775/src/render.js#L9).

Unfortunately there seems to be no way to mark a function parameter as deprecated, only the entire function signature, hence why the overload was used.


src/index.d.ts
```diff
@@ -275,6 +275,11 @@ export namespace h {
 // Preact render
 // -----------------------------------
 
+export function render(
+	vnode: ComponentChild,
+	parent: Element | Document | ShadowRoot | DocumentFragment,
+): void;
+/** @deprecated Will be removed in v11. */
```
rschristian, line 282: Unsure of what we want to suggest in the message. Maybe link to https://gist.github.com/developit/f4c67a2ede71dc2fab7f357f39cff28c until [better docs are written](https://github.com/preactjs/preact/issues/3542#issuecomment-1145445352)?

developit: That would be good, yep.

marvinhagemeister (approved):

marvinhagemeister (approved):
Good catch!

PR #3099, opened 2021-04-02
(restructure) - Remove replace node

JoviDeCroock:

PR #3701, opened 2022-08-30
fix: render()'s TS type still included `replaceNode` in v11

rschristian:
Missed (or perhaps more likely skipped) in #3099


marvinhagemeister (approved):
Good catch 💯

PR #3704, opened 2022-09-02
Various perf and size improvements

JoviDeCroock:
- shorter functions for easy optimization
- getting rid of rendererState
- make mount and patch component more inline

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -3,17 +3,9 @@ import options from './options';
 import { createVNode, Fragment } from './create-element';
 import { patch } from './diff/patch';
 import { DIRTY_BIT, FORCE_UPDATE, MODE_UNMOUNTING } from './constants';
-import { getParentContext, getParentDom } from './tree';
+import { getParentDom } from './tree';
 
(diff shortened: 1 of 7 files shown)


src/component.js
```diff
@@ -12,9 +12,7 @@ export let ENABLE_CLASSES = false;
  * @type {import('./internal').RendererState}
  */
 export const rendererState = {
```
JoviDeCroock, line 14: With something like https://github.com/preactjs/preact/pull/3493 in mind the state could still be a good option @developit

JoviDeCroock:
https://github.com/preactjs/preact/pull/3704/commits/4e31d427804b9758bc2ae42fe8334cdc1c0ccbf0 seems to make `filter_list` ever so slightly slower as opposed to keeping it on the `renderState` I think this could be due to traversing the context upwards on a regular basis with patches

marvinhagemeister (approved):
This is great! Find it a little bit easier to read too without the property access stuff  💯

developit (approved):
Hell yeah

PR #3709, opened 2022-09-02
Cleanup _parent, _dom and __hooks after unmount

JoviDeCroock:
Possible fix for https://github.com/preactjs/preact/issues/3668

The only thing we can’t cleanup during unmount are the `_children` due to this https://github.com/preactjs/preact/blob/master/compat/src/suspense.js#L148 and the `_dom._listeners` due to

```
FAILED TESTS:
    ✖ should allow component to re-suspend using normal suspension mechanics after initial suspended hydration resumes
      Chrome Headless 105.0.5195.52 (Mac OS 10.15.7)

    Error: Uncaught TypeError: Cannot read properties of undefined (reading 'clickfalse') (compat/test/browser/suspense-hydration.test.js:426)
        at  (compat/test/browser/suspense-hydration.test.js:654:23)
```

PR #3710, opened 2022-09-02
Always clone VNodes before using them

developit:
If the perf impact isn't huge, this might be a way to avoid the need to clean up VNodes (#3709).


Thompson1985 (approved):

JoviDeCroock:
Closing: exploratory perf spike tied to the #3709 vnode-cleanup era; v11 internals were rewritten.

PR #3743, opened 2022-09-22
Avoid synchronously adding setState callbacks

### LINKED ISSUE #3742: `componentDidUpdate` affects component state

**Describe the bug**

`componentDidUpdate` affects component state

**To Reproduce**

https://codesandbox.io/s/crimson-worker-1koisq?file=/src/index.js

1. Open Console, it shows `{ count: null, isLoading: false }`
image

2. Comment the line of `componentDidUpdate`, refresh demo page, in console, it shows `{ count: { value: 10 }, isLoading: false }`
image

**Expected behavior**

Whether or not there is a `componentDidUpdate` event, the console should be `{ count: { value: 10 }, isLoading: false }`

JoviDeCroock:
Fixes #3742

The issue in question is caused by our mutable _renderCallbacks and the timing of which we push our new state-setting callback into it.

In the test-scenario we have a component that mounts without children due to a `loading` state variable being true, imagine a spinner or something instead here.
After our component mounts we set this variable to `false` effectively mounting our child-component which has a `componentDidMount` of itself which calls a setState on the parent, due to this `setState` we will synchronously push our `setState` callback onto the `_renderCallbacks` which is the one from the render we are still processing.

Now why do we call it with `componentDidUpdate` and not without? Well we push a `_renderCallback` when we are updating from loading true to false which puts this parent component into the renderQueue which we consume in the inverse order of the tree (child --> parent) this leads us to calling the setState callback when the intermediary render completes (read not the one where we update the parent state).

To solve this we ensure that we only push this callback when we are consuming the render coupled to this dirty state by means of adding a `_stateCallbacks` mechanism that is consumed when the component is handled.

### DIFF
--- a/mangle.json
+++ b/mangle.json
@@ -44,6 +44,7 @@
       "$_force": "__e",
       "$_nextState": "__s",
       "$_renderCallbacks": "__h",
+      "$_stateCallbacks": "_sb",
       "$_vnode": "__v",
       "$_children": "__k",
       "$_pendingSuspensionCount": "__u",
--- a/src/component.js
+++ b/src/component.js
@@ -47,7 +47,9 @@ Component.prototype.setState = function(update, callback) {
 	if (update == null) return;
 
 	if (this._vnode) {
-		if (callback) this._renderCallbacks.push(callback);
+		if (callback) {
+			this._stateCallbacks.push(callback);
+		}
 		enqueueRender(this);
 	}
 };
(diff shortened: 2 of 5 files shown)


marvinhagemeister (approved):
Great work! Impressed with how quickly you found a solution 🙌

zjhch123 (approved):
Thank you very much for your quick fix!

developit:
I'd like a bit of time to noodle on size for this one if we can. Performance numbers look good though.

PR #3747, opened 2022-09-26
Support signal values for html + svg attributes

marvinhagemeister:
This PR updates our JSX type definitions to support passing signals as attributes. In a perfect world we'd be able to do that from `@preact/signals`, but due to the various declaration merging rules and other factors, this would lead to ambiguity and ordering issues. See: https://github.com/microsoft/TypeScript/issues/50808

Whilst we could nudge users to copy & paste a type definition file into their own project, this would be less than ideal for newcomers or users who solely use JavaScript. So after a bit of back and forth and a bit chatter with @developit, we both felt that enhancing the types here is the most DX friendly way for our users.

Looking at the changes you might be curios why I went the verbose way of adding the signals type for every single property value instead of relying on generics or clever intersection types. To be honest: This was the only way TypeScript would pick the Signals type up. Whenever I tried any other approach, TS would drop the whole interface resolution completely. This makes me suspicious of them having special behavior for JSX types or something. Not sure. Either way I think this isn't too bad and workable. The DX improvement for end users is definitely there and imo outweighs a slight annoyance on our part.

Fixes https://github.com/preactjs/signals/issues/106

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -64,11 +64,17 @@ export namespace JSXInternal {
 		cssText?: string | null;
 	}
 
+	export interface SignalLike<T> {
+		value: T;
+		peek(): T;
+		subscribe(fn: (value: T) => void): () => void;
+	}
+
 	export interface SVGAttributes<Target extends EventTarget = SVGElement>
 		extends HTMLAttributes<Target> {
-		accentHeight?: number | string;
-		accumulate?: 'none' | 'sum';
-		additive?: 'replace' | 'sum';
+		accentHeight?: number | string | SignalLike<number | string>;
+		accumulate?: 'none' | 'sum' | SignalLike<'none' | 'sum'>;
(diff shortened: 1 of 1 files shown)


rschristian (approved):
My knowledge of (and willingness to appease) TS beyond basic types is non-existant so I cannot speak for whether there are possible alternatives, but this is working nicely and is super easy to extend in the future so looks great to me!

marvinhagemeister:
FWIW: I found a way with generics that requires less typing on our part, but it's worse compared to the changes in the PR because TS doesn't unwrap the type in IDE hints.

With the changes in this PR for the `checked` attribute TS displays: `checked?: boolean | SignalLike<boolean> | undefined`.

With some generic voodoo, it would display: `checked?: Attr<boolean> | undefined` instead, which kinda masks that you can pass a signal. Personally, I still prefer the way this PR displays it.

JoviDeCroock (approved):

PR #3755, opened 2022-10-03
chore(compat): add package.json to exports

akselander:
# Summary

When I switched from webpack aliasing react to preact, to doing it on package level in an Next.JS codebase I ran into a problem - storybook was throwing error - webpack was trying to access `package.json`, but it wasn't defined in the `exports`. 

Adding this inside of `package.json` helped my case, but if this not is not how we should approach it, please guide me towards a better solution.

### DIFF
--- a/compat/package.json
+++ b/compat/package.json
@@ -44,6 +44,10 @@
 		"./scheduler": {
 			"import": "./scheduler.mjs",
 			"require": "./scheduler.js"
+		},
+		"./package.json": {
+			"import": "./package.json",
+			"require": "./package.json"
 		}
 	}
 }
--- a/compat/package.json
+++ b/compat/package.json
@@ -45,9 +45,6 @@
 			"import": "./scheduler.mjs",
 			"require": "./scheduler.js"
 		},
-		"./package.json": {
-			"import": "./package.json",
-			"require": "./package.json"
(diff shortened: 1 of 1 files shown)


compat/package.json
```diff
@@ -44,6 +44,10 @@
 		"./scheduler": {
 			"import": "./scheduler.mjs",
 			"require": "./scheduler.js"
+		},
+		"./package.json": {
+			"import": "./package.json",
+			"require": "./package.json"
 		}
```
marvinhagemeister, line 51: This can be simplified a bit, since the formats do not differ. See the main `package.json` file:

https://github.com/preactjs/preact/blob/284a8b0d969f8ef3a3eeff0107b0104a52f774ad/package.json#L89

```suggestion
		"./package.json": "./package.json"
```

marvinhagemeister (commented):
Awesome! Thanks a bunch for the PR. Let's simplify it a bit since `.json` doesn't need format specifiers and then it's ready to be merged 👍

compat/package.json
akselander: Thanks, makes sense!

marvinhagemeister (approved):
Awesome, thanks for the PR 🎉

PR #3760, opened 2022-10-09
Adding better support for hooks related to `time-slicing`(concurrent) ⏲️ 

sanjaiyan-dev:
### Adding better support for concurrency⏲️ 

Running the concurrent task in idle period using [`window.requestIdleCallback`](https://developer.chrome.com/blog/using-requestidlecallback/).

> - Adding support for running the concurrental tasks by running it in idle period without blocking other urgent functions. (including `scroll`, `touch`, `click`...)
> - Mocking `useTransition` behaviour using `useState` hook.

_Extremely sorry if I made any mistakes :(_

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -39,15 +39,9 @@ const requestIdleCallbackPreact =
 		self.requestIdleCallback &&
 		self.requestIdleCallback.bind(window)) ||
 	function (cb) {
-		let start = Date.now()
-		return setTimeout(function () {
-			cb({
-				didTimeout: false,
-				timeRemaining: function () {
-					return Math.max(0, 50 - (Date.now() - start))
-				},
-			})
-		}, 3)
+		return queueMicrotask(() => {
+			cb();
+		});
 	}
 
 
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -34,6 +34,23 @@ import {
 	__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED
 } from './render';
 
+const requestIdleCallbackPreact =
+	(typeof self !== 'undefined' &&
+		self.requestIdleCallback &&
+		self.requestIdleCallback.bind(window)) ||
+	function (cb) {
+		let start = Date.now()
+		return setTimeout(function () {
+			cb({
+				didTimeout: false,
+				timeRemaining: function () {
+					return Math.max(0, 50 - (Date.now() - start))
+				},
+			})
+		}, 3)
+	}
+
+
 const version = '17.0.2'; // trick libraries to think we are react
 
 /**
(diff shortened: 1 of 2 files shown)


sanjaiyan-dev:
### Adding better concurrency fallback ⏲️ 

Running the concurrent task in idle period using [`window.requestIdleCallback`](https://developer.chrome.com/blog/using-requestidlecallback/).

> - Adding support for running the concurrental tasks by running it in idle period without blocking other urgent functions. (including `scroll`, `touch`, `click`...)
> - Mocking `useTransition` behaviour using `useState` hook.


_Extremely sorry if I made any mistakes :(_

compat/src/index.js
```diff
@@ -119,15 +130,27 @@ const flushSync = (callback, arg) => callback(arg);
 const StrictMode = Fragment;
 
 export function startTransition(cb) {
-	cb();
+	requestIdleCallbackPreact(() => {
+		cb();
+	});
```
marvinhagemeister, line 135:
```suggestion
	requestIdleCallbackPreact(cb);
```

compat/src/index.js
```diff
@@ -119,15 +130,27 @@ const flushSync = (callback, arg) => callback(arg);
...
 
 export function useDeferredValue(val) {
 	return val;
 }
 
 export function useTransition() {
-	return [false, startTransition];
+	const [isTransitionPending, setIsTransitionPending] = useState(false);
+
+	const startTransitionFunction = (cb) => {
```
marvinhagemeister, line 145: The function identity will change with every render. Let's find a way to keep the reference stable.

compat/src/index.js
```diff
@@ -34,6 +34,17 @@ import {
...
+const requestIdleCallbackPreact =
+	(typeof self !== "undefined" &&
+		self.requestIdleCallback &&
+		self.requestIdleCallback.bind(window)) ||
+	function (cb) {
+		return queueMicrotask(() => {
+			cb();
+		});
+	};
+
```
marvinhagemeister, line 46: I'm not sure if we need to bind the this value explicitly. Maybe we can simplify this a bit?

```suggestion
const requestIdleCallbackPreact =
	typeof requestIdleCallback === 'function'
		? requestIdleCallback
		: queueMicrotask;
```

marvinhagemeister (commented):
This is a great start 🎉 Left some comments, let me know what you think.

compat/src/index.js
sanjaiyan-dev: Hi @marvinhagemeister , Currently i updated to `setTimeout` since it need to be cleared in unmount. (Can you please help me with it)

compat/src/index.js
sanjaiyan-dev: using use callback have any effect on it ?

> The function identity will change with every render. Let's find a way to keep the reference stable.

sanjaiyan-dev:
🤔

JoviDeCroock:
Closing due to inactivity

PR #3769, opened 2022-10-15
Add support for passing event options

### LINKED ISSUE #428: Support passive event listeners
React is trying to figure out how to handle this too, in https://github.com/facebook/react/issues/6436 . Would love to see this added in somehow in preact. It's a fantastic capability for battling jank, and keeping things interactive and buttery smooth.

Explainer doc on passive event listeners: https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md

marvinhagemeister:
- Add support for passing event handler options
- Compat: Automatically make `touchstart`, `touchmove` and `wheel` handlers passive by default, see: https://github.com/facebook/react/blob/main/packages/react-dom-bindings/src/events/DOMPluginEventSystem.js?rgh-link-date=2022-10-15T11%3A29%3A18Z#L432-L447

Example:

```jsx
const fn = () => console.log("clicked");

<button onClick={[fn, { passive: true }]}>click me</button>
```

Fixes #428 .

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -54,7 +54,7 @@ function setStyle(style, key, value) {
  * @param {boolean} isSvg Whether or not this DOM node is an SVG node or not
  */
 export function setProperty(dom, name, value, oldValue, isSvg) {
-	let useCapture;
+	let useCapture, eventOptions;
 
 	o: if (name === 'style') {
 		if (typeof value == 'string') {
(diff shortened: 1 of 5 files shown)


src/diff/props.js
```diff
@@ -90,12 +90,17 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
 		else name = name.slice(2);
 
 		if (!dom._listeners) dom._listeners = {};
+		if (!useCapture && Array.isArray(value)) {
```
JoviDeCroock, line 93: What happens here when the user changes from passive to active? We would need to re-register the event-handler which might be unexpected that we aren't doing that. It did make me wonder whether the `-capture` would be a good fit for the `-passive` as well. The only other options are `once` and `signal` which I am not sure would fit here reliably....

marvinhagemeister: @JoviDeCroock According to MDN only the `capture` option is checked for matching event listeners. See: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#matching_event_listeners_for_removal

marvinhagemeister:
Marking as a draft, because browsers seems to turn scroll events into passive automatically in all browsers. Seems like the chrome DevTools warning is outdated? The passive option was the only reason going for implementing it. Leaving it here in case of future considerations.

Thompson1985 (approved):

JoviDeCroock:
Closing this as stale

PR #3771, opened 2022-10-16
Skip HTML comments during hydration

### LINKED ISSUE #79: Normalized onChange event?
In React, the `onChange` event is normalized across browsers (at least as far back as IE9) so that it fires any time the content of an input changes (e.g. when the user types, backspaces, deletes, etc). Is there a similarly normalized event in preact?

Thanks!

### LINKED ISSUE #3767: Hydration removes and add nodes again when there are HTML comments just before them

**Describe the bug**

Preact unnecessarily removes and adds the DOM nodes that it finds after HTML comments during hydration.

```html
<div id="root">
    <div>I'm going to be preserved in the DOM</div>
    <!-- I'm a comment, so I'll be removed, which is also fine -->
    <div>I'll be removed and added back again</div>
</div>
```

**To Reproduce**

I reproduced the bug in this repository: https://github.com/luisherranz/preact-comments-hydration-bug ([Open with Stackblitz Codeflow](https://stackblitz.com/~/github.com/luisherranz/preact-comments-hydration-bug))

And this is a branch with a [Mutation Observer added to the current hydration tests](https://github.com/preactjs/preact/compare/master...luisherranz:preact:bug/hydrating-html-with-comments-removes-dom-elements).

1. Add a comment to the SSR'ed HTML
2. Monitor changes to the DOM
2. Hydrate it with Preact
3. The nodes after the HTML comment are removed and added again

_There is a video with a longer explanation on [this other issue](https://github.com/WordPress/block-hydration-experiments/issues/79)._

**Expected behavior**

Preact should remove the comments, but it should preserve the rest of the DOM nodes.

marvinhagemeister:
We'd unnecessarily move nodes when the HTML we're going to hydrate contains comments. This is expected because it's sort of a mismatch between the SSR'd tree vs the virtual tree. But it turns out that this feature is needed for streaming SSR as comment nodes are used to indicate hydration boundaries.

Fixes #3767 .

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -325,13 +325,12 @@ function placeChild(
 	if (nextDom !== undefined) {
 		oldDom = nextDom;
 	} else {
-		oldDom = newDom.nextSibling;
-	}
-
-	// Skip over comment nodes. Will be removed right after calling diffChildren
-	// in diff so that the vnode tree matches the DOM tree again.
-	while (oldDom !== null && oldDom.nodeType === 8) {
-		oldDom = oldDom.nextSibling;
+		oldDom = newDom;
+		// Skip over comment nodes. Will be removed right after calling diffChildren
+		// in diff so that the vnode tree matches the DOM tree again.
+		do {
+			oldDom = oldDom.nextSibling;
+		} while (oldDom !== null && oldDom.nodeType === 8);
 	}
 
 	return oldDom;
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -328,6 +328,12 @@ function placeChild(
 		oldDom = newDom.nextSibling;
 	}
 
+	// Skip over comment nodes. Will be removed right after calling diffChildren
+	// in diff so that the vnode tree matches the DOM tree again.
+	while (oldDom !== null && oldDom.nodeType === 8) {
+		oldDom = oldDom.nextSibling;
+	}
+
 	return oldDom;
 }
 
--- a/test/browser/hydrate.test.js
+++ b/test/browser/hydrate.test.js
@@ -50,11 +50,11 @@ describe('hydrate()', () => {
 	beforeEach(() => {
 		scratch = setupScratch();
 		attributesSpy = spyOnElementAttributes();
+		clearLog();
 	});
 
 	afterEach(() => {
 		teardown(scratch);
-		clearLog();
 	});
 
 	it('should reuse existing DOM', () => {
--- a/test/browser/hydrate.test.js
+++ b/test/browser/hydrate.test.js
@@ -92,6 +92,7 @@ describe('hydrate()', () => {
 			scratch
 		);
 		expect(scratch.innerHTML).to.equal('<p><i>0</i><b>1</b></p>');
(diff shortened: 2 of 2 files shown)


test/browser/hydrate.test.js
```diff
@@ -92,6 +92,7 @@ describe('hydrate()', () => {
 			scratch
 		);
 		expect(scratch.innerHTML).to.equal('<p><i>0</i><b>1</b></p>');
+		expect(getLog()).to.deep.equal(['Comment.remove()']);
```
developit, line 95: It's strange that we end up removing the comment. I would have thought it would be skipped and ignored?

andrewiggins: IIUC, the comment will get removed in diffElementNodes here: https://github.com/preactjs/preact/blob/master/src/diff/index.js#L447-L451

developit: Yeah. Ideally we would just walk over them and ignore, but I guess we need to do that in two places due to this being a new type of Node to deal with (vs skipped Text/Element nodes, which we always walk here)

marvinhagemeister: An alternative would be to special case a branch for selective hydration where we manually walk over the comment nodes as part of diff:

```js
if (vnode.type === Island) {
  // We know that island is surrounded with comment nodes
  oldDom = oldDom.nextSibling
}
```

Was thinking some more about comment nodes and I'm not sure how they would behave if we'd just skip over them. Imagine a scenario where we'll support selective hydration and have a non-hydrated subtree marked with comment nodes that is only activated when the user hovers with the mouse over the area or otherwise interacts with it. In that scenario we'd likely be rendering on the client already before all subtrees are hydrated.

This could have an effect on our ordering:

```js
<div>
  <A />
  <B />
  <NonHydrated />
  <C />
</div>
```

Now, if we render on the client and need to re-order nodes due to various conditional rendering, our skip comment logic would likely bias in favor of moving the comments at the end of the childNode list.

I'm wondering if that means that we'd need to move the comment node as part of `NonHydrated` here. That would ensure that the order is always correct.

src/diff/children.js
```diff
@@ -310,7 +310,12 @@ function placeChild(
 	if (nextDom !== undefined) {
 		oldDom = nextDom;
 	} else {
-		oldDom = newDom.nextSibling;
+		oldDom = newDom;
+		// Skip over comment nodes. Will be removed right after calling diffChildren
+		// in diff so that the vnode tree matches the DOM tree again.
+		do {
+			oldDom = oldDom.nextSibling;
+		} while (oldDom !== null && oldDom.nodeType === 8);
```
developit, line 318: I wonder if it would be possible to use a similar (or the same somehow?) check to the one we have for node types here?

https://github.com/preactjs/preact/blob/de08e919860de81ec5f1ee7a37d640eb1496c323/src/diff/index.js#L358-L359

The other one knows what type it's looking for, so maybe it isn't possible? Just trying to think of ways to avoid checking the `nodeType` getter.

JoviDeCroock:
Superseded by https://github.com/preactjs/preact/pull/4128

PR #3776, opened 2022-10-24
Use Object.is in useSyncExternalStore

### LINKED ISSUE #3774: Zustand NaN infinite loop

**Describe the bug**
If you set zustand store state value to `NaN` this will lead to an infinite re-render loop, I think this is due to the fact that `NaN !== NaN`, so Preact always thinks that a new value arrived and does a re-render. Everything works fine if I remove aliases in `webpack.config.js` and use React. I think in some place it is necessary to add an additional check or replace `===` with `Object.is()` function.

**To Reproduce**

https://github.com/zalishchuk/preact-zustand-nan-infinite-loop

1. Clone repository.
2. Install dependencies.
3. Use `yarn start` or `npm run start` to start a development server.
4. Click on the button in the opened window.

**Expected behavior**
No infinite loop, just changing state key value to `NaN`.

zalishchuk:
Fixes #3774

### DIFF
--- a/compat/src/index.js
+++ b/compat/src/index.js
@@ -27,6 +27,7 @@ import { Children } from './Children';
 import { Suspense, lazy } from './suspense';
 import { SuspenseList } from './suspense-list';
 import { createPortal } from './portals';
+import { is } from './util';
 import {
 	hydrate,
 	render,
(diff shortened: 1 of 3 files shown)


marvinhagemeister (commented):
This looks great, thank you so much for making a PR! Let's add a test case for this so that we don't ever regress on it in the future 👍

zalishchuk:
@marvinhagemeister I'm not sure how it would be correct to test the absence of an infinite loop, but I pushed a simple test similar to the original implementation, to test that `getSnapshot` can return `NaN` without causing an infinite loop.

marvinhagemeister:
@zalishchuk That test is perfect!

marvinhagemeister (approved):

PR #3777, opened 2022-10-25
Upgrade workflow actions

andrewiggins:
Upgrade the versions of the actions in our workflows to newer version and take advantage of some new features.

- Specify the our desired version of node in the root package.json using the "volta" key
- Specify all actions to use the node version specified in the volta key of the package.json ([`setup-node` action docs for this feature](https://github.com/actions/setup-node/blob/c96ab56c5b0ebc955a5dd6bb6f1813c2f91b8658/docs/advanced-usage.md#node-version-file))
- Use the `setup-node` built-in cache capability for npm to simplify our workflows


JoviDeCroock (approved):

marvinhagemeister (approved):

andrewiggins:
Hmm it appears that a new feature of npm v7 or v8 is that `prepare` scripts of locally installed dependencies are run when they are installed. That doesn't work for us here cuz I don't do an install at the repo root for each bench and instead build the repo once in the `prepare` job and copy the build output to each bench job (often it's much faster, 1-5s vs 30s for install).

Need to noodle on this some more. Maybe it's just easier to run the install & build in each bench :/ Could also explore installing directly from the tarball produced by the `prepare` job...

marvinhagemeister (approved):

marvinhagemeister:
@andrewiggins That's a bummer. Just grabbing the tarball sounds like the easiest approach for now.

PR #3779, opened 2022-10-26
fix case where keyed children would get removed

### LINKED ISSUE #3766: Preact doesn't preserve the DOM nodes of a keyed list when there are multiple DOM node children

**Describe the bug**

@darerodz discovered in our experiments that Preact recreates the DOM nodes of a keyed list (instead of preserving them) if the component with the key has multiple DOM node children.

It's easier to see than to explain, so here is a video:

https://user-images.githubusercontent.com/3305402/195852098-ddba5424-86fd-4b86-be12-7ffa5f0cfca5.mp4

I've used a text node (`{''}`) in the example because it's the simplest, but it happens with any node. For example:

```js
const User = ({ name, surname }) => {
  return (
    <>
      <span>Name: {name}</span>
      <span>Surname: {surname}</span>
    </>
  );
};
```

**To Reproduce**

- [Stackblitz with reproduction in Preact](https://stackblitz.com/edit/vitejs-vite-ujssbv?file=src%2Fmain.jsx)

1. Create a list
2. Create a component to encapsulate the items
3. Add the key in that component
4. If the component has multiple children, the DOM nodes are not preserved

**Expected behavior**

Preact should preserve the DOM nodes.

_For reference, this is the exact same code using React: [Stackblitz with React](https://stackblitz.com/edit/vitejs-vite-b5dcvx?file=src%2Fmain.jsx)_

JoviDeCroock:
Fixes #3766

When we have a fragment the DOM needs to dig deeper to find the matching elements, what happens is that we  remove 1 element which means when we always skip by 2 we can't find any elements anymore, I know we have went back and forth on removing/not-removing this quite a bit now but this is a really valid case imo.

We can still opt not to do this for unkeyed elements if we find the byte-offset worth which might be a bit expensive as we would have to carry the fact that there's a parent to the Fragment that has a key

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -293,7 +293,7 @@ function placeChild(
 			for (
 				let sibDom = oldDom, j = 0;
 				(sibDom = sibDom.nextSibling) && j < oldChildren.length;
-				j += 2
+				j += 1
 			) {
 				if (sibDom == newDom) {
 					break outer;
--- a/test/browser/fragments.test.js
+++ b/test/browser/fragments.test.js
@@ -689,8 +689,8 @@ describe('Fragment', () => {
 		expect(scratch.innerHTML).to.equal(htmlForFalse);
 		expectDomLogToBe(
 			[
-				'<div>fooHellobeep.insertBefore(<div>beep, <div>foo)',
-				'<div>beepbarHello.appendChild(<div>bar)'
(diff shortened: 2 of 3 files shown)


marvinhagemeister (approved):

andrewiggins:
Huh, I surprised this change fixes that issue. Definitely an important issue to fix ofc. I'm gonna look into this as well to refresh my memory on how this works.

PR #3783, opened 2022-10-29
Add an explicit export default for compatibility with esbuild

Verseth:
Hi, as I tried to switch from react to preact in my project which uses esbuild I encountered an error caused by a lack of explicit default exports.

> ✘ [ERROR] No matching export in "node_modules/react-dom/client.mjs" for import "default"
>
>    frontend/build/index.js:6:7:
>      6 │ import ReactDOM from 'react-dom/client'
>        ╵        ~~~~~~~~
>
>1 error

This was the only thing I had to change to make my app work.
Hope this will make someone's transition more seamless.

### DIFF
--- a/compat/client.mjs
+++ b/compat/client.mjs
@@ -15,3 +15,8 @@ export function hydrateRoot(container, children) {
 	hydrate(children, container)
 	return createRoot(container)
 }
+
+export default {
+	createRoot,
+	hydrateRoot
+}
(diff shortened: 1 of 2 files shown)


marvinhagemeister (commented):
Awesome, that's a good catch! Let's add a quick test [in this file](https://github.com/preactjs/preact/blob/master/compat/test/browser/exports.test.js) so that we ensure that we don't ever regress on this in the future and then it's ready to be merged 🎉

marvinhagemeister (approved):
Sweet! Thank you so much for the PR 👍

PR #3797, opened 2022-11-14
V11 optimizations 1

developit:


jsx-runtime/src/index.js
```diff
@@ -38,23 +41,13 @@ function createVNode(type, props, key, __source, __self) {
...
 		constructor: undefined,
 		_vnodeId: --vnodeId,
 		__source,
 		__self
 	};
 
-	// If a Component VNode, check for and apply defaultProps.
-	// Note: `type` is often a String, and can be `undefined` in development.
-	let defaults, i;
-	if (typeof type === 'function' && (defaults = type.defaultProps)) {
```
marvinhagemeister, line 51: I quite like the idea of making this a compat-only thing 👍

src/component.js
```diff
@@ -91,26 +91,25 @@ Component.prototype.render = Fragment;
...
-	}
+function renderQueuedInternal(internal) {
+	// Don't render unmounting/unmounted trees:
+	if (internal.flags & MODE_UNMOUNTING) return;
+
+	// Don't render trees already rendered in this pass:
+	if (!(internal.flags & DIRTY_BIT)) return;
+
+	const vnode = createElement(internal.type, internal.props);
+	vnode.props = internal.props;
```
marvinhagemeister, line 102: Curious as to why this line is needed. Does the diff logic in v11 check for referential equality on props or something?

marvinhagemeister (approved):
LGTM, quite like the idea of moving `defaultProps` to compat 👍  Left a Q on something, but feel free to merge regardless.

src/component.js
developit: I believe so, yeah. I put it in because the previous code was creating referentially-equal props, but it's possible we don't need it. 

IIRC we have code in the hooks state implementation that is maybe relying on this to differentiate between prop vs state updates.

andrewiggins (approved):

PR #3800, opened 2022-11-15
Setting translate through direct access does not work

### LINKED ISSUE #3799: translate="no" becomes translate="yes"

**Describe the bug**

**To Reproduce**

https://codesandbox.io/s/blissful-hertz-lyyv45?file=/src/index.js

1. render `<b translate="no">translate="no" attribute</b>`
2. but you got `<b translate="yes">translate="no" attribute</b>`

**Expected behavior**

should have "no" value

JoviDeCroock:
Fixes #3799 

This changes our JSX type to reflect the fact that we support Boolean as a value for the property not yes and no.

However in compat we reflect Reacts behavior of supporting yes | no

For the record, React does not support booleans here https://stackblitz.com/edit/stackblitz-starters-tp34wv?description=React%20%20%20TypeScript%20starter%20project&file=src%2FApp.tsx&title=React%20Starter


marvinhagemeister (commented):
Good catch! Let's add a test for that 👍

developit:
I feel like this might be better suited to compat. The correct usage would be `translate={false}`, right?

JoviDeCroock:
@developit it's a valid attribute value so I would say this belongs in core. The attribute is correctly treated with `setAttribute`

marvinhagemeister:
@developit both "yes" and "no" are valid values according to mdn: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/translate .

developit:
@marvinhagemeister those values are for the attribute, not the property.

JoviDeCroock:
Closing due to inactivity

trevoring-okta:
Hi! Is there an update on whether this will be merged and released?

trevoring-okta:
@JoviDeCroock 👋 Bumping my comment above about whether this PR will be merged in at any point? Our team had to do a manual patch of preact using `patch-package` to pull in this exact change that you have here and unblock our use of `translate`

andrewiggins (approved):

PR #3801, opened 2022-11-15
Debug: limit "object as children" error to elements

developit:
This should fix preactjs/preact-render-to-string#254 without incurring the performance overhead of converting primitives in JSX text positions to VNodes.

If this works, we can remove [these lines](https://github.com/preactjs/preact-render-to-string/blob/fac154450e5455446c1cf844f2e7cbb9e0bd4127/src/index.js#L140-L147) from render-to-string.

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -318,7 +318,7 @@ export function initDebug() {
 		// Putting this check in `options.diffed` ensures that
 		// `vnode._children` is set and that we only validate the children
 		// that were actually rendered.
-		if (typeof vnode.type === 'string' && vnode._children) {
+		if (vnode._children) {
 			vnode._children.forEach(child => {
 				if (typeof child === 'object' && child && child.type === undefined) {
 					const keys = Object.keys(child).join(',');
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Not sure why I went the other way. Agree that this is a better approach 👍

debug/src/debug.js
```diff
@@ -318,12 +318,9 @@ export function initDebug() {
 		// Putting this check in `options.diffed` ensures that
 		// `vnode._children` is set and that we only validate the children
 		// that were actually rendered.
-		if (vnode._children) {
+		if (typeof vnode.type === 'string' && vnode._children) {
```
marvinhagemeister, line 321: Wondering if we want to only do these checks on DOM nodes or also components.

developit: Ah good point - I was forgetting that `._children` is the materialized children, not `props.children` as passed to a component. Will remove.

developit:
Alright, I just manually confirmed that this change addresses preactjs/preact-render-to-string#254. Will open a PR there.

PR #3812, opened 2022-11-22
Preserve event handler return values

developit:
I'm not aware of any bugs caused by this, but currently we drop the return value of event handlers rather than returning it to the DOM. This corrects that.

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -147,9 +147,9 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
  * @private
  */
 function eventProxy(e) {
-	this._listeners[e.type + false](options.event ? options.event(e) : e);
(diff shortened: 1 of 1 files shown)


JoviDeCroock (approved):

marvinhagemeister (approved):

andrewiggins:
Perhaps this change is still good to do this but I just realized that in modern browsers the return value is ignored: [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#getting_data_into_and_out_of_an_event_listener:~:text=Event%20listeners%20only%20take%20one%20argument%2C%20the%20Event%20Object%2C%20which%20is%20automatically%20passed%20to%20the%20listener%2C%20and%20the%20return%20value%20is%20ignored.)

PR #3813, opened 2022-11-23
Undefined  can clear checked and value properties for custom elements

objerke:
This pull requests makes it possible to reset the `value` and `checked` properties on a custom element, by passing `undefined` as a prop value. Tests are added with the new behavior.

This is a follow up on https://github.com/preactjs/preact/pull/2472 which slightly altered the behavior of https://github.com/preactjs/preact/pull/2465 in terms of undefined values. Currently an `undefined` value is never set for the `value` and `checked` properties.

```javascript
it(`should clear existing 'value' property with ""`, () => {
    render(<test-component value={{ foo: 'bar' }} />, scratch);

    render(<test-component value={undefined} />, scratch);

    expect(scratch.querySelector('test-component').value).to.equal("");
});
```

We are using preact to render custom elements that uses properties to pass down data. The value can be cleared as a way to indicate an indeterminate state. I think it feels natural to allow the `value` and `checked` properties to function as any other property when it comes to custom elements.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -332,6 +332,8 @@ function diffElementNodes(
 	let oldProps = oldVNode.props;
 	let newProps = newVNode.props;
 	let nodeType = newVNode.type;
+	let isCustomElement = ('' + newVNode.type).includes('-');
+
 	let i = 0;
 
 	// Tracks entering and exiting SVG namespace when descending through the tree.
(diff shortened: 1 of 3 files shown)


Thompson1985 (approved):

JoviDeCroock:
Closing: v11 already coerces null/undefined to empty-string for all `name in dom` props (`src/diff/props.js`), delivering the requested checked/value behavior.

PR #3816, opened 2022-11-30
docs: Removes recommendation for `preact-cli`

rschristian:
At this point we can't really keep the recommendation for `preact-cli` in good faith. It's not in a good state, especially for users who are new to Preact. In lieu of another tool that we want to recommend in its place, the recommendation should be removed as (IMO) it's doing more harm than good.

Going to propose the same for the docs site.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <p align="center">
 <a href="https://preactjs.com" target="_blank">
-	
+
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Good call! We should probably point folks to the vite preset, fresh or astro in the long run.

JoviDeCroock (approved):

developit (approved):

PR #3825, opened 2022-12-10
Remove prevRef from internal

JoviDeCroock:
We can probably golf bytes by working around the keys on the ref object in the refs array.

Removing the lazy initialization saves another 13b


andrewiggins (approved):

JoviDeCroock:
Closing for now as the perf numbers are very negative 😅

developit:
Woot! Looks like conditional allocation has mostly fixed the performance. It increased filesize more though.

Let's revisit this as part of exploring the patch/mount ref issue we talked about today.

marvinhagemeister (approved):

PR #3833, opened 2022-12-22
give the internal a stable id based on both depth and initial vnodeId

JoviDeCroock:
We currently can't remove `_vnodeId` as we rely on [rotating those](https://github.com/preactjs/preact/blob/v11/src/diff/patch.js#L129) to check for strict equality.

This PR adds a stable identifier when we first create an internal, we base this off of the `_vnodeId` and the depth of the internal.

I think this could function as a good replacement to `useId`

TODO: we should later on verify with https://github.com/preactjs/preact/pull/2396 whether this still works correctly with depth ordering


JoviDeCroock:
Closing due to inactivity

PR #3834, opened 2022-12-22
Hoist depth sorting function

marvinhagemeister:
A simple test to see if this moves benchmark numbers at all.


andrewiggins:
I think #3943 supersedes this since it brings in #3921 from master into v11

PR #3838, opened 2022-12-23
apply perf improvements from RTS

JoviDeCroock:
This puts us on par with what we had in the newer RTS versions


server/test/render.test.js
```diff
@@ -860,7 +868,8 @@ describe('render', () => {
 	});
 
 	describe('state locking', () => {
-		it('should set _dirty to true', () => {
+		// TODO: we use flags now
+		it.skip('should set _dirty to true', () => {
```
JoviDeCroock, line 872: Skipped for now due to usage of flags

server/test/render.test.js
```diff
@@ -1079,7 +1088,8 @@ describe('render', () => {
 		});
 	});
 
-	it('should invoke option hooks', () => {
+	// We call these with internals now
+	it.skip('should invoke option hooks', () => {
```
JoviDeCroock, line 1092: Skipped for now as we have no proper way to assert internals

server/test/render.test.js
andrewiggins: Could we open an issue to track updating this or removing it if we determine it no longer makes sense?

andrewiggins (approved):
Woot!

server/test/render.test.js
JoviDeCroock: Will do

marvinhagemeister (approved):

server/test/render.test.js
JoviDeCroock: Opened #3859

PR #3842, opened 2022-12-27
ES6-in-browser compatibility

kungfooman:
This allows developers to use the code without installing even one package.

We all know the npm meme: https://www.youtube.com/watch?v=mQif8VDt7nE

It doesn't need to be like that.

**test.html**:

```html
<html>
    <head>
        <script type="importmap">
            {
                "imports": {
                    "preact": "/preact/src/index.js"
                }
            }
        </script>
    </head>
    <body>        
        <script type="module">
            import { h, Component, render } from 'preact';
            import htm from 'https://unpkg.com/htm?module';
            // Initialize htm with Preact
            const html = htm.bind(h);
            function App(props) {
                return html`<h1>Hello ${props.name}!</h1>`;
            }
            render(html`<${App} name="World" />`, document.body);
        </script>
    </body>
</html>
```

### DIFF
--- a/src/clone-element.js
+++ b/src/clone-element.js
@@ -1,5 +1,5 @@
-import { assign, slice } from './util';
-import { createVNode } from './create-element';
+import { assign, slice } from './util.js';
+import { createVNode } from './create-element.js';
 
 /**
  * Clones the given VNode, optionally adding attributes/props and replacing its children.
--- a/src/component.js
+++ b/src/component.js
@@ -1,7 +1,7 @@
-import { assign } from './util';
-import { diff, commitRoot } from './diff/index';
-import options from './options';
-import { Fragment } from './create-element';
+import { assign } from './util.js';
+import { diff, commitRoot } from './diff/index.js';
+import options from './options.js';
+import { Fragment } from './create-element.js';
 
 /**
  * Base Component class. Provides `setState()` and `forceUpdate()`, which
--- a/src/create-context.js
+++ b/src/create-context.js
@@ -1,4 +1,4 @@
-import { enqueueRender } from './component';
+import { enqueueRender } from './component.js';
 
 export let i = 0;
 
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -1,5 +1,5 @@
-import { slice } from './util';
-import options from './options';
+import { slice } from './util.js';
+import options from './options.js';
 
 let vnodeId = 0;
 
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -1,7 +1,7 @@
-import { diff, unmount, applyRef } from './index';
-import { createVNode, Fragment } from '../create-element';
-import { EMPTY_OBJ, EMPTY_ARR } from '../constants';
-import { getDomSibling } from '../component';
+import { diff, unmount, applyRef } from './index.js';
+import { createVNode, Fragment } from '../create-element.js';
+import { EMPTY_OBJ, EMPTY_ARR } from '../constants.js';
+import { getDomSibling } from '../component.js';
 
 /**
  * Diff the children of a virtual node
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -1,10 +1,10 @@
-import { EMPTY_OBJ } from '../constants';
-import { Component, getDomSibling } from '../component';
-import { Fragment } from '../create-element';
-import { diffChildren } from './children';
-import { diffProps, setProperty } from './props';
-import { assign, removeNode, slice } from '../util';
-import options from '../options';
+import { EMPTY_OBJ } from '../constants.js';
+import { Component, getDomSibling } from '../component.js';
+import { Fragment } from '../create-element.js';
+import { diffChildren } from './children.js';
+import { diffProps, setProperty } from './props.js';
+import { assign, removeNode, slice } from '../util.js';
+import options from '../options.js';
 
 /**
  * Diff two virtual nodes and apply proper changes to the DOM
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -1,5 +1,5 @@
-import { IS_NON_DIMENSIONAL } from '../constants';
-import options from '../options';
+import { IS_NON_DIMENSIONAL } from '../constants.js';
+import options from '../options.js';
 
 /**
  * Diff the old and new properties of a VNode and apply changes to the DOM node
--- a/src/index.js
+++ b/src/index.js
@@ -1,13 +1,13 @@
-export { render, hydrate } from './render';
+export { render, hydrate } from './render.js';
 export {
 	createElement,
 	createElement as h,
 	Fragment,
 	createRef,
 	isValidElement
-} from './create-element';
-export { Component } from './component';
-export { cloneElement } from './clone-element';
-export { createContext } from './create-context';
-export { toChildArray } from './diff/children';
-export { default as options } from './options';
+} from './create-element.js';
+export { Component } from './component.js';
+export { cloneElement } from './clone-element.js';
+export { createContext } from './create-context.js';
+export { toChildArray } from './diff/children.js';
+export { default as options } from './options.js';
--- a/src/options.js
+++ b/src/options.js
@@ -1,4 +1,4 @@
-import { _catchError } from './diff/catch-error';
+import { _catchError } from './diff/catch-error.js';
 
 /**
  * The `option` object can potentially contain callback functions
--- a/src/render.js
+++ b/src/render.js
@@ -1,8 +1,8 @@
-import { EMPTY_OBJ } from './constants';
-import { commitRoot, diff } from './diff/index';
-import { createElement, Fragment } from './create-element';
-import options from './options';
-import { slice } from './util';
+import { EMPTY_OBJ } from './constants.js';
+import { commitRoot, diff } from './diff/index.js';
+import { createElement, Fragment } from './create-element.js';
+import options from './options.js';
+import { slice } from './util.js';
 
 /**
  * Render a Preact virtual node into a DOM element
--- a/src/util.js
+++ b/src/util.js
@@ -1,4 +1,4 @@
-import { EMPTY_ARR } from "./constants";
+import { EMPTY_ARR } from "./constants.js";
 
 /**
  * Assign properties from `props` to `obj`


JoviDeCroock:
The dist files already allow this, let's not alter source which is not intended for this

kungfooman:
> The dist files already allow this, let's not alter source which is not intended for this

Just because you don't intend to, doesn't mean others don't.

The example seems irrelevant, nowadays you also use import-maps for properly named modules and not some strange whatever-paths.

JoviDeCroock:
No need to put things like that, better to show the value as opposed to using package json export maps which is part of the spec. I assume this would use import maps to use the source files?

Do note that the source files are less performant and are bigger in size compared to our published folders... (comments,...). Also using a version we published is generally safer compared to using some source and... if you want to use source on a published version why not utilise the optimised files. 

Note that nothing here speaks against what you're selling, I am just telling you that the dist files are a better utilisation tartet as opposed to source. Is anything maybe not clear how this workflow works?

kungfooman:
> better to show the value as opposed to using package json export maps which is part of the spec.

Which spec? It's a bunch of bundlers that respect this setting, nothing designed by ES6 for the browser. To get rid of any server-side building tool (which is what this simple PR allows) you need to respect how ES6 works in browsers.

> I assume this would use import maps to use the source files?

I just used importmap to have a nice line like this:

```js
import { h, Component, render } from 'preact';
```

However, you can just drop importmap and write it like this:

```js
import { h, Component, render } from '/preact/src/index.js';
```

Which allows every browser with ES6 support to load the dependencies.

I just believe that importmaps is the best spec proposal so far and it has shims for several browsers, so right now it seems to be the best bet: https://github.com/guybedford/es-module-shims

> Do note that the source files are less performant and are bigger in size compared to our published folders... (comments,...).

This PR is mostly for other devs to quickly check out PR's and not waste time in whatever build tool the repo currently demands.

But even for local distribution/coding it is absolutely no problem. Every intranet has dozens of MB/s.

> Note that nothing here speaks against what you're selling, I am just telling you that the dist files are a better utilisation tartet as opposed to source. Is anything maybe not clear how this workflow works?

I perfectly well understand how build workflows work, I wrote my own bundlers and transpilers in Babel. I don't care about dist files, I want the real repository to develop preact and test PR's without filling my SSD with useless bundlers and extra large `node_modules` folders in every fork.

This PR simply makes sure that ES6 module path resolution spec is followed and not some random limited-to-bundlers ad-hoc package.json solutions.

rschristian:
> Which spec? It's a bunch of bundlers that respect this setting

Package exports are not just a bundler concept; [it's baked into Node](https://nodejs.org/api/packages.html#exports) and I imagine other runtimes such as Deno and Bun now.

> This PR is mostly for other devs to quickly check out PR's and not waste time in whatever build tool the repo currently demands.

The problem I foresee here is that the build tooling is an incredibly important part of any library; using the source, rather than the built output, is asking for trouble and discrepancies.

> I want the real repository to develop preact and test PR's without filling my SSD with useless bundlers and extra large node_modules folders in every fork.

I do sympathize, and while it's rather unfortunate, testing and/or benchmarking against the source, instead of the built output, means the results are mostly junk. Some modifications to the source may look and perform great, but, passed through necessary build tooling, end up as a net negative.

While this PR does open some neat possibilities, IMO, we don't want to encourage actually doing this. It's too easy to misuse or for people to think it's a viable strategy in production.

kungfooman:
> The problem I foresee here is that the build tooling is an incredibly important part of any library; using the source, rather than the built output, is asking for trouble and discrepancies.

It appears that you have been using bundlers for an extended period of time; perhaps it would be beneficial to take a break. The real ES6 source is - surprise - the source of truth. If there is any discrepancy in the bundler artifacts, it is the bundler that is at fault.

> I do sympathize, and while it's rather unfortunate, testing and/or benchmarking against the source, instead of the built output, means the results are mostly junk.

I never said to test or benchmark against the real source, this PR just helps people to get work done quicker. If you want to use your time typing useless commands in dozens of shells for everything PR/testing/coding, fine, but don't drag down the productivity of others. This PR doesn't stop your workflow, it simply enables another one.

> While this PR does open some neat possibilities, IMO, we don't want to encourage actually doing this. It's too easy to misuse or for people to think it's a viable strategy in production.

If people think using ES6 modules directly in production is a great idea, then that is on them and who are you actually to judge if that is a good or bad idea. There are reasons to actually do so, for example hacked CDN servers which leave uncountable websites open for abuse at once and webservers can specialize on distributing many yet small files over HTTP/S.

Even NPM itself is a security risk, I'm sure you heard about all kinds stories about NPM distributing malware etc.

rschristian:
> This PR doesn't stop your workflow, it simply enables another one.

One that could perhaps introduce issues or trip up future users, which is my worry here.

> who are you actually to judge if that is a good or bad idea.

Part of the job of a maintainer is to steer users toward happy paths and away from behaviors that will cause issues. The last thing we want is for users to struggle due to behavior we enabled. Additionally, time spent helping users with any issues (if any) arise from this is less time spent on other issues.

It's not necessarily against this but opening up a new method of consuming the library carries some risk and I don't see the benefits as being that great, personally.

kungfooman:
> Part of the job of a maintainer is to steer users toward happy paths and away from behaviors that will cause issues.

Naming security risks and time-wasting and useless dependencies the happy path is the icing on the cake here, thank you for the laugh, sir!

You have clearly exposed the inability to spot security and time wasting issues through the "perfect" use of straw man arguments.

JoviDeCroock:
I would suggest being less rude to people spending their time maintaining open source infrastructure. When multiple people point out issues that you aren't taking the time to even address then why would we spend our time coming to a shared understanding.

I'd love to encourage following the code of conduct here to get to a constructive feedback cycle rather than these passive aggressive passes.

kungfooman:
@JoviDeCroock I replied to both of your straw man arguments, while both of you ignore the reality of issues around npm. Now you even claim that you are patient while not taking one issue into account.

Talking about the beauty of a diverse set of opinions while beating down every other with straw man arguments. You are good at virtue signalling and that's it.

rschristian:
> Naming security risks

While there is indeed _some_ risk in using Node modules, that is risk the authors have made an educated decision to take on.

In the case of our tooling, we use [`microbundle`](https://github.com/developit/microbundle) which is (essentially) Preact's original tooling extracted out into a separate package. It's something we maintain and can address any issues that come up with it or its dependency chain if need be.

> time-wasting and useless dependencies

Please understand that we are people with limited time too, we'd love if we didn't have to build! However, builds are incredibly useful and provide a lot of value to this library; I don't think any of us are willing to hand-write the minified & mangled output that millions of devs rely upon every day. It's a tradeoff we've made.

As such, opening Preact to allow direct usage introduces a consistency issue. It's not pointless time wasting, its necessary to ensure quality.

kungfooman:
> In the case of our tooling, we use [`microbundle`](https://github.com/developit/microbundle) which is (essentially) Preact's original tooling extracted out into a separate package. It's something we maintain and can address any issues that come up with it or its dependency chain if need be.

> With the multitude of bundlers, runtimes, and module resolution techniques, this is a somewhat fragile ecosystem.

Okay, lets add 1 and 1 together here. So this community maintains [`microbundle`](https://github.com/developit/microbundle) and if there ever were an issue using proper ES6 URI's like in this PR (which it never should, because it's the way ES6 works), then you could fix it.

At the same time you claim everything is out of your control and the "fragility" is just too risky. Does this really add up for you?

> However, builds are incredibly useful and provide a lot of value to this library; I don't think any of us are willing to hand-write the minified & mangled output that millions of devs rely upon every day. It's a tradeoff we've made.

You are actually using a straw man *again*. You know exactly that this is not changing anything regarding the build process. It just keeps working. This PR simply opens the door for another one.

What will you argue next? That this PR causes you to write Lisp into punch cards?

rschristian:
> So this community maintains [microbundle](https://github.com/developit/microbundle) and if there ever were an issue using proper ES6 URI's like in this PR (which it never should, because it's the way ES6 works), then you could fix it.

Sorry, you might've misunderstood.

The issue isn't that Microbundle might struggle with your change, but that the wider ecosystem of tools might. If you were unaware, as Jovi mentioned we ship the source code alongside the built artifacts ([can take a peek here](https://www.runpkg.com/?preact@10.11.3/dist/preact.min.js)). The concern is that something or someone can start consuming source code rather than the built, which would not be good.

Frankly, I don't think you understand what a straw man is as we're not distorting any argument. We just don't think your personal concerns over NPM are enough to outweigh the risk and you haven't provided us with anything more.

---

I'm going to leave you with an excerpt from our [Code of Conduct](https://github.com/preactjs/preact/blob/master/CODE_OF_CONDUCT.md) and close this out. I'm sorry if we've frustrated you but this isn't the way to speak to people who are volunteering their time to you.

```
## Our Standards

Examples of behavior that contributes to creating a positive environment include:

* Using welcoming and inclusive language
* Being respectful of differing viewpoints and experiences
* Gracefully accepting constructive criticism
* Focusing on what is best for the community
* Showing empathy towards other community members

Examples of unacceptable behavior by participants include:

* The use of sexualized language or imagery and unwelcome sexual attention or advances
* Trolling, insulting/derogatory comments, and personal or political attacks
* Public or private harassment
* Publishing others' private information, such as a physical or electronic address, without explicit permission
* Other conduct which could reasonably be considered inappropriate in a professional setting
```

kungfooman:
Q: Why did the hypocrite write a Code of Conduct?
A: So they could point the finger at everyone else when they break it!

Jokes aside, thank you for gracefully accepting my viewpoints and for your time and inclusion.

At the end of the day, you are only hindering yourself. Keep strong. :muscle:

developit:
We don't support using Preact directly from `src/`, because the code in that directory is not functionally equivalent to the code we publish to npm. The current primary reason for this divergence is our property mangling configuration, but some of the things we've been working on (removal of Component from core, modular compat, etc) will further separate the two.

To your original question, it is already be possible to use Import Maps with Preact to avoid the need for bundlers or build tools:

```html
<html>
    <head>
        <script type="importmap">
            {
                "imports": {
                    "preact": "/preact/dist/preact.mjs"
                }
            }
        </script>
    </head>
    <body>        
        <script type="module">
            import { h, Component, render } from 'preact';
            import htm from 'https://unpkg.com/htm?module';
            // Initialize htm with Preact
            const html = htm.bind(h);
            function App(props) {
                return html`<h1>Hello ${props.name}!</h1>`;
            }
            render(html`<${App} name="World" />`, document.body);
        </script>
    </body>
</html>
```

Finally, while I would like to add file extensions to our imports, the reason would be to remove an unnecessary reliance on Node-style module resolution - not to enable usage from `src/` (which we cannot do for the above reasons).

PR #3854, opened 2023-01-07
ensure we respect max and min safe integers

JoviDeCroock:
Currently when `jsx` and `createElement` are used for a long time on the server or client we risk going over the max/min safe integer which in essence is a small risk but I think we should account for it as it's a crash risk

### DIFF
--- a/jsx-runtime/src/index.js
+++ b/jsx-runtime/src/index.js
@@ -43,7 +43,7 @@ function createVNode(type, props, key, __source, __self) {
 		key,
 		ref,
 		constructor: undefined,
-		_vnodeId: --vnodeId,
+		_vnodeId: (vnodeId = (vnodeId - 1) | 0),
 		__source,
 		__self
 	};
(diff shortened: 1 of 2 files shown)


jsx-runtime/src/index.js
```diff
@@ -43,7 +43,7 @@ function createVNode(type, props, key, __source, __self) {
 		key,
 		ref,
 		constructor: undefined,
-		_vnodeId: --vnodeId,
+		_vnodeId: (vnodeId = (vnodeId - 1) | 0),
```
andrewiggins, line 46: Note using [bitwise "or" operator restricts `vnodeId` to a 32-bit int](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_OR) (±2\**31) whereas the [safe integer is ±2**53](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger). I wonder what the tradeoff is between doing `| 0` vs `% 2**53-1` (precalcuating and storing `2**53-1`).

JoviDeCroock:
Closing due to inactivity

PR #3860, opened 2023-01-12
Move renderCallbacks to VNode and queue refs as renderCallbacks

andrewiggins:
Ref callbacks can be used for side-effects involving DOM nodes such as setting up IntersectionObservers or measuring width/height of DOM elements for manual positioning. However in Preact today, refs are applied while we are rendering in `diffChildren`. For updates, this is acceptable because the DOM nodes are already attached to the document. When mounting, we apply refs after appending a DOM child to its parent. However, if that DOM node's parent hasn't been mounted yet, then the ref callback is invoked before the DOM node is attached to the document. This PR changes our ref application logic to always apply refs as render/commit callbacks to ensure they are applied after DOM elements have been fully attached to the document.

Resurrect #2098 to fix issue with ref and effect ordering.


andrewiggins:
Note: Memory will increase slightly after this PR because I'm adding a new field to every VNode: `_renderCallbacks` which increases each VNode by ~4 bytes.

In v11 we can reclaim this back because we won't have _renderCallbacks on both Component and Internals. There will only be one on Internals.

JoviDeCroock:
Closing as fixed by https://github.com/preactjs/preact/pull/4054 and https://github.com/preactjs/preact/pull/4130

PR #3862, opened 2023-01-13
Clear css properties when passed undefined

### LINKED ISSUE #3861: Different behavior for keys with undefined properties in style objects using preact/compat
It seems as though there's an issue with how Preact handles style props when values are explicitly `undefined`. It seems to be a little more nuanced than it just leaving `undefined` in though – I think it happens only when the object property was _previously_ defined, and later becomes undefined:

**To Reproduce**

* [Preact demo](https://codesandbox.io/s/preact-undefined-style-properties-5n4giu?file=/src/App.js)
* [React demo](https://codesandbox.io/s/react-undefined-style-properties-sutlv2?file=/src/App.js)

**Expected behavior**

I would expect the `--x` css property to actually be deleted when it becomes `undefined`, as this is what happens on initial render.

andrewiggins:
When `style.setProperty` is given `null` or `''`, the browser effectively clears the value (at least for css custom properties). However, when passed `undefined`, the browser serializes `undefined` to the string`"undefined"` 😢 . This PR modifies our `setStyle` function to handle undefined and convert it to an empty string in this case.

Quick gut check: Is this something we want to support in core or put in compat? It's a fairly small change, but does fall into one of those changes where we are diverging from browser behavior so just wanted to check if that's something we want to put in core.

Fixes #3861

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -35,7 +35,7 @@ export function diffProps(dom, newProps, oldProps, isSvg, hydrate) {
 
 function setStyle(style, key, value) {
 	if (key[0] === '-') {
-		style.setProperty(key, value);
+		style.setProperty(key, value == null ? '' : value);
 	} else if (value == null) {
 		style[key] = '';
 	} else if (typeof value != 'number' || IS_NON_DIMENSIONAL.test(key)) {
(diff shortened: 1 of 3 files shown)


andrewiggins:
P.S. Is this something we can file as a spec bug/change to HTML or CSS? Or is this one of the can of worms not worth opening 😅?

JoviDeCroock (approved):

andrewiggins:
Gonna keep this in core since we already coerce undefined and null to empty string when setting style properties. This change makes our usage of setProperty aligned to what we do for style properties.

PR #3867, opened 2023-01-21
Changes `EventHandler<...>` to have a `this` of type `void`.

MicahZoltu:
Originally, `this` was set to `E['currentTarget']` which, based on the MDN doc linked in #2166, certainly sounds like the correct thing to do.  A couple years later it was changed to `never` in #3147 with no commentary on the PR or commit indicating why.  My guess is that this was changed so arrow functions would work, which I believe are permanently bound and cannot be rebound by the caller.

I believe that `never` is wrong because this code doesn't compile:
```ts
declare const event: JSX.TargetedEvent<HTMLElement, Event>

declare const apple: JSX.EventHandler<JSX.TargetedEvent<HTMLElement, Event>>
apple(event) // error TS2684: The 'this' context of type 'void' is not assignable to method's 'this' of type 'never'.

function Apple(model: Pick<JSX.HTMLAttributes<HTMLElement>, 'onInput'>) {
	model.onInput && model.onInput(event) // error TS2684: The 'this' context of type 'Pick<HTMLAttributes<HTMLElement>, "onInput">' is not assignable to method's 'this' of type 'never'.
}
```
Changing to `void` resolves this error and I believe will work everywhere that never worked.

I removed the comment since it no longer applies with this being `never` or `void`.  It appears that it was just forgotten when the switch from `E['currentTarget']` to `never` was made.

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -418,11 +418,7 @@ export namespace JSXInternal {
 	>;
 
 	export interface EventHandler<E extends TargetedEvent> {
-		/**
-		 * The `this` keyword always points to the DOM element the event handler
-		 * was invoked on. See: https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Event_handlers#Event_handlers_parameters_this_binding_and_the_return_value
-		 */
-		(this: never, event: E): void;
+		(this: void, event: E): void;
 	}
 
 	export type AnimationEventHandler<Target extends EventTarget> = EventHandler<
--- a/test/ts/Component-test.tsx
+++ b/test/ts/Component-test.tsx
@@ -2,12 +2,6 @@ import 'mocha';
 import { expect } from 'chai';
 import { createElement, Component, RenderableProps, Fragment } from '../../';
 
-// Test `this` binding on event handlers
-function onHandler(this: HTMLInputElement, event: any) {
-	return this.value;
-}
-const foo = <input onChange={onHandler} />;
-
 export class ContextComponent extends Component<{ foo: string }> {
 	getChildContext() {
 		return { something: 2 };


robertknight:
> A couple years later it was changed to never in https://github.com/preactjs/preact/pull/3147 with no commentary on the PR or commit indicating why.

@JoviDeCroock can you remember the rationale for the change?

MicahZoltu:
> You can't call a function if it's `this` is set to `never`, since TS will infer the type as `void`, so that definitely seems preferable if we wanted to prevent users from relying on using `this`.

I believe that `void` addresses this problem as `void` can't actually be used for anything, similar to `never`.

JoviDeCroock (approved):

MicahZoltu:
I get `window` for both when the function is not bound to anything.
image

When you bind an arrow function however, the binding is ignored (`this` always points to `globalThis`):
image

The underlying problem here is that JavaScript functions can be rebound.  Just because `this` is the element initially, doesn't mean that will be the case if someone calls the function elsewhere unless the function is explicitly bound.
image
image

The net takeaway here I think is that relying on `this` being the element could be safe, but preact would need to take care to ensure it doesn't accidentally rebind `this` along the way (which is easy to do if you are using classes).

MicahZoltu:
Reviewing the failing test, it appears that the test is incorrect here:
```tsx
// Test `this` binding on event handlers
function onHandler(this: HTMLInputElement, event: any) {
	return this.value;
}
const foo = <input onChange={onHandler} />;
```
Specifically, if we want to say that there are no guarantees that `this` is the element in these callbacks then the test is correctly failing now while it was incorrectly passing previously.  In the test case, if `this` was *not* an element the user would get an error about trying to access a property on undefined.

On the other hand, if we *want* to guarantee that `this` is the element, then we should switch the signature back to `E['currentTarget']`, and do whatever work is necessary to ensure that guarantee is made.

I'm happy to update the PR to "fix the test" (delete it), or switch back to `E['currentTarget']`.

MicahZoltu:
> Let's delete the test, I think we should be good to tell folks not to rely on `this` here

Done, and tests passing.

PR #3868, opened 2023-01-23
add exactOptionalPropertyTypes support to html attributes

deadem:
With [`exactOptionalPropertyTypes`](https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes) option enabled in tsconfig.json, assign undefined value to html attributes shouldn't generate an error.

```ts
const link: string | undefined
...
<a href={link}>Conditional link</a>
```
This code generates an error but should not.

React types declaration (for reference only): https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1917

[Preact playground](https://codesandbox.io/s/boring-dust-7ce1r5?file=/src/App.tsx)

[React playground](https://codesandbox.io/s/determined-hodgkin-75lrr3?file=/src/App.tsx)

This PR continues #3711

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -53,8 +53,8 @@ export type ComponentChild =
 export type ComponentChildren = ComponentChild[] | ComponentChild;
 
 export interface Attributes {
-	key?: Key;
-	jsx?: boolean;
+	key?: Key | undefined;
+	jsx?: boolean | undefined;
 }
 
 export interface ClassAttributes<T> extends Attributes {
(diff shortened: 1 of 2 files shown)


andrewiggins:
To verify my understanding, with this new option turned on, TypeScript no longer "auto adds" `undefined` to the type of a property marked as optional using `?`. Is that correct?

And this PR just adds `undefined` to all DOM properties so that we can pass `undefined` to them when this new option is turned on?

deadem:
> To verify my understanding, with this new option turned on, TypeScript no longer "auto adds" `undefined` to the type of a property marked as optional using `?`. Is that correct?
Yes.

> And this PR just adds `undefined` to all DOM properties so that we can pass `undefined` to them when this new option is turned on?
Yes.

marvinhagemeister (approved):

andrewiggins (approved):

Issue #3737, opened 2022-09-19
Preact adding elements out of order when replacing elements in DOM


jaredkrinke:
- [x] Check if updating to the latest Preact version resolves the issue (No, it doesn't)

**Describe the bug**
Preact seems to be adding HTML elements out of order when updating the DOM, given the right set of before/after elements. In my case, I'm swapping between two pieces of content.

This happens on 10.11.0 and 10.5.8, but does *not* happen on 10.5.7--so it seems like a regression in 10.5.8.

Tested on 10.11.0 in MS Edge Version 105.0.1343.42 on Windows (both 64-bit)

**To Reproduce**

Start with the sandbox:
https://codesandbox.io/s/preact-x-preact-cli-3-starter-vj285y2rn3

IMPORTANT: Upgrade to latest preact version! (Note: the dropdown is very finicky...)

Paste in this code:
```
import { Component, Fragment, render } from "preact";

class App extends Component {
  constructor(props) {
      super(props);
      this.state = { first: true };
  }

  render() {
      if (this.state.first) {
          return <Fragment>
              {<><p>1. Original item first paragraph</p></>}
              <p>2. Original item second paragraph</p>
              <button onClick={() => this.setState({ first: false })}>Click me</button>
          </Fragment>;
      } else {
          return <Fragment>
              <p>1. Second item first paragraph</p>
              {<Fragment>
                  <p>2. Second item second paragraph</p>
                  <div></div>
              </Fragment>}
          <button onClick={() => this.setState({ first: true })}>Click me</button>
          </Fragment>;
      }
  }
}

// Codesandbox workaround.
render(<App />, window.root);
```

1. Observe first set of elements in expected order
2. Click the button
3. Observe the second set of elements in expected order
4. Click the button

**Expected behavior**
The first set of elements should be shown in the same order as in the first step.

**Actual behavior**
The first set of elements is shown in the wrong order!

jaredkrinke:
For the record: in my case (which is fairly convoluted, but from which the repro above was derived), I was able to avoid this issue by adding a `key` attribute to distinguish different pieces of content that I swap between in my container.

PR #3875, opened 2023-02-01
Improve Fragment unmounting while correctly swapping nested fragments

### LINKED ISSUE #3814: Change in v10.11.1 causes inputs to lose focus on DOM changes
- [ ] Check if updating to the latest Preact version resolves the issue (no)

**Describe the bug**
Up to preact v10.11.0 changes in the DOM did not cause a focused input to lose focus. Starting with v10.11.1 the input loses focus.

**To Reproduce**

1. Check out https://github.com/jcestibariz/bug-preact-focus
2. Notice preact is locked to v10.11.0 which does *not* contain the bug
3. Run `npm i` and then `npm start`
4. In the browser click on the input and press backspace
5. Notice the last "tag" is removed but the input keeps focus.
6. Go back to the console and press Ctrl+C to stop the server
7. Run `npm i preact@10.11.1` and then `npm start`
8. Go back to the browser and repeat step 4
9. Notice now the input loses focus
10. You can also check with the latest version (v10.11.3 as of this moment) the bug is still present

**Expected behavior**
The input should not lose focus when DOM changes.

andrewiggins:
## Why does #3814 fail on master?

When transitioning from the tree

```jsx
<div>
  <Fragment>
    <span>0</span>
    <span>1</span>
  </Fragment>
  <input />
</div>
```

to the tree (note `<span>1</span>` was unmounted)

```jsx
<div>
  <Fragment>
    <span>0</span>
  </Fragment>
  <input />
</div>
```

the `master` branch has a bug where it will re-append all the children after the `Fragment`. This re-appending of the `<input>` causes it to lose focus. The reason for this re-appending is that when exiting `diffChildren` for the Fragment, it's `_nextDom` pointer is set to `<span>1</span>`. Since `<span>1</span>` is unmounted it's `parentNode` is `null`. When diffing the `input` element, we hit the `oldDom.parentNode !== parentDom` condition in `placeChild` which re-appends the `<input>` tag and sets oldDom to `null` causing all siblings of `<input>` to re-append.

When diffing components/Fragments, the `_nextDom` pointer should point to where in the DOM tree the diff should continue. So when unmounting `<span>1</span>`, the Fragment's `_nextDom` should point to the `<input>` tag. The previous code in `diffChildren` removed in #3738 was intended to fix this by detecting when `_nextDom` pointed to a node that we were unmounting and reset it to it's sibling. However, that code (copied below) had a correctness bug prompting its removal (#3737). Let's look at this bug.

```js
// Remove remaining oldChildren if there are any.
for (i = oldChildrenLength; i--; ) {
	if (oldChildren[i] != null) {
		if (
			typeof newParentVNode.type == 'function' &&
			oldChildren[i]._dom != null &&
			oldChildren[i]._dom == newParentVNode._nextDom
		) {
			// If the newParentVNode.__nextDom points to a dom node that is about to
			// be unmounted, then get the next sibling of that vnode and set
			// _nextDom to it
			newParentVNode._nextDom = getDomSibling(oldParentVNode, i + 1);
		}

		unmount(oldChildren[i], oldChildren[i]);
	}
}
```

## What caused #3737?

Here is a simplified repro of the issue from #3737:

```jsx
function App() {
	const [condition] = useState(true);
	return this.state.condition ? (
		// Initial render
		<>
			<div>1</div>
			<Fragment>
				<div>A</div>
				<div>B</div>
			</Fragment>
			<div>2</div>
		</>
	) : (
		// Second render: unmount <div>B and move Fragment up
		<>
			<Fragment>
				<div>A</div>
			</Fragment>
			<div>1</div>
			<div>2</div>
		</>
	);
}
```

The first render creates the DOM `1 A B 2` (each wrapped in divs) and when changing the `condition` state, it should rerender to produce `A 1 2` but instead we get `1 A 2`. Why?

When rerendering `App` we unmount `<div>B</div>`, which is a child of a Fragment. This unmounting triggers the call to `getDomSibling` in the code above. `getDomSibling` has a line of code in it that looks like `vnode._parent._children.indexOf(vnode)`. This line of code doesn't work in this situation because when rerendering a component, we only make a shallow copy of the old VNode tree. So when a child of `oldParentVNode` (the `App` component in this case) tries to access it's parent through the `_parent` pointer (e.g. the `Fragment` parent of `<div>A</div>`), it's `_parent` pointer points to the new VNode tree which we are in progress of diffing and modifying. Ultimately, this tree mismatch (trying to access the old VNode tree but getting the in-progress new tree) causes us to get the wrong DOM pointer and leads to incorrect DOM ordering.

In summary, when unmounting `<div>B</div>`, we need `getDomSibling` to return `<div>2</div>` since that is the next DOM sibling in the old VNode tree. Instead, because our VNode pointers are mixed at this stage of diffing, `getDomSibling` doesn't work correctly and we get back the wrong DOM element.

## Why didn't other tests catch this?

Other tests only do top-level render calls (e.g. `render(<App condition={true} />)` then `render(<App condition={false} />)`) which generate brand-new VNode trees with no shared pointers. They did not test renders originating from `setState` calls which go through a different code path and reuse VNode trees which share pointers across the old and new trees.

## The fix

The initial fix for this is to replace `getDomSibling` with a call to `dom.nextSibling` to get the actual next DOM sibling. (I found a situation in which this doesn't work optimally. I'll open a separate PR for that.)

## Final thoughts

One additional thought I have here is that walking through this has given me more confidence in our approach for v11. First, we do unmounts before insertions so we don't have to do this additional DOM pointer checking. Also, by diffing backwards, we ensure that our `_next` pointers are correct when we go to search what DOM element to insert an element before.

Fixes #3814, #3737

### DIFF
--- a/test/browser/focus.test.js
+++ b/test/browser/focus.test.js
@@ -13,6 +13,9 @@ describe('focus', () => {
 	/** @type {() => void} */
 	let rerender;
 
+	/** @type {(newState: Partial<{ before: number[]; after: number[] }>) => void} */
+	let setState;
+
 	/** @type {() => void} */
 	let prepend, append, shift, pop;
 
@@ -26,6 +29,8 @@ describe('focus', () => {
 				before: props.initialBefore || [],
 				after: props.initialAfter || []
 			};
+
+			setState = newState => this.setState(newState);
 
 			prepend = () => {
 				const before = this.state.before;
@@ -56,13 +61,14 @@ describe('focus', () => {
 				});
 			};
 
-			const liHtml = this.props.as == Input ? inputStr : span;
-			getDynamicListHtml = () =>
-				div([
+			getDynamicListHtml = () => {
+				const liHtml = this.props.as == Input ? inputStr : span;
+				return div([
 					...this.state.before.map(liHtml),
 					'<input id="input-0" type="text">',
 					...this.state.after.map(liHtml)
 				]);
+			};
 		}
 
 		render(props, state) {
@@ -375,6 +381,22 @@ describe('focus', () => {
 
 		expect(scratch.innerHTML).to.equal(getDynamicListHtml());
 		validateFocus(input, 'remove sibling before 2');
+	});
+
+	it('should  maintain focus when removing element directly before input', () => {
+		render(
+			<DynamicList initialBefore={[0, 1]} initialAfter={[2, 3]} />,
+			scratch
+		);
+
+		let input = focusInput();
+		expect(scratch.innerHTML).to.equal(getDynamicListHtml());
+
+		setState({ before: [0] });
+		rerender();
+
+		expect(scratch.innerHTML).to.equal(getDynamicListHtml());
+		validateFocus(input, 'remove sibling directly before input');
 	});
 
 	it('should maintain focus when adding input next to the current input', () => {
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -205,6 +205,17 @@ export function diffChildren(
 	// Remove remaining oldChildren if there are any.
 	for (i = oldChildrenLength; i--; ) {
 		if (oldChildren[i] != null) {
+			if (
+				typeof newParentVNode.type == 'function' &&
+				oldChildren[i]._dom != null &&
+				oldChildren[i]._dom == newParentVNode._nextDom
+			) {
+				// If the newParentVNode.__nextDom points to a dom node that is about to
+				// be unmounted, then get the next sibling of that vnode and set
+				// _nextDom to it
+				newParentVNode._nextDom = newParentVNode._nextDom.nextSibling;
+			}
+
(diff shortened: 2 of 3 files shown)


test/browser/focus.test.js
```diff
@@ -377,6 +383,22 @@ describe('focus', () => {
 		validateFocus(input, 'remove sibling before 2');
 	});
 
+	it('should  maintain focus when removing element directly before input', () => {
```
andrewiggins, line 386: This test cases mimics what the repro in #3814 does. I can also add that specific repro if you'd like

marvinhagemeister: Fine be me the way you did it in the PR 👍

marvinhagemeister (approved):

JoviDeCroock (approved):

PR #3879, opened 2023-02-03
use microtick outside of events

### LINKED ISSUE #3724: Issue (with material ui) after changing the default debounceRendering in 10.10
Hi,

I've found another case where the legacy debounceRendering mode was 'more compatible' with React.

Please see codepen below:

https://codesandbox.io/s/adoring-firefly-n9fjg2?file=/src/index.js:322-339

now, the issue is with the Birthday label wrongly animating on component rendering (the animation should not be seen).

To reproduce, reload the codepen with its button until you see the animation (it should be easily visible). In case codepen shows some errors, just reload again, and they should disappear.

Then uncomment the option line in code, try reloading many times, and the animation should not be showing anymore, which is the correct behavior.

Now I'm a bit confused about this and would like to bring this matter to your attention. I've seen that this change has fixed some old issues, but at the same time, it looks like new issues are arising with the changed debounce method. This one, another one in your issue tracker about react-big-calendar drag'n drop.

### LINKED ISSUE #3775: Problem with masked input after v10.9.0
Hi 👋 

I am pretty sure this is not a Preact issue but I would like to present my problem and ask for advice.

**Context**
- my team uses custom-made Preact UI framework and we have `DateInput` component which masks user input to display it as MM/DD/YYYY
- that same input returns `onInput` prop with the whole input event where we format the value to be YYYY-MM-DD bc that's what our BE uses
- this code has been in production for the past two years and we would like to avoid any significant refactoring around this issue
- after updating from 10.9.0 to 10.10.0 or later, date input started misbehaving

**To Reproduce**

Here is a reduced example: https://codepen.io/teodragovic/pen/PoaYyma?editors=0011
(just skip down past utilities)

1. Type numbers into the input
2. You should see flashes of `undefined-0-undefined` before input updates to correct value

**Problem**

Our component is updating `input` value using both `setState` and `event.target`. After updating from microtick to `setTimeout`  value that's set directly gets rendered before `setState` is flushed. We tried to avoid setting `event.target` value by cloning `event` into `onInput` but that caused our E2E cypress tests to fail due to a number of missing non-enumerable properties.

Our current solution is to set the value directly twice which is synchronous enough to avoid flashes of the second value. 

Basically, I'm asking if what we're doing is safe? Are there any other solutions we're missing? It's not imperative that we update from 10.9.0 yet but might as well try to find workarounds.

Thanks!

_EDIT: Changed version number from 1.9.0 to 10.9.0 - @marvinhagemeister_

JoviDeCroock:
Fixes #3775
Fixes #3724

When we are in events we need setTimeout to deal with how events bubble, we want the events to complete before we start our rendering process. This however isn't needed when we are dealing with internal updates coming from effects/...

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -2,6 +2,7 @@ import { assign } from './util';
 import { diff, commitRoot } from './diff/index';
 import options from './options';
 import { Fragment } from './create-element';
+import { inEvent } from './diff/props';
 
 /**
  * Base Component class. Provides `setState()` and `forceUpdate()`, which
@@ -183,6 +184,18 @@ let rerenderQueue = [];
 
 let prevDebounce;
 
+const microTick =
+	typeof Promise == 'function'
+		? Promise.prototype.then.bind(Promise.resolve())
+		: setTimeout;
+function defer(cb) {
+	if (inEvent) {
+		setTimeout(cb);
+	} else {
+		microTick(cb);
+	}
+}
+
 /**
  * Enqueue a rerender of a component
  * @param {import('./internal').Component} c The component to rerender
@@ -196,7 +209,7 @@ export function enqueueRender(c) {
 		prevDebounce !== options.debounceRendering
 	) {
 		prevDebounce = options.debounceRendering;
-		(prevDebounce || setTimeout)(process);
+		(prevDebounce || defer)(process);
 	}
 }
 
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -141,15 +141,29 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
 	}
 }
 
+export let inEvent = false;
+
 /**
  * Proxy an event to hooked event handlers
  * @param {Event} e The event object from the browser
  * @private
  */
 function eventProxy(e) {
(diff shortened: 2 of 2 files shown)


andrewiggins (approved):
Thanks for putting this together!

andrewiggins:
I took a look at the pref regressions in the 02_replace1k benchmark locally and I believe it is caused by a bug in our benchmark. In the change from microtask to setTimeout scheduling in #3608 some assumptions our benchmarks made about when work would be done changed (it previously assumed microtask, etc.). So some measures that previously measured JS execution now just measured the time it takes to scheduled work. As such the change back in this PR isn't comparing apples (preact-master) and oranges (preact-local).

In preact-master, the `run-*` measures only measure the time it takes to schedule rerender and the `duration` measure no longer includes browser layout & paint.

image

In preact-local (this branch), the `run-*` now measure the execution of the rerender and the `duration` measure covers the rerender + browser layout & paint. 

image

I suspect this behavior might change depending on how the browser decides to do scheduling of the `afterFrame` `postMessage` so the results from the GH actions might be slightly different. But at least locally things look comparable so I'm going to go ahead with these changes for now.

PR #3888, opened 2023-02-07
set dimensional attributes through `setAttribute`

### LINKED ISSUE #3887: Images with width=100% don't client side render correctly

**Describe the bug**
When client side rendering an image with width: 100%, the resulting image has a width of zero.

I think the issue is here specifically: https://github.com/preactjs/preact/blob/c483d9606b1ae62468ca07b4e27d478108249b81/src/diff/props.js#L121

It seems as though Preact tries to set `img.width = '100%'` and only tries to do `img.setAttribute('width', '100%')` if that fails. The issue is that `img.width = '100%'` does not cause an exception, but it does coerce the string to an integer and ultimately sets the image width to zero. `img.width` when accessed as as property is meant to be only an integer: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-width

**To Reproduce**
https://codesandbox.io/s/preact-images-csr-only-lrpkbp?file=/src/App.js

Load the page. Expected: the sandbox has an image with width=100%. Actual: the image is not visible since it has width=0

JoviDeCroock:
Fixes #3887 

This also uncovered another bug, when we pass `px` values to these properties it fails to set them as well

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -107,18 +107,7 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
 			// - xlink:href / xlinkHref --> href (xlink:href was removed from SVG and isn't needed)
 			// - className --> class
 			name = name.replace(/xlink(H|:h)/, 'h').replace(/sName$/, 's');
-		} else if (
-			name !== 'width' &&
-			name !== 'height' &&
-			name !== 'href' &&
-			name !== 'list' &&
-			name !== 'form' &&
-			// Default value in browsers is `-1` and an empty string is
-			// cast to `0` instead
-			name !== 'tabIndex' &&
-			name !== 'download' &&
-			name in dom
-		) {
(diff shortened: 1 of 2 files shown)


jpnelson (approved):
Yeah I think this would do the trick!

src/diff/props.js
```diff
@@ -108,6 +108,8 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
 			// - className --> class
 			name = name.replace(/xlink(H|:h)/, 'h').replace(/sName$/, 's');
 		} else if (
+			name !== 'width' &&
```
JoviDeCroock, line 111: We might be able to save bytes with a regex here but not sure  😅

JoviDeCroock:
The regex doesn't really save bytes so I will revert it

marvinhagemeister (approved):

PR #3891, opened 2023-02-10
Fix hydrating textarea with value prop

andrewiggins:
If a component rendered `<textarea value="Test" />`, on the server we'd render `<textarea>Test</textarea>`, but on the client when hydrating, we'd remove the children of `<textarea>` since the component didn't specify any (`textarea value="test" />`). Further, since we don't set attributes or properties on hydration, the value property doesn't get applied either.

This PR fixes this behavior by adding a new hook to compat which checks if the VNode is a textarea and always applying the `value` if the DOM value doesn't match.

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -223,6 +223,25 @@ options._render = function(vnode) {
 	currentComponent = vnode._component;
 };
 
+const oldDiffed = options.diffed;
+/** @type {(vnode: import('./internal').VNode)} */
+options.diffed = function(vnode) {
+	if (oldDiffed) {
+		oldDiffed(vnode);
+	}
+
+	const props = vnode.props;
+	const dom = vnode._dom;
+	if (
+		dom != null &&
+		vnode.type === 'textarea' &&
+		'value' in props &&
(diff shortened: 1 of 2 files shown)


compat/src/render.js
```diff
@@ -223,6 +223,25 @@ options._render = function(vnode) {
...
+
+	const props = vnode.props;
+	const dom = vnode._dom;
+	if (
+		dom != null &&
+		vnode.type === 'textarea' &&
+		'value' in props &&
+		props.value !== dom.value
+	) {
+		dom.value = props.value ?? '';
```
marvinhagemeister, line 241: Currently on a phone so can't check: what does this transpile to in our build chain?

marvinhagemeister (approved):
LGTM, one minor question but feel free to merge

JoviDeCroock (approved):

compat/src/render.js
andrewiggins: Here's a screenshot:
image

Manually doing `dom.value = props.value == null ? '' : props.value;` saves two bytes so I'll do that

Issue #2653, opened 2020-07-28
Preact doesn't work with react-frame-component


churichard:
I'm trying to use Preact with [react-frame-component](https://github.com/ryanseddon/react-frame-component) and running into some issues. Specifically, it seems like when the iframe is unmounted it throws an error: `Cannot read property 'parentNode' of undefined`.

Some additional notes: I think the error is related to how Preact handles portals. The error doesn't occur when I comment out the `createPortal` call on https://github.com/ryanseddon/react-frame-component/blob/9f8f06e1d3fc40da7122f0a57c62f7dec306e6cb/src/Frame.jsx#L109. By default, `this.props.head` is `null`, and Preact doesn't seem to like that.

### Reproduction

Here's a simple CodeSandbox that demonstrates the issue: https://codesandbox.io/s/preact-react-frame-component-y96k8?file=/app.js

Basically when one of the list items is selected, it'll become wrapped in a Frame component and colored red. However, it throws an error when you try to select a different item.

Notice that it works fine if there are no Frames, and it also works fine if you use React as well.

### Steps to reproduce

Click on "Two".

### Expected Behavior

"Two" should be wrapped in an iframe and colored red.

### Actual Behavior

An error is thrown, saying `Cannot read property 'parentNode' of undefined`.

lucasprag:
I'm facing the same problem. I would definitely spend the time and create a PR to react-frame-component if I have some guidance of possible things to look for and if I'm able to move forward with it.

So far I was able to check that `this.props.head` is always `null` for me inside https://github.com/ryanseddon/react-frame-component/blob/9f8f06e1d3fc40da7122f0a57c62f7dec306e6cb/src/Frame.jsx#L109 regardless if I'm using `Frame` with either React or Preact.

andrewpye:
Not sure if it's an ideal solution, but I've found that passing an empty `Fragment` as the `Frame` component's `head` prop prevents the error from being thrown:

```tsx
<Frame head={<Fragment></Fragment>}>
  <div>{title}</div>
</Frame>
```

I've forked the CodeSandbox [here](https://codesandbox.io/s/preact-react-frame-component-forked-7ysfbw?file=/app.js) which shows the demo app working after making the update:

![preact-with-frame-component](https://user-images.githubusercontent.com/5737342/219034479-9a56ce1a-c9b9-4fdc-9f20-91ff5ff65844.gif)

developit:
[...]

PR #3896, opened 2023-02-15
fix react-frame-component by supporting nullish portals

JoviDeCroock:
fixes https://github.com/preactjs/preact/issues/2653

This allows for nullish portals


compat/src/portals.js
```diff
@@ -33,39 +33,38 @@ function Portal(props) {
...
+	// Render our wrapping element into temp.
+	render(
+		createElement(ContextProvider, { context: _this.context }, props._vnode),
+		_this._temp
+	);
+
 	// When we come from a conditional render, on a mounted
 	// portal we should clear the DOM.
-	else if (_this._temp) {
+	if (!props._vnode && _this._temp) {
```
developit, line 67: I think we can just remove this condition and the this.unmount call now. When a null vnode is received, we will render the provider with `null` as it's only child, which produces the same result as calling unmount (because there is no componentDidUnmount method on this provider).

developit:
LGTM but I'd like to check if we can remove that condition entirely.

compat/src/portals.js
JoviDeCroock: That breaks our tests

compat/src/portals.js
```diff
@@ -33,41 +33,34 @@ function Portal(props) {
 
 	// When props.vnode is undefined/false/null we are dealing with some kind of
 	// conditional vnode. This should not trigger a render.
-	if (props._vnode) {
```
rschristian, line 36: nit: Comment should be updated

JoviDeCroock:
Closing due to inactivity

marvinhagemeister (approved):

PR #3898, opened 2023-02-15
fix: Missing `MemoExoticComponent` type in compat

### LINKED ISSUE #3897: Types: Missing `MemoExoticComponent` in `preact/compat`

**Describe the bug**
The React namespace in `preact/compat` is missing a definition for `MemoExoticComponent` so users who rely on that and use `preact/compat`'s types instead of React's run into the issue where their components are typed as `any`.

https://twitter.com/Sawtaytoes/status/1625959909752709121

**To Reproduce**

The following example React library will generate the following types:

```jsx
import { memo } from 'react';

function MyComponent() {
    return <h1>Hello World!</h1>;
}

export const MyMemoComponent = memo(MyComponent);
```

```ts
export const MyMemoComponent: import("react").MemoExoticComponent<typeof MyComponent>;
declare function MyComponent(): JSX.Element;
export {};
```

If we were to alias `react` to `preact/compat` using `"paths"` in a `tsconfig.json` in a consumer of the above library, that would essentially become this:

```ts
export const MyMemoComponent: import("preact/compat").MemoExoticComponent<typeof MyComponent>;
declare function MyComponent(): JSX.Element;
export {};
```

However, as there is no `MemoExoticComponent`, `MyMemoComponent` becomes typed as `any`.

Interestingly, TS doesn't throw an error when an aliased path is missing a member.

**Expected behavior**

Should distribute `MemoExoticComponent`

rschristian:
Closes #3897 

Adapted from [React's type](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/59a571d93be4da23169bc06387f0b8d67a9ed415/types/react/index.d.ts#L829-L831)

Seems to work well, provided I'm not missing anything.

### DIFF
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -111,6 +111,12 @@ declare namespace React {
 		isPureReactComponent: boolean;
 	}
 
+	export type MemoExoticComponent<
+		C extends preact.FunctionalComponent<any>
+	> = preact.FunctionComponent<ComponentProps<C>> & {
+		readonly type: C;
+	};
+
 	export function memo<P = {}>(
 		component: preact.FunctionalComponent<P>,
 		comparer?: (prev: P, next: P) => boolean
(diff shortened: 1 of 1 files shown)


developit (approved):
Nice. This is better than faking the type for sure.

PR #3903, opened 2023-02-18
Trying a few compat optimizations

JoviDeCroock:
- hoist regexes
- remove case insensitive regex

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -120,18 +120,18 @@ options.vnode = vnode => {
 		for (let i in props) {
 			let value = props[i];
 
-			if (IS_DOM && i === 'children' && type === 'noscript') {
+			if (
+				(i === 'value' && 'defaultValue' in props && value == null) ||
 				// Emulate React's behavior of not rendering the contents of noscript tags on the client.
-				continue;
-			} else if (i === 'value' && 'defaultValue' in props && value == null) {
+				(IS_DOM && i === 'children' && type === 'noscript')
+			) {
 				// Skip applying value if it is null/undefined and we already set
 				// a default value
 				continue;
-			} else if (
-				i === 'defaultValue' &&
-				'value' in props &&
-				props.value == null
-			) {
+			}
+
+			let lowerCased = i.toLowerCase();
+			if (i === 'defaultValue' && 'value' in props && props.value == null) {
 				// `defaultValue` is treated as a fallback `value` when a value prop is present but null/undefined.
 				// `defaultValue` for Elements with no value prop is the same as the DOM defaultValue property.
 				i = 'value';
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):

compat/src/render.js
```diff
@@ -142,29 +148,29 @@ options.vnode = vnode => {
...
-			} else if (/^onfocus$/i.test(i)) {
+				lowerCased = i = 'oninput';
+			} else if (ON_FOCUS.test(lowerCased)) {
 				i = 'onfocusin';
-			} else if (/^onblur$/i.test(i)) {
+			} else if (ON_BLUR.test(lowerCased)) {
 				i = 'onfocusout';
-			} else if (/^on(Ani|Tra|Tou|BeforeInp|Compo)/.test(i)) {
+			} else if (ON_ANI.test(i)) {
 				i = i.toLowerCase();
```
andrewiggins, line 163: Could this be replaced with `lowerCased`?

```suggestion
				i = lowerCased;
```

compat/src/render.js
```diff
@@ -114,24 +121,23 @@ options.vnode = vnode => {
...
 				// a default value
 				continue;
-			} else if (
-				i === 'defaultValue' &&
-				'value' in props &&
-				props.value == null
-			) {
+			}
+
+			let lowerCased = i.toLowerCase();
```
andrewiggins, line 139: Ah, is the hypothesis here that doing an upfront lowercase + case-sensitive regex is faster than case-insensitive regex?

andrewiggins: A part of me wonders if for this it'd be faster to take the size penalty and just do full name comparisons (`i === 'onInput'`, or `i.startsWith('onAnim')`)

andrewiggins:
Oh also, do you have some way you are measuring these? It's on my list to build benchmarks for compat eventually lol (but that's a long ways away for now)

compat/src/render.js
JoviDeCroock: Probably ye

JoviDeCroock:
@andrewiggins I measure these with importing compat during the rts benchmark

andrewiggins (approved):

compat/src/render.js
```diff
@@ -11,6 +11,13 @@ export const REACT_ELEMENT_TYPE =
 	0xeac7;
 
 const CAMEL_PROPS = /^(?:accent|alignment|arabic|baseline|cap|clip(?!PathU)|color|dominant|fill|flood|font|glyph(?!R)|horiz|image|letter|lighting|marker(?!H|W|U)|overline|paint|pointer|shape|stop|strikethrough|stroke|text(?!L)|transform|underline|unicode|units|v|vector|vert|word|writing|x(?!C))[A-Z]/;
+const DOUBLE_CLICK = /ondoubleclick/;
+const CHANGE_INPUT = /^onchange(textarea|input)/;
+const ON_ANI = /^on(Ani|Tra|Tou|BeforeInp|Compo)/;
+const ON_BLUR = /^onblur$/;
+const ON_FOCUS = /^onfocus$/;
+const ON_INPUT = /^oninput$/;
```
developit, line 19: Since we're lower-casing preemptively now, I bet we could get away with inlining these as straight equality checks (`lowerCased === 'onfocus'`). Should be faster than the regex.

JoviDeCroock: Good shout, the camel props one is still the one hurting perf the most 😅

PR #3906, opened 2023-02-20
fix case where strict equality leads to a deadlock with c._force

JoviDeCroock:
In a rare case where async context providers get updated we could reach a deadlock scenario with `c._force` https://codesandbox.io/s/amazing-sunset-unlmgg?file=/src/useSubmitHook.ts


marvinhagemeister (approved):
Fantastic work on narrowing it down to a test case 🙌

PR #3909, opened 2023-02-21
pass inEvent so custom debouncers can mirror our logic

### LINKED ISSUE #3902: Pass inEvent argument to a custom debounceRendering function
Is it possible to pass the _inEvent_ flag into the _options.debounceRendering()_ hook or export the flag to the application?

This request is related to the recent change ‘[use microtick outside of events](https://github.com/preactjs/preact/pull/3879)’. With this change the internal _defer()_ call uses either timeout or microtick based on the _inEvent_ flag. We want to mirror this logic in our custom _options.debounceRendering()_ override but our custom override needs to be aware of the value of _inEvent_ flag.

We use the custom _debounceRendering()_ method for automated testing to detect when the rendering queue is complete. We don't want to change the timing approach used by Preact, we just want to inject some additional logic required by our applications.

JoviDeCroock:
fixes #3902

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -282,13 +282,10 @@ interface ContainerNode {
 	insertBefore: Node['insertBefore'];
 	appendChild: Node['appendChild'];
 	removeChild: Node['removeChild'];
-	childNodes: ArrayLike<Node>
+	childNodes: ArrayLike<Node>;
 }
 
-export function render(
-	vnode: ComponentChild,
-	parent: ContainerNode
-): void;
+export function render(vnode: ComponentChild, parent: ContainerNode): void;
 /**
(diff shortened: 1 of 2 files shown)


src/index.d.ts
```diff
@@ -299,10 +296,7 @@ export function render(
 	parent: ContainerNode,
 	replaceNode?: Element | Text
 ): void;
-export function hydrate(
-	vnode: ComponentChild,
-	parent: ContainerNode
-): void;
+export function hydrate(vnode: ComponentChild, parent: ContainerNode): void;
```
JoviDeCroock, line 299: this was prettier

marvinhagemeister (approved):

Issue #3853, opened 2023-01-06
Add ARIA attribute typings


andrewiggins:
**Describe the feature you'd love to see**

React typings have [ARIA attributes defined](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1587) providing stronger typing for the expected values of aria attributes. I think it'd be nice if we provided something similar, give people using TypeScript or people writing JS in an editor that uses TS for completions a nicer experience.

**Additional context (optional)**
Here is an example of the kind of feedback TypeScript can provide with proper aria typings.
https://www.typescriptlang.org/play?ts=4.9.4#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgIilQ3wG4AoNCAOwGd4ATOAXjgB4HgA3OFKYFAFoUAVxgQq4ADZIYSZvloMpmfHAD0APlJA

Perhaps worth exploring if we can reuse anything from TypeScript's lib.dom.d.ts assuming it's typings match what we do.

myasonik:
Anything would be better than nothing of course but for whoever ends up looking at this, a few more thoughts on this:

[lib.dom](https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts#L1998) implements the weakest types where everything is an optional string. That's technically correct and will prevent issues like folks misspelling aria attributes and whatnot but is quite limited. 

React goes one step further and validates the values a little more expecting things like booleans and explicit enums where appropriate. This has the added benefit of validating correct key-value pairs which might be correct HTML but isn't always a correct use of ARIA. 

Most ARIA attributes need to be on specific HTML elements (or elements with a specific role). 
And many attributes should never be used together. This would probably be stricter than many other things are typed but because of a general lack of developer education and testing of assistive tech, I think it would be really valuable here. Using [`aria-autocomplete`](https://w3c.github.io/aria/#aria-autocomplete) as a nice short example: it should only ever be used on HTML elements that either implicitly or explicitly have the roles of combobox, textbox, searchbox. 

I know getting the typings that strict would be a lot more work but it could be a big win for both developer education and accessible software.

OfficerHalf:
Accessibility is super important in general, but this gap in preact's types is one of the reasons I can't yet make the case to switch my employer to preact. We have commitments to hit WCAG AA compliance, which in many cases requires the use of aria attributes, and with a big distributed team accurate types are a must.

[...]

PR #3910, opened 2023-02-22
Add typings for ARIA attributes

andrewiggins:
Related: #3853

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -18,6 +18,8 @@ type Defaultize<Props, Defaults> =
 				Pick<Props, Exclude<keyof Props, keyof Defaults>>
 		: never;
 
+type Booleanish = boolean | 'true' | 'false';
+
 export namespace JSXInternal {
 	export type LibraryManagedAttributes<Component, Props> = Component extends {
 		defaultProps: infer Defaults;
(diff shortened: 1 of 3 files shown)


src/jsx.d.ts
```diff
@@ -1320,9 +1326,315 @@ export namespace JSXInternal {
...
+		| 'tabpanel'
+		| 'term'
+		| 'textbox'
+		| 'timer'
+		| 'toolbar'
+		| 'tooltip'
+		| 'tree'
+		| 'treegrid'
+		| 'treeitem'
+		| (string & {}); // This trick allows any string to be valid here but also provides TS auto-complete behavior for the explicitly listed values here
```
rschristian, line 1632: What's the reason for allowing any string to be valid? Are there use cases for using roles outside of this list? Genuinely curious.

andrewiggins: Tbh, I'm not entirely sure. I've copied this from React. I assume it's to accommodate new roles as the spec changes? I need to check the git history of their types to see the reasoning.

myasonik: I couldn't think of a reason and went back through the aria spec and didn't find anything concrete there either. 

The spec doesn't change very quickly (last version was published December 2017) so having that carve out doesn't seem too pressing IMO. (That said, ARIA 1.2 should be published _soon_ but it's also been soon for like a year now... But that still a 6 year cadence if it gets published this year.) It's also kind of a slippery slope where you could argue any of the attributes should allow for anything because the spec could change.

rschristian: Ah, sorry, missed that. [Doesn't look like a reason was given](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/53361), unless I'm missing part of the conversation.

I'm admittedly not a TS maximalist, so I personally have no issue with a string fallback (working auto-complete is lovely enough), but I wonder if it's better to be strict and maybe catch some typos, even if that means we might need to make alterations in the future? I'm not super knowledgeable about aria stuff though, so maybe there is a use for other roles.

src/jsx.d.ts
```diff
@@ -1320,9 +1326,315 @@ export namespace JSXInternal {
...
+			'none' | 'inline' | 'list' | 'both' | undefined
+		>;
+		/** Indicates an element is being modified and that assistive technologies MAY want to wait until the modifications are complete before exposing them to the user. */
+		'aria-busy'?: Signalish<Booleanish | undefined>;
+		/**
+		 * Indicates the current "checked" state of checkboxes, radio buttons, and other widgets.
+		 * @see aria-pressed @see aria-selected.
+		 */
+		'aria-checked'?: Signalish<
+			boolean | 'false' | 'mixed' | 'true' | undefined
```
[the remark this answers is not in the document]
rschristian: > Meta question: should Booleanish exist? Not just the type, but the concept overall?

IMO, it probably should. It's inevitable that some user(s) will trip up on that and ask why both aren't supported. I think the ship has long since sailed on trying to enforce consistency w/ attributes.

JoviDeCroock (approved):

rschristian (approved):

src/jsx.d.ts
andrewiggins: @myasonik Yea, that makes sense. No point to reduce the usefulness of these typings just cuz anything could change lol. And like you said, the churn is super infrequent.

Thanks for digging that link up @rschristian! I dug a little further and it looks like this was the original thinking ([link](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/53344#pullrequestreview-666294951)):
> 1. It should still allow unknown roles e.g. when new roles are added but types are outdated
> 2. what about fallback roles e.g. role="none presentation"?

For 1, like @myasonik was saying, I'm not worried about it anymore. But the second reason is news to me. I've never heard of "fallback roles" before. Scanning the spec, I can only find [one reference to fallbacks roles in a note](https://www.w3.org/TR/wai-aria-1.1/#role-none-note-none).

Maybe to start off we can just manually add "none presentation" as a possible role value? And in the future deal with other fallback roles as changes to the spec land? At that time we can either manually add more fallback combinations or look at more advanced TS techniques if it becomes complicated.

src/jsx.d.ts
```diff
@@ -1320,9 +1326,315 @@ export namespace JSXInternal {
...
+		>;
+		/**
+		 * Defines the total number of columns in a table, grid, or treegrid.
+		 * @see aria-colindex.
+		 */
+		'aria-colcount'?: Signalish<number | undefined>;
+		/**
+		 * Defines an element's column index or position with respect to the total number of columns within a table, grid, or treegrid.
+		 * @see aria-colcount @see aria-colspan.
```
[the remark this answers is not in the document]
andrewiggins: VSCode seems to handle it properly but GitHub does not. I'll put them on multiple lines to be safe. Good call out :)

src/jsx.d.ts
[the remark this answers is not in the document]
andrewiggins: Oh thanks! Yes! I meant to go back and update this but forgot.

Yea, the handling of boolean type vs string is interesting/complicated. In React, for some properties `false` removes the attributes but for others (e.g. ARIA) `false` is stringified. Personally, I'd prefer to just always say pass the string `"false"` if that's the value the attribute should have but we should match React's behavior here and accept both booleans and strings. There is some nice ergonomics here too.

If you are curious, here is a list of the different attribute behaviors React special cases: https://github.com/facebook/react/blob/c04b18070145b82111e1162729f4776f4d2c6112/packages/react-dom-bindings/src/shared/DOMProperty.js#L18-L49

src/jsx.d.ts
andrewiggins: I went with adding `'none presentation'` for now as that's my vote, but would like to hear what others think as well before merging.

myasonik: Ah, I forgot about that... It's pretty uncommon but, yeah, `role` takes space separated list of roles that you can use as fallbacks (https://www.w3.org/TR/wai-aria-1.1/#host_general_role)

The `none presentation` example given in the spec was just given because `none` was new at the time so `presentation` was an appropriate fallback. But you can theoretically do it with any combination at any arbitrary length. 

Though I can't imagine this being used very often, I do think allowing everything the spec allows is probably important enough to allow `(string & {})`.

PR #3921, opened 2023-02-28
[perf] Hoist render queue sorting fn

developit:
Backport from #3834.


developit:
note: we have no benchmark that exercises the render queue. I'll add one in another PR.

andrewiggins (approved):

marvinhagemeister (approved):

PR #3922, opened 2023-03-01
Fix hook update not being applied when a forced context update was enqueued at the same time

marvinhagemeister:
Fix hook update not being applied when a forced context update was enqueued at the same time. This fixes a regression that was introduced in https://github.com/preactjs/preact/pull/3889 which was released as `10.12.1`.

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -184,7 +184,23 @@ export function useReducer(reducer, initialState, init) {
 
 		if (!currentComponent._hasScuFromHooks) {
 			currentComponent._hasScuFromHooks = true;
-			const prevScu = currentComponent.shouldComponentUpdate;
+			let prevScu = currentComponent.shouldComponentUpdate;
+			const prevCWU = currentComponent.componentWillUpdate;
+
+			// If we're dealing with a forced update `shouldComponentUpdate` will
+			// not be called. But we use that to update the hook values, so we
+			// need to call it.
+			currentComponent.componentWillUpdate = function(p, s, c) {
+				if (this._force) {
+					let tmp = prevScu;
+					// Clear to avoid other sCU hooks from being called
+					prevScu = undefined;
+					this.shouldComponentUpdate(p, s, c);
+					prevScu = tmp;
+				}
+
+				if (prevCWU) prevCWU.call(this, p, s, c);
+			};
 
 			// This SCU has the purpose of bailing out after repeated updates
 			// to stateful hooks.
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -195,7 +195,7 @@ export function useReducer(reducer, initialState, init) {
 					let tmp = prevScu;
 					// Clear to avoid other sCU hooks from being called
 					prevScu = undefined;
-					this.shouldComponentUpdate(p, s, c);
+					updateHookState(p, s, c);
 					prevScu = tmp;
 				}
 
@@ -208,7 +208,13 @@ export function useReducer(reducer, initialState, init) {
 			// state setters, if we have next states and
 			// all next states within a component end up being equal to their original state
 			// we are safe to bail out for this specific component.
-			currentComponent.shouldComponentUpdate = function(p, s, c) {
+			/**
+			 *
+			 * @type {import('./internal').Component["shouldComponentUpdate"]}
+			 */
+			// @ts-ignore - We don't use TS to downtranspile
+			// eslint-disable-next-line no-inner-declarations
+			function updateHookState(p, s, c) {
 				if (!hookState._component.__hooks) return true;
 
 				const stateHooks = hookState._component.__hooks._list.filter(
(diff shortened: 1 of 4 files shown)


src/create-context.js
```diff
@@ -42,7 +42,9 @@ export function createContext(defaultValue, contextId) {
 						// 	enqueueRender(c);
 						// });
 						subs.some(c => {
-							c._force = true;
+							if (!c._hasScuFromHooks) {
```
developit, line 45: Doesn't this disable the sCU bypass for any component with a `setState()` call?

hooks/test/browser/combinations.test.js
```diff
@@ -344,4 +345,57 @@ describe('combinations', () => {
 
 		expect(ops).to.deep.equal(['child effect', 'parent effect']);
 	});
+
+	it('should not block hook updates when context upates are enqueud', () => {
```
mdentremont, line 349: Minor:
```suggestion
	it('should not block hook updates when context updates are enqueued', () => {
```

mdentremont (commented):
The test seems reasonable but unfortunately I don't have enough context on the subject matter to be able to approve/deny - adding a tiny comment though

marvinhagemeister:
@mdentremont Thanks for the suggestions! Yeah it looks like we need to opt for a more elaborate fix, but I'm confident we can do it now that we know why the error occurred in the first place.

marvinhagemeister:
Ok I think I found a better fix.

hooks/src/index.js
```diff
@@ -184,7 +184,23 @@ export function useReducer(reducer, initialState, init) {
...
 		if (!currentComponent._hasScuFromHooks) {
 			currentComponent._hasScuFromHooks = true;
-			const prevScu = currentComponent.shouldComponentUpdate;
+			let prevScu = currentComponent.shouldComponentUpdate;
+			const prevCWU = currentComponent.componentWillUpdate;
+
+			// If we're dealing with a forced update `shouldComponentUpdate` will
+			// not be called. But we use that to update the hook values, so we
+			// need to call it.
+			currentComponent.componentWillUpdate = function(p, s, c) {
```
mdentremont, line 193: Is there any risk that this happens over and over for the same component, leading to more and more wrapping?

hooks/src/index.js
```diff
@@ -184,7 +184,23 @@ export function useReducer(reducer, initialState, init) {
...
+				if (this._force) {
+					let tmp = prevScu;
+					// Clear to avoid other sCU hooks from being called
+					prevScu = undefined;
```
mdentremont, line 197: We're:
- defining an instance var
- clearing the instance var
- calling a method which should have access to the instance var (does it?)
- restoring the instance var

I'm reviewing on my phone so I might be missing some of the flow - but I'm missing how the `prevScu = undefined` is having an impact. Did we mean to do `currentComponent.shouldComponentUpdate = undefined`?

hooks/src/index.js
marvinhagemeister: This branch is guarded by a check a couple of lines earlier which ensures that this code is only executed once per component.

```js
if (!currentComponent._hasScuFromHooks) {
  currentComponent._hasScuFromHooks = true;
  // ...
}
```

hooks/src/index.js
marvinhagemeister: The gist of why the clearing is needed is that we kinda "misuse" `shouldComponentUpdate` to update the hook values, whereas that method is typically expected to be pure. In the patched `shouldComponentUpdate` method in the hooks addon we call the previous `shouldComponentUpdate` method in case the component already had one. The signals addon for example does this.

Now here's the problem: In a forced update we _still_ want to update all hook values, but not call `shouldComponentUpdate`. Doing so would be a violation of the API contract. To work around this temporarily unsetting `prevScu ` ensures that we only call the hooks overwritten method for `shouldComponentUpdate` and not the other ones.

It's a little weird for sure and something we're planning to make cleaner in v11.

JoviDeCroock (approved):

hooks/src/index.js
```diff
@@ -184,7 +184,23 @@ export function useReducer(reducer, initialState, init) {
...
+					this.shouldComponentUpdate(p, s, c);
```
developit, line 198: Idea:

```js
		if (!currentComponent._hasScuFromHooks) {
			currentComponent._hasScuFromHooks = true;
			currentComponent.componentWillUpdate = function(p, s, c) {
				if (this._force) {
					// <snip>
					willUpdate.call(this, p, s, c);
				}
				return prevCWU.call(this, p, s, c);
			};

			/** @this {typeof currentComponent} */
			const willUpdate = function(p, s, c) {
				if (!hookState._component.__hooks) return true;
				// <snip>
			};
			currentComponent.shouldComponentUpdate = willUpdate;
		}
```

hooks/src/index.js
mdentremont: Fair - but I still don't understand how `prevScu = undefined` is doing anything? It's an instance variable, which shouldn't be in play when `this.shouldComponentUpdate(p, s, c);` is invoked?

mdentremont: NVM, I see it now!

PR #3928, opened 2023-03-08
synchronous forceUpdate

KevinDoughty:
The comment for the function `forceUpdate` says:
"Immediately perform a synchronous re-render of the component"

This PR restores that behavior. Does not cause any tests to fail. I cannot find much information on a "double render" problem which sounds dubious to me.

Animation libraries need timely updates and there is no reason to arbitrarily deny that. Please use setState({}) instead.

See:
https://github.com/preactjs/preact/pull/1939

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -67,7 +67,7 @@ Component.prototype.forceUpdate = function(callback) {
 		// shouldComponentUpdate
 		this._force = true;
 		if (callback) this._renderCallbacks.push(callback);
-		enqueueRender(this);
+		renderComponent(this);
 	}
 };
 


KevinDoughty:
My mistake, tests fail. Will try again soon.

marvinhagemeister:
Seeing that [the comment](https://github.com/preactjs/preact/blob/48f55a395e1e9f6756fbf7be51f3138c7e48cec3/src/component.js#L58) was made 5 years ago, it's very likely that it is out of date. In general you don't want to kick off a render whilst another one is currently being processed. That was the main reason for us going with `enqueueRender()` for `.forceUpdate()` if memory serves correctly.

The current scheduling mechanism uses a `microtask` which is the smallest unit of scheduling in the event loop and it runs <1ms. Did you run in a specific problem where that caused problems with an animation?

marvinhagemeister:
Haven't played with the view transitions API yet, but it looks like it might require a synchronous flush of the queue: https://github.com/preactjs/preact/issues/3929

sschultze:
> Haven't played with the view transitions API yet, but it looks like it might require a synchronous flush of the queue: #3929

@marvinhagemeister The flush of the queue actually doesn't have to be synchronous. The callback passed to `startViewTransition` can return a promise, [as described here](https://developer.chrome.com/docs/web-platform/view-transitions/#async-dom-updates-and-waiting-for-content).

But I think that there's a chance that 3rd party libraries will use React's `flushSync` with the View Transition API, so it's probably better to implement this rather than introducing another async function.

KevinDoughty:
Solved my problem without synchronous renders. Thanks.

PR #3939, opened 2023-03-17
For element nodes with a single text child, directly set `textContent`

andrewiggins:
This PR adds a (hopefully) fast path for element VNodes with a single text child. Given how common this kind of tree structure (after all, UI is ultimately about fancy ways to display text lol), I hypothesize this should improve our rendering runtime.

Two significant internal changes that comes out of this are:
1. Single text children no longer are represented as VNodes  
   With this change, single text children no longer get a VNode created representing them. Hopefully this will help contribute to better runtime and memory performance by removing lots of allocations. But it does have implications for the VNode tree (which leads to the next point).
   
2. Elements with a single text child have a `null` `_children` property  
   Since single text childs no longer have a VNode created, the `_children` property of their parents is empty (aka `null`). Inspecting the tree VNode tree just using `_children` would then lead to an incomplete tree. We could fix this by adding the string to the `_children` array, but then that breaks the assumption the codebase has had for a while that `_children` is always `Array<VNode | null>`. But perhaps that is a better approach?
   
For now, I'm going to try this implementation and see if any performance improvements are observable. If so, then we can further pursue what the right thing do here is.

### DIFF
--- a/test/browser/fragments.test.js
+++ b/test/browser/fragments.test.js
@@ -1639,9 +1639,9 @@ describe('Fragment', () => {
 		expect(scratch.innerHTML).to.equal(htmlForFalse);
 		expectDomLogToBe(
 			[
-				'<div>1.insertBefore(<div>3, #text)',
-				'<div>31.insertBefore(<div>4, #text)',
 				'#text.remove()',
+				'<div>.appendChild(<div>3)',
+				'<div>3.appendChild(<div>4)',
 				'<div>2.remove()'
 			],
 			'rendering from true to false'
@@ -1652,12 +1652,7 @@ describe('Fragment', () => {
 
 		expect(scratch.innerHTML).to.equal(htmlForTrue);
 		expectDomLogToBe(
-			[
-				'<div>34.insertBefore(#text, <div>3)',
-				'<div>4.remove()',
-				'<div>3.remove()',
-				'<div>1.appendChild(<div>2)'
-			],
+			['<div>1.appendChild(<div>2)'],
 			'rendering from false to true'
 		);
 	});
--- a/test/browser/fragments.test.js
+++ b/test/browser/fragments.test.js
@@ -82,10 +82,7 @@ describe('Fragment', () => {
 		);
 
 		expect(scratch.innerHTML).to.equal('<span>foo</span>');
-		expectDomLogToBe([
-			'<span>.appendChild(#text)',
-			'<div>.appendChild(<span>foo)'
-		]);
+		expectDomLogToBe(['<div>.appendChild(<span>foo)']);
 	});
 
 	it('should render multiple children via noop renderer', () => {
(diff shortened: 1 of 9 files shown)


test/browser/fragments.test.js
```diff
@@ -1671,11 +1639,9 @@ describe('Fragment', () => {
 		expect(scratch.innerHTML).to.equal(htmlForFalse);
 		expectDomLogToBe(
 			[
-				'<div>.appendChild(#text)',
-				'<div>1.insertBefore(<div>3, #text)',
-				'<div>.appendChild(#text)',
-				'<div>31.insertBefore(<div>4, #text)',
```
andrewiggins, line 1677: Most of the test changes here are just removing calls to `appendChild(#text)`. But this test case invokes something more interesting I think it worth examining.

Here, we are transitioning from a tree like:
```html
<div>1</div>
<div>2</div>
```
to
```html
<div class="wrapper">
  <div>3</div>
  <div>4</div>
</div>
```

Our diff matches `<div>1` with the new `div.wrapper`. So it sees a div with a single text child (`1`) transitioning to a div with multiple children `<div>3` and `<div>4`. So with this update, instead of creating and inserting the new children before the old children, it removes the text node and appends the new children to the now empty div.

test/browser/fragments.test.js
```diff
@@ -1686,13 +1652,7 @@ describe('Fragment', () => {
 
 		expect(scratch.innerHTML).to.equal(htmlForTrue);
 		expectDomLogToBe(
-			[
-				'<div>34.insertBefore(#text, <div>3)',
-				'<div>4.remove()',
-				'<div>3.remove()',
-				'<div>.appendChild(#text)',
-				'<div>1.appendChild(<div>2)'
-			],
```
andrewiggins, line 1695: Here, we are transition from a tree like:
```html
<div class="wrapper">
  <div>3</div>
  <div>4</div>
</div>
```
to
```html
<div>1</div>
<div>2</div>
```

Our diff matches the old `div.wrapper` with the new `<div>1` node. So it sees a div with multiple children transitioning to a div with a single text child. To handle this, it clears away the old children by setting the `textContent` of `div.wrapper` to the new single text child. That change is why you no longer see the calls to `.remove()`. Also since we set `textContent` to the new text child, you don't see the relevant `insertBefore` call of the text node. In one operation we unmount old children and set new children.

We do still call `unmount` on old children so any lifecycle methods are called, but no DOM is removed in those calls since setting `textContent` will take care of that.

andrewiggins:
I think the regression on 02_replace1k is due to changes in how V8 is optimizing the code. Check out the trace logs below:
preact-master:
image

preact-local:
Screenshot 2023-03-17 at 4 38 16 PM

The differences in colors under the red underlined area show that V8's optimization heuristics are treating each of these differently. I don't entirely understand the difference but running this benchmark locally and comparing the average duration of all the runs (warmups + final test run) shows that there isn't really a significant difference between on average between the two:
image

JoviDeCroock (approved):

marvinhagemeister (approved):

JoviDeCroock:
Superseded by #4523

PR #3940, opened 2023-03-18
test optimization for compat

JoviDeCroock:
This replaces our very expensive regex, both to evaluate as well as to execute with arrays of attributes


JoviDeCroock:
Closing as I wanted to show the byte-size impact, not sure if it's worth it for us

PR #3941, opened 2023-03-20
optimize compat by hoisting the vnode handling func

JoviDeCroock:
This seems to significantly improve performance from a V8 Point of View for RTS

```
Text:
compat x 287.468 ops/sec ±2.71 (82 runs sampled)
current x 321.807 ops/sec ±2.38 (86 runs sampled)

SearchResults:    
compat x 735.515 ops/sec ±3.64 (83 runs sampled)
current x 867.643 ops/sec ±2.94 (86 runs sampled)

ColorPicker:      
compat x 3,190.427 ops/sec ±3.62 (89 runs sampled)
current x 4,169.366 ops/sec ±3.11 (89 runs sampled)
```

current is this, compat is the 10.13.1 release

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -108,106 +108,116 @@ let classNameDescriptor = {
 	}
 };
 
+function handleDomVNode(vnode) {
+	let props = vnode.props,
+		type = vnode.type,
+		normalizedProps = {},
+		hasChanged = false;
+
+	for (let i in props) {
+		let original = i,
+			originalValue = props[i],
+			value = props[i];
+
+		if (
+			(i === 'value' && 'defaultValue' in props && value == null) ||
+			// Emulate React's behavior of not rendering the contents of noscript tags on the client.
+			(IS_DOM && i === 'children' && type === 'noscript')
+		) {
+			// Skip applying value if it is null/undefined and we already set
+			// a default value
+			continue;
+		}
+
+		let lowerCased = i.toLowerCase();
+		if (i === 'defaultValue' && 'value' in props && props.value == null) {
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):

andrewiggins (approved):

compat/src/render.js
```diff
@@ -101,113 +101,125 @@ function isDefaultPrevented() {
...
+		Array.isArray(normalizedProps.value)
+	) {
+		// forEach() always returns undefined, which we abuse here to unset the value prop.
+		normalizedProps.value = toChildArray(props.children).forEach(child => {
+			child.props.selected =
+				normalizedProps.value.indexOf(child.props.value) != -1;
+		});
+	}
 
-		if (props.class != props.className) {
```
JoviDeCroock, line 206: What happened before:

- props.class === undefined && props.className !== undefined --> we would make className enumerable and make class === className
- props.class !== undefined && props.className === undefined --> we would make className not enumerable and make className return props.class
- props.class !== undefined && props.className !== undefined --> we would make className enumerable and make class === className

Position 1 and 3 are different because we are just returning the value for `className` rather than the one for class, I don't think this matters.
Position 2 is different because we aren't introducing a getter for `className` to return `class`, I don't think this matters

CC @marvinhagemeister

marvinhagemeister: Thanks for checking. That sounds reasonable. I'm good with merging the PR 👍

developit: Yep that seems reasonable to me. I think the dynamic assignment of `enumerable` here was from when we used to make `class` a descriptor too.

PR #3945, opened 2023-03-21
Upgrade prettier and run on more files

andrewiggins:
The real motivation for this PR is that I noticed our JSON files were formatted using tabs instead of spaces as defined by our `.editorconfig`. We want JSON files to be formatted with spaces so that they are compatible with JS's native `JSON.stringify` which many libraries use to write JSON files. The fix was to remove the `useTabs` and `tabWidth` from the `package.json` `prettier` config cuz it was overriding what we specified in `.editorconfig`.

While I was here messing with prettier I figured I'd do an upgrade, add more extensions, and reformat the repo.

I've added prettier formatting to `mjs,cjs,json,html,md,css,scss` files, on top of our existing config `js,jsx,ts,tsx,yml`. I've also added an npm `format` script which runs prettier on the entire codebase to make manually fixing issues easier (might useful to help PRs work around merge conflicts as they can just take own changes and rerun formatting).

Let me know if you'd like me to make any changes (e.g. not format some extensions, add more files/dirs to .prettierignore, etc.)


package.json
```diff
@@ -1,309 +1,311 @@
...
-		"test/fixtures",
-		"test/ts/",
-		"*.ts",
-		"dist"
-	],
-	"prettier": {
-		"singleQuote": true,
-		"trailingComma": "none",
-		"useTabs": true,
-		"tabWidth": 2
```
andrewiggins, line 188: I removed these cuz they were overriding our .editorconfig

package.json
```diff
@@ -1,309 +1,311 @@
...
+  "eslintIgnore": [
+    "test/fixtures",
+    "test/ts/",
+    "*.ts",
+    "dist"
+  ],
+  "prettier": {
+    "singleQuote": true,
+    "trailingComma": "none",
+    "arrowParens": "avoid"
```
andrewiggins, line 190: Prettier v2 changed this default to "always". I've overridden it to "avoid" to match our existing style, but happy to remove if people would prefer "always". I don't have a strong pref.

JoviDeCroock (approved):

package.json
marvinhagemeister: Personally I like `avoid` more

marvinhagemeister (approved):

PR #3950, opened 2023-03-27
minor code style issues

m-avagyan:


marvinhagemeister (commented):
Thanks for taking the time to make a PR. Whilst I appreciate the effort, I don't think this PR adds value to the code base. Both the non-destructured and destructured variant have the exact same behavior and both are equally readable. It just comes down to personal preference. This in turn poses the question of what's the benefit of merging this PR?

PR #3962, opened 2023-04-06
Wait a major task after triggering GC to measure memory

andrewiggins:
`window.gc` synchronously triggers one Major GC. However that MajorGC asynchronously triggers additional MajorGCs until the usedJSHeapSizeBefore and usedJSHeapSizeAfter are the same. Here, we'll wait a moment (i.e. a major task using postMessage) for some (hopefully all) additional GCs to finish before measuring the memory.

### DIFF
--- a/benches/src/util.js
+++ b/benches/src/util.js
@@ -5,6 +5,12 @@ export { afterFrame };
 
 export const measureName = 'duration';
 
+const majorTask = () =>
+	new Promise(resolve => {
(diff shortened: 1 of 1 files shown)


JoviDeCroock (approved):

andrewiggins:
FYI, microtask is too quick and setTimeout can add extra wait time so I went with post message as it seems to immediately fire after GC has stabilized, judging from the browser traces I looked at.

PR #3963, opened 2023-04-07
chore(workspace): firefox saucelabs not running

JoviDeCroock:
FireFox made a change with regards to their debug port, this was updated in webdriver 7.27, however the `karma-sauce-launcher` seems to not have bumped the dep yet so in this PR we force it so we can make the tests pass again


package-lock.json
```diff
@@ -22302,7 +23146,7 @@
       "requires": {
         "global-agent": "^2.1.12",
         "saucelabs": "^4.6.3",
-        "webdriverio": "^6.7.0"
+        "webdriverio": "7.30.2"
```
JoviDeCroock, line 23149: this is the important place where it's bumped, sauce-launcher will leverage the new 7.30.2

andrewiggins (approved):

PR #3954, opened 2023-03-30
Handle NaN comparison in useEffect dependency array

jayrobin:
## Summary
When comparing changes to `useEffect` dependencies, if any dependency is `NaN` it will result in the effect always being run, even if the value was `NaN` on the previous render. This is due to `NaN !== NaN` in JavaScript.

This change handles the `NaN` case using the `is` utility introduced in #3776. Presumably we don't want to use `Object.is` due to [incompatibility with IE11](https://caniuse.com/mdn-javascript_builtins_object_is) (thanks for the insight @andrewiggins!). Note that I copied the `is` utility, which is also present in `compat/src`, but I'm happy to raise it up to `src/utils` and use it in both places if that's preferred.

## Issue
Note: this will cause Preact to hang due to infinite re-renders
```js
import { render } from 'preact';
import { useState, useEffect } from 'preact/hooks';

function Component({ value }) {
  const [count, setCount] = useState(0);
  useEffect(() => {
    setCount(count + 1);
  }, [value]);
  return <p>{count}</p>;
}

function App() {
  const [value, setValue] = useState(undefined);
  setValue(NaN);

  return <Component value={NaN} />;
}

render(<App />, document.getElementById('app'));
```
**Expected behavior**
Preact runs the effect once, because `value` never changes from `NaN`.

**Actual behavior**
Preact re-renders forever, because it strictly compares `NaN` to `NaN`.

PR #3964, opened 2023-04-07
Throw an error in debug if NaN is in a hook's dependency array

andrewiggins:
Related #3954


JoviDeCroock (approved):

marvinhagemeister (approved):
Great idea! 🙌

PR #3965, opened 2023-04-07
[Compat] Throw an error for too many repeated function component rerenders

### LINKED ISSUE #3959: Component re-renders indefinitely when changing state on render

**Describe the bug**
When changing state on render Preact will re-render indefinitely rather than throwing a "Too many re-renders" error.

**To Reproduce**

[codesandbox](https://codesandbox.io/s/flamboyant-ben-c2ulj1?file=/index.js)

```js
import { render } from 'preact';
import { useState } from 'preact/hooks';

let rerenderCount = 0;
function App() {
  const [, setValue] = useState({});
  console.log('rendering', rerenderCount);
  rerenderCount++;

  if (rerenderCount < 50) {
    setValue({});
  }

  return (
    <div>Counter: {value}</div>
  );
}

render(<App />, window.root);
```

**Expected behavior**

Logs `rendering 0` to `rendering 24`, then throws "Too many re-renders" error ([similar to React](https://legacy.reactjs.org/docs/error-decoder.html/?invariant=301) - [thrown here](https://github.com/facebook/react/blob/main/packages/react-reconciler/src/ReactFiberHooks.js#L703-L708))

**Actual behavior**

Logs `rendering 0` to `rendering 49`. If you remove the `rerenderCount < 50` conditional, it will run indefinitely and crash the browser.

andrewiggins:
Curious to know what others think about this. We could move it to debug if people prefer.

Fixes #3959

### DIFF
--- a/compat/src/render.js
+++ b/compat/src/render.js
@@ -261,14 +261,28 @@ options.vnode = vnode => {
 	if (oldVNodeHook) oldVNodeHook(vnode);
 };
 
-// Only needed for react-relay
+let renderCount = 0;
 let currentComponent;
 const oldBeforeRender = options._render;
 options._render = function (vnode) {
 	if (oldBeforeRender) {
 		oldBeforeRender(vnode);
 	}
-	currentComponent = vnode._component;
+
+	const nextComponent = vnode._component;
+	if (nextComponent === currentComponent) {
+		renderCount++;
+	} else {
+		renderCount = 1;
+	}
+
+	if (renderCount >= 25) {
+		throw new Error(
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):
Personally prefer this in debug but good either way. In react it's only dev as well right?

andrewiggins:
For now I'll move it to debug. We can always add it compat's runtime later if the demand is there for the runtime check.

JoviDeCroock:
Superseded by https://github.com/preactjs/preact/pull/4349

PR #3968, opened 2023-04-12
Add default props in cloneElement.

nnhk23:
**Summary**
This PR added `defaultProps` & a few adjustments how Preact handles undefined properties in `cloneElement`. Currently, any properties in props would be used to override normalizedProps in Preact without any checks. However this is being handled [differently in React](https://github.com/facebook/react/blob/4337c1c00609ec8d7ae399c736e9d37bb159fac5/packages/react/src/ReactElement.js#L514-L518) where there will be a check in place to skip the override when config[propName] (props[i] in our case) is undefined.

**Issue**
Check these 2 code sandbox, the component's color should not be override by an `undefined` property:
- Preact: https://codesandbox.io/s/preact-cloneelement-defaultprops-csr-only-ux43x8?file=/src/index.js
- React: https://codesandbox.io/s/react-cloneelement-defaultprops-csr-only-w3x0ii?file=/src/index.js

### DIFF
--- a/test/browser/cloneElement.test.js
+++ b/test/browser/cloneElement.test.js
@@ -1,3 +1,4 @@
+import { Component } from 'preact';
 import { createElement, cloneElement, createRef } from 'preact';
 
 /** @jsx createElement */
@@ -86,8 +87,15 @@ describe('cloneElement', () => {
 	});
 
 	it('should prevent undefined properties from overriding default props', () => {
-		const element = <div style={{ color: 'blue' }}>thing</div>;
+		class Example extends Component {
+			render(props) {
+				return <div style={{ color: props.color }}>thing</div>;
+			}
+		}
+		Example.defaultProps = { color: 'blue' };
+
+		const element = <Example color="red" />;
 		const clone = cloneElement(element, { color: undefined });
-		expect(clone.props.style.color).to.equal('blue');
+		expect(clone.props.color).to.equal('blue');
 	});
 });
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
This is a great catch! I wasn't aware that `defaultProps` play a role for `cloneElement`. Love the codesandbox comparisons and that you already included a test case in your proposed changes. This is a fantastic PR!

test/browser/cloneElement.test.js
```diff
@@ -84,4 +84,10 @@ describe('cloneElement', () => {
 		const clone = cloneElement(div, { key: 'myKey' });
 		expect(clone.props.key).to.equal(undefined);
 	});
+
+	it('should prevent undefined properties from overriding default props', () => {
+		const element = <div style={{ color: 'blue' }}>thing</div>;
+		const clone = cloneElement(element, { color: undefined });
```
andrewiggins, line 90: Let's update this test to use a Component that defines `defaultProps` to exercise the new code path. Perhaps something like:
```js
class Example extends Component {
  render(props) {
    return <div style={{ color: props.color }}>thing</div>;
  }
}

Example.defaultProps = { color: blue };

const element = <Example color="red" />
const clone = cloneElement(Example, { color: undefined });
```

or somethin like that

nnhk23: this is a good point! Adjusted, thank you

marvinhagemeister (approved):

andrewiggins (approved):
Thanks Kim!!

PR #3970, opened 2023-04-16
Add some JSX attribute types

kidonng:
- [`inert`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert) global attribute
- [`hidden`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) global attribute: `until-found` (new) and `hidden` (missing) value

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -1716,7 +1716,12 @@ export namespace JSXInternal {
 			| string
 			| undefined
 			| SignalLike<number | string | undefined>;
-		hidden?: boolean | undefined | SignalLike<boolean | undefined>;
+		hidden?:
+			| boolean
+			| 'hidden'
+			| 'until-found'
+			| undefined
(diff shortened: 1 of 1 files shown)


src/jsx.d.ts
```diff
@@ -1725,6 +1730,7 @@ export namespace JSXInternal {
 		icon?: string | undefined | SignalLike<string | undefined>;
 		id?: string | undefined | SignalLike<string | undefined>;
 		indeterminate?: boolean | undefined | SignalLike<boolean>;
+		inert?: boolean | undefined | SignalLike<boolean>;
```
rschristian, line 1733:
```suggestion
		inert?: boolean | undefined | SignalLike<boolean | undefined>;
```

Missed the `undefined`, I think

rschristian (approved):
Thanks! We appreciate you taking the time to add these.

PR #3974, opened 2023-04-19
Fix function child not skipped

marvinhagemeister:
Fixes 3 test cases.


marvinhagemeister:
nvm, there are some more cases

PR #3977, opened 2023-04-21
update abb patch

jpnelson:


src/create-context.js
```diff
@@ -52,7 +52,10 @@ export function createContext(defaultValue, contextId) {
 					subs.push(c);
 					let old = c.componentWillUnmount;
 					c.componentWillUnmount = () => {
-						subs.splice(subs.indexOf(c), 1);
+						// Patch: this is a hot path, and taking sub out of the array is much
+						// faster this way since it's an unordered list.
+						subs[subs.indexOf(c)] = subs[subs.length - 1];
+						subs.pop();
```
andrewiggins, line 58: Can you also open a PR for against to `master`? Happy to do that as well if you are busy :)

developit, line 58: not sure if smaller/faster/slower:

```suggestion
						subs[subs.indexOf(c)] = subs.pop();
```

jpnelson: I haven't tested it but I can't see how that'd be any different from having a separate assignment (maybe it'd even get minified to the same thing?) so that's the version I've put into the PR here: https://github.com/preactjs/preact/pull/3979

jpnelson: Yes for sure! https://github.com/preactjs/preact/pull/3979

andrewiggins: I think this suggestion fails if you are removing the last item in the array

developit: indeed it does!

JoviDeCroock:
Closing due to inactivity

PR #3979, opened 2023-04-22
Add a fix for context unmounting, from experimenting with our application

jpnelson:
In testing for our application, this ended up being a fair bit faster than how it was during client routing transitions (that is, when unmounting lots of context subscribers)

I don't know if we need a benchmark specifically for having many context subscribers unmount, but that would be the use case that we ran into with this one.

### DIFF
--- a/src/create-context.js
+++ b/src/create-context.js
@@ -25,7 +25,7 @@ export function createContext(defaultValue, contextId) {
 
 				this.getChildContext = () => ctx;
 
-				this.shouldComponentUpdate = function (_props) {
+				this.shouldComponentUpdate = function(_props) {
 					if (this.props.value !== _props.value) {
 						// I think the forced value propagation here was only needed when `options.debounceRendering` was being bypassed:
 						// https://github.com/preactjs/preact/commit/4d339fb803bea09e9f198abf38ca1bf8ea4b7771#diff-54682ce380935a717e41b8bfc54737f6R358
@@ -52,7 +52,9 @@ export function createContext(defaultValue, contextId) {
 					subs.push(c);
 					let old = c.componentWillUnmount;
 					c.componentWillUnmount = () => {
-						subs.splice(subs.indexOf(c), 1);
+						// This is a hot path so we want to be careful about how we're removing items from the subscriber array to keep this fast.
+						// This is faster than splicing an item out in testing.
+						subs[subs.indexOf(c)] = subs.pop();
 						if (old) old.call(c);
 					};
 				};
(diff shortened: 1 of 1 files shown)


src/create-context.js
```diff
@@ -25,7 +25,7 @@ export function createContext(defaultValue, contextId) {
 
 				this.getChildContext = () => ctx;
 
-				this.shouldComponentUpdate = function (_props) {
+				this.shouldComponentUpdate = function(_props) {
```
jpnelson, line 28: I didn't intend to change the formatting here but it seems as though the precommit hooks are doing that for me. I might have my setup incorrect, or maybe this file just needs these fixes with the config that exists?

marvinhagemeister: All good, the same happens on my end. I guess somehow a commit sneaked through one PR that didn't apply the prettier formatting from the pre-commit hook.

marvinhagemeister (approved):
Love this! It's a clever trick and one of those optimizations where one wonders why nobody thought of that before. I think nobody would say no to a benchmark, but it's totally fine if that's too much work. I'm more than happy to merge the PR as is.

Thank you for filing a PR and for helping to make Preact even better 🙌

src/create-context.js
```diff
@@ -52,7 +52,9 @@ export function createContext(defaultValue, contextId) {
 					subs.push(c);
 					let old = c.componentWillUnmount;
 					c.componentWillUnmount = () => {
-						subs.splice(subs.indexOf(c), 1);
+						// This is a hot path so we want to be careful about how we're removing items from the subscriber array to keep this fast.
+						// This is faster than splicing an item out in testing.
+						subs[subs.indexOf(c)] = subs.pop();
```
andrewiggins, line 57: I think this doesn't work if `c` is the last element in the array, since it'll reassign the last element back to it's original index. I think we have to do this in two lines.

JoviDeCroock (approved):

src/create-context.js
andrewiggins: Try reinstalling node_modules. We recently upgraded our prettier version so maybe that's the reason for the discrepancy? Either way, no biggie.

```
> npm ls prettier
preact@10.13.2 C:\code\github\preactjs\preact
└── prettier@2.8.6
```

andrewiggins (changes requested):
One additional thought

src/create-context.js
marvinhagemeister: oh right, I missed that in my review. Also what happens if there is only one entry in the array?

developit: We could consider using `Set` for this if the Array is a pain, or switching to a sparse array (would require a guard on L45).

Thompson1985 (approved):

JoviDeCroock:
Addressed in https://github.com/preactjs/preact/pull/4526

PR #3980, opened 2023-04-22
Fix debug tests

marvinhagemeister:
This PR fixes the debug tests and a few compat ones. Got the total failure count down to 5. The remaining failures are all for `forwardRef` and it seems like we're putting `ref` back in props at different stages for vnodes vs internals. Haven't found out why though, but I think the existing fixes are already worth PR'ing.

### DIFF
--- a/src/diff/mount.js
+++ b/src/diff/mount.js
@@ -13,7 +13,7 @@ import {
 	MODE_SVG,
 	DIRTY_BIT
 } from '../constants';
-import { normalizeToVNode, createElement, Fragment } from '../create-element';
+import { createElement, Fragment } from '../create-element';
 import { setProperty } from './props';
 import { createInternal, getParentContext } from '../tree';
 import options from '../options';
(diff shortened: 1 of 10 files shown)


src/diff/mount.js
```diff
@@ -237,7 +238,13 @@ export function mountChildren(parentInternal, children, parentDom, startDom) {
 		vnode = children[i];
 
 		// account for holes by incrementing the index:
-		if (vnode == null || vnode === true || vnode === false) continue;
+		if (
+			vnode == null ||
+			vnode === true ||
+			vnode === false ||
+			typeof vnode === 'function'
```
andrewiggins, line 245: Should we do this typeof once and reuse it's result here and line 249?

Same for patch

andrewiggins (approved):
Thanks for looking into this!

src/diff/mount.js
marvinhagemeister: Just checked and storing the result in a variable is 10b larger. I think for now it's fine to leave it as is.

Thompson1985 (approved):

JoviDeCroock:
Closing: targets the throwaway `v11-linked-list-prev-index-nextDom-2` experiment branch, whose linked-list internals were not adopted.

PR #3986, opened 2023-04-25
Set-based alternative to #3979

developit:
Mainly to see if this moves benches to the same extent.


andrewiggins:
Unfortunately I don't think many (any?) of our benchmarks use a lot of context so they may not exercise this code path enough to really see some differences.

Something to keep in mind for our script to capture real world vdom trees. Would be useful to know what nodes are context producers and consumers.

PR #3987, opened 2023-04-28
try local setter for options.vnode

JoviDeCroock:


src/options.js
```diff
@@ -13,4 +15,21 @@ const options = {
...
+		return vnodeHook;
+	},
+	set(newHook) {
+		if (newHook) {
+			let oldHook = vnodeHook;
+			vnodeHook = (vnode) => {
+				newHook(vnode);
+				if (oldHook) oldHook(vnode);
+			};
+		} else {
```
JoviDeCroock, line 29: we need to handle undefined so we can reset the value in our tests

src/options.js
```diff
@@ -13,4 +15,21 @@ const options = {
...
+Object.defineProperty(options, 'vnode', {
+	get() {
+		return vnodeHook;
+	},
+	set(newHook) {
+		if (newHook) {
+			let oldHook = vnodeHook;
+			vnodeHook = (vnode) => {
+				newHook(vnode);
+				if (oldHook) oldHook(vnode);
```
andrewiggins, line 27: Yes!

Though I think we can't auto apply the old hook cuz I think some implementations need to invoke their logic before and/or after the previous hooks. For example, if I want to modify something `preact/hooks` does I may want to invoke my implementation after invoking the previous hook to ensure hooks stuff runs first.

However, I like this idea of if simplifying hook integrations. I think we had previously discussed adding something like a `preact/shared-internals` package or something and exposing a [hook helper like we have in signals](https://github.com/preactjs/signals/blob/6b6af05d433b4d2a8da0be5a65ba5305d49e7b73/packages/preact/src/index.ts#L27-L31) that would always ensure the first argument to the hook is the previous hook. That doesn't need to happen in this PR but just thought I'd write it down cuz I'm not sure we have that written down elsewhere 😅

JoviDeCroock: I updated it, just have one test that seems to be failing, not entirely sure how to remedy it yet will look deeper over the next week.

marvinhagemeister:
I like this PR a lot!

src/options.js
JoviDeCroock: The perf diff does not seem to be working atm

src/options.js
```diff
@@ -9,8 +21,105 @@ import { _catchError } from './diff/catch-error';
  * ctrl+click or cmd+click on mac the type definition below).
  * @type {import('./internal').Options}
  */
-const options = {
-	_catchError
-};
+const options = {};
+
+Object.defineProperty(options, 'vnode', {
```
developit, line 26: I wonder if we could use a helper to define the value field and the options hook property at the same time?

```js
export let vnode = optionsHook('vnode');

/**
 * @template {keyof import('./internal').Options} T
 * @param {T} name
 * @param {any} [value] - initial value for the hook
 * @returns {import('./internal').Options[T]}
 */
function optionsHook(name, value) {
	Object.defineProperty(options, 'unmount', {
		get() {
			return value;
		},
		set(newValue) {
			value = newValue;
		}
	});
	return value;
}
```

JoviDeCroock:
Something weird is going on here, locally the tests fail however in CI they are good 😅 could be a diff between watch and normal runs

src/options.js
JoviDeCroock: @developit the byte-size change is actually very minimal

JoviDeCroock:
Closing as the issues with tests are a bit too much

PR #3996, opened 2023-05-07
fix: Missing `ForwardRefExoticComponent` and `RefAttributes` types in compat

### LINKED ISSUE #3759: (preact/compat types) Add `ForwardRefExoticComponent`

**Describe the bug**

Using `Popper` from `@mui/material` resolves the type to `any` (unresolved) because Popper uses the type `React.ForwardRefExoticComponent`:

```tsx
declare const Popper: React.ForwardRefExoticComponent<Pick<PopperProps, "hidden" | "color" | ...> & React.RefAttributes<HTMLDivElement>>;
```

`react` is path-mapped to `preact/compat` in the tsconfig, but `preact/compat` does not provide `ForwardRefExoticComponent`.

rschristian:
Fixes #3759, though only the immediate issue. `Popper` from `@mui/material` still does not type check correctly -- I added about a half-dozen types in compat and still couldn't get any type errors. We're missing a fair few things I guess, but that lib is such a maze of imports that I couldn't even track down all of what we were missing. Issue for another day I think?

Adapted from React's types [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/8730660c37a33dc0768b3f1e36b679294acc4988/types/react/index.d.ts#L802-L805) and [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/8730660c37a33dc0768b3f1e36b679294acc4988/types/react/index.d.ts#L134-L141)

---

To provide a simpler and more cut down reproduction than the original issue provided, the following example React library will generate the following types:

```tsx
import { forwardRef } from 'react';

export const MyInput = forwardRef<HTMLInputElement, { label: string }>(function MyInput(props, ref) {
    return (
        <label>
            {props.label}
            <input ref={ref} />
        </label>
    );
});
```

```ts
export declare const MyInput: import("react").ForwardRefExoticComponent<{
    label: string;
} & import("react").RefAttributes<HTMLInputElement>>;
```

Aliasing `react` to `preact/compat` using `"paths"` in a `tsconfig.json` would convert the type of `MyInput` into `any` as `ForwardRefExoticComponent` and `RefAttributes` could not be found.


JoviDeCroock (approved):

robertmaier:
This is great :) I ran into the same problem integrating a React library today. Exchanging the types locally works as expected.

@JoviDeCroock Are there any plans to include this in one of the upcoming releases?

JoviDeCroock:
@robertmaier people part of the org can merge their own PR's, I gave the green light by approving it but Ryan can merge himself when he feels like it :) it will then be part of a release later on

rschristian:
Whoops, totally forgot about this. Will merge in a minute.

PR #4010, opened 2023-05-18
Port skew based child diffing

### LINKED ISSUE #2622: diff/patch unnecessarily re-inserts children when removing >1 item and items are class components
Moved from https://github.com/preactjs/preact/issues/2619#issuecomment-657509681

While trying to build an isolated example for https://github.com/preactjs/preact/issues/2619 I've found another thing, which I'm not sure if it is directly related.

Here is a static html to reproduce -

```
<!DOCTYPE html>
<html>
<head>
</head>
<body>

    <div id="app"></div>
    <div id="log"></div>

<script type="module">

    import {h, render, Component} from 'https://unpkg.com/preact@10.4.5/dist/preact.module.js';

    let el = document.getElementById('app'),
        log = document.getElementById('log');


    let observer = new MutationObserver(function(mutations){

        function items(r){
            return r.type == 'childList';
        }

        function out(r){

            let s = '';

            r.addedNodes.forEach(n => {
                s += 'add ' + n.innerText;
            });

            r.removedNodes.forEach(n => {
                s += 'remove ' + n.innerText;
            });

            return s;
        }

        log.innerHTML = mutations.filter(items).map(out);
    });



    class Item extends Component {
        render() {
            return h('span', {}, this.props.children);
        }
    }

//    function item(i){
//        return h(Item, {key: i}, i);
//    }

    function item(i){
        return h('span', {key: i}, i);
    }

    let data = [1,2,3,4,5,6,7,8,9];
    let app = h('div', {}, data.map(item));
    render(app, el);


    observer.observe(el, {childList: true, subtree: true});


//  remove 3 items from the middle
    data.splice(3, 3);
    let app2 = h('div', {}, data.map(item));
    render(app2, el);


</script>
</body>
</html>
```

Here I am rendering a list of items and then removing 3 from the middle. When the item is an html tag the result looks like this 
```
123789
remove 6,remove 5,remove 4
```
which is exactly what I expect. However if the item is a component (uncomment another `function item(i)`  block), the output is different -
```
123789
remove 7,add 7,remove 8,add 8,remove 9,add 9,remove 6,remove 5,remove 4
```
which IMHO is a bug as I would expect exactly the same mutations.

This behavior is present starting from 10.0.0.

### LINKED ISSUE #2783: Unnecessary unmounting/mounting children with keys

### LINKED ISSUE #2987: One element may be rendered in place of another in case of conditional rendering

### LINKED ISSUE #3973: iframe is reloaded when siblings before it in a list are removed

**Describe the bug**
Seems similar to [this previously resolved bug](https://github.com/preactjs/preact/issues/2350)

Basically if you have a list of nodes with one that renders an iframe anywhere besides the first position, if all of the elements before it are removed the iframe is reloaded on the next render.

**To Reproduce**

https://codesandbox.io/s/jolly-sinoussi-n92mru?file=/src/test.js:320-331

1. Go to https://codesandbox.io/s/jolly-sinoussi-n92mru?file=/src/test.js:320-331
2. Click on "Toggle" -- this will remove the test elements from the dom
3. See the iframe reload

**Expected behavior**
The iframe should not reload

JoviDeCroock:
This ports the skew-based child diffing we implemented in v11 to Preact X, it seems to solve a lot of performance issues and generally improves correctness. The original PR for v11 can be found in https://github.com/preactjs/preact/pull/3388

We would be able to benefit from a lot more stuff if we could use `getDomSibling` everywhere like in v11 https://github.com/preactjs/preact/blob/v11/src/diff/children.js#L156

TODO:

- [x] golf bytes (200 --> 60 bytes)
- [x] check why performance goes down (potentially more calls on `reorderChildren`)
- [x] investigate perf drop-off in hydration and text-update

Fixes #3973 
Fixes #2622
Fixes #2783
Fixes #2987

### DIFF
--- a/benches/package-lock.json
+++ b/benches/package-lock.json
@@ -1,6371 +1,6371 @@
 {
-  "name": "preact-benchmarks",
-  "version": "1.0.0",
-  "lockfileVersion": 2,
-  "requires": true,
-  "packages": {
-    "": {
-      "name": "preact-benchmarks",
-      "version": "1.0.0",
-      "license": "MIT",
-      "dependencies": {
-        "afterframe": "^1.0.1"
-      },
-      "devDependencies": {
-        "@kristoferbaxter/async": "^1.0.0",
-        "del": "^7.0.0",
-        "escalade": "^3.1.1",
-        "escape-string-regexp": "^5.0.0",
-        "globby": "^13.1.2",
-        "prompts": "^2.4.2",
-        "sade": "^1.8.1",
-        "strip-ansi": "^7.0.1",
-        "tachometer": "^0.7.0",
-        "v8-deopt-viewer": "^0.2.1"
-      }
-    },
-    "node_modules/@babel/code-frame": {
-      "version": "7.10.4",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.4.tgz",
-      "integrity": "sha512-vG6SvB6oYEhvgisZNFRmRCUkLz11c7rp+tbNTynGqc6mS1d5ATd/sGyV6W0KZZnXRKMTzZDRgQT3Ou9jhpAfUg==",
-      "dev": true,
-      "dependencies": {
-        "@babel/highlight": "^7.10.4"
-      }
-    },
-    "node_modules/@babel/generator": {
-      "version": "7.11.6",
-      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.11.6.tgz",
-      "integrity": "sha512-DWtQ1PV3r+cLbySoHrwn9RWEgKMBLLma4OBQloPRyDYvc5msJM9kvTLo1YnlJd1P/ZuKbdli3ijr5q3FvAF3uA==",
-      "dev": true,
-      "dependencies": {
-        "@babel/types": "^7.11.5",
-        "jsesc": "^2.5.1",
-        "source-map": "^0.5.0"
-      }
-    },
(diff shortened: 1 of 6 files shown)


marvinhagemeister (approved):

marvinhagemeister:
Approving because bench results seem to be an issue with our setup

PR #4022, opened 2023-05-22
[v11] use Text constructor

shoonia:
Use `new Text(text)` instead `document.createTextNode(text)`

MDN: https://developer.mozilla.org/en-US/docs/Web/API/Text

```js
document.createTextNode('text') instanceof Text // true
```

### DIFF
--- a/src/diff/mount.js
+++ b/src/diff/mount.js
@@ -130,7 +130,7 @@ function mountElement(internal, dom, refs) {
 	if (flags & TYPE_TEXT) {
 		if (isNew) {
 			// @ts-ignore createTextNode returns Text, we expect PreactElement
-			dom = document.createTextNode(newProps);
+			dom = new Text(newProps);
 		} else if (dom.data !== newProps) {
 			dom.data = newProps;
 		}
--- a/src/diff/mount.js
+++ b/src/diff/mount.js
@@ -129,7 +129,7 @@ function mountElement(internal, dom, refs) {
 
 	if (flags & TYPE_TEXT) {
 		if (isNew) {
-			// @ts-ignore createTextNode returns Text, we expect PreactElement
+			// @ts-ignore returns Text, we expect PreactElement
 			dom = new Text(newProps);
 		} else if (dom.data !== newProps) {
 			dom.data = newProps;


shoonia:
I can't see some big difference between methods (on my laptop). There are have equal performance
https://www.measurethat.net/Benchmarks/Show/25329/0/documentcreatetextnodetext-vs-new-texttext

dwknippers:
> I can't see some big difference between methods (on my laptop). There are have equal performance https://www.measurethat.net/Benchmarks/Show/25329/0/documentcreatetextnodetext-vs-new-texttext

Noticing about a 8-9% increase in linked benchmark on Firefox 113.0.2 on Windows 10.

developit:
I believe we actually had this landed in v11 at one point (or another branch), but the [benchmark results](https://esbench.com/bench/5b2c4fa6f2949800a0f61db6) were inconclusive:

Screenshot 2023-06-05 at 9 33 50 AM

I'm not against moving to the Text constructor, but I do worry about it regressing when we fix `ownerDocument` (as we'd either have to do `new ownerDocument.defaultView.Text()` or create Text nodes in the wrong document and rely on implicit importNode).

dwknippers:
Most likely this discrepancy in results is due to implementation differences between Firefox and Chromium based browsers.

I'm able to replicate the same 7-8% difference on Firefox on Android and even a 5% increase on chrome on Android using the first benchmark.

Seems best to base these kinds of decisions based on maintainability, readability and the magnitude of difference uncovered by more cross-platform benchmarks.

Uncovered preliminary it seems to provide sufficiently significant benefits compared to the old method, though ultimately thus depends on a more educated overall decision.

dwknippers:
> I believe we actually had this landed in v11 at one point (or another branch), but the [benchmark results](https://esbench.com/bench/5b2c4fa6f2949800a0f61db6) were inconclusive

Running the mentioned benchmark seems to contradict the benchmark listed by shoonia
> I can't see some big difference between methods (on my laptop). There are have equal performance
> https://www.measurethat.net/Benchmarks/Show/25329/0/documentcreatetextnodetext-vs-new-texttext

Using the other benchmark it seems to result in a drop in performance, most likely its safe to close this issue.

PR #4026, opened 2023-05-23
Add W3C aria attributes

shoonia:
- [aria-braillelabel](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-braillelabel)
- [aria-brailleroledescription](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-brailleroledescription)


rschristian (approved):
Thanks, for these and all the other types you've added! Really appreciate it.

PR #4028, opened 2023-05-24
Adds missing Context and Dispatch types to compat

### LINKED ISSUE #4027: Missing types for Dispatch and Context in the React namespace

**Describe the bug**
I've recently started working with a 3rd party library which was designed for react. In their types they use `React.Context` and `React.Dispatch`. In my TS config, I'm aliasing react types to `preact/compat`. But I noticed these specific types were missing from the exported React namespace and it ultimately ends up throwing during my build.

If I go in and patch the preact types, this ends up fixing the problem entirely. A type for Preact.Context already exists but just isn't being exported in the `React` namespace (only ContextType is which doesn't match the name react uses). I didn't see one for Dispatch so that one had to be copied from the react type.

```ts
type Dispatch<A> = (value: A) => void;
export import Context = preact.Context;
```

So I was wondering if there was a specific reason why these two types are not exported under the React namespace. If it was just an oversight, I'd be happy to open up a PR to apply my patch upstream.

Scott-Fischer:
Adds two missing types to the exported React namespace in compat.
Closes #4027

- React.Dispatch (sourced from [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f05d989e1877176af719e03008da70bbc9585c60/types/react/index.d.ts#LL892C1-L892C1))
- React.Context (this one already existed under preact.Context, it just is not being re-exported under the React namespace)

### DIFF
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -30,7 +30,7 @@ declare namespace React {
 	export import useReducer = _hooks.useReducer;
 	export import useRef = _hooks.useRef;
 	export import useState = _hooks.useState;
-	export type Dispatch<A> = (value: A) => void;
+	export type Dispatch<A> = (action: A) => void;
 	// React 18 hooks
 	export import useInsertionEffect = _hooks.useLayoutEffect;
 	export function useTransition(): [false, typeof startTransition];
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -30,6 +30,7 @@ declare namespace React {
 	export import useReducer = _hooks.useReducer;
 	export import useRef = _hooks.useRef;
 	export import useState = _hooks.useState;
+	export type Dispatch<A> = (value: A) => void;
 	// React 18 hooks
 	export import useInsertionEffect = _hooks.useLayoutEffect;
 	export function useTransition(): [false, typeof startTransition];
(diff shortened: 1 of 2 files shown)


compat/src/index.d.ts
```diff
@@ -30,6 +30,7 @@ declare namespace React {
 	export import useReducer = _hooks.useReducer;
 	export import useRef = _hooks.useRef;
 	export import useState = _hooks.useState;
+	export type Dispatch<A> = (value: A) => void;
```
rschristian, line 33:
```suggestion
	export type Dispatch<A> = (action: A) => void;
```

Might be worth extracting to core's types for reuse in submodules, not sure. We do use it in a few different places.

Scott-Fischer: @rschristian Do you mean define the type in `./src/index.d.ts` as part of preact's namespace and re-export in the compat types via `./compat/src/index.d.ts`?

rschristian: Sorry, not core, but to hooks, and importing from there, as it could directly be used in [`useReducer`](https://github.com/preactjs/preact/blob/bd593c49423f1f49a4cfbd46bbe53f8a396aaaa1/hooks/src/index.d.ts#L30)'s types. 

Apologies, was looking at a couple of different issues at once and mixed them up.

Scott-Fischer: Okay sounds good. I'm making this change now then will mark this as ready for review.

JoviDeCroock (approved):

rschristian (approved):
Thanks, looks great!

PR #4030, opened 2023-05-26
10.15.1

JoviDeCroock:
- Add W3C aria attributes (#4026, thanks @shoonia)
- Adds missing Context and Dispatch types to compat (#4028, thanks @Scott-Fischer)
- Fix TS FunctionComponent w/ exactOptionalPropertyTypes (#4025, thanks @rschristian)

### DIFF
--- a/devtools/src/devtools.js
+++ b/devtools/src/devtools.js
@@ -2,7 +2,7 @@ import { options, Fragment, Component } from 'preact';
 
 export function initDevTools() {
 	if (typeof window != 'undefined' && window.__PREACT_DEVTOOLS__) {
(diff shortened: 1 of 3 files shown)


marvinhagemeister (approved):

rschristian:
Not sure when you were planning on merging, but #4028 might be worth waiting a day or two for, if we don't want to finish it off ourselves

Scott-Fischer:
> Not sure when you were planning on merging, but #4028 might be worth waiting a day or two for, if we don't want to finish it off ourselves

https://github.com/preactjs/preact/pull/4028 is now merged in and good to go!

PR #4047, opened 2023-06-15
Server compat should have a default export

MikaFima:
Hi preact,

For now the use of preact/compat do not allow to use `import ReactDOM from 'react-dom/server'`, as it will fail with an error:
image

This PR is here to allow it, without breaking the classical use `import {renderToStaticMarkup} from 'react-dom/server'`

### DIFF
--- a/compat/server.browser.js
+++ b/compat/server.browser.js
@@ -1,4 +1,11 @@
+import { renderToString } from 'preact-render-to-string';
+
 export {
 	renderToString,
 	renderToString as renderToStaticMarkup
 } from 'preact-render-to-string';
+
+export default {
+	renderToString,
+	renderToStaticMarkup: renderToString
+};
--- a/compat/server.mjs
+++ b/compat/server.mjs
@@ -1,4 +1,11 @@
+import { renderToString } from 'preact-render-to-string';
+
 export {
 	renderToString,
 	renderToString as renderToStaticMarkup
 } from 'preact-render-to-string';
+
+export default {
+	renderToString,
+	renderToStaticMarkup: renderToString
+};


MikaFima:
Ok, as you can see with your client.js https://github.com/preactjs/preact/blob/master/compat/client.mjs, you need a default export to be imported with `import ReactDOM from 'react-dom/server'`. 

Do you prefer I add some tests to make it sure both imports work as expected ?

compat/server.browser.js
```diff
@@ -1,4 +1,9 @@
...
+
+export const renderToString = preactRenderToString;
+export const renderToStaticMarkup = preactRenderToString;
+
+export default {
 	renderToString,
-	renderToString as renderToStaticMarkup
-} from 'preact-render-to-string';
+	renderToStaticMarkup
+};
```
rschristian, line 9: What @marvinhagemeister is referring to is that you're not just adding a default export, but changing the style of the existing exports. Simply append the default export change, no need to alter existing exports.

compat/server.js
```diff
@@ -9,7 +9,5 @@ try {
 	);
 }
 
-module.exports = {
-	renderToString: renderToString,
-	renderToStaticMarkup: renderToString
-};
+exports.renderToString = renderToString;
+exports.renderToStaticMarkup = renderToString;
```
rschristian, line 13: These are functionally identical, so I'm not quite sure why this was done?

compat/server.browser.js
MikaFima: Understood, i'm doing it

compat/server.js
MikaFima: I've done it like that to keep consistency with client.js

rschristian: Consistency is a nice-to-have, but lets not change this here. It increases the diff size and messes with the blame unnecessarily.

MikaFima: Okay changes done

MikaFima:
Is that okay for you ?

compat/server.browser.js
```diff
@@ -1,4 +1,11 @@
+import { renderToString as preactRenderToString } from 'preact-render-to-string';
```
rschristian, line 1: nit: No need for the `as`, just keep it as `renderToString`

MikaFima: done

compat/server.mjs
```diff
@@ -1,4 +1,11 @@
+import { renderToString as preactRenderToString } from 'preact-render-to-string';
```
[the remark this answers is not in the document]
MikaFima: done

JoviDeCroock (approved):

PR #4051, opened 2023-06-25
fix: incorrect actDepth calculation in test-utils

ottomao:
I am submitting this pull request to address a bug in the current implementation of the `act` method. Specifically, when the `act` method is called recursively and an inner callback throws an exception, it causes `actDepth` to be calculated incorrectly, leading to issues with subsequent `act` method calls.


marvinhagemeister (approved):
Great catch! This is a well done PR and love the test for it. Thank you so much for this contribution 💯

PR #4054, opened 2023-06-28
fix: setting ref to null after updating it with new element

### LINKED ISSUE #4049: Unexpected null ref

**Describe the bug**

A ref is unexpectedly null when using a particular JSX structure.

**To Reproduce**

[Demo](https://preactjs.com/repl?code=aW1wb3J0IHsgcmVuZGVyIH0gZnJvbSAicHJlYWN0IjsKaW1wb3J0IHsgdXNlU3RhdGUsIHVzZVJlZiwgdXNlRWZmZWN0IH0gZnJvbSAicHJlYWN0L2hvb2tzIjsKCmZ1bmN0aW9uIERlbW8oKSB7CiAgY29uc3QgW3BoYXNlLCBzZXRQaGFzZV0gPSB1c2VTdGF0ZSgxKTsKICBjb25zdCBvbmVSZWYgPSB1c2VSZWYobnVsbCk7CiAgY29uc3QgdHdvUmVmID0gdXNlUmVmKG51bGwpOwogIGNvbnN0IHRocmVlUmVmID0gdXNlUmVmKG51bGwpOwoKICB1c2VFZmZlY3QoKCkgPT4gewogICAgc2V0VGltZW91dCgoKSA9PiBzZXRQaGFzZSgyKSwgMTAwMCk7CiAgfSwgW10pOwoKICB1c2VFZmZlY3QoKCkgPT4gewogICAgY29uc29sZS5sb2coewoJCSAJcGhhc2UsCiAgICAgIG9uZVJlZjogb25lUmVmLmN1cnJlbnQsCiAgICAgIHR3b1JlZjogdHdvUmVmLmN1cnJlbnQsCiAgICAgIHRocmVlUmVmOiB0aHJlZVJlZi5jdXJyZW50LAogICAgfSk7CiAgfSk7CgogIHJldHVybiAoCiAgICA8PgogICAgICB7cGhhc2UgPT09IDEgPyAoCiAgICAgICAgPGRpdj4KICAgICAgICAgIDxkaXYgcmVmPXt0d29SZWZ9PkVsZW1lbnQgdHdvPC9kaXY%2BCiAgICAgICAgICA8ZGl2IHJlZj17dGhyZWVSZWZ9PkVsZW1lbnQgdGhyZWU8L2Rpdj4KICAgICAgICA8L2Rpdj4KICAgICAgKSA6IHBoYXNlID09PSAyID8gKAogICAgICAgIDxkaXYgY2xhc3M9Im91dGVyIj4KICAgICAgICAgIDxkaXYgcmVmPXtvbmVSZWZ9PkVsZW1lbnQgb25lPC9kaXY%2BCiAgICAgICAgICA8ZGl2IGNsYXNzPSJ3cmFwcGVyIj4KICAgICAgICAgICAgPGRpdiByZWY9e3R3b1JlZn0%2BRWxlbWVudCB0d288L2Rpdj4KICAgICAgICAgICAgPGRpdiByZWY9e3RocmVlUmVmfT5FbGVtZW50IHRocmVlPC9kaXY%2BCiAgICAgICAgICA8L2Rpdj4KICAgICAgICA8L2Rpdj4KICAgICAgKSA6IG51bGx9CiAgICA8Lz4KICApOwp9CgpyZW5kZXIoPERlbW8gLz4sIGRvY3VtZW50LmdldEVsZW1lbnRCeUlkKCJhcHAiKSk7Cg%3D%3D).

View the console. I see:

```
{phase: 1, oneRef: null, twoRef: div, threeRef: div}
{phase: 2, oneRef: div, twoRef: null, threeRef: div}
```

**Expected behaviour**

```
{phase: 1, oneRef: null, twoRef: div, threeRef: div}
{phase: 2, oneRef: div, twoRef: div, threeRef: div}
```

It's expected that `oneRef` is `null` in phase 1 (the element isn't rendered), but `twoRef` shouldn't be `null` in phase 2, as the element _is_ rendered.

The bug seems to depend on a few things:

If `<div class="wrapper">` is switched to a fragment, the bug still happens. However, if `<div class="wrapper">` is removed, the bug does not happen.

If `<div class="outer">` is switched to a fragment, the bug does not happen.

JoviDeCroock:
fixes #4049 

As explained in https://github.com/preactjs/preact/issues/4049#issuecomment-1605314450 we did all unsetting/setting logic during diff which meant that if we changed the level of an element we would set --> unset if the level moved lower down.

To counter-act this we execute unsetting first and setting later, which is also inline with how [react does it](https://codesandbox.io/s/blissful-franklin-tq2n95?file=/src/App.js) in React you see a series of unsets and only after you see a series of sets.

Because we queue `useLayoutEffect` to the commit phase we still ensures that refs are set correctly.

### DIFF
--- a/hooks/test/browser/combinations.test.js
+++ b/hooks/test/browser/combinations.test.js
@@ -398,4 +398,83 @@ describe('combinations', () => {
 		});
 		expect(scratch.textContent).to.equal('2');
 	});
+
+	it.skip('parent and child refs should be set before all effects', () => {
+		const anchorId = 'anchor';
+		const tooltipId = 'tooltip';
+		const effectLog = [];
+
+		let useRef2 = sinon.spy(init => {
+			const realRef = useRef(init);
+			const ref = useRef(init);
+			Object.defineProperty(ref, 'current', {
+				get: () => realRef.current,
+				set: value => {
+					realRef.current = value;
+					effectLog.push('set ref ' + value?.tagName);
+				}
+			});
+			return ref;
+		});
+
+		function Tooltip({ anchorRef, children }) {
+			// For example, used to manually position the tooltip
+			const tooltipRef = useRef2(null);
+
+			useLayoutEffect(() => {
+				expect(anchorRef.current?.id).to.equal(anchorId);
+				expect(tooltipRef.current?.id).to.equal(tooltipId);
+				effectLog.push('tooltip layout effect');
+			}, [anchorRef, tooltipRef]);
+			useEffect(() => {
+				expect(anchorRef.current?.id).to.equal(anchorId);
+				expect(tooltipRef.current?.id).to.equal(tooltipId);
+				effectLog.push('tooltip effect');
+			}, [anchorRef, tooltipRef]);
+
+			return (
+				<div class="tooltip-wrapper">
+					<div id={tooltipId} ref={tooltipRef}>
+						{children}
+					</div>
+				</div>
+			);
+		}
+
+		function App() {
+			// For example, used to define what element to anchor the tooltip to
+			const anchorRef = useRef2(null);
+
+			useLayoutEffect(() => {
+				expect(anchorRef.current?.id).to.equal(anchorId);
+				effectLog.push('anchor layout effect');
+			}, [anchorRef]);
+			useEffect(() => {
+				expect(anchorRef.current?.id).to.equal(anchorId);
+				effectLog.push('anchor effect');
+			}, [anchorRef]);
+
+			return (
+				<div>
+					<p id={anchorId} ref={anchorRef}>
+						More info
+					</p>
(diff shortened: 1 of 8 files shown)


marvinhagemeister (approved):
I like the idea of having a ref queue 👍

andrewiggins (approved):
Oh nice! This fixes an issue I've been chasing for a lil while (#3860), but I like this better. It's quite bit a simpler and less impactful to stuff. Nice!

Would you mind if I pushed some tests from #3860 this PR fixes? I can also do a follow up PR if you prefer :)

JoviDeCroock:
@andrewiggins feel free to push tests or a new PR however you prefer

andrewiggins:
Went ahead and pushed the changes to this PR since they are all related and fixed by this change 🎉

PR #4056, opened 2023-06-30
compat PureComponent parity with upstream react type definitions

Blufords:
PureComponent parity with react type definitions, fixes palantir/blueprint#5498 + similar issues in blueprint when using preact, as many of the components extend PureComponent.

cf https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b28468368df0b6ed8d4cf33683cc46087004d4f4/types/react/index.d.ts#L537


compat/src/index.d.ts
```diff
@@ -106,7 +106,7 @@ declare namespace React {
 		component: preact.Component | Element
 	): Element | null;
 
-	export abstract class PureComponent<P = {}, S = {}> extends preact.Component<
+	export abstract class PureComponent<P = {}, S = {}, SS = any> extends preact.Component<
```
JoviDeCroock, line 109: Should we then also accept this on Component to orchestrate typings for `getSnapshotBeforeUpdate` as React does

Blufords: Sorry, I actually have no idea about any of that as I'm pretty new to using React/Preact, just thought I'd contribute a fix to get Blueprint working since it's just a small issue with types.

ricardobeat:
@JoviDeCroock any chance this will be merged? Looking into using Blueprint with Preact as well.

PR #4064, opened 2023-07-08
'Improve' code quality by changing syntax to a more accurate one

lausuarez02:
I've found that the test asserition could be 'improved' by changing the typeof by a to.be.a('function') and i found it to be quite easy to read and understand plus also deleting some unnecessary checks (typeof).

Hope is somehow useful, thanks!

### DIFF
--- a/jsx-runtime/test/browser/jsx-runtime.test.js
+++ b/jsx-runtime/test/browser/jsx-runtime.test.js
@@ -14,10 +14,10 @@ describe('Babel jsx/jsxDEV', () => {
 	});
 
 	it('should have needed exports', () => {
-		expect(typeof jsx).to.equal('function');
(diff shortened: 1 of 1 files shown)


lausuarez02:
Screenshot 2023-07-08 at 17 44 02
I add just in case the screenshot even though i know the actions would run it too :).

marvinhagemeister (changes requested):
I'm afraid I have to reject this PR. The changes proposed here reflect personal preference and have no effect on correctness, performance, readability or makes debugging easier. 

I'm excited that you're taking a look at our codebase. I'd recommend you pick one issue that seems in your ballpark and make a PR to fix that. Those kind of changes are much more useful for the project.

lausuarez02:
I understand thank you so mucho for replying i'm going to take a look to the issues, thanks!

Issue #2323, opened 2020-02-09
Preact 10 more difficult to debug uncaught errors than Preact 8. 


PepsRyuu:
Recently giving Preact 10 a try after using Preact 8 for a very long time, but having massive difficulties working with it cause of the inability to properly debug with it.

Let's say I had something like the following: 

```
class MyComponent extends Component {
    componentDidMount () {
        a + b; // throws an uncaught error
    }

    render () {
        return <div>Hello World</div>
    }
}
```

In Preact 8, this will happen:

![preact8](https://user-images.githubusercontent.com/6637882/74106165-12a37c80-4b5c-11ea-863b-f18d285d36db.PNG)

In Preact 10, this will happen:

![preact10](https://user-images.githubusercontent.com/6637882/74106168-218a2f00-4b5c-11ea-9db8-dfc5b15048ac.PNG)

In Preact 8, it's super easy to see what went wrong, and to inspect variables and try out different snippets of code to try and fix the issue. In Preact 10 because of the ```try/catch``` for supporting ```componentDidCatch```, I can't do that anymore, making it really frustrating to identify the issue and solve it, especially when using ```eval``` source maps where you can't often can't use the stack trace to click on the relevant line.

Is there any way to resolve this? Can Preact's code be re-organised so that there's no try/catch if there's no ```componentDidCatch``` in the component tree? This issue can make fixing code fairly time consuming so would appreciate any thoughts you have.

Thanks! 🙂

developit:
We need some form of exception catching in core in order to support componentDidCatch, but perhaps there's a way to make it conditional on the existence of a `componentDidCatch` somewhere in the tree. That wouldn't solve all cases, but it would at least fix the case where nothing is catching and we're just rethrowing the unmodified error.

FWIW as a short-term fix, if you check the "Pause on caught exceptions" box, the debugger will pause in the same place it used to in 8.x (and then again where you noted it pausing in 10, though you can blackbox `preact.js` to disable that).

wnayes:
React avoids caught exception issues in debug with its [`invokeGuardedCallbackImpl`](https://github.com/facebook/react/blob/master/packages/shared/invokeGuardedCallbackImpl.js). I can understand that including this sort of mechanism might bloat the package. (Well, it could be included for non-minified debug builds only, but I don't know if most Preact users run this sort of build in practice.)

[...]

PR #4067, opened 2023-07-10
reduce stack size of try catch by excluding non components

JoviDeCroock:
Been thinking about solutions for https://github.com/preactjs/preact/issues/2323, wrapping only when `componentDidCatch` is present does not work as we don't render other children then i.e. Suspense would lose the parallel capabilities and componentDidCatch would throw once when two children error...

Thinking about more ways we can improve this atm

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -53,8 +53,8 @@ export function diff(
 
 	if ((tmp = options._diff)) tmp(newVNode);
 
-	try {
-		outer: if (typeof newType == 'function') {
+	outer: if (typeof newType == 'function') {
+		try {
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
2B savings? I'll take it :D

JoviDeCroock:
@marvinhagemeister it will atleast reduce the stack a bit 😅 generally we might need to think of a way where we wrap `_catchError/..` and only report the first stack we encounter, the thread has something about react and how they do it but need to look deeper into that

JoviDeCroock:
I had an idea on what might reduce the stack-size 😅 not including `diffChildren`

PR #4088, opened 2023-08-03
fix(child-diffing): avoid skipping re-orders

### LINKED ISSUE #4087: List with keys sorted multiple times (ascending & descending) is not displayed in correct order

Installed `preact` version: `10.16.0`

Array of string sorted multiple times (ascending & descending) is not displayed in correct order

**To Reproduce**

Here is the link from code sandbox to reproduce: https://codesandbox.io/s/rough-meadow-w36s6x?file=/src/index.js

Here is the code snippet :

```
import { Component, Fragment, render } from "preact";

const keys = {
  Apple: `Apple_1`,
  Orange: `Orange_1`,
  Banana: `Banana_1`,
  Grape: `Grape_1`,
  Kiwi: `Kiwi_1`,
  Cherry: `Cherry_1`
};

class App extends Component {
  order;

  state = { items: ["Apple", "Grape", "Cherry", "Orange", "Banana"] };

  sort() {
    this.order = this.order === "ASC" ? "DESC" : "ASC";
    const items = [...this.state.items].sort((a, b) =>
      this.order === "ASC" ? a.localeCompare(b) : b.localeCompare(a)
    );
    this.setState({ items });
  }

  render(_, { items }) {
    return (
      <Fragment>
        <div>
          {items.map((item) => (
            <div key={keys[item]}>{item}</div>
          ))}
        </div>
        <div>
          <button onClick={() => this.sort()}>Sort</button>
        </div>
      </Fragment>
    );
  }
}

if (typeof window !== "undefined") {
  render(<App />, document.getElementById("root"));
}

```

- [x]  1. Click to sort button to sort items in alphabetical order
- [ ] 2. Click to sort button again to sort items in reverse alphabetical order

Additional information:

- [x] reverse alphabetical order in first click works fine
- [ ] alphabetical order in second click works fine
- [ ] reverse alphabetical order in third click does not work

**Expected behavior**
Items should be displayed in correct order

JoviDeCroock:
When we perform `matchingIndex === i` we risk skipping re-orders as we have a lot of cases where we reset skew to 0 for reverse re-ording. in inverse re-ordering we increase the skew by 1 hence we don't run into this issue, but we can't just decrement the skew to a negative number hence this safety 😅 this PR removes the `hasMatchingIndex` and only checks if we are currently on or next to the skew and appropriately invokes our move logic there.

Fixes #4087

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -41,12 +41,12 @@ jobs:
           artifact: npm-package
           workflow: ci.yml
           required: false
-      - run: mv preact.tgz preact-main.tgz
+      - run: mv preact.tgz preact-master.tgz
       - name: Upload base preact package
         uses: actions/upload-artifact@v3
         with:
           name: bench-environment
-          path: preact-main.tgz
+          path: preact-master.tgz
 
   bench_todo:
     name: Bench todo
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -41,12 +41,12 @@ jobs:
           artifact: npm-package
           workflow: ci.yml
           required: false
-      - run: mv preact.tgz preact-master.tgz
+      - run: mv preact.tgz preact-main.tgz
       - name: Upload base preact package
         uses: actions/upload-artifact@v3
         with:
           name: bench-environment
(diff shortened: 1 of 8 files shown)


marvinhagemeister (approved):
Great work  👍

PR #4090, opened 2023-08-03
Eagerly unmount placeholders

### LINKED ISSUE #4074: iframe is reloaded, when it's a children props in WrapperComponent, and some siblings before children is removed

**Describe the bug**
Nearly the same as https://github.com/preactjs/preact/issues/2350

But this issue only happends in v10.16.0, works well in 10.15.1

The difference from the old issue is that, I pass this `iframe` element with children props into another `WrapperComponent`
And if there is any node removed before the `children`, the `iframe`, which is inside the `children` will get reloaded 

**To Reproduce**

https://codesandbox.io/s/festive-dust-qyr2lc?file=/src/index.js

1. Go to 'https://codesandbox.io/s/festive-dust-qyr2lc?file=/src/index.js'
2. Click on Button 'Focus on me'
3. The iframe is getting reloaded when release the button

**Expected behavior**
The iframe shouldn't get reloaded

andrewiggins:
When we detect that an unkeyed vnode is changing from "something" to a non-renderable (e.g. null or boolean), eagerly unmount that VNode so we can skip over it's DOM elements when diffing the next VNodes. This change prevents us from moving elements around other DOM nodes that are going to be unmounted.

Fixes #4074

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -107,6 +107,15 @@ export function diffChildren(
 		// Terser removes the `continue` here and wraps the loop body
 		// in a `if (childVNode) { ... } condition
 		if (childVNode == null) {
+			oldVNode = oldChildren[i];
+			if (oldVNode && oldVNode.key == null && oldVNode._dom) {
+				if (oldVNode._dom == oldDom) {
+					oldDom = oldDom.nextSibling;
+				}
+
+				unmount(oldVNode, oldVNode, false);
+			}
+
 			continue;
 		}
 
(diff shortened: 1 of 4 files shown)


src/diff/children.js
```diff
@@ -107,6 +107,15 @@ export function diffChildren(
...
 		if (childVNode == null) {
+			oldVNode = oldChildren[i];
+			if (oldVNode && oldVNode.key == null && oldVNode._dom) {
+				if (oldVNode._dom == oldDom) {
+					oldDom = oldDom.nextSibling;
+				}
+
+				unmount(oldVNode, oldVNode, false);
+			}
+
```
andrewiggins, line 118: I need to update this logic to handle fragments with multiple children. `oldDom.nextSibling` may also be unmounted. Gotta come up with a test case first.

andrewiggins: Done

JoviDeCroock (approved):

marvinhagemeister (approved):
I never would've thought of looking at the placeholders specifically. This is great!!

PR #4096, opened 2023-08-14
fix: Don't rename `gotpointercapture` and `lostpointercapture` events

### LINKED ISSUE #4095: Polyfilling pointer-events
First, would like to say thanks for all the work you people put into this.

Now, we're developing a feature relying on pointer-events so that we can handle both mouse and touch events easily.
On old Safari, a [polyfill](https://github.com/jquery-archive/PEP) is required for those events to function.
The polyfill seems to work on many demos but I couldn't get it to work in my project.
I tested multiple directions and eventually got to the conclusion that it's probably Preact that is in my way.
Here are two sandboxes:
[This](https://q38ygd.csb.app/) one is using React and works on Safari 12 with the polyfill
[This](https://f4zl39.csb.app/) one has the exact same component but using Preact, not responsive on Safari 12

Can anyone help explain the issue? Is there anything we can do about it?
Thanks!

rschristian:
Fixes part of #4095 

`onGotPointerCapture` and `onLostPointerCapture` have their names incorrectly altered to `GotPointer` and `LostPointer` respectively with the current state of things. They'll need to be skipped over in some fashion.

I'm a code golf scrub, might be (probably is) a better way to correct these.

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -83,7 +83,11 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
 	}
 	// Benchmark for comparison: https://esbench.com/bench/574c954bdb965b9a00965ac6
 	else if (name[0] === 'o' && name[1] === 'n') {
-		useCapture = name !== (name = name.replace(/Capture$/, ''));
+		if (/PointerCapture$/.test(name)) {
+			useCapture = false;
+		} else {
+			useCapture = name !== (name = name.replace(/Capture$/, ''));
+		}
 
 		// Infer correct casing for DOM built-in events:
 		if (name.toLowerCase() in dom) name = name.toLowerCase().slice(2);
--- a/test/browser/events.test.js
+++ b/test/browser/events.test.js
@@ -95,35 +95,6 @@ describe('event handling', () => {
 		expect(click).to.have.been.calledOnce.and.calledWith(1);
 	});
 
-	// Uniquely named in that the base event names end with 'Capture'
-	it('should support got/lost pointer capture', () => {
-		let ongotpointercapture = sinon.spy(),
-			onlostpointercapture = sinon.spy();
-
-		let firePointerEvent = (on, type) =>
-			on.dispatchEvent(new PointerEvent(type, { pointerId: 1 }));
-
-		render(
-			<div
-				onPointerDown={e => e.target.setPointerCapture(e.pointerId)}
-				onGotPointerCapture={ongotpointercapture}
-				onLostPointerCapture={onlostpointercapture}
-			/>,
-			scratch
-		);
-
-		firePointerEvent(scratch.childNodes[0], 'pointerdown');
-		expect(ongotpointercapture).to.have.been.calledOnce;
-		expect(onlostpointercapture).to.not.have.been.called;
-
-		ongotpointercapture.resetHistory();
-		onlostpointercapture.resetHistory();
-
-		firePointerEvent(scratch.childNodes[0], 'pointerup');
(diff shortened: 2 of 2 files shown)


marvinhagemeister (commented):
LGTM, small code golf nit. Let's add a test and then it's good to go 👍

src/diff/props.js
```diff
@@ -83,7 +83,11 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
 	}
 	// Benchmark for comparison: https://esbench.com/bench/574c954bdb965b9a00965ac6
 	else if (name[0] === 'o' && name[1] === 'n') {
-		useCapture = name !== (name = name.replace(/Capture$/, ''));
+		if (/PointerCapture$/.test(name)) {
+			useCapture = false;
+		} else {
+			useCapture = name !== (name = name.replace(/Capture$/, ''));
+		}
```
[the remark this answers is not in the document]
rschristian: Played with this first actually, but won't it cause issues in IE?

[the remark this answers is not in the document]
marvinhagemeister: Oops, you're right. Didn't realize that this was a new feature 👀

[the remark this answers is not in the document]
rschristian: Well, it's a bit sad that a 6-year old feature is "new" 😅

I missed that Safari gained support for it earlier this year and so was pretty excited to see we now have full ES2017 support in the evergreen browsers. Now just to get rid of that pesky IE...

marvinhagemeister (approved):

test/browser/events.test.js
```diff
@@ -95,6 +95,32 @@ describe('event handling', () => {
...
 	});
 
+	// Uniquely named in that the base event names end with 'Capture'
+	it('should support got/lost pointer capture', () => {
+		let ongotpointercapture = sinon.spy(),
+			onlostpointercapture = sinon.spy();
+
+		render(
+			<div
+				onPointerDown={e => e.target.setPointerCapture(e.pointerId)}
```
rschristian, line 105: Struggling a bit with this test. For some reason, this always results in the following error:

```
Error: Uncaught NotFoundError: Failed to execute 'setPointerCapture' on 'Element': No active pointer with the given id is found. (test/browser/events.test.js:14011)
        at fireEvent (test/browser/events.test.js:16:6)
        at Context.<anonymous> (test/browser/events.test.js:112:3)
```

I don't quite understand how the pointer is inactive? From reading the spec, it'll implicitly become inactive when `pointerup` or `pointercancel` fire, but neither should be happening here (at least, not before it has a chance to set).

rschristian: Oh, `PointerEvent`s might be their own thing? Funny how you see parts of the docs only after pushing... Will need to try that.

rschristian: Hm, apparently you need to hard-code the pointer ID when using `dispatchEvent`? Weird. Doesn't seem to trigger `got/lostPointerCapture` though.

rschristian: Gave up on actually testing these events, browser behavior is too odd when using `dispatchEvent`. Things don't fire correctly.

Ensuring that `addEventListener` is called with the correct names should be enough anyhow, that's where we were going wrong.

JoviDeCroock (approved):

marvinhagemeister (approved):

developit:
Code golf suggestion:

```js
useCapture = name !== (name = name.replace(/(PointerCapture)$|Capture$/, '$1'));
```

Issue #4099, opened 2023-08-15
[compat types] - issue with the element's type like button, that gets any instead of ButtonHTMLAttributes


samsam-ahmadi:
- [x] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
The issue is with some conflicts type of compat with React types, when other libraries trying to using the react types inner like: styled-components. 

For some elements react has special type like this line: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f701ddcb5bec7d89238b8f1b9d362a20ff9c8e94/types/react/v17/index.d.ts#L2048

So as you can see there is an error for h1 but for button there isn't and it takes `any`

![Screenshot 2023-08-15 at 15 14 27](https://github.com/preactjs/preact/assets/18388480/71f2762b-bfab-4ef8-8e82-794d45a416e2)

So i checked the styled-component that give the button this type. 

`button: Styled<"web", "button", import("react").DetailedHTMLProps<import("react").ButtonHTMLAttributes<HTMLButtonElement>, HTMLButtonElement>, import("../types").BaseObject>;`

But after using aliases of compat, it will be something like this: 

`button: Styled<"web", "button", **any**, import("../types").BaseObject>;`

Because styled component trying to access to ButtonHTMLAttributes, but this type doesn't exist in the compact so it takes any. 

**To Reproduce**

If possible, please provide a CodeSandbox/Codepen that demonstrates the issue. You can use the following template: https://codesandbox.io/s/preact-x-preact-cli-3-starter-vj285y2rn3

I can't do it, Because for displaying the errors needs more typescript configuration but i explain it here: 

1. create a vite + typescript + preact+ styled-components
2. npm i -D vite-plugin-checker 
3. Add this config to vite plugins. => `checker({ typescript: true })`
4. Also this one
5. `    preact({
        babel: {
          plugins: [
            [
              "babel-plugin-styled-components",
              {
                pure: true,
                displayName: false,
                fileName: false,
              },
            ],
          ],
        },
      }),`

Then add ThemeProvider and styled.d.ts like this codesandbox. 
https://codesandbox.io/s/interesting-hooks-5m8kvr?file=/src/App.tsx

After that If you create those components in the screenshot you can see h1 has correct type but button doesn't have. 
 

**Expected behavior**
What should have happened when following the steps above?

victor-williams:
Has this issue been resolved?

samsam-ahmadi:
[...]

PR #4100, opened 2023-08-15
type[compat]: add html attributes types #4099

samsam-ahmadi:
This PR is for: #4099


compat/src/index.d.ts
```diff
@@ -68,12 +68,858 @@ declare namespace React {
...
+
+	// this list is "complete" in that it contains every SVG attribute
+	// that React supports, but the types can be improved.
+	// Full list here: https://facebook.github.io/react/docs/dom-elements.html
+	//
+	// The three broad type categories are (in order of restrictiveness):
+	//   - "number | string"
+	//   - "string"
+	//   - union of string literals
+	interface SVGAttributes<T> extends AriaAttributes, DOMAttributes<T> {
```
rschristian, line 559: Any reason why we're not simply reexporting Preact's types here?

https://github.com/preactjs/preact/blob/7cd48335e709f0b32d83070ae320f2010943c310/src/jsx.d.ts#L79

compat/src/index.d.ts
```diff
@@ -68,12 +68,858 @@ declare namespace React {
...
+		nodeintegration?: boolean | undefined;
+		partition?: string | undefined;
+		plugins?: boolean | undefined;
+		preload?: string | undefined;
+		src?: string | undefined;
+		useragent?: string | undefined;
+		webpreferences?: string | undefined;
+	}
+
+	interface ReactSVG {
```
rschristian, line 865: Same here, looks like we're duplicating `InstrinsicElements` definitions?

samsam-ahmadi: @rschristian Yes we can do this.
I just copied them from react type.

compat/src/index.d.ts
rschristian: We want to avoid duplication as much as possible, copying huge segments from React's types is probably undesirable

samsam-ahmadi: Thanks for reviewing and comments. 
I updated the PR and removed the duplicated code and checked others.

rschristian: Sorry, I wasn't saying that they should be removed outright, but that we should be re-exporting core's definitions where at all possible.

Most of these are on the JSX namespace in core and therefore do need to be reexported if users want to import the type directly.

samsam-ahmadi: Hmm, at first i thought this types are from `IntrinsicElements`. but they not. These types generated by react. 

For example we have `HTMLButtonElement` that comes from mdn, but here we have a new interface of `ButtonHTMLAttributes`

So i don't know is it possible to generate all them by a simple code?

src/jsx.d.ts
```diff
@@ -2021,7 +2021,69 @@ export namespace JSXInternal {
 			| SignalLike<number | string | undefined>;
 	}
 
-	export interface IntrinsicElements {
+	export interface SVGElements {
```
rschristian, line 2024:
```suggestion
	export interface SVGInstrinsicElements {
```

or maybe `IntrinsicSVGElements`

test-utils/src/index.js
```diff
@@ -35,14 +35,17 @@ export function act(cb) {
...
-				})
+				return result.then(
+					() => {
+						--actDepth;
+					},
+					e => {
+						--actDepth;
+						throw e;
+					}
+				);
```
rschristian, line 46: Should be reverted, keep the diff focused

test-utils/test/shared/act.test.js
```diff
@@ -389,7 +389,7 @@ describe('act', () => {
 			act(() => {
 				tryRenderBroken();
 			});
-		}
```
rschristian, line 392: Same with the changes here

rschristian (approved):
Looks very useful, thanks for doing this! We really appreciate it.

Might be nice to move the majority of these into core in the future, as I can see even non-compat users maybe wanting these, but that's an idea for another day.

samsam-ahmadi:
Thanks @rschristian, 
I updated the branch and it's ready to merge.

samsam-ahmadi:
Hi @rschristian,  Any update for this one? How we can get the approval to can merge it?

baltarifcan:
Will this be merged anytime soon?

rschristian (approved):
I'd like to apologize for this PR, evidently I missed a couple of notifications and it then fell off my radar entirely -- I'm really sorry.

We certainly appreciate any & all contributions, and the time contributors invest into them.

Thanks again.

PR #4101, opened 2023-08-16
fix: Ensure all WAI-ARIA 1.2 roles are supported in our types

### LINKED ISSUE #4098: Update AriaRole type to WAI-ARIA 1.2
Hello,

I have a problem with [AriaRole](https://github.com/preactjs/preact/blob/main/src/jsx.d.ts#L1591) type.
It includes a list of all [WAI-ARIA 1.1 role definitions](https://www.w3.org/TR/wai-aria-1.1/#role_definitions) which makes it impossible to use a `meter` role for the HTML element without workarounds like type casting.

Is it possible to update `AriaRole` and use a list of roles from [WAI-ARIA 1.2](https://www.w3.org/TR/wai-aria-1.2/#role_definitions)?

rschristian:
Closes #4098

I pulled the list directly from the spec docs the issue opener kindly provided.

Here's a little script one could use in their browser console to pull the whole list as a type def from the spec docs:

```js
const rolesParent = document.querySelector('#index_role');
const roles = rolesParent.querySelectorAll('dt > a > code');
`type AriaRole = ${[...roles].map(role => `'${role.innerText}'`).join(' | ')} | 'none presentation'`;
```


JoviDeCroock (approved):

src/jsx.d.ts
```diff
@@ -1630,34 +1642,47 @@ export namespace JSXInternal {
...
+		| 'time'
 		| 'timer'
 		| 'toolbar'
 		| 'tooltip'
 		| 'tree'
 		| 'treegrid'
 		| 'treeitem'
+		| 'widget'
+		| 'window'
 		| 'none presentation';
```
rschristian, line 1686: This is a bit of a weird one, do we need to support it?

[`none` is meant to be an alias of `presentation`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/none_role), so I don't understand why someone would set both intentionally.

rschristian: Looks intentional: https://github.com/preactjs/preact/pull/3910/commits/8223d002e3784c2a2bcfe725d8ccd2ce17aebafb

Good enough for me!

rschristian: Turns out it's a fallback: https://www.w3.org/TR/wai-aria-1.1/#presentation-0

> Note regarding the ARIA 1.1 none role.[§](https://www.w3.org/TR/wai-aria-1.1/#role-presentation-note-none)

> In ARIA 1.1, the working group introduced none as a synonym to the presentation role, due to author confusion surrounding the intended meaning of the word "presentation" or "presentational." Many individuals erroneously consider role="presentation" to be synonymous with aria-hidden="true", and we believe role="none" conveys the actual meaning more unambiguously.

> **Until implementations include sufficient support for role="none", web authors are advised to use the presentation role alone role="presentation" or redundantly as a fallback to the none role role="none presentation".**

PR #4105, opened 2023-08-18
Fix double unmount

### LINKED ISSUE #4104: Unbalanced number of mount/unmount calls for a child node when the node is added and removed

**Describe the bug**
If a child node has a ref callback the callback is called once for mount and twice for unmount operation. 

The issue was introduced in 10.17.0 as a result of the following fix: https://github.com/preactjs/preact/issues/4074

**To Reproduce**
https://codesandbox.io/s/vigilant-chebyshev-yvyrtv?file=/src/index.js
I cloned the code from the original issue #4074 and added a ref callback to the div element.

1. Go to https://codesandbox.io/s/vigilant-chebyshev-yvyrtv?file=/src/index.js
2. Click on Button 'Focus on me'
3. Check the console panel. See that ref callback is called once when <div> element is added to the DOM and twice when it is removed from the DOM.

**Expected behavior**
The number of mount and unmount calls is the same.
You can switch to version 10.16.0 to verify that mount/unmount calls were balanced.

developit:
This should fix #4104.

I believe the issue is that `unmount()` is triggered for the VNode at `oldChildren[i]`, but that VNode remains in `oldChildren`. When we hit the end of `diffChildren()`, a loop invokes `unmount()` on every non-null vnode in `oldChildren`, which still includes the eagerly-unmounted VNode.

I didn't have a chance to add a test as I'm doing this from GH web, if someone gets around to it before me we can likely just use the example from #4104.

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -115,6 +115,7 @@ export function diffChildren(
 				}
 
 				unmount(oldVNode, oldVNode, false);
+				oldChildren[i] = null;
 			}
 
 			continue;
--- a/test/browser/placeholders.test.js
+++ b/test/browser/placeholders.test.js
@@ -417,4 +417,61 @@ describe('null placeholders', () => {
 			'<div>Test4.remove()'
 		]);
 	});
+
+	it('should only call unmount once when removing placeholders (#4104)', () => {
+		/** @type {(state: { value: boolean }) => void} */
+		let setState;
+		const Iframe = () => {
+			// Using a div here to make debugging tests in devtools a little less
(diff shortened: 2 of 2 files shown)


andrewiggins (approved):
Added a test.

Something fun I noticed: You can't observe this bug using lifecycles cuz on the first unmount, we clear `_component` which prevents re-entry into the part of `unmount` that invokes `componentWillUnmount`.

marvinhagemeister (approved):
Nice catch 👍

JoviDeCroock (approved):

PR #4110, opened 2023-08-20
Fix #4109

### LINKED ISSUE #4109: Replace for-of loop with standard for loop
Just noticed that we have a `for-of` loop in Preact debug which causes some array iterator polyfills to be included. Since it's only in `preact/debug` and not the core library itself it's kinda ok, but since we already know that this is an array we can make `preact/debug` a little smaller.

So let's rewrite that `for-of` loop to a standard `for` loop.

https://github.com/preactjs/preact/blob/0afad8cdcc106a017bcb3f7243661cd40ef0f5ab/debug/src/debug.js#L365

Popovkov57:
Hi 

I replaced for-of loop with standard for loop in debug.js to fixe the issue 4109.

Best regards.

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -362,7 +362,8 @@ export function initDebug() {
 				for (let i = 0; i < hooks.length; i += 1) {
 					const hook = hooks[i];
 					if (hook._args) {
-						for (const arg of hook._args) {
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Awesome, thank you for making a PR for this 👍

PR #4112, opened 2023-08-23
fix: TS component type

rschristian:
https://preact.slack.com/archives/C3NMBSJKH/p1692741376856869

Trying to match [React's types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/05094b6df4fb6f4404333725f13d63dd8aabeb34/types/react/index.d.ts#L563) though something isn't quite right yet (as the failing tests show). Need to play with this some more.


rschristian:
Closing for tracking in #4113,

PR #4121, opened 2023-08-31
fix: avoid registering non-function event-handlers on custom-elements

### LINKED ISSUE #4085: Uncaught Error: Component's "only-when-overflow" property should be a function, but got [boolean] instead

**Describe the bug**

I'm using a custom web component called `my-tooltip` and there's a boolean attribute called `only-when-overflow`.When using the `only-when-overflow` attribute i'm getting the below error.

```
Uncaught Error: Component's "only-when-overflow" property should be a function, but got [boolean] instead
<my-tooltip only-when-overflow="true" for="a1a1">..</z-tooltip>

```

**To Reproduce**

We usually use a attribute which starts with `on` to handle events eg: `onClick, onKeyPress` which requires function. Here I'm using a boolean attribute which starts with `on` and the error thrown

JoviDeCroock:
Fixes #4085

Prevent us from registering functions on custom-elements when the property starts with `on`

### DIFF
--- a/test/browser/render.test.js
+++ b/test/browser/render.test.js
@@ -130,6 +130,24 @@ describe('render()', () => {
 		expect(scratch.firstChild).to.have.property('nodeName', 'X-BAR');
 	});
 
+	it('should not register non-function event handlers', () => {
+		const proto = Element.prototype;
+		proto.addEventListener.resetHistory();
+		render(
+			<x-bar
+				only-overflow={true}
+				onChange={() => {
+					console.log('Change');
+				}}
+			/>,
+			scratch
+		);
+
+		expect(proto.addEventListener).to.have.been.calledOnce;
+		expect(scratch.childNodes).to.have.length(1);
+		expect(scratch.firstChild).to.have.property('nodeName', 'X-BAR');
+	});
+
 	it('should support the form attribute', () => {
 		render(
 			<div>


JoviDeCroock:
The failure here is interesting, basically we have a fallthrough case now that we shield for the value being a function in either the old or new case. When we are hydrating and have the test where we perform `onClick={null}` then we have no oldValue (`undefined`) and no newValue (`null`) so we fall to the `removeAttribute` case

developit:
I worry that this will prevent us from supporting [EventListener Objects](https://www.w3.org/TR/DOM-Level-2-Events/events.html#Events-EventListener) in the future, in addition to the issue you noted with set/unset.

An idea I had while looking at this:  what if we hoisted the `in` check up above `on*` handler detection?

```
if (name === 'style') {
  // set/diff style
  return;
}
if (
  !isSvg &&
  name !== 'width' &&
  name !== 'height' &&
  name !== 'href' &&
  name !== 'list' &&
  name !== 'form' &&
  name !== 'tabIndex' &&
  name !== 'download' &&
  name !== 'rowSpan' &&
  name !== 'colSpan' &&
  name in dom
) {
  try {
    dom[name] = value == null ? '' : value;
    return;
  } catch (e) {
    // property assignment failed, fall back to attribute
  }
}
if (name[0] === 'o' && name[1] === 'n' && name.indexOf('-') === -1) {
  // if we got here, we know `onxyz` is not a property defined on `dom` and there was no dash
  // set/diff event listener
  return;
}

if (typeof value === 'function') return;

// fall through to setting as attribute
if (value != null && (value !== false || name[4] === '-')) {
  dom.setAttribute(name, value);
} else {
  dom.removeAttribute(name);
}
```

marvinhagemeister (approved):

JoviDeCroock:
@developit the issue with that approach is that you'll fall into the `on` case for `only-overflow` for instance. Checking whether we are dealing with a custom-element on that branch seems wrong as we can't support event-handlers on custom-elements like that. We could do a check like the event not containing a dash but if they then leverage `onlyoverflow` it would also fall in this bucket 😅

PR #4125, opened 2023-09-01
fix: ordering issue with Component re-render

### LINKED ISSUE #4123: Ordering of children in a Fragment changes in re-renders
**Describe the bug**

After updating from 10.16.0 to 10.17.1, when re-rendering, the ordering of children inside a Fragment seems to change.

This sounds _suspiciously_ similar to #4122 - to the point I was just going to add a little "me too" to that issue... But that also seems to happen on older versions of preact & this doesn't rely on preact-compat at all, so I'm not sure if it's unrelated, or has the same underlying root cause.

**To Reproduce**

Here's a (rather contrived) example with a Fragment (After ~1 second a hook/reducer on a timer will cause a re-render and the content will sit above the header).

https://codesandbox.io/s/dreamy-benji-cm79jk?file=/src/containers/job.js

Either downgrading preact, or Changing the ```Fragment``` to just a ```div``` as a workaround works flawlessly.

JoviDeCroock:
When we have a Fragment as the top-level return from a render function we assign the underlying dom-children with the rendering function as their `_parent`. This however means that when we run `getDomSibling` like we do in our `null` placeholder unmount case https://github.com/preactjs/preact/blob/main/src/diff/children.js#L114 that we will move onto the wrapping root-fragment here which can refer to a stale dom-node, in this case we should leverage the potentially set `_nextDom` property to ensure we transfer `oldDom` back into a valid state.

fixes #4123

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -111,6 +111,7 @@ export function diffChildren(
 			oldVNode = oldChildren[i];
 			if (oldVNode && oldVNode.key == null && oldVNode._dom) {
 				if (oldVNode._dom == oldDom) {
+					oldVNode._parent = oldParentVNode;
 					oldDom = getDomSibling(oldVNode);
 				}
 
@@ -200,6 +201,7 @@ export function diffChildren(
 				(matchingIndex !== skewedIndex ||
 					oldVNode._children === childVNode._children)
 			) {
+				console.log('reorder');
 				oldDom = reorderChildren(childVNode, oldDom, parentDom);
 			} else if (
 				typeof childVNode.type != 'function' &&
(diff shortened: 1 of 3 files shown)


src/diff/children.js
```diff
@@ -111,6 +111,7 @@ export function diffChildren(
 			oldVNode = oldChildren[i];
 			if (oldVNode && oldVNode.key == null && oldVNode._dom) {
 				if (oldVNode._dom == oldDom) {
+					oldVNode._parent = oldParentVNode;
```
JoviDeCroock, line 114: I am currently trying to figure out why this part matters but it looks like `oldVNode._parent` refers to the new version of the parent-vnode here

marvinhagemeister (approved):
Great catch!!

gpoitch:
Hoping for a release with this fix when you guys get a chance, thanks! 🙏

PR #4126, opened 2023-09-03
fix: add timer to event handler so we can check whether it was attached during the current propagation

### LINKED ISSUE #3927: Event fired twice [bug only appears in latest version v10.13]

The issue appears in latest v10.13 only (v10.12.1 and before seem to be working fine).

**Describe the bug**

I have a simple component with two states. When no option has been selected, the component should show a list of options. Clicking on one option should switch the component in the second state that only shows the selected option.

When preact v10.13 is used, clicking on an option seems to fire a click even on both alternative states of the component, hence the component does not work as expected.

**To Reproduce**

Here is a codesandbox link with preact v10.13 that exhibits the bug: https://codesandbox.io/s/condescending-rui-5own5r

Here is the code:

```js
function Combobox({ options }) {
  const [option, setOption] = useState(undefined);
  if (option !== undefined) {
    return (
      <div style={{ border: "1px solid #ccc", margin: "10px", padding: "10px", cursor: "pointer" }}
        onClick={() => {
          console.log("Clear option");
          setOption(undefined);
        }}
      >
        {option}
      </div>
    );
  } else {
    return (
      <div style={{ border: "1px solid #ccc", margin: "10px", cursor: "pointer" }}>
        {options.map((option) => {
          return (
            <div
              style={{ padding: "10px" }}
              onClick={() => {
                console.log("Set option", option);
                setOption(option);
              }}
            >
              {option}
            </div>
          );
        })}
      </div>
    );
  }
}

const root = document.getElementById("root") ?? document.getElementById("app");

render(<Combobox options={["option 1", "option 2"]} />, root);
```

1. Clicking on "option1" or "option2" should normally change the component, but nothing happens
2. You can check in the console that the click event is being fired twice on both alternative states of the component, toggling it back to the initial state.

As mentioned earlier, when selecting preact v10.12.1 or below, the component works as expected.

Here is a codesandbox link using preact v10.12.1 that works fine: https://codesandbox.io/s/confident-keldysh-r4f5ow

Bug is also present when copy/pasting the code on the preact repl (https://preactjs.com/repl)

JoviDeCroock:
Browsers leave microticks in-between propagating an event up, this can be seen in research performed [here](https://www.jovidecroock.com/blog/browser-timings). Our first attempt at fixing this was using `setTimeout` which caused other issues to occur. This fix is a bit different rather than changing our scheduling we place a timestamp on the event-handler so we can check whether the event is appropriate to fire.

This bug can occur when we switch in a dom-node i.e. when we go from

```jsx
<div><div onClick

to

<div onClick><div
```

when the inner div changes the dom-structure then we would perform the patch before the event could bubble up.

> This is entirely inspired by [Vue.js](https://github.com/vuejs/core/blob/main/packages/runtime-dom/src/modules/events.ts#L90-L101)

fixes #3927

You can see this approach working in [the following sandbox](https://codesandbox.io/s/amazing-rhodes-xpj62v?file=/src/preact.mjs) - due to this being related to browser timings I could not create a test

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -83,7 +83,8 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
 	}
 	// Benchmark for comparison: https://esbench.com/bench/574c954bdb965b9a00965ac6
 	else if (name[0] === 'o' && name[1] === 'n') {
-		useCapture = name !== (name = name.replace(/(PointerCapture)$|Capture$/, '$1'));
+		useCapture =
+			name !== (name = name.replace(/(PointerCapture)$|Capture$/, '$1'));
 
 		// Infer correct casing for DOM built-in events:
 		if (name.toLowerCase() in dom) name = name.toLowerCase().slice(2);
@@ -94,8 +95,11 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
 
 		if (value) {
 			if (!oldValue) {
+				value._attached = Date.now();
 				const handler = useCapture ? eventProxyCapture : eventProxy;
 				dom.addEventListener(name, handler, useCapture);
+			} else {
+				value._attached = oldValue._attached;
 			}
 		} else {
 			const handler = useCapture ? eventProxyCapture : eventProxy;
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
As discussed in slack I think this is a great fix. Our benchmark setup doesn't account for ticks iirc so it always prefers sync code.

PR #4129, opened 2023-09-09
chore: fix some wrong annotation types

ulivz:
Fixed some wrong annotation types


rschristian:
I believe the types are correct as-is, actually.

Did you run into an issue where something that should fall under `ComponentChild` wasn't working correctly?

rschristian:
Going to close this out, if you do have an issue w/ `ComponentChild` let us know and we can take a look.

PR #4135, opened 2023-09-19
chore: Update link & verbiage for reproduction in issue template

rschristian:
This is the result of a plain `create-preact` project initialization.

I figured the fewer the files the better, as it would (hopefully) result in cleaner repros, so I cut out the routing. I do have a [separate template](https://stackblitz.com/edit/create-preact-starter-routing) if we want to keep that around though, I'm not opinionated. Same thing goes for ESLint, figured the faster startup time might be preferable but happy to add it in too if we want it.


marvinhagemeister (approved):
Nice!

JoviDeCroock (approved):

Issue #4136, opened 2023-09-19
It is not possible to remove the `role` attribute which results in invalid HTML


SantosGuillamot:
- [x] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
It seems impossible to toggle between having a role attribute (`dialog` for example) and removing the attribute. For example, I want to have a `role="dialog"` when a menu is open, but remove the role attribute when it is closed. It currently returns an empty value. This affects accessibility because having an empty role is not valid HTML. You can check that using an [HTML validator](https://validator.w3.org/nu/#textarea):

![Screenshot 2023-09-19 at 16 20 43](https://github.com/preactjs/preact/assets/34552881/14e1551f-2ac2-4e0a-96a6-6e4e0edb0188)

This seems to be the behavior in Preact with the following options:

* Using `null`: It keeps the `role` attribute in the element.
* Using an empty string: It keeps the `role` attribute in the element.
* Using false: It sets `role="false"`.

I also tried using another role value as `none`, but it changes the accessibility tree.

I made a quick video using [this CodeSandbox](https://codesandbox.io/s/lucid-goldberg-fj6s5c?file=/src/index.js) explaining that better:

https://github.com/preactjs/preact/assets/34552881/62827984-b687-4deb-be2d-0850c3c9592e

I've tested it in React, and it seems possible to remove the `role` attribute by using `null`. I used [this other CodeSandbox](https://codesandbox.io/s/festive-https-m2vdpv?file=/src/index.js) as shown in this other video:

https://github.com/preactjs/preact/assets/34552881/0dc54edc-29e3-4612-a88d-14caa6b0b6cb

**To Reproduce**

_**Preact behavior**_

Using [this CodeSandbox](https://codesandbox.io/s/lucid-goldberg-fj6s5c?file=/src/index.js), toggle between the different options to see how the `role` attribute of the `div` containing the `<pre>` element changes.

_**React behavior**_

Using [this CodeSandbox](https://codesandbox.io/s/festive-https-m2vdpv?file=/src/index.js), toggle between the different options to see how the `role` attribute of the `div` containing the `<pre>` element changes.

**Expected behavior**
It should be possible to remove the `role` attribute as it happens in React.

DAreRodz:
In the browsers where `Element.role` exists, this condition inside `setProperty` becomes true:

https://github.com/preactjs/preact/blob/7dc66bc8835f36d4e4c9f878ae0fd3b39bbda8ce/src/diff/props.js#L126

[...]

PR #4137, opened 2023-09-20
fix: remove `role` attribute when it's `null` or `undefined`

DAreRodz:
Fixes https://github.com/preactjs/preact/issues/4136.

Most browsers now implement the attributes defined in the [ARIAMixin interface](https://www.w3.org/TR/wai-aria-1.2/#ARIAMixin), which includes [`Element.role`](https://caniuse.com/mdn-api_element_role). In such a case, the current code assigns an empty string to the `role` property when it's `null` or `undefined` instead of removing it.

This PR prevents that so the `role` prop can be removed as normal.


JoviDeCroock (approved):
Awesome, thank you!

Issue #3149, opened 2021-05-12
Cannot read property 'insertBefore' of undefined when using `Suspense` and `lazy`


pt-hoang:
- [x] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
`Cannot read property 'insertBefore' of undefined` when using `Suspense` and `lazy`. Reason being `originalParent` is undefined in `Suspense`'s `removeOriginal` function

Works if checks `originalParent` before calling `insertBefore` on it
![image](https://user-images.githubusercontent.com/84064700/117931561-e748a800-b331-11eb-9783-0872dfa215ef.png)

**To Reproduce**

If possible, please provide a CodeSandbox/Codepen that demonstrates the issue. You can use the following template: https://codesandbox.io/s/immutable-grass-ziq4b?file=/src/index.js

In codesandbox

**Expected behavior**
No error

developit:
I think someone else ran into this as well. Not 100% sure what the root cause is.

Nedomas:
I have the same problem when state above suspense changes. Did you find the solution @pt-hoang ? I understand this is preact only bug?

JoviDeCroock:
Seems like a duplicate of https://github.com/preactjs/preact/issues/3105 let's track there

Issue #3105, opened 2021-04-06
Suspense: DOMException: Failed to execute 'insertBefore' on 'Node'


LXSMNSYC:
- [ ] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
If there are multiple components that suspended at the same time and tried to resolve at the same time, there's a chance that only one of the components will successfully re-render while the rest stays in their fallback state. I'm not sure the specificity but I think one of the reasons is that Preact fails to match the component vdom to the actual dom node.

Once it fails to resolve, it throws this error:
![image](https://user-images.githubusercontent.com/4783372/113679194-172cdc00-96f2-11eb-9e28-b06a5a8e615f.png)

This may also happen if one of the components, that isn't suspending, updates during the resolution timeframe.

A softfix for this is to wrap Suspending components with a container like a simple `<div>`.

**To Reproduce**

1. Go to https://codesandbox.io/s/github/LXSMNSYC/preact-suspense-poc
2. Click on Reload button rapidly.
3. Check Console for error message.

**Expected behavior**

Should've been able to continuously retain the same behavior prior to the error.

benjamincombes:
I have the same problem, just tried Preact for the first time, and I get this error anytime I start my app... Unfortunately, the softfix does not work for me, so I just can't use Preact until it's fixed.

JoviDeCroock:
This issue seems fixed with Preact 10.8.2 https://codesandbox.io/s/misty-cloud-u5et4t, closing for now but feel free to reopen if the issue persists for you.

Brisstor:
Hey there! The issue appeared with Preact 10.11.2

peacepostman:
Hello ! I'm still having this problem with `10.11.3`

alfredsgenkins:
As a possible fix to already compiled bundle, the following can be used:
```js
webpackConfig.module.rules.push({
    test: new RegExp('preact/compat'),
    loader: require.resolve('./preact-patch'),
});
```
And the loader contents:
```
module.exports = function patchPreact(rawSource) {
    const originalCode = `t.__e&&r.insertBefore(t.__e,t.__d)`;
    const newCode = `t.__e&&r&&r.insertBefore(t.__e,t.__d)`;
    return rawSource.replace(originalCode, newCode);
};
```

But indeed, why is not the `originalParent` checked? https://github.com/preactjs/preact/issues/3149

BennyAlex:
[...]

PR #4141, opened 2023-09-24
suspense fix: Cannot read property 'insertBefore' of undefined

zakstucke:
Please see #3149 & #3105 for context. 

Migrating from React to Preact, our suspense boundaries are failing everywhere with this problem still on latest Preact. 

Looking into those specific issues and not finding a solution that worked, I delved into the `suspense.js` file and played around. The only solution I could find was adding a check for `originalParent` at the top level of `removeOriginal()`. 

I've created a test from the repro in #3149, their way was the only way i could find to simply reproduce.

The change to `suspense.js` fixes the test. 

I'm currently patching preact dist code with this change to make our app work, if this change is wrong and can't be merged, any way of getting round the problem would be greatly appreciated, it's the only thing I've found to work!

### DIFF
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -70,7 +70,7 @@ function detachedClone(vnode, detachedParent, parentDom) {
 }
 
 function removeOriginal(vnode, detachedParent, originalParent) {
-	if (vnode) {
+	if (vnode && originalParent) {
 		vnode._original = null;
 		vnode._children =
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
Great catch! Thanks for making a PR for this 👍

zakstucke:
Awesome thanks @marvinhagemeister! :tada:

PR #4144, opened 2023-09-29
Fix: "Improper nesting of paragraph" for span under a paragraph

cheeaun:
I upgraded to Preact v10.18.0 and got quite a number of warnings. There's one warning that seems wrong:

```
Improper nesting of paragraph. Your <p> should not have span as child-elements.
```

I looked at the code and the regex is:

```
/address|article|aside|blockquote|details|div|dl|fieldset|figcaption|figure|footer|form|h1|h2|h3|h4|h5|h6|header|hgroup|hr|main|menu|nav|ol|p|pre|search|section|table|ul/
```

Testing it on browser console:

Screenshot 2023-09-29 at 12 19 28 PM

Turns out it returns true because `span` contains `p` 🫣

Changes made:
- Add start, end and round brackets to regex so that it does exact matching.
  - I'm not sure if the non-exact matching was intentional, but I ran the tests (`npm run test:karma`), everything seems passing.
- Add test specifically for span under a paragraph.
  - There could be other tags that might be wrongly matched, but as of now, I only know `span` has this issue.

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -489,7 +489,7 @@ function isTableElement(type) {
 }
 
 const ILLEGAL_PARAGRAPH_CHILD_ELEMENTS =
-	/address|article|aside|blockquote|details|div|dl|fieldset|figcaption|figure|footer|form|h1|h2|h3|h4|h5|h6|header|hgroup|hr|main|menu|nav|ol|p|pre|search|section|table|ul/;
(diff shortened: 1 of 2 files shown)


andrewiggins (approved):
Thanks for the fix!

JoviDeCroock (approved):

PR #4157, opened 2023-10-09
golf: Switch `===` to `==` in a few places where not needed

rschristian:
Sort of a follow-up to #4152, I don't believe there's a need for `===` in `.constructor` or `typeof` checks, and locally I saw a 3-4 byte reduction that I'd like to check.


JoviDeCroock (approved):
Great spot!

PR #4160, opened 2023-10-18
fix: prevent invalid DOM nesting false positives

marvinhagemeister:
When Preact is rendering into a container in the middle of the page we attempted to walk outside the render root via the DOM tree. Problem is that that doesn't work on the server where there is no DOM, which lead to us spamming the terminal with false positive warnings.

So this skips the DOM validation if we reach the render root and thereby partially reverts https://github.com/preactjs/preact/pull/4043 . For SPAs this should work as the offending elements very likely have another DOM parent node anyway. For rendering into an existing DOM tree, this skips validation if we reach the render root eagerly. This seems like an ok tradeoff to make.

Fixes https://github.com/denoland/fresh/issues/1773, fixes https://github.com/preactjs/preact-render-to-string/issues/320

### DIFF
--- a/debug/test/browser/component-stack.test.js
+++ b/debug/test/browser/component-stack.test.js
@@ -79,7 +79,7 @@ describe('component stack', () => {
 		render(<Bar />, scratch);
 
 		let lines = getStack(errors).split('\n');
-		expect(lines[0].indexOf('td') > -1).to.equal(true);
+		expect(lines[0].indexOf('tr') > -1).to.equal(true);
 		expect(lines[1].indexOf('Thrower') > -1).to.equal(true);
 		expect(lines[2].indexOf('Bar') > -1).to.equal(true);
 	});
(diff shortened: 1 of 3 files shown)


debug/test/browser/component-stack.test.js
```diff
@@ -79,7 +79,7 @@ describe('component stack', () => {
 		render(<Bar />, scratch);
 
 		let lines = getStack(errors).split('\n');
-		expect(lines[0].indexOf('td') > -1).to.equal(true);
+		expect(lines[0].indexOf('tr') > -1).to.equal(true);
```
marvinhagemeister, line 82: This generates two warnings, which are both correct imo. The difference is that the order is only swapped because we do the validation based on the order in `options.diffed` which is the reverse of what we had before.

JoviDeCroock (approved):

PR #4164, opened 2023-10-23
Add the `exportparts` global attribute

shoonia:
MDN: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/exportparts


marvinhagemeister (approved):
Sweet, thanks for making a PR 👍

PR #4162, opened 2023-10-21
Rework children diffing to run in multiple phases

andrewiggins:
Rework children diffing to run in multiple phases:

1. Find matches for all new VNodes in the old VNode tree, and determine which VNodes moved
2. Removed any unmatched VNodes in the old tree from the DOM
3. Diff, move, and insert all VNodes in the new tree

TODO:
- [ ] See if this fixes the test case in #4159

PR #4166, opened 2023-10-24
Always clear `_nextDom` field on VNodes

andrewiggins:
While working on #4162, I discovered some situations where we fail to clear `_nextDom` meaning we hang on to possibly detached HTMLElements. This fixes a bug in diffChildren so we always clear `_nextDom` for components/fragments and clears `_nextDom` in `commitRoot` since root VNodes don't necessarily go through `diffChildren` to get cleared.

I've added some code to teardown to run through the VNode tree and verify that `_nextDom` is null/undefined.

I also found a reference to `_nextDom` in Suspense. However in all of our tests it is null so I replaced that call to `insertBefore(node, null)` with a call to `appendChild`.

### DIFF
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -81,7 +81,7 @@ function removeOriginal(vnode, detachedParent, originalParent) {
 		if (vnode._component) {
 			if (vnode._component._parentDom === detachedParent) {
 				if (vnode._dom) {
-					originalParent.insertBefore(vnode._dom, vnode._nextDom);
+					originalParent.appendChild(vnode._dom);
 				}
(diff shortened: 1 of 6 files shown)


JoviDeCroock (approved):
Nice find, the many updates perf is weird though

andrewiggins:
Yea, agreed. Looking into it. Given how large of a range it is, I wonder if we are catching a new GC in the measured duration with this update

andrewiggins:
Wow. The fact that 7d03e4b2388b970b55f0954c1e9c6c70b61a0389 fixed the many_updates regression is utterly baffling to me

PR #4172, opened 2023-10-26
Improve place child

andrewiggins:
NOTE: Might be easier to review by adding `?w=1` to the URL

When unmounting DOM, we need to move `oldDom`/`_nextDom` past the unmounted DOM so the diff can continue by comparing the next VNodes to DOM nodes that are mounted. We did this in the unmount loop but we missed a couple other places where this needs to happen as well.

Let's say we have the following components:
```jsx
function Conditional({ condition }) {
  return condition ? (
      <>
        <p>2</p>
        <p>3</p>
      </>
    ) : null;
  }

function App() { return <>... <Conditional/> ... </>; }
```

Here, Conditional will switch from a group of DOM nodes and `null`. When doing this, diffChildren would hit our null placeholder code which does update oldDom past the unmounted DOM, but it doesn't update `_nextDom`, so `App` wouldn't see this move and try to continue diffing from an unmounted DOM node.

Further, even if we set `_nextDOM`, diffChildren on App wouldn't see it because `Conditional` returned `null`. Since Conditional returned `null`, it would return `null` for `newDom`,  meaning diffChildren on App would skip over all the code that was previously under a `if (newDom)` check, including the code that reads `_nextDom`. This PR also fixes that issue so even if a component returns `null`, we update the skewed index, and will read `_nextDom` if necessary.

Also, the test "should efficiently unmount & mount components that conditionally return null with null first child" contains an additional edge case where Conditional switching to null, matches it's previous null first child. We need to ensure we detect we still need to move oldDom and update `_nextDom`.

Previously we attempted to handle all of these cases by checking `oldDom.parentNode !== parentDom` in `placeChild`. However, in our entire test case, this was only occurred when oldDom was unmounted. So this update prevents unmounted oldDom from ever reaching `placeChild` removing the need for that condition entirely, dramatically simplifying `placeChild`.

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -120,6 +120,13 @@ export function diffChildren(
 			if (oldVNode && oldVNode.key == null && oldVNode._dom) {
 				if (oldVNode._dom == oldDom) {
 					oldDom = getDomSibling(oldVNode);
+
+					if (typeof newParentVNode.type == 'function') {
+						// If the parent VNode is a component/fragment, make sure its diff
+						// continues with a DOM node that is still mounted in case this loop
+						// exits here because the rest of the new children are `null`.
+						newParentVNode._nextDom = oldDom;
+					}
 				}
 
 				unmount(oldVNode, oldVNode, false);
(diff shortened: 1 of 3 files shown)


src/diff/children.js
```diff
@@ -304,10 +310,8 @@ export function toChildArray(children, out) {
 }
 
 function placeChild(parentDom, newDom, oldDom) {
-	if (oldDom == null || oldDom.parentNode !== parentDom) {
-		parentDom.insertBefore(newDom, null);
-	} else if (newDom != oldDom || newDom.parentNode == null) {
```
andrewiggins, line 309: The `newDom.parentNode == null` condition can also be removed cuz it would never happen, or if did would result in a DOMException.

For this condition to be hit, `newDom` would have to be equal to `oldDom` cuz of the check prior to this one. So in that scenario, if `newDom.parentNode == null`, then `oldDom.parentNode == null`. If oldDom is unmounted (since it has no parentNode), then calling `parentDom.insertBefore(newDom, oldDom)` would error cuz oldDom is unmounted and you can't insert a node before another node that is unmounted.

andrewiggins:
For some additional context, the new Fragment tests I added are actually from our Suspense test suite. The suspense tests started failing, so I pulled out a simplified repro from those tests into its own test so it was easier to debug and see what's going on.

andrewiggins:
@marvinhagemeister This PR removes usages of `.parentNode` in our diff! Hopefully this will help unblock some of the Fresh stuff we talked about before

marvinhagemeister (approved):
Yooo this is fantastic!!! From quickly trying it out it looks like it resolves the issue I was running into with Fresh 👍 

Also good call on making Fragment tests out of some Suspense tests.

JoviDeCroock (approved):

PR #4173, opened 2023-10-27
Improve internal JSDoc types

andrewiggins:
This PR improves the internal JSDoc types of the core source. It puts all internal types into the global namespace so you no longer need to do `import('../internal')` in the core source files as well as adds some additional JSDoc and ts-expect-error to improve typing.

I've added a "tsc" script that runs the TypeScript type checker against the core source files to validate their types. I don't know if we want to keep this as a merge check, but figured we'd give it a shot.

### DIFF
--- a/jsconfig-lint.json
+++ b/jsconfig-lint.json
@@ -1,0 +1,4 @@
+{
+  "extends": "./jsconfig.json",
+  "include": ["src/**/*"]
+}
--- a/package.json
+++ b/package.json
@@ -133,6 +133,7 @@
     "test:karma:bench": "cross-env PERFORMANCE=true COVERAGE=false BABEL_NO_MODULES=true karma start karma.conf.js --grep=test/benchmarks/**.js --single-run",
     "benchmark": "npm run test:karma:bench -- no-single-run",
     "lint": "run-s eslint",
+    "lint:tsc": "tsc -p jsconfig-lint.json",
     "eslint": "eslint src test debug compat hooks test-utils",
     "format": "prettier --write \"**/*.{js,jsx,mjs,cjs,ts,tsx,yml,json,html,md,css,scss}\"",
     "format:check": "prettier --check '**/*.{js,jsx,mjs,cjs,ts,tsx,yml,json,html,md,css,scss}'"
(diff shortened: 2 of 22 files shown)


package.json
```diff
@@ -132,7 +132,8 @@
     "test:karma:test-utils": "cross-env PERFORMANCE=false COVERAGE=false BABEL_NO_MODULES=true karma start karma.conf.js --grep=test-utils/test/shared/**.js --no-single-run",
     "test:karma:bench": "cross-env PERFORMANCE=true COVERAGE=false BABEL_NO_MODULES=true karma start karma.conf.js --grep=test/benchmarks/**.js --single-run",
     "benchmark": "npm run test:karma:bench -- no-single-run",
-    "lint": "run-s eslint",
+    "lint": "run-s eslint tsc",
+    "tsc": "tsc -p jsconfig-lint.json",
```
JoviDeCroock, line 136: Let's add this to CI now that it passes so we can keep that

JoviDeCroock (approved):

jsconfig-lint.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "./jsconfig.json",
+  "include": ["src/**/*"]
+}
```
rschristian, line 4: Just curious: what's the need for this? Wouldn't the base config also include `src/**/*`, given the default value for `include` is "everything" and `src` hasn't been explicitly excluded?

andrewiggins: Yup, that's right. It does. I created this file to *only* include `./src/**/*` and so other src directories (like compat and hooks) and test directories are not included.

Running `tsc -p jsconfig.json` results in >5k errors. I wanted to start small and just enforce type checks on the root source folder for now.

In other words - this is the config we'll use for CI where we only enforce type correctness on the root src folder, not nested src folders, yet.

rschristian: > Running tsc -p jsconfig.json results in >5k errors.

Oh, lol. That makes sense. I forgot that I turn on my blinders when it comes to TS in this repo.

package.json
andrewiggins: Yup! It now runs as part of the `lint` command: https://github.com/preactjs/preact/actions/runs/6664355219/job/18111803276?pr=4173#step:5:21

marvinhagemeister (approved):
Ohh this is lovely! Was always annoying me a little that we had to type `import("./internal")` all the time but didn't know a better solution. Glad to see this being improved 👍

andrewiggins:
Note, you may have to do import in sub packages. I haven't looked into those yet. If anyone else wants to look at those feel free to.

PR #4175, opened 2023-10-27
Allow handling errors in getSnapshot of useSyncExternalStore & add more tests

andrewiggins:
* If getSnapshot throws an error while checking for an update, allow components to handle it by rerendering the component so that getSnapshot can throw again during render.
* Move useSyncExternalStore tests into their own file
* Copy relevant tests from React's test suite. I removed tests related to useSyncExternalStoreWithSelector since React doesn't support this and engineers have to use the shim package.
* Add test for handling updates to store before subscribing in useEffect


JoviDeCroock (approved):

marvinhagemeister (approved):
Nice!

PR #4177, opened 2023-10-30
feat: support precompiled JSX transform

marvinhagemeister:
This PR adds support for the new precompile JSX transform that will ship with the next deno version. The new deno version will come out this week.

The idea behind the precompile JSX transform is that instead of allocating a full vnode object + props for every element, we precompile HTML nodes during the JSX transpilation step to strings. During rendering we then need to mostly only concatenate strings together.

Demo: https://github.com/marvinhagemeister/deno-jsx-precompile-preact-demo

See https://github.com/preactjs/preact-render-to-string/pull/322 .


jsx-runtime/src/encode.js
```diff
@@ -0,0 +1,36 @@
+const ENCODED_ENTITIES = /["&<]/;
+
+/** @param {string} str */
+export function encodeEntities(str) {
```
marvinhagemeister, line 4: This function is just copied over from `preact-render-to-string` https://github.com/preactjs/preact-render-to-string/blob/17f839c47309d7de35cc407c25b016a0dcc325c8/src/util.js#L11

JoviDeCroock (approved):
I am so excited for this!

jsx-runtime/src/utils.js
```diff
@@ -0,0 +1,102 @@
...
+		if (i !== last) out += str.slice(last, i);
+		out += ch;
+		// Start the next seek/buffer after the entity's offset:
+		last = i + 1;
+	}
+	if (i !== last) out += str.slice(last, i);
+	return out;
+}
+
+const IS_NON_DIMENSIONAL = new Set([
```
JoviDeCroock, line 38: Should we use `export const IS_NON_DIMENSIONAL =/acit|ex(?:s|g|n|p|$)|rph|grid|ows|mnc|ntw|ine[ch]|zoo|^ord|itera/i;` from our codebase?

PR #4180, opened 2023-10-30
Rework children diffing to run in multiple phases

### LINKED ISSUE #4156: <iframe> reloads its contents after removing one of its previous siblings when used with react-slate

**Describe the bug**

In a web application with slate, slate-react and Preact, if I have a node before an iframe and then I remove that node, the iframe will be reloaded. This seem to be caused by a complete re-render of the Slate's editor's content.

This behavior is not present in Preact 10.15.1 and before, and it looks like it was introduced in 10.16.0 and keep happening after that version.

**To Reproduce**

1. Go to https://stackblitz.com/edit/create-preact-starter-iza5ww?file=src%2Findex.jsx
2. Open result in another tab
3. Try to delete the second node (the paragraph with the text "Delete this line to see the iframe reload"). Or press Enter after or before the paragraph to create a new one and delete the newly created node.
4. The iframe will refresh.

**Note:** Compare with https://stackblitz.com/edit/create-preact-starter-i9u6lv?file=src%2Findex.jsx, where the same code uses Preact 10.15.1 and doesn't result in a iframe refresh. Also, [it doesn't happen in React](https://stackblitz.com/edit/stackblitz-starters-4s5n2h?file=src%2FApp.tsx).

**Expected behavior**
The iframe shouldn't refresh if I update its previous siblings.

andrewiggins:
Restart of #4162 on top of lastest main.

Rework children diffing to run in multiple phases:

1. First, find matches for all new VNodes in the old VNode tree, and determine which VNodes moved
2. Second, remove any unmatched VNodes in the old tree from the DOM
3. Finally, diff, move, and insert all VNodes in the new tree

A couple additional changes to highlight
* fd710b8beb3703a4aafca5b87fc6f65c8c81e827 Add new _flags VNode field to capture some diffing state (whether an oldVNode was matched and whether to insert a newVNode)
* b6a10729b7931bb125f25ce54c0b0456370ea40e Combine `placeChild` and `reorderChildren` into a single `insert` function & simplify insertion logic in diffChildren
* cd0127fe41a97ce07c3efed7c6e838ee5112ed7c Simplify some _nextDom logic by always setting & clearing it
* 6eef30b3723201b37cb717700bcc9ba795b962a0 Fold _hydrating field into _flags field (+12 B)

Fixes #4156

### DIFF
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -1,4 +1,5 @@
 import { Component, createElement, options, Fragment } from 'preact';
+import { MODE_HYDRATE } from 'preact/src/constants';
 import { assign } from './util';
 
 const oldCatchError = options._catchError;
@@ -34,7 +35,7 @@ options.unmount = function (vnode) {
 	// most likely it is because the component is suspended
 	// we set the vnode.type as `null` so that it is not a typeof function
 	// so the unmount will remove the vnode._dom
-	if (component && vnode._hydrating === true) {
+	if (component && vnode._flags & MODE_HYDRATE) {
 		vnode.type = null;
 	}
 
@@ -166,7 +167,7 @@ Suspense.prototype._childDidSuspend = function (promise, suspendingVNode) {
 	 * to remain on screen and hydrate it when the suspense actually gets resolved.
 	 * While in non-hydration cases the usual fallback -> component flow would occour.
 	 */
-	const wasHydrating = suspendingVNode._hydrating === true;
+	const wasHydrating = (suspendingVNode._flags & MODE_HYDRATE) === MODE_HYDRATE;
 	if (!c._pendingSuspensionCount++ && !wasHydrating) {
 		c.setState({ _suspended: (c._detachOnNextRender = c._vnode._children[0]) });
 	}
@@ -204,7 +205,7 @@ Suspense.prototype.render = function (props, state) {
 	/** @type {import('./internal').VNode} */
 	const fallback =
 		state._suspended && createElement(Fragment, null, props.fallback);
-	if (fallback) fallback._hydrating = null;
+	if (fallback) fallback._flags &= ~MODE_HYDRATE;
 
 	return [
 		createElement(Fragment, null, state._suspended ? null : props.children),
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -167,8 +167,7 @@ Suspense.prototype._childDidSuspend = function (promise, suspendingVNode) {
 	 * to remain on screen and hydrate it when the suspense actually gets resolved.
 	 * While in non-hydration cases the usual fallback -> component flow would occour.
 	 */
-	const wasHydrating = (suspendingVNode._flags & MODE_HYDRATE) === MODE_HYDRATE;
-	if (!c._pendingSuspensionCount++ && !wasHydrating) {
+	if (!c._pendingSuspensionCount++ && !(suspendingVNode._flags & MODE_HYDRATE)) {
 		c.setState({ _suspended: (c._detachOnNextRender = c._vnode._children[0]) });
 	}
 	promise.then(onResolved, onResolved);
--- a/src/component.js
+++ b/src/component.js
@@ -122,12 +122,11 @@ export function getDomSibling(vnode, childIndex) {
 function renderComponent(component) {
 	let oldVNode = component._vnode,
 		oldDom = oldVNode._dom,
-		parentDom = component._parentDom;
+		parentDom = component._parentDom,
+		commitQueue = [],
+		refQueue = [];
 
 	if (parentDom) {
-		let commitQueue = [],
-			refQueue = [];
-
 		const newVNode = assign({}, oldVNode);
 		newVNode._original = oldVNode._original + 1;
 
--- a/src/component.js
+++ b/src/component.js
@@ -2,6 +2,7 @@ import { assign } from './util';
 import { diff, commitRoot } from './diff/index';
 import options from './options';
 import { Fragment } from './create-element';
+import { MODE_HYDRATE } from './constants';
 
 /**
  * Base Component class. Provides `setState()` and `forceUpdate()`, which
@@ -136,10 +137,10 @@ function renderComponent(component) {
 			oldVNode,
 			component._globalContext,
 			parentDom.ownerSVGElement !== undefined,
-			oldVNode._hydrating != null ? [oldDom] : null,
+			oldVNode._flags & MODE_HYDRATE ? [oldDom] : null,
 			commitQueue,
 			oldDom == null ? getDomSibling(oldVNode) : oldDom,
-			oldVNode._hydrating,
+			(oldVNode._flags & MODE_HYDRATE) === MODE_HYDRATE,
 			refQueue
 		);
 
--- a/src/component.js
+++ b/src/component.js
@@ -140,7 +140,7 @@ function renderComponent(component) {
 			oldVNode._flags & MODE_HYDRATE ? [oldDom] : null,
 			commitQueue,
 			oldDom == null ? getDomSibling(oldVNode) : oldDom,
-			(oldVNode._flags & MODE_HYDRATE) === MODE_HYDRATE,
+			!!(oldVNode._flags & MODE_HYDRATE),
 			refQueue
 		);
 
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -269,7 +269,7 @@ function constructNewChildrenArray(newParentVNode, renderResult, oldChildren) {
 		childVNode._parent = newParentVNode;
 		childVNode._depth = newParentVNode._depth + 1;
 
-		let skewedIndex = i + skew;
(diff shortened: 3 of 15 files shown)


src/diff/children.js
```diff
@@ -40,27 +46,148 @@ export function diffChildren(
...
 	// as EMPTY_OBJ._children should be `undefined`.
 	/** @type {VNode[]} */
 	let oldChildren = (oldParentVNode && oldParentVNode._children) || EMPTY_ARR;
 
-	let oldChildrenLength = oldChildren.length,
-		remainingOldChildren = oldChildrenLength,
-		newChildrenLength = renderResult.length;
+	let newChildrenLength = renderResult.length;
+
+	// TODO: Could we drop oldDom all together and just use _nextDom?
```
JoviDeCroock, line 65: We could check this in a follow-up PR

JoviDeCroock (approved):
This looks amazing, I have a follow up with the test in #4181

src/diff/children.js
andrewiggins: Agreed. I gave it a quick attempt and it got messy really quickly. So not gonna do that in this PR. Will clean up the comment before checking in

andrewiggins:
Okay, I think I've figured out the performance slowdown in many_updates. It appears we have hit a perf cliff with the number of properties on VNode.

I first tried some code reorganizations to see if it was a runtime difference (those experiments can be found on the `two-phase-diffChildren-2-deopts` branch). However none of them particularly moved the needle a lot. So I dug into the tracing analysis & glanced at a couple traces to see what was different.

To find the analysis tables shown below, go to the linked "Benchmark Debug" workflow run, view the logs for the "Bench many_updates" job, and expand the "Analyze logs if present" step. To view the trace timelines, download the `many_updates_logs` artifact of the linked workflow run and upload it to [ui.perfetto.com](https://ui.perfetto.dev). Reminder, tables shown below with the title "Count ..." have the wrong unit/label in the "Avg time" column. The column should be labelled "Avg count" and it shouldn't have a `ms` unit. Known issue in tachometer. Will maybe fix one day 😅 

Below are some screenshots from tracing and analyzing a commit from two-phase-diffChildren-2-deopts branch (before a rebase and before playing with the number of VNode fields). You can see the results yourself here: https://github.com/preactjs/preact/actions/runs/6739103123.

image

image

image

image

image

image

The first thing I noticed from examining a sample trace timeline of `preact-local` running the `many_updates` is that we have a lot of MinorGCs happening synchronously while V8 is running the benchmark. Looking at the timelines for `preact-main`, that almost never happens and MinorGCs almost always happen as a task after paint. 

Looking at the analysis tables also confirms some other changes related to GC. Notice how while the "Count of MinorGCs" is basically the same between the "preact-main" and "preact-local" (this branch), the amount of time spent in MinorGC ("Sum of MinorGC") is much larger. Also, the amount of heap used after a MinorGC ("MinorGC usedHeapSizeAfter") is much larger! even though after a MajorGC the usedHeapSize is basically the same (couple hundred kilobytes different).

The biggest change to GC that I think this PR introduces is adding a field to VNode. So I tested again by removing the `_hydrating` field (since many_updates doesn't need it) and the benchmark appears to return to normal. That workflow run can be found here: https://github.com/preactjs/preact/actions/runs/6739355022/job/18320720488

image

image

image

image

So I'm going to explore ways to perhaps reduce the number of VNode fields so fix this regression. I'm considering folding `_hydrating` into a `_flags` but removing a VNode field could be a breaking change, so would be curious to get other's thoughts here.

JoviDeCroock:
@andrewiggins 

Removing internal only properties like `_hydrating` seem okay to me, as they are internal `_hydrating`. I don't know whether there are many others that we can remove without drastic changes to i.e. how we track DOM

src/component.js
```diff
@@ -137,10 +137,10 @@ function renderComponent(component) {
 			oldVNode,
 			component._globalContext,
 			parentDom.ownerSVGElement !== undefined,
-			oldVNode._hydrating != null ? [oldDom] : null,
+			oldVNode._flags & MODE_HYDRATE ? [oldDom] : null,
 			commitQueue,
 			oldDom == null ? getDomSibling(oldVNode) : oldDom,
-			oldVNode._hydrating,
+			(oldVNode._flags & MODE_HYDRATE) === MODE_HYDRATE,
```
JoviDeCroock, line 143: Can be shortened to

```suggestion
			!!(oldVNode._flags & MODE_HYDRATE),
```

compat/src/suspense.js
```diff
@@ -166,7 +167,7 @@ Suspense.prototype._childDidSuspend = function (promise, suspendingVNode) {
 	 * to remain on screen and hydrate it when the suspense actually gets resolved.
 	 * While in non-hydration cases the usual fallback -> component flow would occour.
 	 */
-	const wasHydrating = suspendingVNode._hydrating === true;
+	const wasHydrating = (suspendingVNode._flags & MODE_HYDRATE) === MODE_HYDRATE;
 	if (!c._pendingSuspensionCount++ && !wasHydrating) {
```
JoviDeCroock, line 171: Saves some bytes

```suggestion
	if (!c._pendingSuspensionCount++ && !(suspendingVNode._flags & MODE_HYDRATE)) {
```

src/diff/index.js
```diff
@@ -45,11 +50,11 @@ export function diff(
 	if (newVNode.constructor !== undefined) return null;
 
 	// If the previous diff bailed out, resume creating/hydrating.
-	if (oldVNode._hydrating != null) {
-		isHydrating = oldVNode._hydrating;
+	if (oldVNode._flags & MODE_SUSPENDED) {
+		isHydrating = (oldVNode._flags & MODE_HYDRATE) === MODE_HYDRATE;
```
JoviDeCroock, line 54:
```suggestion
		isHydrating = !!(oldVNode._flags & MODE_HYDRATE);
```

src/diff/index.js
```diff
@@ -253,7 +258,7 @@ export function diff(
 			c.base = newVNode._dom;
 
 			// We successfully rendered this VNode, unset any stored hydration/bailout state:
-			newVNode._hydrating = null;
+			newVNode._flags &= RESET_MODE;
```
JoviDeCroock, line 261: This can be removed, I removed it in my PR and everything worked correctly

src/diff/children.js
```diff
@@ -40,25 +40,139 @@ export function diffChildren(
...
+	/** @type {number} */
+	let i;
+	/** @type {VNode} */
+	let childVNode;
+	/** @type {VNode} */
+	let oldVNode;
+
+	const newChildrenLength = renderResult.length;
+	const oldChildrenLength = oldChildren.length;
+	let remainingOldChildren = oldChildren.length;
```
JoviDeCroock, line 173:
```suggestion
	let oldChildrenLength = oldChildren.length,
		remainingOldChildren = oldChildrenLength
```

JoviDeCroock (approved):

src/diff/index.js
andrewiggins: Oh interesting. Would've expected this to error. Might suggest we are missing a test case. Gonna look into this to see understand that better.

andrewiggins (commented):
Thanks for all the code golf suggestions <3

src/diff/index.js
andrewiggins: I think imma keep this cuz there is a situation where we'll preserve the MODE_HYDRATE, etc. flags causing rerenders of the component to go down the hydration code paths. Given how hydration works today, it doesn't appear to lead to incorrect DOM, just unnecessarily running through the hydration code paths.

Specifically the test I added in my last commit reproduces what I'm thinking. Normally with Suspense, when the Suspense rerenders on promise resolution, we'll generate a new VNode using createElement for the component that suspended, which will have `0` for `_flags`. Once that rerender completes successfully, the `0` for `_flags` will become the new tree, effectively "clearing" the flags on the suspending component/VNode.

However, if that suspending component is memoized, then when Suspense rerenders on promise resolution, we won't generate a new VNode for the suspending component using createElement. Instead will trigger a forceUpdate on the suspending component, which will generate VNode using `Object.assign` in `renderComponent`. The newVNode for the suspending component will have the same flags as the oldVNode, i.e. MODE_HYDRATE. Once it successfully rerenders, the newVNode will become part of the tree, preserving the MODE_HYDRATE in the new tree. As long as the component doesn't get any rerenders from the parent (which is likely since this component is memoized), it will only rerender by going through `renderComponent` invoking the same behavior as described here (i.e. creating a newVNode by copying oldVNode therefore inheriting flags).

However, since with MODE_HYDRATION set, we just set oldDom to newVNode._dom, it usually just works. We also set excessDomChildren to `[oldDom]` which is what invokes some hydration code paths. Presumably there is a situation where setting `oldDom` to `newVNode._dom` would be incorrect, but I haven't come up with one yet. But to be safer, Imma leave this line here.

JoviDeCroock:
The text_update regression makes me wonder whether it would be worth having a specialized path but not sure 😅 can look deeper after we merge

PR #4186, opened 2023-10-31
Revert "Add Benchmark Debug workflow"

andrewiggins:
Reverts preactjs/preact#4185

Reverting as this breaks PR Benchmarks. Workflows that call the CI workflow will default to 'main' branch now instead of the default checkout behavior, meaning the benchmark job on PRs will now just do a A/A comparison since they will build the 'main' branch instead of the PR merge commit.

Below is a screenshot of the benchmark action on a PR rebased to included this change. Notice how the checkout job has 'main' specified as the checkout ref.

image

Will play around with this more in a fork to get it right.


andrewiggins:
Found fixes. Gonna roll-forward with them.

PR #4187, opened 2023-10-31
Fix benchmark debug action

andrewiggins:
Couple issues found in the "Benchmark Debug" & related changes to "CI" workflow:

- CI `ref` input should default to empty string so by default it checkouts whatever the default commit of the event that triggered it (e.g. merge commit for pull_request event, pushed commit for push event)
- Add `npm run analyze` command for benchmarks that include logs to analyze
- There is a bug in GitHub workflow parser that doesn't properly handle number-like inputs being passed as an input to a called workflow. Comment out timeout input for now
- Fix ordering of tarballing and clearing directory.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on:
         description: 'Branch or tag ref to check out'
         type: string
         required: false
-        default: 'main'
+        default: ''
       artifact_name:
         description: 'Name of the artifact to upload'
         type: string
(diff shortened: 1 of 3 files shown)


andrewiggins:
Successful benchmark debug workflow runs here:

https://github.com/preactjs/preact/actions/runs/6714073169
https://github.com/preactjs/preact/actions/runs/6714075055

JoviDeCroock (approved):

PR #4191, opened 2023-11-04
Remove the _isHydrating property for a flag

JoviDeCroock:


andrewiggins:
Oh whoops! Just seeing this sorry. Hmm you got this working without a MODE_SUSPENDING flag... Might use that if it saves some bytes

Issue #4194, opened 2023-11-06
10.18.2 regression:  Cannot read properties of undefined


zakstucke:
- [ ] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
Our dashboard starts fails when upgrading from 18.1 to 18.2. Looks like suspense boundary regressions. 
I haven't got an easy repro or capacity to find one right now, just opening for visibility on the issue.

<img width="827" alt="Screenshot 2023-11-06 at 13 23 22" src="https://github.com/preactjs/preact/assets/44890343/51f226d5-df37-4b69-ba81-02dcd641a4e6">

JoviDeCroock:
@zakstucke any luck on the reproduction yet?

zakstucke:
@JoviDeCroock sad to say not yet! Like with the last suspense issue I helped with a while back, they seem to pop up in very specific circumstances and so all seems to work fine in a simple repo...

I've just checked with the latest version: `10.19.3` where issue still persists, last working version `10.18.1`.

<img width="776" alt="image" src="https://github.com/preactjs/preact/assets/44890343/d84c9883-dc2f-46dc-b496-9528b26c3444">

zakstucke:
Although looking at this new screenshot errs seem slightly different, now in signals, although this might be unrelated and from changes in my code.

JoviDeCroock:
@zakstucke I wonder if it's https://github.com/preactjs/preact/pull/4182/files we should check for the existence of oldVNode where does the stacktrace lead you in preact?

zakstucke:
@JoviDeCroock @marvinhagemeister sorry to say that fix wasn't it!

But I've been useless for long enough, I've spent some time this morning trimming code until I got to a simple repro. 

Repro made with `npm preact init` can be found here:
https://github.com/zakstucke/preact-4194-repro

This repro fails with `10.19.5`, works with `10.18.1`, and pretty sure it's the root of this issue.

`cd my-preact-app && npm i && npm run dev`

Going to `http://localhost:5173` will give the following error:
<img width="844" alt="image" src="https://github.com/preactjs/preact/assets/44890343/88576c01-2b37-4855-ac6c-28ee2beb0988">

The repro is effectively:
```ts
import { render } from 'preact';
import { Suspense } from "react";

import { Tab } from "@headlessui/react";

import { TestComp, TestWrapper } from "./lazyBarrier";

export function App() {
	return (
		<Suspense fallback={<p>Loading...</p>}>
[...]

PR #4196, opened 2023-11-07
Potential Suspense fix

JoviDeCroock:
I have not been able to reproduce #4194 however my thinking is that we replace `_children` erroneously in some cases 😅 

@zakstucke mind either providing a reproduction or testing this


zakstucke:
@JoviDeCroock just locally packaged and installed this branch, still failing with same errors :(

PR #4198, opened 2023-11-07
Do typeof string check before looking for String constructor

andrewiggins:
In `constructNewChildrenArray` we look for the `String` constructor to support #4151 (i.e. strings created with `new String`). However, doing this check first leads to a megamorphic deopt in `constructorNewChildrenArray` since V8 has many different internal types for strings, leading to a megamorphic access for `childVNode.constructor` in the common case (normal strings and VNodes).

This PR adds back the `typeof childVNode == 'string'` check to catch normal strings first before falling back to `childVNode.constructor == String` to check for manually constructed strings.

The below screenshot shows the results of DeoptExplorer running many_updates benchmark on main. We hit a deopt on the `childVNode.constructor` due to the number of maps it hits (V8 has many different internal string representations leading to different maps at this location).

image

Adding back the `typeof childVNode == 'string'` removes the deopt for the common case:

image

To reduce size impact I've included the following golfs:
* Store matching oldVNode in variable in constructNewChildrenArray (-4 B)
* Inline insertion checks into skew block (-3 B)

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -288,6 +288,11 @@ function constructNewChildrenArray(newParentVNode, renderResult, oldChildren) {
 			if (matchingIndex == -1) {
 				skew--;
 			}
+
+			// If we are mounting a DOM VNode, mark it for insertion
+			if (typeof childVNode.type != 'function') {
+				childVNode._flags |= INSERT_VNODE;
+			}
 		} else if (matchingIndex !== skewedIndex) {
 			if (matchingIndex === skewedIndex + 1) {
 				skew++;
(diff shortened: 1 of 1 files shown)


andrewiggins:
Re: 1c06449badd9cfb6f2c1aaaf5f5ce7ff2553eb27 Add back typeof string check in diffChildren (+5 B)

The below screenshot shows the results of DeoptExplorer running many_updates on main. We hit a deopt on the `childVNode.constructor` due to the number of maps it hits (V8 has many different internal string representations leading to different maps at this location).

image

Adding back the `typeof childVNode == 'string'` removes the deopt for the common case:

image

JoviDeCroock (approved):

marvinhagemeister (approved):
Great catch!

PR #4199, opened 2023-11-07
Only run unmounting loop if any children remain to unmont (+2 B)

andrewiggins:
Since we are now looping through children 3 times, I'm hoping skipping a loop on update scenarios provides a small improvement


marvinhagemeister (approved):

andrewiggins:
👀 
image

PR #4200, opened 2023-11-08
Inline `diffProps` function

andrewiggins:
Inline `diffProps` to take advantage of the props loop to read values and use later in the diff. This avoids duplicate megamorphic reads for props like `dangerouslySetInnerHTML`, `children`, `value`, and `checked` since we can reuse reading the value in the props loop for handling elsewhere.

### DIFF
--- a/.github/workflows/single-bench.yml
+++ b/.github/workflows/single-bench.yml
@@ -8,7 +8,7 @@ on:
         type: choice
         options:
           - 02_replace1k
-          - 03_update10k
+          - 03_update10th1k_x16
           - 07_create10k
           - filter_list
           - hydrate1k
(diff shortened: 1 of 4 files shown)


marvinhagemeister (approved):

developit:
Do you think the bench regression is due to the bench change? Or is it due to the second loop? I like the idea of using the first loop to store the incoming `checked` and `value` values and then replacing the second loop with two conditionals.

andrewiggins:
The changes to the bench are me trying the debug the regression. Will try the storing in a variable to see if that improves it at all

andrewiggins:
Yup, that fixes it

PR #4209, opened 2023-11-13
Call options.vnode on rerender

andrewiggins:
Should fix issues in Prefresh, specifically preactjs/prefresh#528


JoviDeCroock (approved):

JoviDeCroock:
@marvinhagemeister is it possible that this messed with devtools as well?

Issue #4208, opened 2023-11-12
Missing typescript types for MathML


scarf005:
- [x] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
types for [MathML](https://developer.mozilla.org/en-US/docs/Web/MathML) are missing in preact. I've searched through [preact documentation](https://preactjs.com) and [repository](https://github.com/search?q=repo%3Apreactjs%2Fpreact%20mathml&type=code) but wasn't able to find related information.

**To Reproduce**

[![Stackblitz](https://img.shields.io/badge/Stackblitz-fff?style=for-the-badge&logo=Stackblitz&logoColor=1389FD)](https://stackblitz.com/edit/preact-tsx-starter-m8gibr?file=App.tsx)

1. Add `<math>` and related MathML tag.
2. The tag errors as `Property 'math' does not exist on type 'JSX.IntrinsicElements'`.

**Expected behavior**

![image](https://github.com/preactjs/preact/assets/54838975/adac47e0-e224-42fb-ac22-bb9185678d22)

mathml is an [widely supported (92.44%) feature](https://caniuse.com/mathml). it'd be great to use them without type errors in preact.

fekete965:
I would be happy to work on this. :)

marvinhagemeister:
Go for it, no need to ask.

PR #4214, opened 2023-11-16
Feature/add mathml definitions

fekete965:
This PR is trying to address the following issue: https://github.com/preactjs/preact/issues/4208
The MathML interfaces were based on the MDN documentation: https://developer.mozilla.org/en-US/docs/Web/MathML
The high commit count could look intimidating, but each commit contains a small change by introducing a new MathML tag.


fekete965:
I will double-check each element and the corresponding definition this week.

fekete965:
I think the PR is good for a review 🙈

marvinhagemeister (approved):
Sweet, thank you so much for adding this 👍

PR #4219, opened 2023-11-26
Fix access to undefined exception

billti:
Running some code locally and I was regularly hitting an error `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'is')`. I tracked it down to the line fixed in this PR. 

It would appear that sometimes `newVNode.props` may be undefined, so guarding against that.

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -212,7 +212,7 @@ function constructNewChildrenArray(newParentVNode, renderResult, oldChildren) {
 				null,
 				null
 			);
-		} else if (childVNode._depth > 0) {
+		} else if (childVNode.constructor === undefined && childVNode._depth > 0) {
 			// VNode is already in use, clone it. This can happen in the following
 			// scenario:
 			//   const reuse = <div />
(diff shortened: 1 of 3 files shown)


JoviDeCroock:
Mind giving some way to reproduce this issue as props are defaulted to an object https://github.com/preactjs/preact/blob/main/src/create-element.js

billti:
It was in the middle of a larger codebase and only happened sporadically, but I'll try and conjure up a minimal repo.

I assumed it could be null as a few lines later (427) it seems to check if it could have been null on the old node (via `oldProps = oldVNode.props || EMPTY_OBJ;`).

JoviDeCroock:
A repro or test case would be needed either way so we can prevent regressions

marvinhagemeister:
Took the liberty of updating the PR. The observed issue is a symptom of another cause in that recent refactors lead to invalid vnodes being passed to `createVNode`.

billti:
Great! Thanks for the 'correct' fix, and super-fast turnaround. I love the project and use it whenever I can, so thanks for your work on it.

andrewiggins (approved):

PR #4220, opened 2023-11-27
fix: onSubmit event type

marvinhagemeister:
The `submit` event dispatches a `SubmitEvent` class, not just `Event`. See https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit_event


lucacasonato (approved):
❤️ This makes my life so much better :0

JoviDeCroock (approved):

PR #4226, opened 2023-12-07
fix: incorrect oninput event type

marvinhagemeister:
Noticed that the event type for `onInput` is too generic. The browser passes an `InputEvent` there. Also noticed that I made an error in the way I typed `onSubmit` in https://github.com/preactjs/preact/pull/4220


marvinhagemeister:
Not sure what coveralls is up to. This doesn't touch any of the files it complains about.

JoviDeCroock (approved):

PR #4229, opened 2023-12-13
Improve types of hook source

andrewiggins:
Improve the typing in hooks source file and add it to the jsconfig for linting on each PR.

One big change, that others may have opinions about, is that here I used `unknown` for types where we explicitly don't know what it is because it is user provided (e.g. the return type of the callback given to `useMemo`). Meaning, usages of `any` in hooks are places where we could improve the types, but I couldn't be bothered cuz it wasn't straightforward. Something to consider improving for later lol.

I wanted to to do this cuz I was recently looking at a compat bug in hooks and was working on a fix but refactoring things in hooks was slightly more difficult given I couldn't trust the red squiggles to be meaningful or not. After this PR they are hopefully more helpful (or at least mean you should double check this before adding an `any` override lol).

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -1,4 +1,4 @@
-import { options } from 'preact';
+import { options as _options } from 'preact';
 
 /** @type {number} */
 let currentIndex;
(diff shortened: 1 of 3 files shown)


JoviDeCroock (approved):

marvinhagemeister (approved):
Nice, I'm all for stronger typing. Going with `unknown` is the right move imo 👍

PR #4230, opened 2023-12-14
Update README.md with code fragments formatted with Prettier

Michael77:
When reading through your README I noticed some crazy spacing in the code examples. I thought huh, that seems awfully odd. So I checked my Github settings and the width is set to 4, but I can see the examples are rendering what appears to be 8 character width tabs. I'm not sure why that would be.

I decided to try running it through Prettier (v3.1.1) and compare the result. Prettier is using spaces which means the result will look the same across all browsers. Many devs use Prettier for formatting so I don't think it would be out of line to propose this change, even if it isn't your favourite choice of formatting.

I'm only proposing this for the front page (README) code examples because when devs view the source for most other files I would assume they are opening them in their IDE via jump to declaration (which then would apply their editors set tab width).

Before:
Screenshot from 2023-12-13 18-32-33

After:
image

Also:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ With Preact, you create user interfaces by assembling trees of components and el
 To get started using Preact, first look at the render() function. This function accepts a tree description and creates the structure described. Next, it appends this structure to a parent DOM element provided as the second argument. Future calls to render() will reuse the existing tree and update it in-place in the DOM. Internally, render() will calculate the difference from previous outputted structures in an attempt to perform as few DOM operations as possible.
 
 ```js
-import { h, render } from 'preact';
+import { h, render } from "preact";
 // Tells babel to use h for JSX. It's better to configure this globally.
 // See https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#usage
 // In tsconfig you can specify this with the jsxFactory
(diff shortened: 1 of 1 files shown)


rschristian:
> I checked my Github settings and the width is set to 4, but I can see the examples are rendering what appears to be 8 character width tabs. I'm not sure why that would be.

This is a GitHub bug/lack of feature, settings don't yet apply to markdown code. There's a few open requests for it.

Tabs are more accessible and this should be addressed on GitHub's end, not ours. Replacing tabs with spaces that match your spacing preferences isn't a good change to land. Thanks though

---

Not quite sure I understand why that FireFox image is relevant?

Michael77:
Thank you for your reply and considering my proposal.  Hopefully Github will address this as I also consider tabs to be optimal for indentation for the reason you mentioned. The formatting Prettier uses is not my personal preference, it just happens to be widely used and renders the resulting code highly portable for presentation purposes due to the fixed indentation size.

> Not quite sure I understand why that FireFox image is relevant?

Sharing browser specs is common in order to repro, if necessary.

rschristian:
I hadn't noticed the different tab spacing per browser, that might be worth making an issue upstream (if it doesn't already exist). Weird indeed.

> The formatting Prettier uses is not my personal preference

Preact uses tabs everywhere, so any deviation from that is to a personal preference is more of a what I meant.

Issue #4161, opened 2023-10-19
Event bubbling can break in test conditions


jramanat:
- [x] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
In test scenarios, on a fast machine, it's possible for preact rendering and a subsequent programmatic call to occur within the span of a single millisecond (yay!).  However, that can trip up the logic in https://github.com/preactjs/preact/pull/4126 because the _dispatched and _attached flags in the dispatched events and associated event listener props all end up getting the same value from Date.now(), resulting in any bubbling listeners getting short-circuited.

**To Reproduce**

1. Go to https://codesandbox.io/s/infallible-https-2z9kls?file=/src/index.js and view the console.log

The test case loops 50 times, rendering a div containing an input, both with a 'focusin' listener, and immediately calling focus on the input.  The listener on the input element is always called (since the new logic always invokes on the first listener), but on my machine, the listener on the div is usually not called.

**Expected behavior**
If you flip the preact version back to 10.17.1, you'll see 50 perfectly interleaved calls to the input + div.

Not sure if the spirit of the original fix could be maintained by just using an incrementing counter rather than Date.now() which I think would solve this case as well?

ritschwumm:
it might be worse - afaik the resolution of `Date.now()` is more granular depending on an number of factors. in firefox it can e.g. be up to 100ms (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now) and it might depend on the underlying operating system, too.

i used a tuple of a Date millisecond and an auto-incrementing 32 bit int (`number|0`)  in a similar case. of course this is not a "real" solution because as soon as someone manages to do 2^32 operations within the same date granule, the problem will come back. it's much more unlikely, though.

jramanat:
Thanks for the additional context!  I can confirm that running my testcase in Firefox with privacy.resistFingerPrinting enabled,  the bubbling almost always fails to work (48+ of the 50 runs).

JoviDeCroock:
Those are all very valid points, I am going to give it some thought whether we can have a better solution to this 😅 basically something where we can indicate that a dom-node/event-handler was created later than when this event started bubbling up.

NatteeB:
@JoviDeCroock Is there any chance that Date.now() can be replaced with performance.now()? This change will not resolve the issue completely but it will improve the chances of the correct behavior.
[...]

PR #4232, opened 2023-12-15
Fix event bubbling can break in test conditions (#4161)

nbalatsk-oracle:
This should fix [#4161](https://github.com/preactjs/preact/issues/4161).

The idea is to use a global counter that will increment on event dispatch only, so it’s less likely that the globalCounter will overflow and the order of _attached/_dispatched is going to be guaranteed.

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -13,6 +13,11 @@ function setStyle(style, key, value) {
 	}
 }
 
+let globalCounter = Number.MIN_SAFE_INTEGER;
+const getCounter = increment => {
+	return increment ? ++globalCounter : globalCounter;
+};
+
 /**
  * Set a property value on a DOM node
  * @param {PreactElement} dom The DOM node to modify
@@ -63,7 +68,7 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
 
 		if (value) {
 			if (!oldValue) {
-				value._attached = Date.now();
+				value._attached = getCounter();
 				const handler = useCapture ? eventProxyCapture : eventProxy;
 				dom.addEventListener(name, handler, useCapture);
 			} else {
@@ -134,7 +139,7 @@ function eventProxy(e) {
 		if (!e._dispatched) {
 			// When an event has no _dispatched we know this is the first event-target in the chain
 			// so we set the initial dispatched time.
-			e._dispatched = Date.now();
+			e._dispatched = getCounter(true);
 			// When the _dispatched is smaller than the time when the targetted event handler was attached
 			// we know we have bubbled up to an element that was added during patching the dom.
 		} else if (e._dispatched <= eventHandler._attached) {
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -13,10 +13,14 @@ function setStyle(style, key, value) {
 	}
 }
 
-let globalCounter = Number.MIN_SAFE_INTEGER;
-const getCounter = increment => {
-	return increment ? ++globalCounter : globalCounter;
-};
+// Force counters to 32-bit unsigned integer
+let globalCounter = 0 >>> 0;
+let cachedCounter = 0 >>> 0;
+const p = Promise.resolve();
+const getCounter = () =>
+	cachedCounter ||
+	(p.then(() => (cachedCounter = 0 >>> 0)),
+	(cachedCounter = ++globalCounter >>> 0));
 
 /**
  * Set a property value on a DOM node
@@ -138,11 +142,11 @@ function eventProxy(e) {
 		 */
 		if (!e._dispatched) {
 			// When an event has no _dispatched we know this is the first event-target in the chain
-			// so we set the initial dispatched time.
-			e._dispatched = getCounter(true);
-			// When the _dispatched is smaller than the time when the targetted event handler was attached
+			// so we set the initial dispatched counter value.
+			e._dispatched = globalCounter;
+			// If the _dispatched is smaller than the counter when the targetted event handler was attached
 			// we know we have bubbled up to an element that was added during patching the dom.
-		} else if (e._dispatched <= eventHandler._attached) {
+		} else if (e._dispatched < eventHandler._attached) {
 			return;
 		}
 		return eventHandler(options.event ? options.event(e) : e);
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -13,14 +13,17 @@ function setStyle(style, key, value) {
 	}
 }
 
-// Force counters to 32-bit unsigned integer
-let globalCounter = 0 >>> 0;
-let cachedCounter = 0 >>> 0;
+// BigInt is used for the counter to avoid overflow.
+// It is supported by all major browsers except IE and Opera mini
+// where the plain number is used as fallback.
+// eslint-disable-next-line no-undef
+const zero = typeof window != 'undefined' && window.BigInt ? BigInt(0) : 0;
+let globalCounter = zero;
+let cachedCounter = zero;
 const p = Promise.resolve();
 const getCounter = () =>
 	cachedCounter ||
-	(p.then(() => (cachedCounter = 0 >>> 0)),
-	(cachedCounter = ++globalCounter >>> 0));
+	(p.then(() => (cachedCounter = zero)), (cachedCounter = ++globalCounter));
 
 /**
  * Set a property value on a DOM node
@@ -72,6 +75,14 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
 
 		if (value) {
 			if (!oldValue) {
+				// Note that the counter is incremented when an event listener is newly attached
+				// for the first time during a render cycle. All event listeners that are newly
+				// attached during that render cycle will share the same counter value.
+				// The value will be compared to the _dispatched added to the event itself
+				// and the event will propagate when _attached is less or equal to _dispatched.
+				// Potentially the event handler might be shared between multiple nodes, but
+				// as long as the order of _attached and _dispatched values is correct
+				// the event bubbling will work.
 				value._attached = getCounter();
 				const handler = useCapture ? eventProxyCapture : eventProxy;
 				dom.addEventListener(name, handler, useCapture);
--- a/src/internal.d.ts
+++ b/src/internal.d.ts
@@ -120,7 +120,7 @@ declare global {
 	}
 
 	export interface PreactEvent extends Event {
-		_dispatched?: number;
+		_dispatched?: number | bigint;
 	}
 
 	// We use the `current` property to differentiate between the two kinds of Refs so


src/diff/props.js
```diff
@@ -13,6 +13,11 @@ function setStyle(style, key, value) {
 	}
 }
 
+let globalCounter = Number.MIN_SAFE_INTEGER;
+const getCounter = increment => {
+	return increment ? ++globalCounter : globalCounter;
```
JoviDeCroock, line 18: I don't think it's safe to assume that the user has a certain amount of events, if we however memoize this like the example I gave in the issue that would become a lot safer as then the assumption becomes amount of re-renders instead of amount of events.

JoviDeCroock (commented):
Have you been able to test this with the issues that were closed by https://github.com/preactjs/preact/pull/4126

nbalatsk-oracle:
Yes, I tested issue [#3927](https://github.com/preactjs/preact/issues/3927) that was closed by #4126.

src/diff/props.js
ritschwumm: maybe force globalCounter to be a 32 bit int using `globalCounter | 0`? as far as i know javascript VMs, it would be a tad bit faster.

additionally it would at least lead to predictable overflows whereas a regular double would start doing interesting things at around 2^52.

nbalatsk-oracle: Take a look at the update. Used the counter [technique mentioned in the issue](https://github.com/vuejs/core/blob/main/packages/runtime-dom/src/modules/events.ts#L82-L83), forced counters to be unsigned int with  >>> 0.

ritschwumm: do i miss something that makes `>>> 0` better than `| 0`? it's longer and halves the available range...

nbalatsk-oracle: I agree that it is uglier, but if I understand it correctly the shift operator converts a number to an  unsigned 32 bit int with the range of [0 to 4294967295]. The bitwise OR converts a number to a signed 32 bit int with the range of [-2147483648 to 2147483647]. 
When we start the counter with 0 the unsigned int doubles the range. Right?
You can check it with code like this:
```
const x = 0b1111111111111111111111111111111 >>> 0; // 2^31 -1 as unsigned int32
const y = 0b1111111111111111111111111111111 | 0; // 2^31 -1 as signed int32
console.log(x, y);
console.log(x + 1 >>> 0, y + 1 | 0);
```
I've read about the signed/unsigned manipulation in [this article](https://james.darpinian.com/blog/integer-math-in-javascript). I wonder if we want to use those manipulations at all?

ritschwumm: i think i misunderstood what `>>>` does and unsigned should have exactly the same range size as signed (both are 32 bits in the end...). i definitely would use integers instead of floating point, it just feels a lot better :)

nbalatsk-oracle:
Hi guys, I've refreshed my branch with the latest code. Let me know if you have any comments.

src/diff/props.js
JoviDeCroock: The thing bothering me most about this approach is when it rolls over (rare case) we might have event-handlers registered with MAX_INT_32 and 0 which all of a sudden would incorrectly bail out of a bubbled click 😅 thinking about alternatives like registering event-handlers during the `commit` phase but not sure as of just yet

JoviDeCroock (approved):
The thing bothering me most about this approach is when it rolls over (rare case) we might have event-handlers registered with MAX_INT_32 and 0 which all of a sudden would incorrectly bail out of a bubbled click 😅 thinking about alternatives like registering event-handlers during the `commit` phase but not sure as of just yet because `commit` is still triggered synchronously so the `microTick` delay between bubbling could make the bug re-appear.

The initial solution we took was to delay state-setting by a tick rather than a microtick which would successfully make it trigger _after_ bubbling completes https://www.jovidecroock.com/blog/browser-timings however that had its own set of issues 

We can see this working in the original issue: https://codesandbox.io/p/sandbox/wandering-cdn-4s53z4?file=%2Fsrc%2Fpreact.js%3A726%2C1 Also has still fixed the original issue of the event timings https://codesandbox.io/p/sandbox/amazing-rhodes-xpj62v?file=%2Fsrc%2Findex.jsx%3A10%2C6 

WDYT @marvinhagemeister @developit @andrewiggins

sschultze:
@JoviDeCroock I have seen that you have gone for the 32-bit integer solution, as opposed to just counting from `Number.MIN_SAFE_INTEGER` upwards.

With 32-bit integers, there is a slight chance that it rolls over if many events are fired and if you have a system that runs for a long time (kiosk systems?). If I understand the code correctly, every event attached before MaxInt32 wouldn't be fired after rolling over.

Counting upwards from `Number.MIN_SAFE_INTEGER` however, you would practically never reach `Number.MAX_SAFE_INTEGER`. Even with 10,000 events a second, this would take 57,123 years.

JoviDeCroock:
@sschultze this isn't my PR just my comment 😅

src/diff/props.js
```diff
@@ -124,22 +133,22 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
...
-		if (!e._dispatched) {
-			// When an event has no _dispatched we know this is the first event-target in the chain
-			// so we set the initial dispatched time.
-			e._dispatched = Date.now();
-			// When the _dispatched is smaller than the time when the targetted event handler was attached
-			// we know we have bubbled up to an element that was added during patching the dom.
-		} else if (e._dispatched <= eventHandler._attached) {
-			return;
-		}
+  if (this._listeners) {
```
JoviDeCroock, line 136: indentation seems off

sschultze:
@JoviDeCroock Sorry 😅

sschultze:
@nbalatsk-oracle @JoviDeCroock What about using BigInt for the counter if the browser supports it? BigInts have an arbitrary size and will never overflow, and also no strange things happen if they get very large, giving some peace of mind. 

The performance cost should be negligible. On my computer, 10 million increments of a BigInt take 151ms.

Initialization for the counter could go like this:

`const counter = window.BigInt ? BigInt(0) : 0;`

The `++` operand and the comparison would work fine on both `number` and `bigint`, so no need for another case distinction.

nbalatsk-oracle:
@ritschwumm @JoviDeCroock do we want to use BigInt for the counter as suggested by @sschultze? Do you any concerns for performance?

src/diff/props.js
nbalatsk-oracle: @JoviDeCroock Fixed indentation

ritschwumm:
@nbalatsk-oracle can't the rollover be made negligible by just comparing for equality with the previous value? then the only problem left is when 2^32 events happen "at the same instant"...

sschultze:
Probably another issue, but isn't it problematic that, even in the original code, the `_attached` property is added to the event handler function? The same event handler might be added to multiple nodes, potentially leading to internal mess-up. Even I re-use event handlers in my code (e.g. for list items differentiated by a `data-key` attribute), and my UI is not that special.

I also noticed that in the minified version, `_attached` is shortened to `u`. Couldn't it happen that there's a conflict between the minified name and the "official" properties of the function?

jramanat-oracle:
> @nbalatsk-oracle can't the rollover be made negligible by just comparing for equality with the previous value? then the only problem left is when 2^32 events happen "at the same instant"...

@ritschwumm I don't think that's how the current proposed code changes are working.  The idea is that the counter is incremented when an event listener is newly attached for the first time during a render cycle.  All event listeners that are newly attached during that render cycle will share the same counter value; all event listeners attached during previous renders will have lower counter values, all event listeners attached in future renders will have higher counter values.  As a result, event listeners attached during the 0th render cycle appear indistinguishable from event listeners attached during the 2^32nd render cycle, hence the "rollover" issue.

nbalatsk-oracle:
> Probably another issue, but isn't it problematic that, even in the original code, the `_attached` property is added to the event handler function? The same event handler might be added to multiple nodes, potentially leading to internal mess-up. Even I re-use event handlers in my code (e.g. for list items differentiated by a `data-key` attribute), and my UI is not that special.

@sschultze Is it really a problem? It seems that it does not really matter that the handler is shared as long as the _attached/_dispatched order is correct. Even if _attached is updated by some DOM mutation (new elements are added that share the same handler), the event will not propagate to those updated elements. What is the use-case where we want it?  And if you trigger an event on one of those older elements the event should work because the order of the _attached/_dispatched pair is still correct. Right?

sschultze:
@nbalatsk-oracle A thought experiment: If it would indeed be 100% safe to share event handlers between nodes, and still always get the correct behavior here, then it would - as a consequence - also be safe to share the same one event handler for everything in the application. Meaning there's really only one event handler function for everything in the application in this theoretical scenario.

In this scenario, the `_attached` property would always be set in the same function object.

If this would still lead to the correct behavior, than `_attached` could be a (module-)global variable, and not be attached to the event handler functions in the first place.

What I want to say by this is:

- either event handlers can really be shared between nodes and still get 100% correct behavior, in which case `_attached` could also be made a (module-)global variable
- or neither the original nor the corrected workaround provide a 100% solution.

In any case, especially if we get the overflow problem away (e.g. with the `bigint` proposal), the new code definitely provides better results that the old code relying on `Date`. (The `Date`-based workaround not only makes problems in the test scenario, but also if the users changes their system date/time while the app is running.)

nbalatsk-oracle:
Updated the code to use BigInt for the counters to avoid overflow and added a few comments to _attached/_detached usage.

jramanat-oracle:
@JoviDeCroock @developit @andrewiggins @marvinhagemeister Just curious if you might be able to provide any indication on the likelihood of moving forward with the approach in this PR or if there are other approaches to this problem that you might like to see investigated?  Our product is currently blocked from moving forward from preact 10.17.1 due to https://github.com/preactjs/preact/issues/4161 so we're very interested in finding some acceptable resolution.  Happy to explore other options if there's a preference, thanks!

src/diff/props.js
```diff
@@ -13,6 +13,18 @@ function setStyle(style, key, value) {
 	}
 }
 
+// BigInt is used for the counter to avoid overflow.
+// It is supported by all major browsers except IE and Opera mini
+// where the plain number is used as fallback.
+// eslint-disable-next-line no-undef
+const zero = typeof window != 'undefined' && window.BigInt ? BigInt(0) : 0;
```
JoviDeCroock, line 20:
```suggestion
const zero = Number.MIN_SAFE_INTEGER;
```

src/internal.d.ts
```diff
@@ -120,7 +120,7 @@ declare global {
 	}
 
 	export interface PreactEvent extends Event {
-		_dispatched?: number;
+		_dispatched?: number | bigint;
```
JoviDeCroock, line 123:
```suggestion
		_dispatched?: number;
```

PR #4243, opened 2023-12-29
ensure we are able to support capture events from compat

### LINKED ISSUE #4237: [preact/compat] onTouchMoveCapture not processed

**Describe the bug**

I noticed the react-remove-scroll library is not working with preact/compat on a mobile device, because the event `onTouchMoveCapture` (and I think some others xxxCapture) is not processed at all.

**To Reproduce**

https://stackblitz.com/edit/create-preact-starter-qdpzqk?file=src%2Findex.jsx

Steps to reproduce the behavior: Open the stackblitz in a new window, then open Chrome Dev tools, then simulate a mobile device : no log is issued when I scroll the view. If I change to `onTouchMove`, it's working fine

Likely cause : the event method name is fully lowercased in [compat/src/render.js](https://github.com/preactjs/preact/blob/main/compat/src/render.js#L159), before the regex [testing if the event ends with Capture](https://github.com/preactjs/preact/blob/main/src/diff/props.js#L54), fails. (Because it's case sensitive, and `ontouchmovecapture` does not end with `Capture`)

JoviDeCroock:
Resolves #4237 

In compat we always lower-case our events, this prevents us from supporting events ending with `Capture` (which is our heuristic to make the events passive). We could stop doing this but the change would be a lot bigger as opposed to performing a case insensitive check.

One concern I have here would be custom-elements 😅

### DIFF
--- a/compat/test/browser/events.test.js
+++ b/compat/test/browser/events.test.js
@@ -2,7 +2,8 @@ import { render } from 'preact';
 import {
 	setupScratch,
 	teardown,
-	createEvent
+	createEvent,
+	supportsPassiveEvents
 } from '../../../test/_util/helpers';
 
 import React, { createElement } from 'preact/compat';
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
Nice catch!

PR #4244, opened 2023-12-29
chore: refocus element after a diff

### LINKED ISSUE #3242: Focus is lost during list re-order

**Describe the bug**
Focus is lost during list re-order.
When a list re-renders, if the focused item is moved to later in the source order, it loses focus.

**To Reproduce**
As shown in this [Code Sandbox](https://codesandbox.io/s/preact-focus-losing-bug-txtwj?file=/src/index.js)
If a list is re-ordered: focus is properly preserved when moving up in the source order, but not down in the source order.
(All items have keys that are stable and not indexed based)
preact-reorder-focus-bug

1. Go to [the Code Sandbox](https://codesandbox.io/s/preact-focus-losing-bug-txtwj?file=/src/index.js)
2. Click on an item in the list
3. Move the item up with the up arrow key, and observe the focus remains on the correct item.
3. Move the item down with the down arrow key, and observe the focus is lost.

**Expected behavior**
The item should maintain focus no matter which direction in the source order it moves.

**Similar issue**
This is similar to: [Input loses focus when conditionally rerendering other divs #540](https://github.com/preactjs/preact/issues/540).
However, since this issue is not about conditional rendering, only list rendering, and because that issue is closed whereas this issue effects the latest tested version (`10.5.14`), I believe this warrants it's own ticket.

### LINKED ISSUE #4235: Focus lost when moving child with `key`

## Describe the bug
Moving children with a `key` attribute can cause a focused element in the child DOM to lose focus.

## To Reproduce

The following app demonstrates the issue on https://preactjs.com/repl

```jsx
import { render } from 'preact';
import { useState, useEffect } from 'preact/hooks';

function Counter() {
	const [flipped, setFlipped] = useState(false);

	useEffect(() => {
		let id = setInterval(() => {
			setFlipped(Math.random() > .5)
		}, 1000)
		return () => clearInterval(id)
	})

	return flipped ? <>
		<div key="one"><input defaultValue="one" /></div>
		<div key="two"><input defaultValue="two" /></div>
	</> : <>
		<div key="two"><input defaultValue="two" /></div>
		<div key="one"><input defaultValue="one" /></div>
	</> 
}

render(<Counter />, document.getElementById('app'));
```

Click on one of the `<input>` fields and start typing. Notice that when the element gets moved, you might lose focus. (I noticed that it only happens when the node is moved *down* the DOM, not *up*. You can tell that the DOM node was not re-rendered: the `input` field keeps its value.

https://github.com/preactjs/preact/assets/6933510/f8bca58c-beec-482f-a07e-b39fe14858c0

## Expected behavior
The element should maintain focus. In `react-dom`, this works correctly. Try the following app in https://playcode.io/react

```jsx
import React, {useEffect, useState} from 'react';

export function App(props) {
	const [flipped, setFlipped] = useState(false);

	useEffect(() => {
		let id = setInterval(() => {
			setFlipped(Math.random() > .5)
		}, 1000)
		return () => clearInterval(id)
	})

	return flipped ? <>
		<div key="one"><input defaultValue="one" /></div>
		<div key="two"><input defaultValue="two" /></div>
	</> : <>
		<div key="two"><input defaultValue="two" /></div>
		<div key="one"><input defaultValue="one" /></div>
	</> 

}
```

https://github.com/preactjs/preact/assets/6933510/bf59bd99-2ee8-4609-a85f-88e9eed229d2

JoviDeCroock:
This adds back the focussing of elements, this now only happens after committing the root rather than performing this check on every phase of `child-diffing`. This PR mainly checks whether it's worth the byte-size and to start a discussion

Fixes #4235 
Fixes #3242


developit:
perf impact here is not worthwhile IMO, since it impacts performance on every diff regardless of whether focus needed to be restored.

I'd love to come up with a way to have the diff avoid losing focus in this case, but the problem I see is that it would always be specific to focus state. It would never fix things like `<video>` playback stopping or `<iframe>` being reloaded. The problem is that it's impossible to determine whether one DOM element is "more stateful" than any other DOM element.

Here's a particularly fun case where, no matter what we do, at least one of the iframes will be reloaded:

```jsx
function Swapper() {
  const [swapped, swap] = useReducer(s => !s, false);
  
  const a = <iframe key="a" src="/a.html" />;
  const b = <iframe key="b" src="/b.html" />;

  return <>
    <button onClick={swap}>Swap</button>
    {swapped ? [b, a] : [a, b]}
  </>;
}
```

JoviDeCroock:
This could be perf neutral but there are a lot of considerations like in-render blurring as well as cross-frame focus 😅 just wanted to check the perf after these changes

Issue #3515, opened 2022-04-10
type check doesn't run for styled-components


miyacoz:
- [ ] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
*before starting, i'm honestly not sure if this's a bug or it's me doing something wrong, especially about typescript settings with preact, so i want you to check the configure in the repro code below first.*

hi, i found tsc somehow didn't return a type error that i expected for components using `styled-components` in my private preact project.  
it checks non-`styled` components properly and returns a result as intended, also the issue doesn't happen in a react project with the same source (though the settings are kinda different as you see in my repro code).

i know preact isn't 100% compatible with react in typing as well, so am kinda reckoning it causes the problem but haven't dug down preact's .d.ts and else.

**To Reproduce**

here's the repro code: [miyacoz/preact-styled-repro](https://github.com/miyacoz/preact-styled-repro)  
please note there're `master` branch with preact where the problem is happening, and `react` branch with react where everything's fine (= tsc properly emitting type errors).

1. clone the repro code
2. `yarn start` then tsc starts
3. confirm tsc doesn't point out `<Styled>` having an error

also compare the result with react one (don't forget `yarn install` after checking out).

**Expected behavior**
tsc properly reports type errors.  
in my repro code, it points out type errors with both `<Styled>` and `<Component>`.

rschristian:
You have `"skipLibCheck"` enabled in your `tsconfig.json` which hides the fact that a type cannot be formed to check against.

If you remove the `"paths"` from your `tsconfig.json`, the correct types will be found to check against. Not sure if there was a reason you added them, but it looks like compat's types might not match up well.

rschristian:
> i've enabled skipLibCheck intentionally because some of the libraries i'm using emit errors like these

And that'd definitely be the right thing to do, however, `skipLibCheck` just ignores the errors in `node_modules/`; it doesn't create a working type from an error.

It seems that `preact/compat` doesn't include all the fields on React that you/libraries might want to use, so manually setting type paths causes an issue. These fields *can* be found in preact's typings however, and are automatically picked up upon otherwise.

> paths is required in my project because i'm using parcel as the bundler, and i set it up according to [your documentation](https://preactjs.com/guide/v10/getting-started/#aliasing-react-to-preact).

Not sure I follow? Our docs don't instruct to use `paths` in your `tsconfig.json` w/ parcel.

[...]

PR #4271, opened 2024-02-02
ts: improve preact compat types adding a few missing @types/react used by styled-components

jduthon:
Hello preact team!

Recently, we ran into some TypeScript issues in a project where we use preact with styled-components.

Looking a bit more in depth at the issues, they seemed to stem from a few missing types from @types/react, this PR aims to add those types in.

With those changes, we got a project with preact+styled-components working correctly (PS: this is a good first sign, but the project isn't that huge, so there could be things that don't work in 100% of the situations)

It would likely fix this [issue](https://github.com/preactjs/preact/issues/3515)

### DIFF
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -47,6 +47,7 @@ declare namespace React {
 	export import Component = preact.Component;
 	export import FunctionComponent = preact.FunctionComponent;
 	export import ComponentType = preact.ComponentType;
+	export import ComponentClass = preact.ComponentClass;
 	export import FC = preact.FunctionComponent;
 	export import createContext = preact.createContext;
 	export import createRef = preact.createRef;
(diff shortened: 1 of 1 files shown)


compat/src/index.d.ts
```diff
@@ -55,6 +56,8 @@ declare namespace React {
 	export import cloneElement = preact.cloneElement;
 	export import ComponentProps = preact.ComponentProps;
 	export import ReactNode = preact.ComponentChild;
+	export import ReactElement = preact.VNode;
```
jduthon, line 59: `VNode` seemed to be the closest thing in preact to ReactElement from what I could gather, but this might be wrong.

rschristian: Seems right, `VNode` looks to cover all the same properties (and a couple more).

rschristian (approved):

JoviDeCroock (approved):

compat/src/index.d.ts
```diff
@@ -169,6 +172,12 @@ declare namespace React {
 		| MutableRefObject<T | null>
 		| null;
 
+	export type ComponentPropsWithRef<
```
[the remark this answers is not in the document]
akornatskyy: this cause issue #4293.

Issue #4279, opened 2024-02-14
Problem with rendering child component on props changes


coder4u:
**Describe the bug**
There was a problem with the rendering of components when updating props.
If you pass the state of a parent component as a prop to a child component, and then update the state in the parent component, the child component is not rerendered.

**To Reproduce**
Preact - https://codepen.io/coder4u/pen/YzgdQEL

React without rerender issue - https://codepen.io/coder4u/pen/qBvLXOy

1. Go to pen.
2. Open console.
3. You will see that parent state not synced with child props in render methods.

**Expected behavior**
Child component must rerender when it props changed.

JoviDeCroock:
Looks like this got broken between 10.18.1 and 10.18.2

dmitrage:
Did some debugging, maybe it could help.

When `BrokenChild` is rendered for the first time its `vnode._original == 3` (++vnodeId).
After `setState` in `BrokenChild.componentDidUpdate` it [increments](https://github.com/preactjs/preact/blob/65310c6c6f8445b531c47b188431385a80a1d104/src/component.js#L130) to `4`. Note that global `vnodeId == 3` at this time.
When `App` component re-renders after `setTimeout` it creates a new vnode for `BrokenChild` with `_original == 4` (++vnodeId).
Then rendering bails out because `oldVNode._original == newVNode._original`.

Before the [mentioned change](https://github.com/preactjs/preact/pull/4171) `_original` was changed on a temporary copy of the oldVNode but now it is changed on the newVNode.

JoviDeCroock:
Yeah, I got to a similar conclusion, we might just need to change the `vnodeId` back after diffing 😅

PR #4280, opened 2024-02-15
fix: correctly restore _original

JoviDeCroock:
We have to restore `_original` to the `oldVNode` value else the counter in `createElement` gets outdated. We did this implicitly before https://github.com/preactjs/preact/pull/4171 by copying the oldVNode rather than the new one.

fixes https://github.com/preactjs/preact/issues/4279

### DIFF
--- a/test/browser/components.test.js
+++ b/test/browser/components.test.js
@@ -340,6 +340,50 @@ describe('Components', () => {
 			expect(scratch.innerHTML).to.equal('<p>B</p>');
 		});
 
+		it('should update children props correct', () => {
+			let update, update2;
+			class Counter extends Component {
+				constructor(props) {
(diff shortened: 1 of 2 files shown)


test/browser/components.test.js
```diff
@@ -340,6 +340,50 @@ describe('Components', () => {
 			expect(scratch.innerHTML).to.equal('<p>B</p>');
 		});
 
+		it('should update children props correct', () => {
```
JoviDeCroock, line 343:
```suggestion
		it('should update children props correctly in subsequent renders', () => {
```

marvinhagemeister (approved):

marvinhagemeister (approved):
Great find!

PR #4281, opened 2024-02-16
fix: address scenario where we would crash when replacing a matched vnode with null

### LINKED ISSUE #4194: 10.18.2 regression:  Cannot read properties of undefined

### LINKED ISSUE #4274: Components keep being added when updating state

I'm having a problem where a component is repeatedly added to `document.body`, even though it's only used once in the component. Below is a simplified version of the code, where `<B />` keeps getting added repeatedly. This issue doesn't happen with React. Am I missing something?

```typescript
import { render } from "preact"
import { useEffect, useState } from "preact/hooks"

const B = () => <div>B</div>

const Test = () => {
    const [state, setState] = useState(true)
    useEffect(() => {
        const timer = setInterval(() => {
            setState((s) => !s)
        }, 250)
        return () => { clearInterval(timer) }
    }, [])

    if (state) {
        return <div>
            <B />
            <div></div>
        </div>
    } else {
        const cond = false
        return <div>
            <div></div>
            {cond && <div></div>}
            <B />
            <div></div>
        </div>
    }
}

render(<Test />, document.body)
```

**To Reproduce**, you can copy-paste the above code to https://preactjs.com/repl to see that the text "B" is continuously added.

**Expected behavior**
Only a single instance of `<B />` should be shown.

Edit: I've further simplified the code.

JoviDeCroock:
Fixes #4274
Fixes #4194

This change adds one crucial change to our child-diffing, when an oldVNode has already been matched we won't unmount it

### DIFF
--- a/hooks/test/browser/combinations.test.js
+++ b/hooks/test/browser/combinations.test.js
@@ -477,4 +477,44 @@ describe('combinations', () => {
 			'anchor effect'
 		]);
 	});
+
+	it.only('should not crash', () => {
+		const B = () => <div>B</div>;
+
+		let update;
+		const Test = () => {
+			const [state, setState] = useState(true);
+			update = () => setState(s => !s);
+
+			if (state) {
+				return (
+					<div>
+						<B />
+						<div />
+					</div>
+				);
+			}
+			return (
+				<div>
+					<div />
+					{null}
+					<B />
+				</div>
+			);
+		};
+		render(<Test />, scratch);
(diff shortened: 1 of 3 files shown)


marvinhagemeister (approved):
Nice one!!

andrewiggins:
Oh, nice catch!

PR #4287, opened 2024-02-21
chore: add reconcile bug test

marvinhagemeister:
@squidfunk ran into this case and I think it might be related to one I'm running into with Fresh. With his reproduction case I was able to narrow it down further to a minimal test case. Haven't had the time to look into this yet.


JoviDeCroock (approved):

squidfunk:
Thanks for fixing this so quickly! ❤️

Issue #3899, opened 2023-02-16
[preact/compact] Suppport react-is


clyfish:
**Describe the feature you'd love to see**
Some libraries use `react-is` to check types of react element.
Will `@preact/compat` support `react-is`?
**Additional context (optional)**
https://www.npmjs.com/package/react-is
```
// for example
import { isMemo, isFragment } from 'react-is'
```

marvinhagemeister:
It's very unlikely that we'll ever support `react-is`. It relies on very react specific internals that no other third party library needs. If there is a need for something like `react-is` we'd rather add another entry to our package so that `react-is` can be aliased to that.

So far we haven't come across libraries checking for the type of react element. Can you share the ones you found?

clyfish:
@marvinhagemeister [antd](https://www.npmjs.com/package/antd) and [react-dnd](https://www.npmjs.com/package/react-dnd) depend on [react-is](https://www.npmjs.com/package/react-is), and [more](https://www.npmjs.com/browse/depended/react-is).
```
$ npm ls react-is
├─┬ antd@5.2.2
│ └─┬ rc-util@5.28.0
│   └── react-is@16.13.1
└─┬ react-dnd@16.0.1
  └─┬ hoist-non-react-statics@3.3.2
    └── react-is@16.13.1 deduped
```

ziongh:
Great that ```isFragment``` is implemented! But ```isMemo``` is not yet implemented, and that breaks the Antd (and other) packages.

JoviDeCroock:
`isMemo` has been added, going to close this PR now

PR #4302, opened 2024-03-13
feat: added isMemo to compact to allow compatibility with react-is dependant libraries

ziongh:
Add support for react-is: #3899


marvinhagemeister (approved):
Thanks 💯

Issue #4285, opened 2024-02-17
preact/compat's SetStateAction is different from React's SetStateAction


tobycm:
- [ ] Check if updating to the latest Preact version resolves the issue

**Describe the bug**

preact/compat's [SetStateAction](https://github.com/preactjs/preact/blob/main/hooks/src/index.d.ts#L5) is different from React's [SetStateAction](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/ts5.0/index.d.ts#L1727)

**Expected behavior**
It should be the same, unless the internal mechanics are different (unlikely because the transpiled JS works fine).

rschristian:
Can you clarify what the actual problem is? A type being different shouldn't be an issue on its own.

tobycm:
The type checking throws an error when it shouldn't because the transpiled code works fine.

rschristian:
But can you give an actual example of this? That's what I'm asking for.

There's many ways in which your code could work just fine while not adhering to correct types.

tobycm:
Example:

React: https://codesandbox.io/p/devbox/condescending-wilbur-9gf73z

Preact: https://codesandbox.io/p/devbox/k4mvdw

PR #4306, opened 2024-03-17
align state updater type

JoviDeCroock:
Fixes https://github.com/preactjs/preact/issues/4285


marvinhagemeister (approved):

rschristian (approved):

evtn:
A bit late to the party, but changing public type definitions within patch versions is not ideal, I just had spent an hour trying to figure out why did my project stopped passing typecheck on CI

PR #4310, opened 2024-03-18
Integrate the new benchmarks repo and update 

andrewiggins:
This PR integrates the preactjs/benchmarks repo as a git submodule for us to run locally.

I added some instructions for setting up the submodule in the CONTRIBUTING file, but they can probably be expanded.

### DIFF
--- a/.github/workflows/run-bench.yml
+++ b/.github/workflows/run-bench.yml
@@ -135,10 +135,15 @@ jobs:
         with:
           name: results
           path: benchmarks/out/results/${{ inputs.benchmark }}.json
+      - name: Calculate log artifact name
+        id: log-artifact-name
+        run: |
+          NAME=$(echo "${{ inputs.benchmark }}" | sed -r 's/[\/]+/_/g')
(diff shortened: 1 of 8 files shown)


.github/workflows/run-bench.yml
```diff
@@ -32,36 +32,90 @@ jobs:
...
+          submodules: 'recursive'
       - uses: actions/setup-node@v4
         with:
           node-version-file: 'package.json'
           cache: 'npm'
           cache-dependency-path: '**/package-lock.json'
+
+      # Setup pnpm
+      - name: Install pnpm
+        uses: pnpm/action-setup@v2
```
JoviDeCroock, line 45: Can we update this to latest, same with the cache one

JoviDeCroock (approved):

.github/workflows/run-bench.yml
andrewiggins: Definitely

marvinhagemeister (approved):

PR #4318, opened 2024-03-22
fix: check whether `oldDom` is present in the DOM

JoviDeCroock:
Fixes https://github.com/preactjs/preact/issues/4194

This was a bit of a tricky one and haven't been able to reproduce this in tests but the base case looks like the following

```
Suspense
  Component --> Fragment-like
    Component --> Fragment-like
      some-dom-node
  LazyComponent --> Fragment-like
    LazyComponent --> Fragment-like
      some-dom-node
```

The main issue is that in nested fragment-like children we can have a very deep chain of components where the last one eventually resolves to unmounting the children. This is problematic because we bubble back up to the `_parent` where the `_nextDom`/`oldDom` is already set but can be unmounted by a later `Fragment-like` child.

This effectively means that _nextDom on all parents is going to be invalid, we however can't course-correct as we've already higher-up in the [stack assigned oldDom](https://github.com/preactjs/preact/blob/main/src/diff/children.js#L61) which now in-turn can become invalid by the [first call to diff](https://github.com/preactjs/preact/blob/main/src/diff/children.js#L85).

My naive solution for this is to check `isConnected` and get a new DOM-Sibling, this re-checks the children and shouldn't come up with an old-node as we are post-diff. Open to any other solutions here.

For debugging I copied the sandbox attached in the issue in our `demo` and logged that way.

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -116,7 +116,7 @@ export function diffChildren(
 			childVNode._flags & INSERT_VNODE ||
 			oldVNode._children === childVNode._children
 		) {
-			if (!newDom && oldVNode._dom == oldDom) {
+			if (oldDom && !oldDom.isConnected) {
 				oldDom = getDomSibling(oldVNode);
 			}
 			oldDom = insert(childVNode, oldDom, parentDom);
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -116,6 +116,7 @@ export function diffChildren(
 			childVNode._flags & INSERT_VNODE ||
 			oldVNode._children === childVNode._children
 		) {
+			// @ts-expect-error olDom should be present on a DOM node
 			if (oldDom && !oldDom.isConnected) {
 				oldDom = getDomSibling(oldVNode);
 			}


marvinhagemeister (approved):

bagratinho:
@JoviDeCroock I'm experiencing the insertBefore error on the latest preact on a list that gets frequently changed. Unfortunately I cant provide a stable reproduction steps. 
I have seen your description of this fix and I'm curious what type of structures of component should I avoid to get a workaround on that error, particularly what you refer under Fragment-like?
Is having nested fragments inside my list item components can be the reason I have this error thrown sometimes?

JoviDeCroock:
@bagratinho I'd love to have a reproduction but yes, nested fragments or components that just return children are fragment-like.

andrewburgess:
I started seeing errors similar to what @bagratinho is mentioning as well, and seems to be related to Google Chrome's auto translation.

I put together a minimal reproduction that seems to trigger in the REPL env as well

```javascript
import { render, h } from 'preact';
import { useState, useEffect } from 'preact/hooks';

const Button = (props) => {
    return (
        <div>
            <p>hello {props.label} there</p>
        </div>
    );
};

function App() {
    const [, setValue] = useState(0);

    useEffect(() => {
        const interval = setInterval(() => setValue((v) => v + 1), 1000);

        return () => clearInterval(interval);
    }, []);

    return (
        <>
            <h1>Hello</h1>
            <Button label="Next" />
        </>
    );
}

render(<App />, document.getElementById('app'));
```
preact

I narrowed it down to this tag for when the issue started happening https://github.com/preactjs/preact/releases/tag/10.20.2 so that's why I'm assuming this commit is the root cause.

Let me know if you want me to migrate this to an issue for better tracking

JoviDeCroock:
Thank you both for the input, I will try and look at the translate example. If you find one without it as the first message here implies please share 🙏

JoviDeCroock:
@andrewburgess I would encourage you to try out #4409 if possible

andrewburgess:
@JoviDeCroock checked it out and seems to solve the issue, thanks for looking into it!

PR #4321, opened 2024-03-23
Some byte improvements

JoviDeCroock:


src/diff/props.js
```diff
@@ -13,6 +13,21 @@ function setStyle(style, key, value) {
 	}
 }
 
+const reserved = [
```
JoviDeCroock, line 16: removing this saves another 11 bytes

marvinhagemeister (approved):

Saschaym (commented):
src/diff/index.js

Saschaym (commented):
src/diff/index.js

PR #4322, opened 2024-03-23
fix: use a virtual clock instead of Date.now() for event dispatch times

jviide:
This is an alternative solution for issue #4161. It takes the general idea of #4232, but modifies the logic with an extra trick: odd counter values are reserved for `e._dispatched` and even values for `value._attached`. This way we can tell whether an event was dispatched between two handler attachments, and vice versa. This in turn allows incrementing the counter only when needed without scheduling a microtask for it.

Fixes https://github.com/preactjs/preact/issues/4161.

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -13,6 +13,23 @@ function setStyle(style, key, value) {
 	}
 }
 
+// A "virtual clock" to solve issues like https://github.com/preactjs/preact/issues/3927.
+// When the DOM performs an event it leaves micro-ticks in between bubbling up which means that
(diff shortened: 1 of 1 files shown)


JoviDeCroock (approved):
Checked both issues

- https://codesandbox.io/p/sandbox/amazing-rhodes-xpj62v?file=%2Fsrc%2Findex.jsx
- https://codesandbox.io/p/sandbox/interesting-wilson-qlrcqm?file=%2Fsrc%2Findex.js%3A1%2C34

Which works correctly

PR #4325, opened 2024-03-25
Use Set

shoonia:


src/diff/props.js
```diff
@@ -26,6 +26,21 @@ function setStyle(style, key, value) {
 // per second for over 280 years before the value reaches Number.MAX_SAFE_INTEGER (2**53 - 1).
 let eventClock = 0;
 
+let p = new Set([
```
JoviDeCroock, line 29: Sets aren't supported in IE11 which Preact X still supports

shoonia: Are there any plans to stop supporting IE11?

JoviDeCroock: That would require a new major version

Issue #4202, opened 2023-11-09
Do these types belong to HTMLAttributes?


iivvaannxx:
- [X] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
Looking at [these](https://github.com/preactjs/preact/blob/51771f7864d288c7f6bd00fd138cf5de59dd5404/src/jsx.d.ts#L2082) type declarations, I can see some attributes inside the `HTMLAttributes` interface that are commented as 'Standard HTML Attributes' but many of them aren't standard at all. 

In my particular case I spent a lot of time trying to setup a type with Typescript that used `HTMLAttributes` for a couple of elements (`HTMLInputElement` and `HTMLTextAreaElement`) which discriminated attributes from one or another depending on a prop value. I was expecting the `rows` attribute to be present only on the `HTMLTextAreaElement` and not in the `HTMLInputElement` because `rows` is a native element of a `textarea` and not of an `input`. I assumed this and slammed my head onto a wall when I figured out that the attribute `rows` wasn't being discriminated because it was present as a 'standard' element in the general `HTMLAttributes` interface.

Is it intended to be like this because something that I am missing? (I am not very familiar with preact). Every other major framework (including React, Astro and Svelte) don't include this 'standard' attribute.

![image](https://github.com/preactjs/preact/assets/47715589/a1142ce0-aa7e-4599-8bea-a09c0da05efa)

**To Reproduce**

I don't think an actual minimal repo isn't necessary in this case, but I have [this example](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgCQCoFkAyBBGMrABGArjAKYDOcAhlWlrvkaZQApRnUDG8AvnADMoEEHADkYDtxgB6LiLDUYYgFChIsRCgw48BEuSq1tDPc0MAlTjzj8hI8VJ6qVMAJ5gycCxADuFdmt4AF4TXSYDNicYAB56TABJADswUgBRABsyEDIkmAA+AG0xYX8xAF1XDy8ffytpOFD4xn0WCnqeOJ1k1JhM7NyC4tKKCpUVIA) of the Typescript Playground which tries to grab the 'rows' attribute from an input element using both the react typings and the preact ones. While the preact ones are resolved, the react ones are not (as I would expect).

**Expected behavior**

The 'rows' key of the `HTMLAttributes<HTMLInputElement>` should not be resolved (how it happens with the react types).

iivvaannxx:
[...]

PR #4333, opened 2024-04-01
Add missing PropsWithRef type to fix #4293

### LINKED ISSUE #4293: Typescript issues in 10.19.4+ with @mui/material
It works as expected with `preact` version `10.19.3`. The issues are reproducable with `10.19.4` - `10.19.6`.

### Issue 1

```tsx
<Card component="form" />
```

cause:

```
  Overload 1 of 2, '(props: never): Element | null', gave the following error.
    Type 'string' is not assignable to type 'never'.
  Overload 2 of 2, '(props: DefaultComponentProps<CardTypeMap<{}, "div">>): Element | null', gave the following error.
    Type '{ component: string; }' is not assignable to type 'IntrinsicAttributes & CardOwnProps & CommonProps & Omit<HTMLAttributes<HTMLDivElement>, "children" | ... 7 more ... | "raised">'.
      Property 'component' does not exist on type 'IntrinsicAttributes & CardOwnProps & CommonProps & Omit<HTMLAttributes<HTMLDivElement>, "children" | ... 7 more ... | "raised">'.ts(2769)
```

### Issue 2

another issue is related to `Button`:

```tsx
<Button>
  test
  <Icon />
</Button>
```

cause:

```
  Overload 2 of 3, '(props: { component: React.ElementType; } & ButtonOwnProps & Omit<ButtonBaseOwnProps, "classes"> & CommonProps & Omit<any, "className" | "style" | ... 23 more ... | "startIcon">): Element | null', gave the following error.
    Type 'Element' is not assignable to type 'string'.ts(2769)
```

### Issue 3

and another one:

```tsx
<Button>
  test
  {disabled && <Icon />}
</Button>
```

cause:

```
  Overload 2 of 3, '(props: { component: React.ElementType; } & ButtonOwnProps & Omit<ButtonBaseOwnProps, "classes"> & CommonProps & Omit<any, "className" | "style" | ... 23 more ... | "startIcon">): Element | null', gave the following error.
    Type 'false | Element | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.ts(2769)
```

ryanlchan:
#4271 introduces a bug where some types are incorrectly resolved.

Compared to the original types/react [implementation](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/41aa59f8e323a93b98ecea69fcf67a7a0a2cfa67/types/react/index.d.ts#L1643C5-L1645C43), #4271's is missing an enclosing PropsWithRef in the return statement. The problems appear to resolve after restoring that line.

Fixes #4293

### DIFF
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -178,7 +178,7 @@ declare namespace React {
 		C extends ComponentType<any> | keyof JSXInternal.IntrinsicElements
 	> = C extends new (props: infer P) => Component<any, any>
 		? PropsWithoutRef<P> & RefAttributes<InstanceType<C>>
-		: ComponentProps<C>;
+		: PropsWithRef<ComponentProps<C>>;
 
 	export function flushSync<R>(fn: () => R): R;
 	export function flushSync<A, R>(fn: (a: A) => R, a: A): R;


compat/src/index.d.ts
```diff
@@ -178,7 +178,7 @@ declare namespace React {
 		C extends ComponentType<any> | keyof JSXInternal.IntrinsicElements
 	> = C extends new (props: infer P) => Component<any, any>
 		? PropsWithoutRef<P> & RefAttributes<InstanceType<C>>
-		: ComponentProps<C>;
+		: PropsWithRef<ComponentProps<C>>;
```
rschristian, line 181: Bear with me as I'm doing this from my phone, but I don't see `PropsWithRef` defined anywhere?

ryanlchan: Ah - that's totally my miss. I'm still trying to figure out problems with other MUI components on another branch and didn't see that I'd missed copying the type here. Unfortunately, after copying, it doesn't appear to resolve the issue.  I'll keep digging. Sorry for that!

ryanlchan: After further investigation, I think this change + the types/react PropsWithRef definition does fix the original issue.

However, it also revealed a related problem: several MUI component props refer to React types that are not currently supported by compat (e.g. React.InputHTMLAttributes). The simple fix would be to alias those to HTMLAttributes, which captures all valid HTML attributes already (per #4202). A more React-y way would be to add React's stricter tag-based props there's quite a lot of them and it this seems like a big change in typing complexity over the current state.

@rschristian do you have a preference?

rschristian: Aliasing to the looser HTMLAttributes sounds fine for the time being, if you're willing.

Indeed, supporting all of those per-tag attributes would be a PITA, though (eventually) desirable I guess.

PR #4337, opened 2024-04-11
feat: optimise jsx runtime

JoviDeCroock:
Similar to React in v19 we can avoid cloning props

### DIFF
--- a/jsx-runtime/src/index.js
+++ b/jsx-runtime/src/index.js
@@ -30,14 +30,15 @@ function createVNode(type, props, key, isStaticChildren, __source, __self) {
 	// We'll want to preserve `ref` in props to get rid of the need for
 	// forwardRef components in the future, but that should happen via
 	// a separate PR.
-	let normalizedProps = {},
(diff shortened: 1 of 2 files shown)


jsx-runtime/src/index.js
```diff
@@ -27,17 +27,19 @@ const isArray = Array.isArray;
...
-		} else {
-			normalizedProps[i] = props[i];
+	if ((ref = normalizedProps.ref)) {
+		normalizedProps = {};
+		for (i in props) {
+			if (i != 'ref') {
+				normalizedProps[i] = props[i];
+			}
 		}
 	}
```
developit, line 44: I think this might be slightly faster:

```suggestion
	if ('ref' in normalizedProps) {
		normalizedProps = {};
		for (i in props) {
			if (i != 'ref') {
				ref = props[i];
			} else {
				normalizedProps[i] = props[i];
			}
		}
	}
```

developit (approved):

PR #4339, opened 2024-04-15
provide event handler types

JoviDeCroock:
This provides typings that are exported from `preact` that can be used to type functions that will be fed into event handlers.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -395,3 +395,28 @@ export interface Context<T> {
 export interface PreactContext<T> extends Context<T> {}
 
 export function createContext<T>(defaultValue: T): Context<T>;
+
+type EventHandler<E extends Event<any>> = (event: E) => void;
+export type ClipboardEventHandler<T = Element> = EventHandler<
+	ClipboardEvent<T>
+>;
+export type CompositionEventHandler<T = Element> = EventHandler<
+	CompositionEvent<T>
+>;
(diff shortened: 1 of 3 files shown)


src/index.d.ts
```diff
@@ -395,3 +395,28 @@ export interface Context<T> {
...
 export function createContext<T>(defaultValue: T): Context<T>;
+
+type EventHandler<E extends Event<any>> = (event: E) => void;
+export type ClipboardEventHandler<T = Element> = EventHandler<
+	ClipboardEvent<T>
+>;
+export type CompositionEventHandler<T = Element> = EventHandler<
+	CompositionEvent<T>
+>;
+export type DragEventHandler<T = Element> = EventHandler<DragEvent<T>>;
```
JoviDeCroock, line 406: Should we instead export them from JSX?

test/ts/dom-attributes-test.tsx
```diff
@@ -45,3 +45,20 @@ const booleanishTest = (
...
+		<div>
+			<input onInput={onChange} />
+			<input
+				onInput={e => {
+					return e.currentTarget.value;
+				}}
+			/>
+		</div>
+	);
+};
```
marvinhagemeister, line 64: Let's add the `onScroll` issue you mentioned as a test.

marvinhagemeister (approved):

JoviDeCroock:
I realised these are already in our jsx types 🤦

PR #4341, opened 2024-04-18
respect default value

### LINKED ISSUE #4340: `defaultValue` incorrectly (?) diffed against prerendered HTML

## Describe the bug

Rendering an input with `defaultValue` set runs into a bit of an issue with prerendered HTML as the HTML element will have `value` set (as there's no `defaultValue` attribute), but when Preact goes to diff it against JSX containing `<input defaultValue="..."/>`, Preact will clear out `value` and set `defaultValue`. If only `defaultValue` were set this wouldn't be an issue, but explicitly clearing out `value` means `defaultValue` doesn't do much of anything (to my knowledge).

Fiddled with it a bit, but haven't found a decently golfed solution yet. Need some way to skip setting `value` or set `value` & `defaultValue` when `defaultValue` is incoming (I think).

Bit of a weird situation that can be solved by using `value` tbh.

Related: https://github.com/denoland/fresh/issues/2406

## To Reproduce

[Reproduction](https://preactjs.com/repl?code=aW1wb3J0IHsgaHlkcmF0ZSwgcmVuZGVyIH0gZnJvbSAncHJlYWN0JwoKZnVuY3Rpb24gQXBwKCkgewogICAgcmV0dXJuIDxpbnB1dCBkZWZhdWx0VmFsdWU9ImJhciIgLz47Cn0KCmNvbnN0IGVudHJ5ID0gZG9jdW1lbnQuZ2V0RWxlbWVudEJ5SWQoJ2FwcCcpOwplbnRyeS5pbm5lckhUTUwgPSAnPGlucHV0IHZhbHVlPSJmb28iIC8%2BJwoKcmVuZGVyKDxBcHAgLz4sIGVudHJ5KTsKCmNvbnN0IGlucHV0ID0gZG9jdW1lbnQucXVlcnlTZWxlY3RvcignaW5wdXQnKTsKY29uc29sZS5sb2coaW5wdXQudmFsdWUpOwpjb25zb2xlLmxvZyhpbnB1dC5kZWZhdWx0VmFsdWUpOw%3D%3D)

1. Open repro
2. Notice input is empty
3. Check console to see that while the `defaultValue` prop is set, `value` is now an empty string.

## Expected behavior
Like if the component was rendered w/out pre-existing HTML, the `value` should be correctly set so the input contains the right content.

JoviDeCroock:
Fixes #4340 

We could also implement this for `defaultChecked`

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -446,8 +446,9 @@ function diffElementNodes(
 				if (
 					(i == 'value' && 'defaultValue' in newProps) ||
 					(i == 'checked' && 'defaultChecked' in newProps)
-				)
+				) {
 					continue;
+				}
 				setProperty(dom, i, null, value, isSvg);
 			}
 		}
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -525,7 +525,6 @@ function diffElementNodes(
 					// again, which triggers IE11 to re-evaluate the select value
 					(nodeType === 'option' && inputValue !== oldProps[i]))
 			) {
-				console.log('setting', i);
 				setProperty(dom, i, inputValue, oldProps[i], false);
 			}
 
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -443,7 +443,11 @@ function diffElementNodes(
 			} else if (i == 'dangerouslySetInnerHTML') {
 				oldHtml = value;
 			} else if (i !== 'key' && !(i in newProps)) {
-				if (i == 'value' && 'defaultValue' in newProps) continue;
+				if (
+					(i == 'value' && 'defaultValue' in newProps) ||
+					(i == 'checked' && 'defaultChecked' in newProps)
+				)
+					continue;
 				setProperty(dom, i, null, value, isSvg);
 			}
 		}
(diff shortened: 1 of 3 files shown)


src/diff/index.js
```diff
@@ -524,6 +525,7 @@ function diffElementNodes(
 					// again, which triggers IE11 to re-evaluate the select value
 					(nodeType === 'option' && inputValue !== oldProps[i]))
 			) {
+				console.log('setting', i);
```
JoviDeCroock, line 528:
```suggestion
```

test/browser/render.test.js
```diff
@@ -1,5 +1,5 @@
 import { setupRerender } from 'preact/test-utils';
-import { createElement, render, Component, options } from 'preact';
+import { createElement, render, Component, options, hydrate } from 'preact';
```
JoviDeCroock, line 2:
```suggestion
import { createElement, render, Component, options } from 'preact';
```

marvinhagemeister (approved):
Sweet 🎉

src/diff/index.js
```diff
@@ -443,7 +443,11 @@ function diffElementNodes(
 			} else if (i == 'dangerouslySetInnerHTML') {
 				oldHtml = value;
 			} else if (i !== 'key' && !(i in newProps)) {
-				if (i == 'value' && 'defaultValue' in newProps) continue;
+				if (
+					(i == 'value' && 'defaultValue' in newProps) ||
+					(i == 'checked' && 'defaultChecked' in newProps)
+				)
+					continue;
```
marvinhagemeister, line 450: tiny nit, let's add proper braces here. I'm fine with leaving them out when it's a single line condition like before

PR #4345, opened 2024-04-22
feat: Add compat/client types

### LINKED ISSUE #3845: Missing types for react-dom/client package in compat

**Describe the bug**
When we switched to Preact using compat, we got the following TS error:

`Could not find a declaration file for module 'react-dom/client'`. I've looked around but although this module does exist and everything functions like it should, I couldn't find type declarations for it.

I solved it for now by adding a custom declarations file with the proper types, but it would be nice if this is included in Preact itself.

**To Reproduce**

I can't get preact/compat to work in codesandbox unfortunately. Simplest reproduction steps would be to alias Preact in a simple React 18 TS app.

**Expected behavior**
I would have expected that there would be types for react-dom/client in preact/compat, where I at most would have to add another entry to the "paths" in my tsconfig.json

rschristian:
Closes #3845 (but for real this time)

Semi-related: is it intentional that Preact can take a [`ComponentChild`](https://github.com/preactjs/preact/blob/e2c8a4564ef75e544735a96f88c250002f99f1ff/src/index.d.ts#L301) as a render arg whilst compat can only take [`VNode<any>`](https://github.com/preactjs/preact/blob/e2c8a4564ef75e544735a96f88c250002f99f1ff/compat/src/index.d.ts#L94-L95)? Doesn't seem like there's an immediately obvious issue w/ compat rendering a string, etc., as it's passed straight through to the core `render()`, more or less.

### DIFF
--- a/compat/client.d.ts
+++ b/compat/client.d.ts
@@ -1,0 +1,11 @@
+import * as preact from '../src';
+
+export function createRoot(container: preact.ContainerNode): {
+	render(children: preact.VNode<any>): void;
+	unmount(): void;
+};
+
+export function hydrateRoot(
+	container: preact.ContainerNode,
+	children: preact.VNode<any>
+): typeof createRoot;
--- a/compat/package.json
+++ b/compat/package.json
@@ -25,6 +25,7 @@
       "require": "./dist/compat.js"
     },
     "./client": {
+      "types": "./client.d.ts",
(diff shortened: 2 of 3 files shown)


JoviDeCroock (approved):

JoviDeCroock:
You are right yes, that should be possible

PR #4349, opened 2024-04-26
feat: debug throw on too many rerenders

rschristian:
Rebase of @andrewiggins' work in #3965, and moves to `/debug` as mentioned.

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -247,11 +247,31 @@ export function initDebug() {
 		if (oldBeforeDiff) oldBeforeDiff(vnode);
 	};
 
+	let renderCount = 0;
+	let currentComponent;
 	options._render = vnode => {
 		if (oldRender) {
 			oldRender(vnode);
 		}
 		hooksAllowed = true;
+
+		const nextComponent = vnode._component;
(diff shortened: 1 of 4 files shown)


debug/src/debug.js
```diff
@@ -247,11 +247,31 @@ export function initDebug() {
...
+		if (nextComponent === currentComponent) {
+			renderCount++;
+		} else {
+			renderCount = 1;
+		}
+
+		if (renderCount >= 25) {
+			throw new Error(
+				`Too many re-renders. This is limited to prevent an infinite loop ` +
+					`which may lock up your browser. The component causing this is: ${getDisplayName(
```
rschristian, line 268: I added the `...which may lock up your browser.` bit just to emphasize this isn't something one could safely ignore, especially as we'll only address it in dev (for most users, anyhow. I doubt many are using `/debug` in prod).

JoviDeCroock (approved):

tcoz:
Note that this PR appears to break some render intense components such as the MUI-X Datagrid. Setting preact debug to false in vite.config is a workaround for now.

PR #4350, opened 2024-04-26
fix: Incorrect "missing `transform-jsx-source`" warning

### LINKED ISSUE #3818: @babel/plugin-transform-react-jsx-source warning appears regardless of configuration

**Describe the bug**
Throwing an error within a Preact application with `preact/debug` enabled always logs the following warning in the dev console:

```
Add @babel/plugin-transform-react-jsx-source to get a more detailed component stack. Note that you should not add it to production builds of your App for bundle size reasons.
```

Adding `@babel/plugin-transform-react-jsx-source` to my babel configuration does not stop the warning.

This warning is logged from [here](https://github.com/preactjs/preact/blob/master/debug/src/component-stack.js#L92-L94) in `component-stack.js`. https://github.com/preactjs/preact/blob/d4089df1263faab9b980a3493a4c7e986f254f8e/debug/src/component-stack.js#L90-L95

Looking at that file, I don't believe that `hasBabelPlugin` is ever set to `true` before this warning gets logged. Unfortunately, I don't know how to detect the presence or absence of a babel plugin, otherwise I'd suggest a patch!

**To Reproduce**

(I could not reproduce this in the provided codesandbox environment, maybe because it's using `react-error-overlay`?)

Example repo: https://github.com/imidom/preact-warning-example

1. Create a new Preact application using the latest version of `preact-cli`: `npx preact-cli create default lol2` (or following instructions for a webpack - babel build)
2. Put an undefined variable into `<App />` like the following:
   ```
   ...
   <Header />
	{eekAnUndefinedVariableThatWillCauseAnException}
   <Router>
   ...
   ```
3. Run the application: `npm run dev`
4. Load the application with your dev console open
image

**Expected behavior**
The error is logged without an accompanying warning about a missing babel plugin if the babel plugin has, in fact, been loaded. Though I was not able to see any difference in output with the plugin installed or removed.

rschristian:
Fixes #3818

The Preact-CLI usage from the issue is unfortunately borked and not fixable really: CLI has users export their root component which it (CLI) then renders using `h()` -- no JSX source info is going to be added to that. That creates a situation in which there might not have been any vnodes with `.__source` set to flow through before `getOwnerStack()` is called.

### DIFF
--- a/debug/src/component-stack.js
+++ b/debug/src/component-stack.js
@@ -58,7 +58,7 @@ export function getCurrentVNode() {
  * location of a component. In that case we just omit that, but we'll
  * print a helpful message to the console, notifying the user of it.
  */
-let shouldShowJsxSourceWarning = true;
+let showJsxSourcePluginWarning = true;
 
 /**
  * Check if a `vnode` is a possible owner.
@@ -87,12 +87,12 @@ export function getOwnerStack(vnode) {
 		const source = owner.__source;
 		if (source) {
 			acc += ` (at ${source.fileName}:${source.lineNumber})`;
-		} else if (shouldShowJsxSourceWarning) {
+		} else if (showJsxSourcePluginWarning) {
 			console.warn(
 				'Add @babel/plugin-transform-react-jsx-source to get a more detailed component stack. Note that you should not add it to production builds of your App for bundle size reasons.'
 			);
 		}
-		shouldShowJsxSourceWarning = false;
+		showJsxSourcePluginWarning = false;
 
 		return (acc += '\n');
 	}, '');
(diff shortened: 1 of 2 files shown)


debug/src/component-stack.js
```diff
@@ -87,12 +87,12 @@ export function getOwnerStack(vnode) {
...
 		if (source) {
 			acc += ` (at ${source.fileName}:${source.lineNumber})`;
-		} else if (!hasBabelPlugin) {
-			hasBabelPlugin = true;
+		} else if (showJsxSourcePluginWarning) {
 			console.warn(
 				'Add @babel/plugin-transform-react-jsx-source to get a more detailed component stack. Note that you should not add it to production builds of your App for bundle size reasons.'
 			);
 		}
+		showJsxSourcePluginWarning = false;
```
rschristian, line 95: The idea being that at this point we've either seen `.__source` or printed the warning, and in either case, we're now covered. `hasBabelPlugin` previously would only ever be set to `true` once the warning was shown.

JoviDeCroock (approved):

PR #4364, opened 2024-04-30
feat: Support MathML namespace

### LINKED ISSUE #4363: Incorrect namespacing of MathML elements

**Describe the bug**
Preact doesn't correctly namespace MathML elements.

**To Reproduce**

1. Go to https://preactjs.com/repl?code=aW1wb3J0IHugcmVuZGVyIH0gZnJvbSAncHJlYWN0JzsKaW1wb3J0IHsgdXNlU3RhdGUgfSBmcm9tICdwcmVhY3QvaG9va3MnOwoKZnVuY3Rpb24gQ291bnRlcigpIHsKCXJldHVybiAoCgkJPD4KCTxtYXRoPjxtcm93PjEyMzwvbXJvdz48L21hdGg%2BCgk8c3ZnPjx0ZXh0PjEyMzwvdGV4dD48L3N2Zz4KCQk8Lz4KCSk7Cn0KCnJlbmRlcig8Q291bnRlciAvPiwgZG9jdW1lbnQuZ2V0RWxlbWVudEJ5SWQoJ2FwcCcpKTsK
2. In dev tools select the `<math>` element, then run `$0.namespaceURI` and you will see `http://www.w3.org/1999/xhtml`
3. If you then select the `<svg>` element, and repeat, you will correctly see `http://www.w3.org/2000/svg`

**Expected behavior**

I would expect Preact to handle MathML element namespacing correctly.

See https://w3c.github.io/mathml-core/ for the latest MathML spec.

rschristian:
Closes #4363

~~Probably needs a proper refactor of `isSvg` to a `namespace` enum, but I need to play around with this a bit to see what we can skip & what we need.~~ Done

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -136,12 +136,7 @@ function renderComponent(component) {
 			newVNode,
 			oldVNode,
 			component._globalContext,
-			component._parentDom.namespaceURI == 'http://www.w3.org/2000/svg'
-				? 2
-				: component._parentDom.namespaceURI ==
-				  'http://www.w3.org/1998/Math/MathML'
-				? 3
-				: 1,
+			component._parentDom.namespaceURI,
 			oldVNode._flags & MODE_HYDRATE ? [oldDom] : null,
 			commitQueue,
 			oldDom == null ? getDomSibling(oldVNode) : oldDom,
--- a/src/component.js
+++ b/src/component.js
@@ -136,7 +136,11 @@ function renderComponent(component) {
 			newVNode,
 			oldVNode,
 			component._globalContext,
-			component._parentDom.ownerSVGElement !== undefined ? 2 : 1,
+			component._parentDom.namespaceURI == 'http://www.w3.org/1999/xhtml'
+				? 1
+				: component._parentDom.namespaceURI == 'http://www.w3.org/2000/svg'
+				? 2
+				: 3,
 			oldVNode._flags & MODE_HYDRATE ? [oldDom] : null,
 			commitQueue,
 			oldDom == null ? getDomSibling(oldVNode) : oldDom,
--- a/src/component.js
+++ b/src/component.js
@@ -136,11 +136,12 @@ function renderComponent(component) {
 			newVNode,
 			oldVNode,
 			component._globalContext,
-			component._parentDom.namespaceURI == 'http://www.w3.org/1999/xhtml'
-				? 1
-				: component._parentDom.namespaceURI == 'http://www.w3.org/2000/svg'
+			component._parentDom.namespaceURI == 'http://www.w3.org/2000/svg'
 				? 2
-				: 3,
+				: component._parentDom.namespaceURI ==
+				  'http://www.w3.org/1998/Math/MathML'
+				? 3
+				: 1,
 			oldVNode._flags & MODE_HYDRATE ? [oldDom] : null,
 			commitQueue,
 			oldDom == null ? getDomSibling(oldVNode) : oldDom,
--- a/src/component.js
+++ b/src/component.js
@@ -136,7 +136,7 @@ function renderComponent(component) {
 			newVNode,
 			oldVNode,
 			component._globalContext,
-			component._parentDom.ownerSVGElement !== undefined,
+			component._parentDom.ownerSVGElement !== undefined ? 2 : 1,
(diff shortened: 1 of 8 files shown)


src/component.js
```diff
@@ -136,7 +136,7 @@ function renderComponent(component) {
 			newVNode,
 			oldVNode,
 			component._globalContext,
-			component._parentDom.ownerSVGElement !== undefined,
+			component._parentDom.ownerSVGElement !== undefined ? 2 : 1,
```
rschristian, line 139: Earlier MathML specs had a `.ownerMathElement` API that seems to have disappeared... Need to look into this more.

JoviDeCroock: Should exist according to https://www.w3.org/Math/DOM/mathml2/mathml-dom_ecma.html

rschristian: That's MathML 2, which was replaced with [MathML Core](https://www.w3.org/TR/mathml-core/)

rschristian: Testing `.namespaceURI` or `instanceof MathMLElement` is all we get at the moment.

rschristian:
This seems to be the best mix of size & perf I could come up with

lukewarlow (commented):
I'm not super familiar with Preact internals but this change looks good overall to me. It touches the various places I found when reporting the issue.

Perhaps some tests that handle conditional rendering would ensure context isn't lost in those cases?

rschristian:
Yeah that's fair -- will try to add later (for SVGs too)

src/component.js
```diff
@@ -136,7 +136,12 @@ function renderComponent(component) {
 			newVNode,
 			oldVNode,
 			component._globalContext,
-			component._parentDom.ownerSVGElement !== undefined,
+			component._parentDom.namespaceURI == 'http://www.w3.org/2000/svg'
```
JoviDeCroock, line 139: wouldn't it be more size efficient to use the old check here, embedding these strings can be costly on bytes

src/render.js
```diff
@@ -41,7 +41,11 @@ export function render(vnode, parentDom, replaceNode) {
 		vnode,
 		oldVNode || EMPTY_OBJ,
 		EMPTY_OBJ,
-		parentDom.ownerSVGElement !== undefined,
+		parentDom.namespaceURI == 'http://www.w3.org/2000/svg'
```
JoviDeCroock, line 44: Same remark here about using the old check for svg

JoviDeCroock (approved):
Looks great, thank you for tackling this! I left two size-nits

src/component.js
rschristian: I don't think so, as we already have the string for `createElementNS()` (and it's unavoidable there)

rschristian: Using only `.namespaceURI` cuts down on a few bytes too.

src/render.js
rschristian, line 44: Note for myself for tomorrow: double check we can't just pass the namespace through here, the enum might be unnecessary? Could maybe merge the `createElement` and `createElementNS` calls then (assuming the latter doesn't perform worse for HTML elements)?

src/diff/props.js
```diff
@@ -98,7 +98,7 @@ export function setProperty(dom, name, value, oldValue, isSvg) {
 			);
 		}
 	} else {
-		if (isSvg) {
+		if (namespace == 'http://www.w3.org/2000/svg') {
```
lukewarlow, line 101: Is this intentionally a double equals?

rschristian: Shaves 1b from `preact.module.js.br` :p

Should be safe

src/diff/index.js
```diff
@@ -375,8 +375,11 @@ function diffElementNodes(
 	let inputValue;
 	let checked;
 
-	// Tracks entering and exiting SVG namespace when descending through the tree.
-	if (nodeType === 'svg') isSvg = true;
+	// Tracks entering and exiting namespaces when descending through the tree.
+	if (nodeType === 'svg') namespace = 'http://www.w3.org/2000/svg';
+	else if (nodeType === 'math')
+		namespace = 'http://www.w3.org/1998/Math/MathML';
+	else if (!namespace) namespace = 'http://www.w3.org/1999/xhtml';
```
rschristian, line 382: Switching this to a ternary shaves off -6b or so but performs a fair bit worse.

PR #4366, opened 2024-04-30
perf(core): test perf of skipping some lifecycle hooks for perf

JoviDeCroock:
The main thinking behind this is that we are accessing the `_component` megamorphic object less resulting in better performance.

This leaves four accessors behind

- `contextType` - our provider is a functional component
- `shouldComponentUpdate` - our provider component uses this
- `componentWillUnmount` - our provider component uses this
- `componentDidUpdate` - this is used in the `useState` hook


developit (approved):
nice find! this also more closely matched what we have in 11 IIRC

PR #4372, opened 2024-05-05
fix: Syntax in `compat/client` for IE11 support

### LINKED ISSUE #4371: `createRoot` uses an unsupported syntax for IE11
- [ X ] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
The `createRoot` uses the "method definition" syntax which is [not supported in IE11](https://caniuse.com/mdn-javascript_functions_method_definitions):
https://github.com/preactjs/preact/blob/d18ac9f697e3604495d6496e4973c0af555ae7ba/compat/client.js#L3-L12

rschristian:
Closes #4371

Not that this case is problematic, but might be worth considering passing all package exports through Microbundle in the future to catch this sort of thing.

### DIFF
--- a/compat/client.js
+++ b/compat/client.js
@@ -2,10 +2,10 @@ const { render, hydrate, unmountComponentAtNode } = require('preact/compat');
 
 function createRoot(container) {
 	return {
-		render(children) {
+		render: function (children) {
 			render(children, container);
 		},
-		unmount() {
+		unmount: function () {
 			unmountComponentAtNode(container);
 		}
 	};
--- a/compat/client.js
+++ b/compat/client.js
@@ -2,9 +2,11 @@ const { render, hydrate, unmountComponentAtNode } = require('preact/compat');
 
 function createRoot(container) {
 	return {
+		// eslint-disable-next-line
 		render: function (children) {
 			render(children, container);
 		},
+		// eslint-disable-next-line
 		unmount: function () {
 			unmountComponentAtNode(container);
 		}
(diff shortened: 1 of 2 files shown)


compat/client.js
```diff
@@ -2,10 +2,10 @@ const { render, hydrate, unmountComponentAtNode } = require('preact/compat');
 
 function createRoot(container) {
 	return {
-		render(children) {
+		render: function (children) {
```
rschristian, line 5:
```suggestion
		// eslint-disable-next-line
		render: function (children) {
```

compat/client.mjs
```diff
@@ -2,10 +2,10 @@ import { render, hydrate, unmountComponentAtNode } from 'preact/compat';
 
 export function createRoot(container) {
 	return {
-		render(children) {
+		render: function (children) {
```
rschristian, line 5:
```suggestion
		// eslint-disable-next-line
		render: function (children) {
```

compat/client.js
```diff
@@ -2,10 +2,10 @@ const { render, hydrate, unmountComponentAtNode } = require('preact/compat');
...
 			render(children, container);
 		},
-		unmount() {
+		unmount: function () {
```
rschristian, line 8:
```suggestion
		// eslint-disable-next-line
		unmount: function () {
```

compat/client.mjs
```diff
@@ -2,10 +2,10 @@ import { render, hydrate, unmountComponentAtNode } from 'preact/compat';
...
 			render(children, container);
 		},
-		unmount() {
+		unmount: function () {
```
rschristian, line 8:
```suggestion
		// eslint-disable-next-line
		unmount: function () {
```

marvinhagemeister (approved):

PR #4378, opened 2024-05-10
feat: Add popover types

rschristian:
Reference:

* [popover](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover)
* [popovertarget](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#popovertarget)
* [popovertargetaction](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#popovertargetaction)

---

`popover="hint"` hasn't yet [landed in the spec](https://github.com/whatwg/html/pull/9778) and as such is perhaps questionable. It seems likely that it will and I figure jumping the gun on implementation might be slightly better than needing to be pinged to add it later (as it's not a spec I'm watching), but happy to remove it if it's undesirable.

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -2333,6 +2333,27 @@ export namespace JSXInternal {
 		placeholder?: string | undefined | SignalLike<string | undefined>;
 		playsInline?: boolean | undefined | SignalLike<boolean | undefined>;
 		playsinline?: boolean | undefined | SignalLike<boolean | undefined>;
+		popover?:
+			| 'auto'
+			| 'hint'
+			| 'manual'
+			| boolean
+			| undefined
+			| SignalLike<'auto' | 'hint' | 'manual' | boolean | undefined>;
+		popovertarget?: string | undefined | SignalLike<string | undefined>;
+		popoverTarget?: string | undefined | SignalLike<string | undefined>;
+		popovertargetaction?:
+			| 'hide'
+			| 'show'
+			| 'toggle'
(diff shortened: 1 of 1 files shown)


rschristian:
One sec, I need to check the camel cased versions, that might be no good as the JS property is slightly different

src/jsx.d.ts
```diff
@@ -2333,6 +2333,27 @@ export namespace JSXInternal {
...
+			| SignalLike<'auto' | 'hint' | 'manual' | boolean | undefined>;
+		popovertarget?: string | undefined | SignalLike<string | undefined>;
+		popoverTarget?: string | undefined | SignalLike<string | undefined>;
+		popovertargetaction?:
+			| 'hide'
+			| 'show'
+			| 'toggle'
+			| undefined
+			| SignalLike<'hide' | 'show' | 'toggle' | undefined>;
+		popoverTargetAction?:
```
lukewarlow, line 2351: This will set the property whereas popoverTarget will set the attribute due to the fact the property is actually popoverTargetElement (and takes an element reference rather than an idref string). 

This mismatch might be confusing to people?

lukewarlow: I've just remembered I think the property for action reflects to the content attribute so it might be fine?

rschristian: Yep, should actually be okay because of that.

JoviDeCroock (approved):

PR #4386, opened 2024-05-15
chore: migrate to biome

JoviDeCroock:
This is an initial attempt of migrating us to biome, a lot of formatting changes will happen because this will use the prettier v3 default settings 😅 wanted to see whether folks are interested in this, we can now also run `biome migrate eslint` or use oxlint for that.


marvinhagemeister (approved):

g-plane:
Hey, Biome doesn't support HTML and CSS/SCSS at this moment. Would you like to migrate to dprint? We can use dprint with [dprint-plugin-biome](https://github.com/dprint/dprint-plugin-biome) for JavaScript, [Malva](https://github.com/g-plane/malva) for CSS/SCSS and [markup_fmt](https://github.com/g-plane/markup_fmt).

PR #4390, opened 2024-05-19
chore: migrate husky v9

castrogarciajs:
Hello preact team, 

I took the trouble to update the popular pre-commit library `husky` in preact to its latest version in order to keep preact up to date with the latest releases, thanks for reading.


.husky/pre-commit
```diff
@@ -0,0 +1 @@
+npx lint-staged
```
rschristian, line 1: Have you set this to be executable? I know that's a common issue from other repos.

rschristian (approved):

.husky/pre-commit
castrogarciajs: Of course, Husky hooks are prepared to avoid permitting issues.

image

rschristian: I'd rather you just made the file executable. `prepare` is going to run extremely often and we want to minimize what's in it.

Issue #4391, opened 2024-05-20
Boolean values not handled as expected for `popover` prop


robertknight:
**Describe the bug**

The latest release adds the `popover` attribute with `boolean` as a supported type (see https://github.com/preactjs/preact/pull/4378). When setting this prop to a boolean value, I'd expect it to add or remove the `popover` attribute, but not set a value, like eg. the `disabled` prop. Instead it adds `popover="true"` or `popover="false"`, stringifying the boolean value, like `aria-` attributes. This results in a warning from Chrome that the value is not allowed.

A workaround is to use the values `"auto"` and `undefined` to substitute for boolean true/false.

Tested with Preact v10.22.0.

**To Reproduce**

test.html:

```html
<html>
  <body>
    <div id="app">
    <script type="module">
      import { render, h } from 'https://unpkg.com/preact?module';
      function App() {
        return h('div', { popover: true });
      }
      render(h(App), document.getElementById('app'));
    </script>
  </body>
</html>
```

1. Open test.html in Chrome
2. Observe warning in the logs: `Found a 'popover' attribute with an invalid value. <div popover=​"true">​</div>​`

**Expected behavior**

The above should render `<div popup></div>`.

PR #4393, opened 2024-05-21
fix(core): support popover boolean attribute

JoviDeCroock:
Fixes https://github.com/preactjs/preact/issues/4391

When doing tests it looks like `setAttribute('popover', '')` is the only one that results in proper behavior here. We should double check what we do in RTS for this as well

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -116,6 +116,7 @@ export function setProperty(dom, name, value, oldValue, namespace) {
 			name != 'rowSpan' &&
 			name != 'colSpan' &&
 			name != 'role' &&
+			name != 'popover' &&
 			name in dom
 		) {
 			try {
@@ -135,7 +136,7 @@ export function setProperty(dom, name, value, oldValue, namespace) {
 		if (typeof value == 'function') {
 			// never serialize functions as attribute values
 		} else if (value != null && (value !== false || name[4] === '-')) {
-			dom.setAttribute(name, value);
+			dom.setAttribute(name, name === 'popover' && value === true ? '' : value);
 		} else {
 			dom.removeAttribute(name);
 		}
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):

rschristian (approved):

rschristian:
RTS looks correct, as it matches other boolean attributes for the most part.

Not sure if it's smaller/faster, but we can set the prop, but we'd need to coerce it to the right value first, as attribute reflection uses the provided prop value (whereas the DOM property will coerce the value to auto/manual internally itself -- shame attribute reflection doesn't use the result of this coercion)

test/browser/render.test.js
```diff
@@ -472,6 +472,21 @@ describe('render()', () => {
...
+		render(<div popover="auto" />, scratch);
+		expect(scratch.innerHTML).to.equal("<div popover=\"auto\"></div>");
+	});
+
+	it('should support popover true boolean', () => {
+		render(<div popover />, scratch);
+		expect(scratch.innerHTML).to.equal("<div popover=\"\"></div>");
+	});
+
+	it('should support popover true boolean', () => {
```
rschristian, line 485:
```suggestion
	it('should support popover false boolean', () => {
```

src/diff/props.js
```diff
@@ -135,7 +136,7 @@ export function setProperty(dom, name, value, oldValue, namespace) {
 		if (typeof value == 'function') {
 			// never serialize functions as attribute values
 		} else if (value != null && (value !== false || name[4] === '-')) {
-			dom.setAttribute(name, value);
+			dom.setAttribute(name, name === 'popover' && value === true ? '' : value);
```
rschristian, line 139:
```suggestion
			dom.setAttribute(name, name == 'popover' && value == true ? '' : value);
```

Will save 2-6b and is pretty safe/reasonable?

robertknight:
Thanks for the quick turnaround @JoviDeCroock!

I wonder if there is a more general way to test whether attributes behave like this?

PR #4395, opened 2024-05-22
chore(compat): Expose stream render from `preact-render-to-string`

Austaras:


compat/server.js
```diff
@@ -9,7 +9,18 @@ try {
 	);
 }
 
+var renderToPipeableStream;
+try {
+	const mod = require('preact-render-to-string/stream-node');
+	renderToPipeableStream = mod.default || mod.renderToPipeableStream || mod;
+} catch (e) {
+	throw Error(
+		'renderToPipeableStream() error: using outdated version of "preact-render-to-string" dependency.'
```
rschristian, line 18:
```suggestion
		'renderToPipeableStream() error: update "preact-render-to-string" dependency to at least 6.5.0.'
```

PR #4403, opened 2024-05-31
compat: Improve React compatibility for `Ref` type.

maxbrieiev:
In React, the `Ref` type corresponds to the `Ref` type in preact core. However, `preact/compat` re-exports the `Ref` type from `preact/hooks` instead of from core.

The following snippet does not produce a type error in React, but it does when using preact:

```typescript
export const baz = (
  ref: React.ForwardedRef<HTMLElement>
): React.Ref<HTMLElement> => ref;
```

In preact it produces`Type 'ForwardedRef<HTMLElement>' is not assignable to type 'Ref<HTMLElement>'.`

This pull requests fixes this issue.

### DIFF
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -18,7 +18,6 @@ declare namespace React {
 	export import PropRef = _hooks.PropRef;
 	export import Reducer = _hooks.Reducer;
 	export import Dispatch = _hooks.Dispatch;
-	export import Ref = _hooks.Ref;
 	export import SetStateAction = _hooks.StateUpdater;
 	export import useCallback = _hooks.useCallback;
 	export import useContext = _hooks.useContext;
@@ -50,6 +49,7 @@ declare namespace React {
 	export import ComponentClass = preact.ComponentClass;
 	export import FC = preact.FunctionComponent;
 	export import createContext = preact.createContext;
+	export import Ref = preact.Ref;
 	export import createRef = preact.createRef;
 	export import Fragment = preact.Fragment;
 	export import createElement = preact.createElement;
--- a/compat/test/ts/forward-ref.tsx
+++ b/compat/test/ts/forward-ref.tsx
@@ -24,3 +24,7 @@ export const Bar = React.forwardRef<HTMLDivElement, { children: any }>(
 		return <div ref={ref}>{props.children}</div>;
 	}
 );
+
+export const baz = (
+	ref: React.ForwardedRef<HTMLElement>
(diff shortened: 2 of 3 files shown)


compat/src/index.d.ts
```diff
@@ -18,7 +18,6 @@ declare namespace React {
 	export import PropRef = _hooks.PropRef;
 	export import Reducer = _hooks.Reducer;
 	export import Dispatch = _hooks.Dispatch;
-	export import Ref = _hooks.Ref;
```
rschristian, line 21: It's ok if you want to call it out-of-scope for this PR, but should `_hooks.Ref` even exist? Should hooks also pull the `Ref` type from core?

maxbrieiev: The type that corresponds to `_hooks.Ref` is called [`RefObject` in React](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6c7875bb2df2e55876097b8aa15c2aa6644641f0/types/react/index.d.ts#L150).

preact also defines `RefObject` type in core, which is similar to both `_hooks.Ref` and React's `RefObject`. So I can assume that hooks should pull `RefObject` type from core.

rschristian: Yes, that was what I was asking. Seems like the root issue is a diverging type in `/hooks` rather than anything here.

maxbrieiev: So is there anything left to be done in this pull request, or is it good to go?

rschristian: Well ideally we'd fix & unify all ref types, as I'd guess `PropRef` might cause similar problems, but if you don't want to fix it I suppose it can be done separately

maxbrieiev: No problem, I will look deeper into how we can unify these types. I just wanted to clarify the scope of this issue.

maxbrieiev: > Well ideally we'd fix & unify all ref types

I have tried to address the usage of a `Ref` type in hooks. Please, check.

> I'd guess `PropRef` might cause similar problems

`PropRef` is marked as deprecated, and there are no usages of it, except in one test.

rschristian: > PropRef is marked as deprecated, and there are no usages of it, except in one test.

Moreso that it's a rebranding of `MutableRef`, which, to my knowledge, shouldn't really exist either?

But whatever, doesn't matter much.

rschristian (approved):
Cheers, thanks for going beyond too! It'll hopefully correct similar mismatches in the future.

PR #4407, opened 2024-06-13
feat: preserve comments for hydration to support OOO streaming of Suspense boundaries

jacob-ebey:


jacob-ebey:
NVM, this doesn't address the core issue.

PR #4414, opened 2024-06-19
test performance of reducing prototype access

JoviDeCroock:
This tests whether we can give a minor performance boost with a size tradeoff by reducing prototype access, this currently breaks tests due to us not registering the calls anymore.

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -2,7 +2,7 @@ import { EMPTY_OBJ } from './constants';
 import { commitRoot, diff } from './diff/index';
 import { createElement, Fragment } from './create-element';
 import options from './options';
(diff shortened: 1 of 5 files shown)


src/render.js
```diff
@@ -12,6 +12,7 @@ import { slice } from './util';
  * existing DOM tree rooted at `replaceNode`
  */
 export function render(vnode, parentDom, replaceNode) {
+	setup();
```
JoviDeCroock, line 15: This would be needed to avoid us reaching into globals while we renderToStream or renderToString

JoviDeCroock:
Looks to be negligible

Issue #4416, opened 2024-06-20
we need syntax for choosing setting a property vs attribute on elements


trusktr:
- [ ] Check if updating to the latest Preact version resolves the issue

**Describe the bug**

It is not possible to guess 100% of the time whether an element user needs to set a property, or an attribute.

Does this set an attribute, or a property?

```jsx
return <some-element foo={[1, 2, 3]} />
```

Answer: it depends, sometimes Preact sets the `.foo` property, sometimes Preact sets the `foo` attribute. The user has no control over this.

The implementation of an element can vary (and may be out of the Preact user's control):

- some elements handle properties only
- some elements handle attributes only
- some elements handle both
- even if an element handles one or the other, or both
  - a user might want to set the attribute
  - or the property
  - or both
- user might want to add or remove an attribute based on a boolean (regardless if they also want to set the corresponding JS property or not)

It is not possible for Preact to always guess correctly which of the above cases the user will want, and the Preact user needs control.

**To Reproduce**

...

**Expected behavior**

Users need a way to express whether they want to set an attribute, or a property, or both. We also need to be able to add/remove attributes based on booleans.

Lit, Solid.js, and others have (sometimes only partial) solutions to this. For more details on the solutions other frameworks have, see this new test proposal for `custom-elements-everywhere`:

- https://github.com/webcomponents/custom-elements-everywhere/issues/2352

**Possible solution idea**

Solid.js has JSX syntax such as `attr:foo=...` and `prop:foo=...` for setting an attribute vs a property, but no boolean support. Pota, a library that is built on Solid.js, has its own JSX and `html` syntax that adds `bool:foo=...` support. Etc.

Preact could choose to adopt a similar approach, which is valid JSX syntax.

An alternative `html` template string tag for Preact could also provide Lit-like syntax (`.foo=...`, `foo=...`, and `?foo=...`).

rschristian:
> Answer: it depends, sometimes Preact sets the `.foo` property, sometimes Preact sets the `foo` attribute. The user has no control over this.

While I don't disagree with the premise/request, and have written up minor forks/patches that add this, there are quite defined rules. It's not random or a guess: if the provided label has a DOM property, and isn't from a [small list of excluded labels](https://github.com/preactjs/preact/blob/4c20c23c16dd60f380ce9fe98afc93041a7e1562/src/diff/props.js#L107-L119), it gets set as a property, else, attr.

trusktr:
[...]

Issue #3790, opened 2022-11-10
Regression: boolean attributes are no longer removed if value is false if attr name contains a dash


rejhgadellaa:
- [X] Check if updating to the latest Preact version resolves the issue

**Describe the bug**

Preact 10.11.1 introduced the following change:

> Fix falsy data attributes not working
https://github.com/preactjs/preact/releases/tag/10.11.1

Related issue #3717 and PR #3720

**To Reproduce**

With preact 10.11.0, the following snippet:

```jsx
<div data-has-toolbar={hasToolbar} />
```

would result in either a `<div />` with or without the attribute:

```jsx
<div data-has-toolbar="true" />
<div />
```

but starting in 10.11.1, it results in:

```jsx
<div data-has-toolbar="true" />
<div data-has-toolbar="false" />
```

**Expected behavior**

I was relying on the old behaviour for style rules so, at least in my use-case, this change could be considered a regression. The following CSS selector now always applies:

```css
div[data-has-toolbar] { /* ... */ }
```

where previously it would only apply if the prop's value was `true`.

The old behaviour feels more like HTML spec for boolean attributes/properties, but I'm not sure this applies to `data-` attributes?

For example, I think `element.dataset.hasToolbar` will always return a string where boolean properties-derived-from-attributes like `element.required` and `-.disabled` will actually return a boolean in javascript.

I can also see how it's worthwhile following React's behaviour here. They've actually recently fixed an issue and adopted Preact's behaviour for boolean attributes without dashes: https://github.com/facebook/react/issues/9230

krskrs:
Hi,
in my case it wasn't a regression, but a fix, as it was causing issues with React MUI which don't occur in react.

rejhgadellaa:
@krskrs yeah I can see how this fixes compat with React but at the same time introduces issues for projects that rely on the old behaviour. My main reason for opening this ticket is to point that out and to make sure the team is aware that they didn't just fix a bug (unless the old behaviour truly was unintentional - the fix may still break sites but at least then those sites just relied on the effects of the bug, not actual intended behaviour)

rburgst:
see https://github.com/preactjs/preact/issues/3717

JoviDeCroock:
@rejhgadellaa We are aware of this and I want to apologise for it, I did not account for this breaking behavior but thinking about it both unit tests and manual dom lookups could have suffered because of it.

trusktr:
Related, more syntax control would solve this:

[...]

PR #4419, opened 2024-06-22
feat: Add support for `prop:` & `attr:` directives

rschristian:
First-pass impl of #4416

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -35,7 +35,9 @@ let eventClock = 0;
  * @param {string} namespace Whether or not this DOM node is an SVG node or not
  */
 export function setProperty(dom, name, value, oldValue, namespace) {
-	let useCapture;
+	let useCapture,
+		split = name.split(':');
+	name = name.replace(/(.+):/, '');
 
 	o: if (name === 'style') {
 		if (typeof value == 'string') {
@@ -104,6 +106,8 @@ export function setProperty(dom, name, value, oldValue, namespace) {
 			// - className --> class
 			name = name.replace(/xlink(H|:h)/, 'h').replace(/sName$/, 's');
 		} else if (
+			split[0] == 'prop' ||
+			split[0] != 'attr' &&
 			name != 'width' &&
 			name != 'height' &&
 			name != 'href' &&
--- a/babel.config.js
+++ b/babel.config.js
@@ -31,7 +31,7 @@ module.exports = function (api) {
 		],
 		plugins: [
 			'@babel/plugin-proposal-object-rest-spread',
-			'@babel/plugin-transform-react-jsx',
+			['@babel/plugin-transform-react-jsx', { throwIfNamespace: false }],
 			'babel-plugin-transform-async-to-promises',
 			['babel-plugin-transform-rename-properties', { rename }]
 		],
--- a/test/browser/render.test.js
+++ b/test/browser/render.test.js
@@ -474,17 +474,17 @@ describe('render()', () => {
 
 	it('should support popover auto', () => {
 		render(<div popover="auto" />, scratch);
-		expect(scratch.innerHTML).to.equal("<div popover=\"auto\"></div>");
+		expect(scratch.innerHTML).to.equal('<div popover="auto"></div>');
 	});
 
 	it('should support popover true boolean', () => {
 		render(<div popover />, scratch);
-		expect(scratch.innerHTML).to.equal("<div popover=\"\"></div>");
+		expect(scratch.innerHTML).to.equal('<div popover=""></div>');
 	});
 
 	it('should support popover false boolean', () => {
 		render(<div popover={false} />, scratch);
-		expect(scratch.innerHTML).to.equal("<div></div>");
+		expect(scratch.innerHTML).to.equal('<div></div>');
 	});
 
 	// Test for preactjs/preact#4340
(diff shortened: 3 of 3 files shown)


src/diff/props.js
```diff
@@ -35,7 +35,9 @@ let eventClock = 0;
  * @param {string} namespace Whether or not this DOM node is an SVG node or not
  */
 export function setProperty(dom, name, value, oldValue, namespace) {
-	let useCapture;
+	let useCapture,
+		split = name.split(':');
```
marvinhagemeister, line 39: The `.split()` always causes an array allocation here. I wonder if we can avoid the allocation by only checking if there even is a prefix to begin with and if so do string slicing.

```ts
let prefix = "";
// Avoid doing anything at all if there is zero chance of
// it being a prefix name.
if (name[0] === "a" || name[0] === "p") {
  let idx = name.indexOf(":")
  if (idx > -1) {
    // Only slices here, should be faster than array allocation
    prefix = name.slice(0, idx)
    name = name.slice(idx + 1)   
  }
}
```

src/diff/props.js
```diff
@@ -104,20 +106,22 @@ export function setProperty(dom, name, value, oldValue, namespace) {
...
-			// cast to `0` instead
-			name != 'tabIndex' &&
-			name != 'download' &&
-			name != 'rowSpan' &&
-			name != 'colSpan' &&
-			name != 'role' &&
-			name != 'popover' &&
-			name in dom
+			split[0] == 'prop' ||
+			(split[0] != 'attr' &&
```
marvinhagemeister, line 110: Not sure if it will make a difference: Usually it's faster to have the common case upfront. The `prop` prefix being there is probably not the common case.

src/diff/props.js
rschristian: Yeah, good call.

I don't love the checking if first letters (we've had sone collateral damage with the "on" check) though maybe that's the best way to do it if we did want these props.

I assume Babel does support leading symbols in the JSX props (haven't checked) so perhaps following Lit with `.foo` and what not is easier for us.

src/diff/props.js
rschristian: Very true.

This was inserted quick just so I could check the perf & side, truth be told, but will swap it to the bottom.

src/diff/props.js
marvinhagemeister: yeah in this case the first letter check is fine I think since there are not many DOM attributes starting with `a` or `p`. It's only used as an early bail out in this case here. Having a unique first character would definitely be a better choice from a detection perspective. We could leave default as attribute and then you add some prefix like a dot or `@` to mark it as a prop.

rschristian: > We could leave default as attribute and then you add some prefix like a dot or @ to mark it as a prop.

If we want this to be purely additive, and I think we do, we'll need symbols/directives for both. If someone wants to set the `value` attribute instead of the prop (as problematic as this is), they'll need some way to express this as our current setup doesn't allow it.

rschristian:
I'll mark this as "ready", though I'm happy to still discuss whether we even want to do this and how it should interact with other things like RTS.

IMO, it is a nice bit of power to give users though it is a bit of a deviation and might confuse some.

src/diff/props.js
```diff
@@ -35,7 +35,15 @@ let eventClock = 0;
  * @param {string} namespace Whether or not this DOM node is an SVG node or not
  */
 export function setProperty(dom, name, value, oldValue, namespace) {
-	let useCapture;
+	let useCapture, prefix;
+	if (name[0] == 'a' || name[0] == 'p') {
```
JoviDeCroock, line 39: Would it be easier/mre performant by adding `&& name[5] == ':'` here? This would allow us to remove `idx` and replace it with 5 by default

rschristian: That it would be

developit:
@rschristian I like this direction, but I would love to avoid using namespaces for the prefixes. I believe vue uses something like `<foo .value="bar">` for properties and `<foo @value="bar">` for attributes. JSX doesn't support that though.

Interestingly, JSX does allow for two suffixes: `$` and `-`. The `-` suffix makes a decent fit for attributes because `-` at the end of an identifier is invalid so it's an uncommon property name (`obj['foo-']`):

```jsx
<div prop$="foo" />;
<div attr-="bar" />;
```

The other tidbit I wanted to add to this discussion is that we do technically already have a "force to use attribute" syntax, which is to uppercase the name (`<div ATTR="bar" />`). Unless a Custom Element happens to have an all-upper-case property (which would be extremely weird), it'll always hit `setAttribute('NAME')` which is case-insensitive and results in `attributeChangedCallback('name')`.

rschristian:
> The other tidbit I wanted to add to this discussion is that we do technically already have a "force to use attribute" syntax, which is to uppercase the name (`<div ATTR="bar" />`).

I'm fully aware, but it's a hard thing to push users toward; as stable as it is, it's an obvious hack that most raise an eyebrow at when it's suggested.

There's a few threads recently in particular w/ users uneasy after I've suggested using capitalization to get around it, hence, this PR.

---

It's a shame directives require some opt-in, I don't love how unclear `$` & `-` as suffixes are comparatively. Not as easy to spot either.

Thompson1985 (approved):

lilnasy:
https://github.com/preactjs/preact/issues/3790 (boolean attributes) was closed in favor of https://github.com/preactjs/preact/issues/4416. Is there a plan for it?

rschristian:
Not at this time. `data-` attributes, being user defined, I don't think can be safely treated as boolean attributes? It's impossible for the framework to know whether you expect to use it as an enumerated attribute or a boolean.

lilnasy:
What if there was a `bool:` directive?

rschristian:
I don't think we want to get into every possibility; I'm not sure even explicit prop/attr directives will land as-is.

If you want boolean, probably going to be on you to ensure they're added/removed when necessary.

PR #4426, opened 2024-06-28
fix(hooks): memo being too lazy with repeated renders

### LINKED ISSUE #4422: [preact/hooks] useMemo keeps recalculating even when the dependency is not changed

**Describe the bug**
When using the useMemo hook to calculate a value and then using this value to conditionally update the state during render, the app indefinitely re-renders. Something like shown below

```js
const [value, setValue] = useState(false);

const data = useMemo(() => {
  return { data: 'some data' };
}, [value]);

const [prevData, setPreviousData] = useState(data);

if (prevData !== data) {
  // This causes the infinite loop
  setPreviousData(data);
}
```

Preact version - 10.21.0

**To Reproduce**

Sandbox - https://codesandbox.io/p/sandbox/usememo-behavior-preact-w3rlrv

1. Go to the sandbox and open the console
2. Click on the 'Set to true' button
3. See error (and the console log of the memoized value changes)

After some debugging, it seems like the internalHookState of the useMemo hook is not updated with the new dependency args if a state update happens during the render cycle. And the hook always ends up comparing the dependency array with the very first dependency array. This is resulting in a recalculation of the hook every render and the component re-renders indefinitely.

**Expected behavior**
The useMemo hook should remember the last provided dependency array and should only recalculate the value when there is an actual change in the dependency array.

React seems to be handling this case as expected - https://codesandbox.io/p/sandbox/usememo-behavior-react-l3cxhs

JoviDeCroock:
This now equalises our behavior with React, we need to ensure that when we repeat a render we throwaway queued effects, this is done in `_render`. We however want to maintain achieved hook-states for i.e. `useMemo` as they are synchronous anyway, React seems to achieve this by disposing the WIP but retaining hooks-state. This PR makes us behave similarly.

- [Test case](https://codesandbox.io/p/sandbox/boring-wilson-9p6hyj?file=%2Fsrc%2Findex.js%3A12%2C16)
- [Repeated renders](https://stackblitz.com/edit/vitejs-vite-tylqxa?file=src%2FApp.tsx,src%2Fmain.tsx,src%2FMemo.tsx&terminal=dev)

Fixes #4422

### DIFF
--- a/hooks/test/browser/combinations.test.js
+++ b/hooks/test/browser/combinations.test.js
@@ -477,4 +477,44 @@ describe('combinations', () => {
 			'anchor effect'
 		]);
 	});
+
+	it('should not loop infinitely', () => {
+		const actions = [];
+		let toggle;
+		function App() {
+			const [value, setValue] = useState(false);
+
+			const data = useMemo(() => {
+				actions.push('memo');
+				return {};
+			}, [value]);
+
+			const [prevData, setPreviousData] = useState(data);
+			if (prevData !== data) {
+				setPreviousData(data);
+			}
+
+			actions.push('render');
+			toggle = () => setValue(!value);
+			return <div>Value: {JSON.stringify(value)}</div>;
+		}
+
+		act(() => {
+			render(<App />, scratch);
+		});
+		expect(actions).to.deep.equal(['memo', 'render']);
+		expect(scratch.innerHTML).to.deep.equal('<div>Value: false</div>');
+
+		act(() => {
+			toggle();
+		});
+		expect(actions).to.deep.equal([
+			'memo',
+			'render',
+			'memo',
+			'render',
+			'render'
+		]);
+		expect(scratch.innerHTML).to.deep.equal('<div>Value: true</div>');
+	});
 });
--- a/hooks/test/browser/useMemo.test.js
+++ b/hooks/test/browser/useMemo.test.js
@@ -151,12 +151,13 @@ describe('useMemo', () => {
 		act(() => {
 			set('bye');
 		});
-		expect(calls.length).to.equal(5);
+		expect(calls.length).to.equal(6);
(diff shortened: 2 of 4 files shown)


hooks/test/browser/useMemo.test.js
```diff
@@ -151,12 +151,13 @@ describe('useMemo', () => {
 		act(() => {
 			set('bye');
 		});
-		expect(calls.length).to.equal(5);
+		expect(calls.length).to.equal(6);
 		expect(calls).to.deep.equal([
 			'doing memo',
 			'render hi',
 			'doing memo',
-			'render bye', // We expect a missing "doing memo"  here because we return to the previous args value
```
JoviDeCroock, line 159: Justification for change can be found [here](https://stackblitz.com/edit/vitejs-vite-tylqxa?file=src%2FApp.tsx,src%2Fmain.tsx,src%2FMemo.tsx&terminal=dev)

hooks/test/browser/useMemo.test.js
```diff
@@ -172,6 +173,8 @@ describe('useMemo', () => {
 			if (v === 0) {
 				set(v + 1);
 			}
+
+			console.log('hi', res);
```
JoviDeCroock, line 177:
```suggestion

```

hooks/test/browser/combinations.test.js
```diff
@@ -477,4 +477,44 @@ describe('combinations', () => {
 			'anchor effect'
 		]);
 	});
+
+	it('should not loop infinitely', () => {
```
JoviDeCroock, line 481: How react behaves https://codesandbox.io/p/sandbox/boring-wilson-9p6hyj?file=%2Fsrc%2Findex.js%3A12%2C16

marvinhagemeister (approved):

PR #4432, opened 2024-07-06
Update ComponentChild types

sant123:
The `object` type is pretty wide and does not follow ReactNode types, so this change is to reflect the type from React.

Follows https://github.com/preactjs/preact/pull/1116

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -49,7 +49,8 @@ export type ComponentChild =
 	| bigint
 	| boolean
 	| null
-	| undefined;
+	| undefined
+	| Function;
 export type ComponentChildren = ComponentChild[] | ComponentChild;
 
 export interface Attributes {
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -45,7 +45,7 @@ export type ComponentChild =
 	| VNode<any>
 	| string
 	| number
-	| Iterable<ComponentChild>
+	| Array<ComponentChild>
 	| bigint
 	| boolean
 	| null
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -43,9 +43,9 @@ export type Ref<T> = RefObject<T> | RefCallback<T> | null;
 
 export type ComponentChild =
 	| VNode<any>
-	| object
 	| string
 	| number
+	| Iterable<ComponentChild>
 	| bigint
 	| boolean
 	| null


rschristian:
Looks like this breaks a few TS tests.

> does not follow ReactNode types, so this change is to reflect the type from React.

For what it's worth, we don't aim to match React's types in core necessarily, only in compat, and even that is more about avoiding conflict than matching in strictness. We're different frameworks with different implementations, we do accept different types in some circumstances.

sant123:
> Looks like this breaks a few TS tests.
> 
> > does not follow ReactNode types, so this change is to reflect the type from React.
> 
> For what it's worth, we don't aim to match React's types in core necessarily, only in compat, and even that is more about avoiding conflict than matching in strictness. We're different frameworks with different implementations, we do accept different types in some circumstances.

Fair enough, for the failed tests the issue is when passing functions as children in components (that is found in Context.Consumer). I have uploaded a fix for that so the tests should be passing.

src/index.d.ts
```diff
@@ -43,13 +43,14 @@ export type Ref<T> = RefObject<T> | RefCallback<T> | null;
 
 export type ComponentChild =
 	| VNode<any>
-	| object
 	| string
 	| number
+	| Iterable<ComponentChild>
```
marvinhagemeister, line 48: I don't think we support any iterables as component children. Can't find code in our source that explicitly invokes the iterator protocol.

sant123: So do you think is better the Array type here?

src/index.d.ts
```diff
@@ -43,13 +43,14 @@ export type Ref<T> = RefObject<T> | RefCallback<T> | null;
...
 	| string
 	| number
+	| Array<ComponentChild>
 	| bigint
 	| boolean
 	| null
-	| undefined;
+	| undefined
+	| Function;
 export type ComponentChildren = ComponentChild[] | ComponentChild;
```
rschristian, line 54: I wonder, do we really need to alter the type here? I do generally prefer `ComponentChild` being singular, with `ComponentChildren` being used if an array of children is needed.

If it's a compat issue, we can simply alter the [`ReactNode`](https://github.com/preactjs/preact/blob/f7f9d9bf4309d09bcd0a432acd72787b7d9baecb/compat/src/index.d.ts#L66) type instead?

Thompson1985 (approved):

PR #4437, opened 2024-07-07
fix(hydration): address hydration bug for fragment with missmatched dom-nodes

JoviDeCroock:
At the moment mainly looking for input here, the issue being that for non-fragment-like diffing this could be problematic, in theory we would have to unmount the missmatched DOM-node but we aren't sure what position we are in... This is also a problem pretty exclusive to Fragment like returns because if I'd wrap the Suspense boundary in a DOM tag this would be a non-issue.

The current solution will unmount all excessDomChildren once we have a missmatched DOM-node type.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -402,6 +402,14 @@ function diffElementNodes(
 				break;
 			}
 		}
+
+		if (
+			dom == null &&
+			excessDomChildren.length == 1 &&
+			excessDomChildren[0] != null
+		) {
+			removeNode(excessDomChildren[0]);
+		}
 	}
 
 	if (dom == null) {
@@ -416,9 +424,6 @@ function diffElementNodes(
 		);
 
 		// we created a new parent, so none of the previously attached children can be reused:
-		// TODO: this here is problematic because we could be missmatching a fragment with a node
-		// this happens a lot in Suspense with a fallback where after hydration completes we are hydrating
-		// the content and not the fallback.
 		excessDomChildren = null;
 
 		// we are creating a new node, so we can assume this is a new subtree (in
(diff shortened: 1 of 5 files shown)


src/diff/index.js
```diff
@@ -402,6 +402,14 @@ function diffElementNodes(
...
 			}
 		}
+
+		if (
+			dom == null &&
+			excessDomChildren.length == 1 &&
+			excessDomChildren[0] != null
+		) {
+			removeNode(excessDomChildren[0]);
+		}
```
JoviDeCroock, line 412:
```suggestion
		if (dom == null) {
		  for (i = 0; i < excessDomChildren.length; i++) {
			  value = excessDomChildren[i];
			  if (value) {
			    removeNode(value);
			    excessDomChildren[i] = null;
			  }
			}
		}
```

src/diff/index.js
```diff
@@ -402,6 +402,14 @@ function diffElementNodes(
 				break;
 			}
 		}
+
+		if (
+			dom == null &&
+			excessDomChildren.length == 1 &&
+			excessDomChildren[0] != null
+		) {
+			removeNode(excessDomChildren[0]);
```
JoviDeCroock, line 411: For Resumed hydration this could be safe as lazy/suspending components return a single VNode, however we are talking about dom-children here, which means that when the VNode returned is a Fragment returning multiple DOM-nodes our heuristic would only match the first DOM-node. Which I guess might also be an issue in how we resume hydration at https://github.com/preactjs/preact/blob/main/src/diff/index.js#L55

marvinhagemeister (approved):

PR #4438, opened 2024-07-08
fix(hydration): Suspender returns Fragment with multiple DOM-nodes

JoviDeCroock:
This is an issue that presents itself when a lazy boundary returns a Fragment-like which results in an issue with our current `excessDomChildren` restoration. At this point in time we just take `oldDom` which amounts to us restoring The first of said Fragment-like wrapper. The current solution is a bit of a "quick one" to highlight a solution to the problem.

The current solution however would introduce an issue when we have 

```
<>
  <Suspense>
    <Lazy /> --> <div>hello</div><div>world</div>
  <Suspense>
  <div>baz</div> --> because we only mark excessDomChildren[indexOfOldDom] as null we could accidentally match <div>world</div> here
</>
```

This might however be acceptable, because React fails here as well

- React 17: https://codesandbox.io/p/sandbox/empty-sun-dmj892?file=%2Fsrc%2FApp.js%3A16%2C25
- React 18: https://codesandbox.io/p/sandbox/empty-sun-forked-28pq95?file=%2Fsrc%2Findex.js%3A7%2C1

### DIFF
--- a/compat/test/browser/suspense-hydration.test.js
+++ b/compat/test/browser/suspense-hydration.test.js
@@ -169,6 +169,43 @@ describe('suspense hydration', () => {
 		});
 	});
 
+	it('Should hydrate a fragment with multiple children and an adjacent node correctly', () => {
+		scratch.innerHTML = '<div>Hello</div><div>World!</div><div>Baz</div>';
+		clearLog();
+
+		const [Lazy, resolve] = createLazy();
+		hydrate(
+			<>
+				<Suspense>
+					<Lazy />
+				</Suspense>
+				<div>Baz</div>
+			</>,
+			scratch
+		);
+		rerender(); // Flush rerender queue to mimic what preact will really do
+		expect(scratch.innerHTML).to.equal(
+			'<div>Hello</div><div>World!</div><div>Baz</div>'
+		);
+		expect(getLog()).to.deep.equal([]);
+		clearLog();
+
+		return resolve(() => (
+			<>
+				<div>Hello</div>
+				<div>World!</div>
+			</>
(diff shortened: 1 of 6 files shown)


compat/test/browser/suspense-hydration.test.js
```diff
@@ -139,6 +139,73 @@ describe('suspense hydration', () => {
...
+		)).then(() => {
+			rerender();
+			expect(scratch.innerHTML).to.equal('<div>Hello</div><div>World!</div>');
+			expect(getLog()).to.deep.equal([]);
+
+			clearLog();
+		});
+	});
+
+	it('Should hydrate a fragment with multiple children and an adjacent node correctly', () => {
```
JoviDeCroock, line 172: React seems to also fail on this

```suggestion
    // This is an expected fail case, we should encourage folks to not have adjacent
    // nodes of the same type, or return a single wrapping dom node from a lazy boundary.
	it.skip('Should hydrate a fragment with multiple children and an adjacent node correctly', () => {
```

src/diff/index.js
```diff
@@ -277,6 +277,7 @@ export function diff(
 				newVNode._flags |= isHydrating
 					? MODE_HYDRATE | MODE_SUSPENDED
 					: MODE_HYDRATE;
+				newVNode._excess = [...excessDomChildren];
```
JoviDeCroock, line 280: An alternative to copying here would be to mark certain nodes i.e. we mark it with the oldVNode signature and then when we resume hydration in the initial `src/diff/index.js` check we unlock it by checking

marvinhagemeister (approved):

Issue #4442, opened 2024-07-12
RFC: Hydration 2.0


JoviDeCroock:
With Streaming Server Side Chunks as well as rendering to string asynchronously becoming more popular it's becoming apparent that our hydration algorithm isn't quite sufficient anymore. There's a set way in which it will behave well for these resumed hydration cases but a lot of innocent-looking footguns that can lead to opting out of hydration or in some cases UI bugs.

The goal of hydration is to re-use the HTML that is returned to us from the server and just attach the event-listeners so the UI is usable to the consumer again, this is the case for simple full-on hydration. We later on introduced the concept of resumed hydration so we could hydrate `Suspense` boundaries etc, the idea was that when a child throws a Promise because the JS chunk is being downloaded, ... we'd pause hydration, mark that VNode as in need for hydration, save the DOM-pointer we're on and continue. When the Promise resolved we could continue from the VNode with the DOM pointer and continue our hydration.

The benefit of this resumed hydration approach is that we can have parts of the DOM ready for interaction before all chunks (be that data or JS) are ready on the client. There's one upgrade over this which is streaming the HTML in as then we can even forego the latency introduced by those Promises on the server. This RFC explicitly does not look too deep into streaming, we touch on it in the bonus section, however we aim to solve the `renderToStrincAsync()` into resumed `hydrate()` workflow.

The main problem tackled in this RFC resides in the storing of the DOM-pointer to continue hydration from. When we are storing the DOM-pointer, we aren't aware what content that suspended node will return when it ends up resolving, i.e. will this render function return null, 1 DOM-node or multiple.

### Suspending render returns null

When we have the following case

```jsx
const X = () => {
  // Some logic to i.e. create your google-analytics script
  return null;
}

const SuspendingX = lazy(() => /* importing x */)
```

Then we'll suspend as soon as we try to render `SuspendingX`, however we'll reserve a DOM-pointer for this that isn't usable for any of its siblings anymore while this component will end up returning `null` so when we see `<div><SuspendingX /><p>hello world</p></div>` we'll reserve the `<p>` tag to hydrate `<SuspendingX />` and end up duplicating `<p>hello world</p>` because we think it's a hydration mismatch.

Related issue https://github.com/preactjs/preact/pull/4075
### Offset creation with multiple DOM-children

This builds on the previous case but instead of returning null it will return multiple DOM-nodes.

```jsx
const Y = () => {
  return <Fragment><p>Foo</p><p>Bar</p></Fragment>;
}

[...]

PR #4444, opened 2024-07-15
chore(hydration): Test approach to comment markers

JoviDeCroock:
Relates to https://github.com/preactjs/preact/issues/4442
Implementation from https://github.com/preactjs/preact-render-to-string/pull/376

This implements a hydration approach with the comment markers, one thing that's noteworthy is that `oldDom` in our `diffChildren` algorithm becomes stale.

We diff the wrapping element of the Suspense and non-suspenseful child, we get to an oldDom which is the first child-node of said series. When we suspend we don't move oldDom, we can try and move it inside of `diff` but this won't affect the continued `diffChildren` so we are stuck inserting that node in front of our oldDom.

This has been solved in https://github.com/preactjs/preact/pull/4444/commits/647d13865b3c9730f17f8ce68644145925a226d7 through correctly skipping comment nodes while placing children.

I think the last part of this research would be to set up a comprehensive test suite for both `renderToStringAsync` as well as `renderToStream` so we can have fixtures for this behaviour, what do you all think?

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -283,20 +283,36 @@ export function diff(
 				let shouldFallback = true,
 					commentMarkersToFind = 0,
 					done = false;
+
 				newVNode._component._excess = [];
 				for (let i = 0; i < excessDomChildren.length; i++) {
 					let child = excessDomChildren[i];
 					if (child == null || done) continue;
 
+					// When we are inside of the boundaries we set
+					// the excess child to null.
 					if (commentMarkersToFind > 0) {
 						excessDomChildren[i] = null;
 					}
 
+					// When we encounter a boundary with $s we are opening
+					// a boundary, this implies that we need to bump
+					// the amount of markers we need to find before closing
+					// the outer boundary.
+					// We exclude the open and closing marker from
+					// the future excessDomChildren but any nested one
+					// needs to be included for future suspensions.
 					if (child.nodeType == 8 && child.data == '$s') {
+						if (commentMarkersToFind > 0) {
+							newVNode._component._excess.push(child);
+						}
 						commentMarkersToFind++;
 						shouldFallback = false;
 					} else if (child.nodeType == 8 && child.data == '/$s') {
 						commentMarkersToFind--;
+						if (commentMarkersToFind > 0) {
+							newVNode._component._excess.push(child);
+						}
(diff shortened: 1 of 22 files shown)


src/component.js
```diff
@@ -100,7 +100,11 @@ export function getDomSibling(vnode, childIndex) {
 	for (; childIndex < vnode._children.length; childIndex++) {
 		sibling = vnode._children[childIndex];
 
-		if (sibling != null && sibling._dom != null) {
+		if (
+			sibling != null &&
+			sibling._dom != null &&
+			sibling._dom.nodeType !== 8
+		) {
```
JoviDeCroock, line 107:
```suggestion
		if (sibling != null && sibling._dom != null) {
```

JoviDeCroock:
Two open todo's on this one

- Extract the streaming fixes (skipping comments code in children and index.js)
- Move `_excess` to live on `_component`

src/diff/index.js
```diff
@@ -273,13 +281,42 @@ export function diff(
...
-				// excessDomChildren.length = 0;
+
+				let found = excessDomChildren.find(
+					child => child && child.nodeType == 8 && child.data == '$s'
+				);
+
+				newVNode._dom = oldDom;
+				if (found) {
+					let commentMarkersToFind = 1,
+						index = excessDomChildren.indexOf(found) + 1;
```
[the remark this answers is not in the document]
JoviDeCroock: Would appreciate a proof read here 😅 https://github.com/preactjs/preact/pull/4444/commits/38e1a604b53a83edc2a03744646e8a5517207301

marvinhagemeister (approved):

src/diff/index.js
```diff
@@ -277,10 +280,53 @@ export function diff(
...
+
+				newVNode._component._excess = [];
+				for (let i = 0; i < excessDomChildren.length; i++) {
+					let child = excessDomChildren[i];
+					if (child == null || done) continue;
+
+					// When we are inside of the boundaries we set
+					// the excess child to null.
+					if (commentMarkersToFind > 0) {
+						excessDomChildren[i] = null;
```
JoviDeCroock, line 295: This should happen in all three if branches

Thompson1985 (approved):

JoviDeCroock:
Merged in v11

PR #4454, opened 2024-07-24
Fix issue in testing libraries where there might not be a DOM node

JoviDeCroock:
Some hooks based testing libraries work without a DOM node so it will be undefined all the time, this loosens the check

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -42,8 +42,8 @@ function getDomChildren(vnode) {
 function getClosestDomNodeParentName(parent) {
 	if (!parent) return '';
 	if (typeof parent.type == 'function') {
-		if (parent._parent === null) {
-			if (parent._dom !== null && parent._dom.parentNode !== null) {
+		if (parent._parent == null) {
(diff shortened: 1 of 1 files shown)


debug/src/debug.js
```diff
@@ -42,8 +42,8 @@ function getDomChildren(vnode) {
 function getClosestDomNodeParentName(parent) {
 	if (!parent) return '';
 	if (typeof parent.type == 'function') {
-		if (parent._parent === null) {
-			if (parent._dom !== null && parent._dom.parentNode !== null) {
+		if (parent._parent == null) {
+			if (parent._dom != null && parent._dom.parentNode !== null) {
```
marvinhagemeister, line 46: Just for consistency: We should probably loosen the `.parentNode` check as well

```suggestion
			if (parent._dom != null && parent._dom.parentNode != null) {
```

marvinhagemeister (approved):
I'm a bit worried that we'll have to widen our types to include `undefined` in other places in the future as well, but for now this seems fine.

Issue #4457, opened 2024-07-27
preact demo won't run


gengjiawen:
cd demo && npm i && yarn dev

```
  ✘ [ERROR] The JSX syntax extension is not currently enabled

    zustand.js:19:2:
      19 │     <div>

```

PR #4458, opened 2024-07-27
fix demo build

gengjiawen:
fix https://github.com/preactjs/preact/issues/4457

cc @marvinhagemeister


rschristian:
Yeah unfortunately there's a circular reference there that Vite can't work out from the source files, will need to use the built.

demo/index.jsx
```diff
@@ -140,7 +140,7 @@ class App extends Component {
 						<Link href="/zustand" activeClassName="active">
 							zustand
 						</Link>
-						<Link href="/redux-toolkit" activeClassName="active">
+						<Link href="/demo/redux-toolkit" activeClassName="active">
```
rschristian, line 143: I don't think this should change?

```suggestion
						<Link href="/redux-toolkit" activeClassName="active">
```

gengjiawen:
> Yeah unfortunately there's a circular reference there that Vite can't work out from the source files, will need to use the built.

looks fixable via a hack.

demo/package-lock.json
```diff
@@ -44,11 +44,12 @@
       }
     },
     "node_modules/@babel/code-frame": {
```
rschristian, line 46: I don't think we need to change the lock file at all.

compat/src/render.js
```diff
@@ -297,7 +296,7 @@ export const __SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED = {
 			useEffect,
 			useId,
 			useImperativeHandle,
-			useInsertionEffect,
+			useInsertionEffect: useLayoutEffect,
```
rschristian, line 299: I'd rather not split that alias across two modules, no. Especially not for the demo.

gengjiawen: It will make the vite dev and build work.

demo/package-lock.json
gengjiawen: It's a minimal npm audit fix. mainly for the vite and babel security issue.

compat/src/render.js
rschristian: And I get that, but the demo is very minor (I don't think we even use it for anything -- frankly, I'd be up for removal as it's a bit of a mess) and shouldn't change source.

The issue is that if we do provide an implementation for `useInsertionEffect`, now there's two separate places that need to be updated which I don't love.

demo/package-lock.json
rschristian: I think we should only change what's necessary. Keeps the diff sane over the long-term.

compat/src/render.js
gengjiawen: > And I get that, but the demo is very minor

I really like it since it very easy to tinker with preact source. Anyway, I will leave you guys decide with PR.

.github/workflows/ci.yml
```diff
@@ -45,6 +45,9 @@ jobs:
         run: |
           npm run lint
           npm run test:unit
+      - name: test Demo
+        run: |
+          cd demo && npm i && npm run build
```
rschristian, line 50: Vite being able to build the app tells us very little (beyond that we have correct file paths, as you've spotted & corrected). Whether these modules actually work or not wouldn't be caught at all.

This is more of a test of Vite than Preact, I don't think it makes sense to be part of the CI as-is. If we did want to do an ecosystem CI (and that might be a very good idea), it would need proper tests and what have you. `/demo` just isn't that at the moment. I don't see the value-add in testing it.

compat/src/render.js
JoviDeCroock: I mean, let's find a different solution that doesn't involve changes to our packages 😅 because at this point it points at a bug in vite

.github/workflows/ci.yml
JoviDeCroock: I agree, let's not add arbitrary things to our CI that don't involve anything Preact related. This demo tests more that vite bundles this than anything else, so it would be better fit for preset-vite

compat/src/render.js
rschristian: I certainly see the value, and do think we need to set up something to facilitate that better (I'll be looking into it). But there's a fair number of issues in this approach, which is why I'm fairly apprehensive about encouraging it. Anything that uses the plugin API (outside of this repo, anyways) won't function correctly.

JoviDeCroock: We could just do what I did initially and make a `react-hooks` file with these implementations

rschristian: @JoviDeCroock Not necessarily a bug, but a cyclical reference. index -> render, render -> index. I think Microbundle/rollup does throw warnings because of this (but it can handle it alright as it's just bundling). Vite, serving the unbundled individual modules in the browser, falls over (but again, non-representative as no one is running Preact from source in reality, so...)

rschristian: Took another look: while the compat issue can be reasonably fixed by externalizing the React hook implementations to a new module as mentioned, core has a couple cyclical references that I don't think can be untangled without merging modules or doing an internal barrel file sorta thing. I don't think either are a worthwhile approach.

Best solution is to insert `"preact": "../",` into `/demo/package.json` so that it uses built output instead.

Thompson1985 (approved):

compat/src/render.js
prust: > no one is running Preact from source in reality

FWIW, I just did (I wanted a no-build workflow that wasn't minified for development, so I can get familiar with the Preact source during dev). Firefox & Chrome both choked on this cyclical reference. This change, plus adding `.js` extensions to the imports, was all I've needed so far to run Preact from source (IMO it's worth the effort).

rschristian: > FWIW, I just did (I wanted a no-build workflow that wasn't minified for development, so I can get familiar with the Preact source during dev).

There certainly are some situations where this works, with that being one of them, however, this breaks expectations of any lib using the plugin API (which at least the first-party ecosystem _heavily, heavily_ relies upon). As such, we don't recommend it in prod as it'll lead to nasty surprises in all likelihood.

prust: @rschristian: Good to know -- thank you!

JoviDeCroock:
Fixed in #4551, thank you for starting this 🙌

gengjiawen:
> Fixed in #4551, thank you for starting this 🙌

Thx for doing this. A co-author would be perfect in your new commit :)

Issue #4464, opened 2024-07-31
Wrong typing for HTML translate property


teodragovic:
**Describe the bug**

[Per spec], translate property value is defined as empty string, "yes" or "no" but [Preact](https://github.com/preactjs/preact/blob/main/src/jsx.d.ts#L2488) expects boolean which produces TS error.

**To Reproduce**

Add `translate="no"` or `translate="yes"` to any HTML element in Preact app.

rschristian:
(Just to copy what I wrote on the PR...)

The translate _attribute_ is an empty string, "yes", or "no": [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/translate)
The translate _property_ is a boolean: [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/translate)

Personally, I'd be up for supporting both given JSX is already an arbitrary mix of attributes & properties, but this won't be just a types change.

teodragovic:
Ah, well that was confusing lol. Ok, I'll close PR and ticket. Thanks for the clarification @rschristian

rschristian:
It's especially fun when you set `htmlElement.translate = "no"` and it becomes `"yes"`/`true` :)

PR #4465, opened 2024-07-31
Fix typings for HTML translate attribute

teodragovic:
Fixes https://github.com/preactjs/preact/issues/4464

Wasn't sure if `boolean` is better removed or left in (I think "empty string" part from the specs would be same as `translate={true}`)?


rschristian:
Unfortunately, I don't think this is correct, see https://github.com/preactjs/preact/pull/3800/files

PR #4471, opened 2024-08-06
Add TypeScript support for Container.contains

sjoerdmulder:
Recently the code started using the `contains` method on the Node interface but this is not exposed in the types causing a runtime crash when implementing your own `ContainerNode `


JoviDeCroock (approved):
Thank you!

PR #4478, opened 2024-08-20
fix: Always set props in a CE's 'observedAttributes' as attributes

hesxenon:


src/diff/props.js
```diff
@@ -117,6 +117,7 @@ export function setProperty(dom, name, value, oldValue, namespace) {
 			name != 'colSpan' &&
 			name != 'role' &&
 			name != 'popover' &&
+			!dom.constructor.observedAttributes?.includes(name) &&
```
rschristian, line 120: I'd be surprised if this wasn't breaking and I don't think it's inline with our existing flow here.

Preact doesn't make any attempt to guess what the user wants; we set props if the name is recognized (with a few exceptions), else, it's an attribute. This breaks that by attempting to guess the user wants an attribute instead.

#4419 is probably a better option, takes the guesswork out for Preact and gives the user more power.

hesxenon:
On the contrary - preact is breaking things _now_ because it guesses what the user wants _now_. It _presumes_ that when I write web components I want preact to take care of the IDL side of things only, which is not what I want when I put a (content) attribute name into the observedAttributes array. This array is by definition there to inform the browser that you are interested in changes to these content attributes and in combination with the [MDN glossary on IDL](https://developer.mozilla.org/en-US/docs/Glossary/IDL) this means (as I understood it) that the IDL attribute should reflect the content attribute, not the other way around. In other words:

```jsx
class MyCustomElement extends HTMLElement {
  static observedAttributes = ["href", "disabled", "foobar"];
  
  href;

  disabled;
  
  foobar;

  connectedCallback() {
    console.log(this.href, this.disabled, this.fobar);
  }
}
customElements.define("my-custom-element", MyCustomElement);

// render with preact
<my-custom-element href="/some/url" disabled foobar="bazqux"></my-custom-element>
```
will (confusingly) lead to the following HTML:
```html
<my-custom-element href="/some/url"></my-custom-element>
```
but it will log all three properties.

This is also out of line with the following example
```jsx
// given this element in the dom
<input value="foo" />

const input = document.querySelector("input")
input.value = "bar"
console.log(input.value, input.getAttribute("value")) // logs bar, foo, since IDL doesn't update content on purpose
```

Long story short: the current behaviour of preact breaks webcomponents that try to follow IDL and Content rules. This PR would fix that. Components that currently indicate they are interested in an _attribute_ by specifying its name in observed _Attributes_ that then expect this _attribute_ to be treated like a prop are, well... let's say they're a mystery to me and imho they're going against the spec.

hesxenon:
Converted to draft because regardless of the outcome tests are still missing

rschristian:
I understand your reasoning, but I'd lean towards saying this is unsupported behavior. Preact, as a JS framework, prefers properties, falling back to attributes only when we need to. This might not work great with all APIs, but is easier to reason about and has some large perf advantages.

There's absolutely a gap in that Preact users can't always easily choose which is set (you can get away with some fiddling of capitalization usually), but I'm not sure this is the best way to address the immediate issue or that gap as a whole. It's better if users can explicitly opt-in to attributes than adding a guess that they might want an attribute in that particular case, IMO.

>```js
> console.log(input.value, input.getAttribute("value")) // logs bar, foo, since IDL doesn't update content on purpose
>```

Setting the `value` attribute of an input is incredibly dangerous and shouldn't be done (by JS, that is -- it's fine if you're using it as a default from HTML), I'm not quite sure what this is supposed to show.

hesxenon:
> preact is a js framework

and how many users do you know that even know about the IDL vs. Content gap? I'd say preact is primarily seen as a "rendering soluition" and falling into the mentioned gap while the MDN docs explicitly mention `observedAttributes` as a way to well, observe _attributes_ shouldn't just be dismissed by saying "js (IDL) first, tough luck" imho :sweat_smile:.

My point is (and that's what that example was supposed to show) that there is also a gap between content and IDL and that's by design and should be taken into consideration by webcomponent authors following "existing rules". And following those rules leads to incompatibilities with preact (and react for that matter) which basically leads to "you can use preact/react as long as you stay in the ecosystem" all while stating "we're closer to the platform than the rest".

rschristian:
> and how many users do you know that even know about the IDL vs. Content gap?

Oh boy, just take a look at our issues/discussions.

Not many do, and I'm very much of the opinion the gap itself is an inherent flaw in the web, but... that's the way it is, and pretending it doesn't exist doesn't help anyone.

> MDN docs explicitly mention `observedAttributes` as a way to well, observe attributes

It's a way for web components to observe attributes. We're not web components however, nor do we necessarily facilitate all behaviors as we cannot whilst staying in our perf & size targets. It's always a balancing act.

> shouldn't just be dismissed by saying "js (IDL) first, tough luck" imho 😅.

Not tough luck, but you should use a workaround until it can be explicitly set.

hesxenon:
> and pretending it doesn't exist doesn't help anyone.

sad but true

> we're not web components

true, but you claim to be compatible with webcomponents by staying closer to the platform - which webcomponents _can_ I use with preact with this behaviour then that aren't also written with preact/react in mind (which is the antithesis to webcomponents...)?

> use a workaround

I'd be happy to know which workaround this could be?

> perf and size

This one line shouldn't break perf and surely doesn't break size? _And_ it shouldn't break existing projects since for those to break they'd have to use webcomponents that explicitly go against the mentioned gap.

Honestly I'm finding myself between a rock and a hard place here since I'm now faced with either changing the webcomponents to consider being used within react/preact _or_ dropping react/preact entirely _or_ using a custom preact fork (wasn't able to adapt react as easily) for the time being.

rschristian:
> I'd be happy to know which workaround this could be?

The prop name has to fail that check -- that's it. Capitalization, using other names, etc.

A silly example of this is `<input Value="foo" />`  -- `Value` doesn't exist upon a DOM element, therefore it's set as an attribute.

> This one line shouldn't break perf and surely doesn't break size?

Keep in mind this is a very hot path; every single prop being set or updated passes through there.

Some code golf wins in this repo are -4b, for reference. +32b certainly isn't the end of the world though.

My point was more that full support of everything one could want is a non-goal, not that this particular PR was. There are some things we knowingly don't support as the perf & byte cost isn't worth it.

> And it shouldn't break existing projects since for those to break they'd have to use webcomponents that explicitly go against the mentioned gap.

Not necessarily; they'd just have to use web components that react, but don't intend on having those attribute set from themselves. A bit uncommon, but it's not going against anything. That might not be too big of a group, and so it might be fine to push out, but it is breaking behavior.

hesxenon:
I am not against the "sigil" approach - but this is a much smaller change, both internally and externally?

The proposed workaround then is to design the webcomponent with preact in mind - since uppercasing the content attribute while lowercasing the IDL introduces complexity in the webcomponent for the purpose of making it compatible with a framework :/

Plus it means that the IDL attribute does not technically reflect the content attribute, which _does_ go against the definition of IDL attributes?

> but don't intend on having those attribute set from themselves

Keeping in mind the existing "native elements" like select, input, div etc - they all set their IDL attributes according to the content attributes - that's what "IDL reflects content" means, no? And the IDL attribute is named like the content attribute with very few exceptions that mostly relate to camelCasing kebab-cased attributes.

> I don't think use of that API sends the signal that you apply here.

I agree - in isolation it does not send the signal that I apply here. But in combination with the definition of IDL attributes and existing implementations it does, doesn't it? IDL attributes should reflect their content counterparts (and may transform them), observedAttributes means "I'm interested in changes to the content attribute" and existing implementations update their IDL attributes if the content attribute changes.

But I think I've found a "loophole" here. Quoting from MDN:
> The IDL attribute is always going to use (but might transform) the underlying content attribute to return a value when you get it and is going to save something in the content attribute when you set it

"is going to save something in the content attribute when you set it" being the key part here. That sounds to me like webcomponents _should reflect the IDL change back into the content_... I don't see yet how to prevent infinite loops here but that's hardly preacts problem...

I'd still be interested on your thoughts about that and how that relates to the aforementioned existing implementations - e.g. why `document.querySelector("input").value = "foobar"` does set the IDL attribute but does not update the content attribute

Thompson1985 (approved):

Thompson1985 (approved):

Thompson1985 (approved):

Thompson1985 (approved):

Thompson1985 (approved):

hesxenon:
since this PR has been approved, can I get some pointers regarding tests? I don't think this should be marked as ready to merge without tests?

marvinhagemeister:
FYI: The person that approved it has no approval rights and is a bot. They've been going over the preact and graphqljs repo and approving every open PR. It's a spam bot.

hesxenon:
interesting... a bug in github? Or how does an account without rights
approve this? Anyway, I still think this would be the way to go

JoviDeCroock:
Closing: draft with no tests; forcing observedAttributes props to attributes leans toward unsupported behavior per maintainer feedback, and it has stalled.

PR #4485, opened 2024-08-30
Golf down children

JoviDeCroock:


src/diff/children.js
```diff
@@ -62,13 +62,7 @@ export function diffChildren(
 
 	for (i = 0; i < newChildrenLength; i++) {
 		childVNode = newParentVNode._children[i];
-		if (
-			childVNode == null ||
-			typeof childVNode == 'boolean' ||
-			typeof childVNode == 'function'
```
JoviDeCroock, line 68: We already handle these during the new array construction

src/diff/children.js
```diff
@@ -185,6 +179,7 @@ function constructNewChildrenArray(newParentVNode, renderResult, oldChildren) {
 			typeof childVNode == 'function'
 		) {
 			childVNode = newParentVNode._children[i] = null;
+			continue;
```
JoviDeCroock, line 182: Was repeated on line 230-234

src/diff/children.js
```diff
@@ -299,9 +289,8 @@ function constructNewChildrenArray(newParentVNode, renderResult, oldChildren) {
 
 				// Move this VNode's DOM if the original index (matchingIndex) doesn't
 				// match the new skew index (i + new skew)
-				if (matchingIndex !== i + skew) {
-					childVNode._flags |= INSERT_VNODE;
-				}
+				// In the former two branches we know that it matches after skewing
+				childVNode._flags |= INSERT_VNODE;
```
JoviDeCroock, line 293: Is always going to be truthy now!

marvinhagemeister (approved):

Issue #2737, opened 2020-09-01
Add SSR hydration mismatch errors


JamesBurnside:
In React when hydrating a server side rendered page using `ReactDom.hydrate` this API will produce a warning in debug builds when the html generated client side did not match the html generated server side, for example:
![image](https://user-images.githubusercontent.com/2684369/91904341-9c7e7880-ec59-11ea-96dc-24289d69a3b5.png)

Preact should show the same warnings so developers know when a mismatch has occured.

betula:
I consider this a significant point.
Please Preact - this is sorely lacking!

PR #4490, opened 2024-09-12
Warn our user for hydration mismatches

JoviDeCroock:
One of the things I have been battling while SSR'ing Preact is that I don't know where the mismatch occurs, this for one gives me an error and tells me what the mismatch was.

Another idea that I have been floating here and would love feedback on is to remove the `isHydrating` as a plain value and move to a ref, the reason for this being that currently we opt-out subtrees from hydration but this is particularly dangerous as siblings to the mismatched node can still be in hydration mode which will cause us to duplicate the DOM if mismatches continue happening.

Related to #4442 
Resolves https://github.com/preactjs/preact/issues/2737

### DIFF
--- a/debug/src/debug.js
+++ b/debug/src/debug.js
@@ -582,3 +582,11 @@ export function serializeVNode(vnode) {
 		children && children.length ? '>..</' + name + '>' : ' />'
 	}`;
 }
+
+options._hydrationMismatch = (newVNode, excessDomChildren) => {
+	const { type } = newVNode;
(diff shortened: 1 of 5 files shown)


marvinhagemeister (approved):
Great idea!

rschristian:
Should we say this addresses #2737? With React throwing those in prod (to my knowledge), not sure if we want to say that `/debug` addresses that or not.

Sorry for the noise, rethought it once I linked it.

JoviDeCroock:
@rschristian that's a great shout! Thanks for linking that

PR #4505, opened 2024-09-22
fix: Correct some missing & incorrect SVG types

rschristian:


src/jsx.d.ts
```diff
@@ -198,11 +198,6 @@ export namespace JSXInternal {
 			| SignalLike<number | string | undefined>;
 		attributeName?: string | undefined | SignalLike<string | undefined>;
 		attributeType?: string | undefined | SignalLike<string | undefined>;
-		autoReverse?:
-			| number
-			| string
-			| undefined
-			| SignalLike<number | string | undefined>;
```
rschristian, line 205: This simply does not exist as far as I can tell.

This type was originally copied from React's types, and it was [added to React's types as part of a big batch](https://github.com/DefinitelyTyped/DefinitelyTyped/commit/78edcdb985e172601d7788919fecb9d0ea16252f). Best guess, this was incorrectly picked up from `rotate="auto-reverse"`.

JoviDeCroock (approved):

PR #4514, opened 2024-09-28
Use a ref to track the hydrating state

JoviDeCroock:
The main reason for this change is so that we discontinue hydration as soon as a mismatch is detected. For deferred subtrees this won't be a problem but for synchronous trees it's problematic that we can have the following

```
<Fragment>
  <span>Hello world</span>
</Fragment>
<span>Foo</span>
```

If the span mismatches we'll still try to hydrate Foo which could mean missing props/... This is more a proposal in follow up from https://github.com/preactjs/preact/pull/4490

One thing I need to check is whether this opts out of hydration for a mismatched text node as that would be undesirable.

Might even be worth to do it differently to throw and use a regular render 😅

### DIFF
--- a/test/browser/hydrate.test.js
+++ b/test/browser/hydrate.test.js
@@ -238,6 +238,22 @@ describe('hydrate()', () => {
 		expect(scratch.innerHTML).to.equal('<div class="foo">bar</div>');
 	});
 
+	it('should bail for adjacent trees', () => {
+		scratch.innerHTML = '<main><div>Hello world</div></main>';
+		hydrate(
+			<main>
+				<Fragment>
+					<span>Foo</span>
+				</Fragment>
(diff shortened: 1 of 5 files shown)


test/browser/hydrate.test.js
```diff
@@ -238,6 +238,22 @@ describe('hydrate()', () => {
...
 	});
 
+	it('should bail for adjacent trees', () => {
+		scratch.innerHTML = '<main><div>Hello world</div></main>';
+		hydrate(
+			<main>
+				<Fragment>
+					<span>Foo</span>
+				</Fragment>
+				<div test-id="x">Hello world</div>
```
JoviDeCroock, line 248: On main this prop would not be set for instance

Thompson1985 (approved):

PR #4520, opened 2024-10-01
ci: Unify workflows into singular pipeline

rschristian:
Currently, we run 3 separate pipelines for each PR (assuming the PR touches source JS, we do have a bit of filtering in play):
  - CI - builds, lints & tests, packages Preact
  - Benchmarks - CI workflow, sets up source & current Preact tarballs, runs benches
  - Compressed Size

As such, we're running the CI workflow twice (once directly, once called from Benchmarks) which, while not necessarily problematic as it's done in parallel, can be a bit noisy in the case of build/test failures (each will notify of failure separately) and is duplicated work at the very least. 3 separate pipelines is also a bit messy if you need to jump into the 'Actions' tab to check/compare different runs -- far from a major problem, but less than ideal IMO.

---

The idea here is to unify the aforementioned workflows into a singular pipeline; we'll get an overview of all jobs at once, make the control flow a bit more obvious & easier to edit, and skip some duplicated work. `ci.yml` will therefore be the orchestrator, triggering the runs of `build-test.yml`, `benchmarks.yml`, & `size.yml`, they'll no longer be entrypoints/pipelines themselves.

Workflow overview diagram

[Workflow run in pic](https://github.com/preactjs/preact/actions/runs/11137485735)


rschristian:
The required status checks in the repo will need to be altered if this is to land; `Build & Test` is no longer ran directly, so the status will never be reported.

JoviDeCroock (approved):

Thompson1985 (approved):

Issue #2618, opened 2020-07-11
App statistics: Help us make Preact even faster


marvinhagemeister:
# :rocket:  Making Preact even faster

Good news! We're experimenting with various ways to speed up Preact even more :100: To do so we have various enhancements and optimizations planned for our reconciler. So far we've come up with multiple different approaches that are battling against each other to claim the speed title or be less memory intensive.

Although we do have a small selection of apps to test the various optimizations, we want to make sure that we optimize the things that actually matter most to everyone. We some intuitions what that might be, but it's always better to have some solid data!

For this reason we've added a new "Statistics" panel to the Preact Devtools extensions. It allows you to measure statistics for our renderer and will help us get a picture of which optimizations will matter the most.

## 👋 How you can help

To start you need to have the [Preact Devtools](https://preactjs.github.io/preact-devtools/) extension installed. Only the latest version (`1.1.0`) has the "Statistics"-tab. The extension is currently waiting approval and you can see the status for your browser below:

- [ ] Chrome: Pending Review
- [ ] Edge: Pending Review
- [ ] Firefox: Pending Review

Now we're ready to go!

1. Open Preact Devtools and go to the "Statistics" tab
2. Click the red "Record" button or the "Reload and record stats" button right next to it
3. Navigate through your app like a typical user would
4. Post a screenshot of the "Statistics"-Panel in this thread.

Here is an example of what it looks like after measuring [preactjs.com](https://preactjs.com) for a while by clicking through various pages:

![Screenshot from 2020-07-11 16-23-18](https://user-images.githubusercontent.com/1062408/87226271-ebfba300-c392-11ea-9ca3-75bf30308436.png)

marvinhagemeister:
Here are some statistics for the "People browser" in our demo app:

![Screenshot from 2020-07-11 17-19-16](https://user-images.githubusercontent.com/1062408/87227366-c2467a00-c39a-11ea-97df-33c90daef82c.png)

srdjan:
<img width="821" alt="Screen Shot 2020-07-11 at 11 47 54 AM" src="https://user-images.githubusercontent.com/61190/87228003-67962980-c36c-11ea-9891-3ce8b67d63d6.png">

marvinhagemeister:
Stats from the virtualized table from #2619 :

![Screenshot from 2020-07-11 22-46-21](https://user-images.githubusercontent.com/1062408/87233514-8cb88580-c3c8-11ea-927e-8f272e87004e.png)

[...]

PR #4523, opened 2024-10-03
Single text node hot path

JoviDeCroock:
Supersedes https://github.com/preactjs/preact/pull/3939

Similar to 3939 we are seeing some operations take longer as rather than mutating the text-node we'll be swapping positions instead. The impact seems higher in skew based diff though so not sure whether the performance impact will be as high as in 3939.

The replace_1k difference is similar to https://github.com/preactjs/preact/pull/3939#issuecomment-1474518958 

This could also be problematic with signals as text-updates won't automatically update afaik

Looks like this should not break signals actually
image

From https://github.com/preactjs/preact/issues/2618 we can derive that there are actually a lot of cases where the single child type is a text-node.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -486,7 +486,6 @@ function diffElementNodes(
 			}
 		}
 
-		// If the new vnode didn't have dangerouslySetInnerHTML, diff its children
 		if (newHtml) {
 			// Avoid re-applying the same '__html' if it did not changed between re-render
 			if (
@@ -499,6 +498,15 @@ function diffElementNodes(
 			}
 
 			newVNode._children = [];
+		} else if (
+			typeof newChildren === 'string' &&
+			typeof oldProps.children === 'string'
+		) {
+			if (newChildren !== oldProps.children) {
+				oldVNode._children[0]._dom.data = newChildren;
+				newVNode._children = oldVNode._children;
+				oldVNode._children = null;
+			}
 		} else {
 			if (oldHtml) dom.innerHTML = '';
 
--- a/test/browser/fragments.test.js
+++ b/test/browser/fragments.test.js
@@ -1325,6 +1325,7 @@ describe('Fragment', () => {
 		);
 	});
 
+	// TODO
 	it('should support moving Fragments between beginning and end', () => {
 		const Foo = ({ condition }) => (
 			<ol>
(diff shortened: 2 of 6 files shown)


Thompson1985 (approved):

Thompson1985 (approved):

rschristian (approved):
Re:Andre's comment about the "regression" in replace1k, have you by chance tried giving it more warmup runs? Say bumping up to 25 like a few of the other benches?

That very well might have been in one of your force pushes, just thought I'd ask.

benches/src/02_replace1k.html
```diff
@@ -35,7 +35,7 @@
 				const elementSelector = 'tr:first-child > td:first-child';
 
 				// MUST BE KEPT IN SYNC WITH WARMUP COUNT IN benches/scripts/config.js
-				const WARMUP_COUNT = 5;
+				const WARMUP_COUNT = 25;
```
rschristian, line 38: Per that comment, might be a problem here.

Benchmarks are only ran by the CI if that particular commit contains changes to `**/src/**.js`... that's not great. Will see if I can fix that to skip bogus comments needing to be added & reverted.

rschristian:
Looking at the results, the warmup might close the gap _a little_ on average, but not totally: https://github.com/preactjs/preact/actions/runs/11304050028/job/31441949188

Can probably revert/drop that.

marvinhagemeister (approved):
Very impressed with the findings. Really good work!

src/diff/index.js
```diff
@@ -499,6 +498,22 @@ function diffElementNodes(
...
+				while (oldVNode._children && (i = oldVNode._children.pop())) {
+					// Setting textContent on the dom element will unmount all DOM nodes
+					// of the previous children, so we don't need to remove DOM in this
+					// call to unmount
+					unmount(i, oldVNode, true);
+				}
+
+				newVNode._children = [
+					// @ts-expect-error
+					createVNode(null, (dom.textContent = newChildren), null, null)
```
developit, line 512: We might want to consider using `dom.appendChild(document.createTextNode(newChildren))` and manually removing previous children. Otherwise, rendering a single text child will blow away non-preact-rendered children from a parent element.

src/diff/index.js
```diff
@@ -499,6 +498,22 @@ function diffElementNodes(
...
+				];
+				// @ts-expect-error
+				newVNode._children[0]._dom = dom.firstChild;
+			}
```
developit, line 516:
```suggestion
		} else if (typeof newChildren === 'string' && typeof oldProps.children === 'string') {
			if (newChildren !== oldProps.children) {
				oldVNode._children[0]._dom.data = newChildren;
				newVNode._children = oldVNode._children;
				oldVNode._children = [];
			}
```

test/browser/fragments.test.js
```diff
@@ -1325,6 +1325,7 @@ describe('Fragment', () => {
 		);
 	});
 
+	// TODO
```
JoviDeCroock, line 1328:
```suggestion
```

JoviDeCroock:
Closing due to the memory regression

Issue #4344, opened 2024-04-19
Preact 10.20 + @mui/x-data-grid v7 causes datagrid to appear multiple times in dev-mode, but not in preview-mode


vsacom:
- [X] Check if updating to the latest Preact version resolves the issue
Using version: 10.20.2

**Describe the bug**

Environment: Preact 10.20.2 + Typescript 5.4.5 + Vite 5.2.9 + @mui/x-data-grid 7.2.0.

Using vite+preact in development mode, the x-data-grid appears multiple times on the page.
With the same setup, running vite --preview, the x-data-grid appears once as expected.

Bug report has been opened at @mui/mui-x, where maintainer has requested to open ticket here. Link to ticket: https://github.com/mui/mui-x/issues/12829

**To Reproduce**

Codesandbox: https://codesandbox.io/p/devbox/datagridpro-test-7-2-0-47rlrc

Sample source code based on the codesandbox:
[datagrid_720_preact.zip](https://github.com/preactjs/preact/files/15039590/datagrid_720_preact.zip)

With the sample source code, run:
1. npm install
2. npm run dev -> The output should have just one datagrid, but has multiple overlapping.
3. npm run preview --> The output appears as expected.

**Expected behavior**
The datagrid should appear just once

rschristian:
Looks to be using `NaN` in a dependency array, might be related to https://github.com/preactjs/preact/pull/3954

JoviDeCroock:
@rschristian maybe we should convert it to a `console.warn` rather than throwing an error, in libraries users can't change much but if it works they should just be warned.

PR #4527, opened 2024-10-12
refactor: Warn on NaN in dep arrays instead of throwing

rschristian:
Closes: #4344

It'll fix the issue of debug throwing & killing the app, but whether or not this works correctly is still very usage-dependent.

Will depend on #4528, as our benchmarks require a successful asset upload from `main` to be able to run, and its CI currently fails.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -19,7 +19,7 @@ jobs:
         uses: andrewiggins/download-base-artifact@v3
         with:
           artifact: npm-package
-          workflow: build-test.yml
+          workflow: ci.yml
           required: true
       - run: mv preact.tgz preact-main.tgz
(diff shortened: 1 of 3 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         uses: andrewiggins/download-base-artifact@v3
         with:
           artifact: npm-package
-          workflow: build-test.yml
+          workflow: ci.yml
```
rschristian, line 22: The CI should now be completely on track again.

Sorry that was so hairy, the fact that our setup runs both the PR & main workflows (as they exist on both branches) simultaneously made this a bit weird to figure out all the details on.

JoviDeCroock (approved):

PR #4528, opened 2024-10-12
ci: Ensure paths-filter action works on push events

rschristian:
It seems `dorny/paths-filter` (which we use for the `CI` workflow) specifically fails on `push` events, [unable to find the Git repo](https://github.com/preactjs/preact/actions/runs/11210372482/job/31439621624). Adding [an explicit checkout stage seems to be the fix](https://github.com/dorny/paths-filter/issues/212#issuecomment-1960976719).

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -20,7 +20,7 @@ jobs:
         with:
           artifact: npm-package
           workflow: build-test.yml
-          required: false
+          required: true
       - run: mv preact.tgz preact-main.tgz
       - name: Upload locally build & base preact package
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         with:
           artifact: npm-package
           workflow: build-test.yml
-          required: false
+          required: true
       - run: mv preact.tgz preact-main.tgz
```
rschristian, line 24: This is very much required, as the very next step (as well as other steps later in the flow) require that asset to exist. Therefore, setting `required: false` just moves the error lower, to the attempt to move the (non-)extracted asset.

JoviDeCroock (approved):

Issue #4439, opened 2024-07-09
Suspense / ErrorBoundary content rendered twice when hydrating with render


PhilippMi:
- [x] Check if updating to the latest Preact version resolves the issue

**Describe the bug**

We use `ErrorBoundary` and `lazy` from [preact-iso](https://github.com/preactjs/preact-iso/blob/1337db0f2807ae362fb5e71e9cbc8ced55979d31/src/lazy.js) for code-splitting our application but still being able to server-side render them in advance. With the upgrade to >10.19.0, we noticed that all our lazy components are rendered twice when first server-side rendered.

One important thing to note is that we use `render` instead of `hydrate` for hydrating our server-side rendered HTML because the server-side markup differs from what we render on the client. The bug does not appear when using `hydrate`. But changing that behaviour in our application would take an immense effort.

Switching from `ErrorBoundary` to `Suspense` did not solve the bug.

**To Reproduce**

1. Go to https://codepen.io/phmitterer/pen/eYwYrpB 
2. Server-side rendering is simulated and "Hello, World!" is rendered
3. After 1 second, client-side hydration kicks in and renders a second "Hello, World!"

Here is the same bug with `Suspense` instead of `ErrorBoundary`: https://codepen.io/phmitterer/pen/oNrNOVJ

**Expected behavior**

"Hello, World!" is only rendered once.

JoviDeCroock:
Hmm, this seems related to https://github.com/preactjs/preact/pull/4180 what seems to happen is that we opt out of resumed hydration due to leveraging `render`. When we initiate render we correctly re-use the initial components until we hit the lazy-boundary, it bubbles up the Promise to the error-boundary/suspense and then fails to recover the correct `oldDom` instead we are looking at either nothing or a disconnected node (the fallback) which is not dissimilar to what we've seen in other issues relating to Suspense.

I agree that we def should not duplicate the UI, I currently can't look at this but this can probably be reproduced in tests.

PR #4529, opened 2024-10-12
We should not always set to hydration

JoviDeCroock:
Fixes https://github.com/preactjs/preact/issues/4439

Currently we try to set everything to `hydrate` mode which feels like a mistake, we should only set the hydration flag when we are hydrating. I feel like we intended to set the SUSPENDED flag. This will make it so that we try to render the fallback rather than falling back to a non-suspended hydration mode, falling back to non-suspended hydration will clobber the DOM as we fall back to no excessDomChildren which is an inherent hydration mismatch.

This is so because we skip [this condition](https://github.com/preactjs/preact/blob/main/src/diff/index.js#L53)

### DIFF
--- a/compat/test/browser/suspense-hydration.test.js
+++ b/compat/test/browser/suspense-hydration.test.js
@@ -15,6 +15,7 @@ import {
 } from '../../../test/_util/helpers';
 import { ul, li, div } from '../../../test/_util/dom';
 import { createLazy, createSuspenseLoader } from './suspense-utils';
+import { render } from 'preact';
 
 /* eslint-env browser, mocha */
 describe('suspense hydration', () => {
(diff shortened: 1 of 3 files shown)


compat/test/browser/suspense-hydration.test.js
```diff
@@ -15,6 +15,7 @@ import {
 } from '../../../test/_util/helpers';
 import { ul, li, div } from '../../../test/_util/dom';
 import { createLazy, createSuspenseLoader } from './suspense-utils';
+import { render } from 'preact';
```
JoviDeCroock, line 18:
```suggestion
```

rschristian (approved):

PR #4530, opened 2024-10-12
ci: Remove saucelabs

rschristian:
Let me know if anyone is opposed to this, but it's been failing for the last 10 months straight, with 4 working commits, then another 5 months of failures which is where the actions history runs out.

Looks like `karma-sauce-launcher` has been abandoned w/ no changes in the last 2 years, and is 4 majors behind the current `saucelabs` version. Per NPM, the karma launcher down to half the downloads per week compared to last year, so I'm guessing everyone else is bailing off it too.


JoviDeCroock (approved):

package.json
```diff
@@ -217,6 +217,7 @@
     "benchmark": "^2.1.4",
     "chai": "^4.1.2",
     "check-export-map": "^1.3.1",
+    "core-js": "^3.38.1",
```
rschristian, line 220: We rely on `core-js` in our polyfill tests:

https://github.com/preactjs/preact/blob/c9846f13393baef888629c219367ba23c5f2e739/test/polyfills.js#L2

Undeclared dep, hence the additon here.

PR #4531, opened 2024-10-13
ci: Fix PR reporter

rschristian:
This should correct the PR reporter

There's no elegant way to watch jobs within a workflow, so we'll have to settle for mirroring CI's filter here to ensure we're only running the PR reporter when the benchmarks will be ran (`**/src/**/*.js`)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,8 +22,9 @@ jobs:
       - uses: dorny/paths-filter@v3
         id: filter
         with:
+          # Should be kept in sync with the filter in the PR Reporter workflow
           filters: |
-            jsChanged: '**/src/**.js'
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -22,8 +22,9 @@ jobs:
       - uses: dorny/paths-filter@v3
         id: filter
         with:
+          # Should be kept in sync with the filter in the PR Reporter workflow
           filters: |
-            jsChanged: '**/src/**.js'
+            jsChanged: '**/src/**/*.js'
```
rschristian, line 27: As we found in https://github.com/preactjs/preact/pull/4523/commits/ad1fdf50c8ce83b53dcd9a9a33a24fd7abcc654b

PR #4538, opened 2024-10-28
Support `{ handleEvent() {} }` object interface as a listener

### LINKED ISSUE #2382: Support for handleEvent DOM interface
### Reproduction

```jsx
class Foo {
  click(e) {
	this.setState({ clicked: true })
  }

  handleEvent(e) {
    // nicety: `this` is the component
    this[e.type](e)
  }
  render() {
    return <button onClick={this} />
  }
}
```

### Steps to reproduce

Click the button.

### Expected Behavior

handleEvent is used and doesn't throw an error.

### Actual Behavior

`this.l[e.type](P.event ? P.event(e) : e)` throws an error, as `this.l[e.type]` is `this`.

lilnasy:
- Implements and closes #2382
- ~~Currently a draft because I couldn't get tests to run.~~
- Introduces a 13 byte increase in brotli'd bundle size (4247b -> 4260b)
- [x] Update typescript declarations
- [x] Tests for new behavior

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -1480,9 +1480,13 @@ export namespace JSXInternal {
 	export type TargetedPictureInPictureEvent<Target extends EventTarget> =
 		TargetedEvent<Target, PictureInPictureEvent>;
 
+	export type EventHandlerObject<E extends TargetedEvent> = {
+		handleEvent(e: E): unknown
+	}
+
 	export type EventHandler<E extends TargetedEvent> = {
 		bivarianceHack(event: E): void;
-	}['bivarianceHack'];
+	}['bivarianceHack']  | EventHandlerObject<E>;
 
 	export type AnimationEventHandler<Target extends EventTarget> = EventHandler<
 		TargetedAnimationEvent<Target>
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -166,7 +166,7 @@ function createEventProxy(useCapture) {
 			} else if (e._dispatched < eventHandler._attached) {
 				return;
 			}
-			return eventHandler(options.event ? options.event(e) : e);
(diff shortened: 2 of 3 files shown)


marvinhagemeister:
To run the tests:

1. Run `npm install` (other package managers won't work)
2. Run `npm run test:karma:watch` to run the tests.

lilnasy:
Thanks, I had to install chrome using playwright but I could run the tests with that command!

src/jsx.d.ts
```diff
@@ -1480,9 +1480,13 @@ export namespace JSXInternal {
...
 		TargetedEvent<Target, PictureInPictureEvent>;
 
+	export type EventHandlerObject<E extends TargetedEvent> = {
+		handleEvent(e: E): unknown
+	}
+
 	export type EventHandler<E extends TargetedEvent> = {
 		bivarianceHack(event: E): void;
-	}['bivarianceHack'];
+	}['bivarianceHack']  | EventHandlerObject<E>;
```
lilnasy, line 1489: This worked for my projects, but I'm still cautious of changing an exported type. It might be necessary to leave this intact and instead add to the attribute interfaces one-by-one.

marvinhagemeister (approved):
Sweet, thanks!

src/jsx.d.ts
rschristian: This seems to cause issues elsewhere (see #4581), though I'm not quite sure what it's even doing? If you have a moment, would love an explanation. A lot of this TS stuff flies right over my head

lilnasy: I'll reply in the issue.

rschristian: Cheers, and no rush! If that's incorrect usage then fair enough too! I'm very much unfamiliar with what this is doing.

PR #4541, opened 2024-10-31
Re-export ErrorInfo type

jose-torres-marin:
This type is useful when creating an ErrorBoundary. See #3452


rschristian (approved):
Thanks!

PR #4546, opened 2024-11-08
types: Move per-element type interfaces into core and more strictly type IntrinsicElements

### LINKED ISSUE #4544: Types: Make per-element types consistent
At the moment, core doesn't discriminate attribute/property types by HTML element, instead, adding them to a huge singular interface. Compat has recently added most per-element interfaces for types, though the HTML elements all still use the singular interface from core. Strangely, core has ended up with MathML element interfaces though.

We should make this more consistent. Per-element types shouldn't impact maintainability for the better or worse really, but would appeal to many who want stricter types for their elements (see #4202). Moving all the attribute types over to core and attaching `InstrinsicElements` up to them seems like a good move at this point, though there are some risks of type breakages in the ecosystem.

rschristian:
Closes #4544 and #4202

Someone kindly contributed type interfaces for most elements to compat, so this PR takes those, moves them into core, and alters `IntrinsicElements` to use those more specific interfaces rather than the massive & generic `HTMLAttributes` that we used to have. That interface still exists, however, it's been renamed to `AllHTMLAttributes`, matching React. The new `HTMLAttributes` is strictly limited to global attributes/props, only the things that are valid & make sense everywhere.

As such, with this PR `<div src="/foo.jpg">` will now correctly result in a type error. I'm slightly concerned about breakages because of this but I think it's an unavoidable bandaid we'll have to rip off at some point.

This is a fairly big change and I expect we will have some issues raised. There's 240 or so attributes now spread across a couple dozen interfaces. I've done my best to ensure this is correct but there's a lot of surface area here.

### DIFF
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -149,7 +149,6 @@ declare namespace React {
 	export import TimeHTMLAttributes = JSXInternal.TimeHTMLAttributes;
 	export import TrackHTMLAttributes = JSXInternal.TrackHTMLAttributes;
 	export import VideoHTMLAttributes = JSXInternal.VideoHTMLAttributes;
-	export import WebViewHTMLAttributes = JSXInternal.WebViewHTMLAttributes;
 
 	// Events
 	export import TargetedEvent = JSXInternal.TargetedEvent;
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):
WOW, this is amazing

marvinhagemeister (approved):
This is great!

titoBouzout:
Nice, I have updated the table to include the new types 
https://potahtml.github.io/namespace-jsx-project/index.html

rschristian:
Awesome, thanks! Jovi & I were both looking at that, will take another look soon-ish to see which types we're missing (that we do support).

Issue #3219, opened 2021-07-06
lazy produces TypeScript error: `JSX element type 'Component' does not have any construct or call signatures`


birtles:
- [ ] Check if updating to the latest Preact version resolves the issue (it does not -- 10.5.13 is fine, but 10.5.14 is broken for me)

**Describe the bug**

As of 10.5.14 code like the following:

```ts
const MyComponent = lazy(() =>
  import('./my-component').then((m) => {
    return {
      default: m.MyComponent,
    };
  })
);
```

may begin to error with: `TS2604: JSX element type 'MyComponent' does not have any construct or call signatures`

#3139 changed the type of `lazy` as follows:

```diff
- export function lazy<T>(loader: () => Promise<{ default: T }>): T;
+ export function lazy<T>(loader: () => Promise<{ default: T } | T>): T;
```

From what I can tell, in the above code fragment, TypeScript tries to infer the type of `T` by matching the return value of the function against the newly added `T` instead of `{ default: T }` and hence resolves the return type of `lazy` as `{ default: React.FunctionComponent<Props> }` _not_ `React.FunctionComponent<Props>` as it previously did.

For the time being it seems possible to work around this by explicitly providing the generic type parameter such as follows:

```ts
import type { Props as MyComponentProps } from './my-component';

const MyComponent = lazy<React.FunctionComponent<MyComponentProps>>(() =>
  import('./my-component').then((m) => {
    return {
      default: m.MyComponent,
    };
  })
);
```

JoviDeCroock:
This has been implemented in #3139

birtles:
> This has been implemented in #3139

Is that right? #3139 is the regressing issue, not the fix.

JoviDeCroock:
You are completely right

PR #4547, opened 2024-11-09
Ensure we unwrap custom .then()

JoviDeCroock:
Fixes https://github.com/preactjs/preact/issues/3219

Before this fix we would return `{ default: JSX.Element }` which is obviously wrong 😅


rschristian (approved):
Neat type and TIL! Somehow I've never ran across `infer` before

Issue #4425, opened 2024-06-27
JSX type errors on React Bootstrap 2.x & Preact 10.x


TechQuery:
- [x] Check if updating to the latest Preact version resolves the issue

## Describe the bug

Every React Bootstrap component with Preact has a similar error as shown below:

```text
'Container' cannot be used as a JSX component.
  Its type 'BsPrefixRefForwardingComponent<"div", ContainerProps>' is not a valid JSX element type.
    Type 'BsPrefixRefForwardingComponent<"div", ContainerProps>' is not assignable to type 'FunctionComponent<any>'.
      Type 'ComponentChild' is not assignable to type 'VNode<any>'.
        Type 'string' is not assignable to type 'VNode<any>'.ts(2786)
```

## To Reproduce

- Real World scaffold: https://github.com/idea2app/React-MobX-Bootstrap-ts
- Minimal reproduce: https://codesandbox.io/p/devbox/preact-react-bootstrap-llqsv9?file=%2Fsrc%2Findex.tsx%3A5%2C14

### Steps to reproduce the behavior

1. `git clone https://github.com/idea2app/React-MobX-Bootstrap-ts`
2. `npm i`
3. `npx tsc --noEmit`

## Expected behavior

No type errors as React does.

rschristian:
Please provide a _minimal_ reproduction, it's not fair to ask maintainers to browser through a full project.

TechQuery:
> Please provide a _minimal_ reproduction, it's not fair to ask maintainers to browser through a full project.

That repository is not a full project, it's just a scaffold with some simple example pages, but almost every page & component has the same type errors...

rschristian:
Anything more than a few files in `src` is non-minimal. Please clear it out to the minimum that is required to demonstrate the issue.

TechQuery:
> [This is the example StackBlitz we provide in the bug template](https://stackblitz.com/edit/create-preact-starter), which is just about ideal for a web reproduction.

Thanks, but Parcel 2.x is my real bundler, not Vite: https://codesandbox.io/p/devbox/preact-react-bootstrap-llqsv9?file=%2Fsrc%2Findex.tsx%3A5%2C14

rschristian:
Build tooling isn't likely to matter in a types issue, but regardless, it's just an example of scope that's appropriate. But thanks, the minimal repro does help.

[...]

Issue #4113, opened 2023-08-23
Types: Unable to return `ComponentChild` from component


rschristian:
- [X] Check if updating to the latest Preact version resolves the issue

**Describe the bug**

[From Slack](https://preact.slack.com/archives/C3NMBSJKH/p1692741376856869)

**To Reproduce**

[TS Playground Link](https://www.typescriptlang.org/play?ssl=11&ssc=8&pln=1&pc=1#code/PQKgBAAgVgzgHmAxgJwKYEMAuqCiAbVAW1QDtMwRgAoAS0IAcB7ZcgbyrDAGFGHGTSmLgAsaeACYAaDkjRZcBYmWkBfMADNkvMAHJ6cxJh0BuKlUT8Y5ACqorYALxgAFKySiJaEmBUAuMG6IHuJe-jx8AmQiYuI+AJSOAHzuMV6mVAA8tlaJMgCMAEwAzJnA2ZiJQA)

The React equivalent, using `ReactNode`, works just fine. [TS Playground Link](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAJQKYEMDGMA0cDejUYByEAJknAL5wBmUEIcA5FATIwNwBQnaEAdgGd4AFSRC4AXjgAKPGgAWwADYkWfSgC5ccBctVI+W5OhjEylAJSSAfDsUq1XTgB5RQ65zhwAjACYAzC4A9G4w1kA)

I believe we need to change [this line](https://github.com/preactjs/preact/blob/099c38c6ef92055428afbc116d18a6b9e0c2ea2c/src/index.d.ts#L90) to return a `ComponentChild` [as React does](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/05094b6df4fb6f4404333725f13d63dd8aabeb34/types/react/index.d.ts#L563) but this seems to break a whole bunch of our types.

augustjk:
@rschristian The change you proposed of updating return type of `FunctionComponent` to include `ComponentChild` does work but only on TS >= 5.1 due to this change https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-1.html#decoupled-type-checking-between-jsx-elements-and-jsx-tag-types

[...]

Issue #3815, opened 2022-11-24
Type 'VNode<any> | VNode<any>[]' is not assignable to type 'VNode<any> | null'


zim32:
- [x ] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
When using htm with typescript I have types incompatibility

```
import {FunctionalComponent, h} from 'preact'
import htm from 'htm'

const html = htm.bind(h)

const TestComponent: FunctionalComponent<{ foo: number }> = ({ foo }) => {
    return html`<div>${foo}</div>`
}
```

Here typescript complains that 

`TS2322: Type '({ foo }: RenderableProps<{ foo: number; }, any>) => VNode<any> | VNode<any>[]' is not assignable to type 'FunctionalComponent<{ foo: number; }>'.   Type 'VNode<any> | VNode<any>[]' is not assignable to type 'VNode<any> | null'.     Type 'VNode<any>[]' is missing the following properties from type 'VNode<any>': type, props, key`

**Expected behavior**
No type error.

Doing 

```
return html`<div>${foo}</div>` as VNode<any>
```

resolves this issue but I think it will be better to change htm or FunctionalComponent declaration.

PR #4548, opened 2024-11-10
Fork types for TS 5.1 and beyond

JoviDeCroock:
In TypeScript 5.1 there was a change to the JSX types, to support this properly we'll need to fork our types in pre and post this minor version.

I'm pretty unsure whether we can add tests for this in our repository 😅 

Resolves https://github.com/preactjs/preact/issues/4113
Resolves https://github.com/preactjs/preact/issues/4425
Resolves https://github.com/preactjs/preact/issues/3611
Resolves https://github.com/preactjs/preact/issues/3815
Enables https://github.com/preactjs/preact/issues/4481
Supersedes https://github.com/preactjs/preact/pull/3482

### DIFF
--- a/package.json
+++ b/package.json
@@ -254,7 +254,7 @@
     "sade": "^1.8.1",
     "sinon": "^9.2.3",
     "sinon-chai": "^3.7.0",
-    "typescript": "^4.9.5",
+    "typescript": "^5.6.3",
     "undici": "^4.12.0"
   },
   "volta": {
--- a/package.json
+++ b/package.json
@@ -9,9 +9,16 @@
   "umd:main": "dist/preact.umd.js",
   "unpkg": "dist/preact.min.js",
   "source": "src/index.js",
+  "typesVersions": {
+    "<=5.0": {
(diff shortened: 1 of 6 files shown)


rschristian (approved):
Think we just need to bump our TS devDep to 5.1 and we should be good

package.json
```diff
@@ -246,7 +254,7 @@
     "sade": "^1.8.1",
     "sinon": "^9.2.3",
     "sinon-chai": "^3.7.0",
-    "typescript": "^4.9.5",
+    "typescript": "^5.6.3",
```
rschristian, line 257: One last minor thing: do we want to jump the this all the way to 5.6? One somewhat useful feature of using a less-than-current version of TS is that it's a warning light when we/a contributor goes to use a newer feature that'd require end users upgrade their TS version just to consume our types.

Right now we have an actual minimum (on the v5.1+ file) of 5.1 but that can easily & accidentally change without us realizing it if the TS version is set to 5.6 here.

Just a thought though, up to you if we want to set this as the new minimum.

Issue #4406, opened 2024-06-11
`package.json#exports.browser` does not adhere to Node module type semantics


mcous:
- [ ] Check if updating to the latest Preact version resolves the issue

**Describe the bug**

Hello! I'm coming here from testing-library/svelte-testing-library#381, where a user reported interop issues when using `preact` together with our `svelteTesting` Vite plugin. Our plugin makes one significant change to the Vite config: it adds `browser` to `resolve.conditions` to ensure Svelte's browser code is imported.

This also causes Preact's `browser` export - `./dist/preact.module.js` - to be used:

```ts
{
  "exports": {
    ".": {
      "types": "./src/index.d.ts",
      "browser": "./dist/preact.module.js",
      "umd": "./dist/preact.umd.js",
      "import": "./dist/preact.mjs",
      "require": "./dist/preact.js"
    }
  }
}
```

The bug is:

- `preact.module.js` is ESM
- Preact's `package.json` does not specific `type`, so `"type": "commonjs"` is implied
- Because `preact.module.js` ends in `.js` rather than `.mjs`, it is treated as CommonJS

At worst, this produces a fairly inscrutable error:

```
SyntaxError: Named export 'options' not found. The requested module 'preact' is a
CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'preact';
const { options } = pkg;
```

With enough fiddling, I can get Vite to complain in this manner:

```
Module /Users/mc/cloned/vitest-preact-svelte/node_modules/preact/dist/preact.module.js:2 
seems to be an ES Module but shipped in a CommonJS package. You might want to create an 
issue to the package "preact" asking them to ship the file in .mjs extension or add
"type": "module" in their package.json.
```

If I go in and manually rename files to `.module.mjs` and update the `package.json`, everything works happily.

**To Reproduce**

I can put together a minimal reproduction if required - just let me know. There is also this repro from the original reporter: https://github.com/molily/vitest-preact-svelte

**Expected behavior**

I should be able to load the `browser` copy of Preact in Vitest without issue

mcous:
Thanks for the detailed response and additional tickets! Apologies that my cursory search didn't turn them up. I definitely see (and agree with much of) where you're coming from here. 

Out of curiosity, what are the differences between the `preact.module.js` and `preact.mjs`? It doesn't look to me like there are any, based on a quick diff.

[...]

PR #4549, opened 2024-11-10
V11 2

JoviDeCroock:


rschristian:
#4406 would be another easy thing to fit in, swapping out our (pretty much unused & problematic) `package.json##browser` condition for `package.json##module`

src/util.js
```diff
@@ -1,28 +1,5 @@
...
 export const isArray = Array.isArray;
-
-/**
- * Assign properties from `props` to `obj`
- * @template O, P The obj and props types
- * @param {O} obj The object to copy properties to
- * @param {P} props The object to copy properties from
- * @returns {O & P}
- */
-export function assign(obj, props) {
```
developit, line 12: We were using this because it's faster, probably the reason the benchmarks show a drop.

src/diff/index.js
```diff
@@ -302,7 +300,9 @@ export function diff(
 					newVNode._dom = oldDom;
 				} else {
 					for (let i = excessDomChildren.length; i--; ) {
-						removeNode(excessDomChildren[i]);
+						if (excessDomChildren[i]) {
+							excessDomChildren[i].remove();
+						}
```
developit, line 305:
```suggestion
						const child = excessDomChildren[i];
						if (child) child.remove();
```

src/diff/index.js
```diff
@@ -544,7 +544,7 @@ function diffElementNodes(
 			// Remove children that are not part of any vnode.
 			if (excessDomChildren != null) {
 				for (i = excessDomChildren.length; i--; ) {
-					removeNode(excessDomChildren[i]);
+					if (excessDomChildren[i]) excessDomChildren[i].remove();
```
developit, line 547:
```suggestion
					const child = excessDomChildren[i];
					if (child) child.remove();
```

src/diff/index.js
```diff
@@ -648,8 +643,8 @@ export function unmount(vnode, parentVNode, skipRemove) {
 		}
 	}
 
-	if (!skipRemove) {
-		removeNode(vnode._dom);
+	if (!skipRemove && vnode._dom != null && typeof vnode.type != 'function') {
```
developit, line 646: it's interesting that we weren't checking the type here before, and I'm not sure we should? Currently unmount was doing:

```js
unmount(vnode, removeDom = true) {
  if (IS_COMPONENT(vnode)) {
    // unmount (but do not remove) all children:
    for (child of vnode.children) unmount(child, removeDom: false)
    if (removeDom) vnode.dom.remove()
  }
  else if (removeDom === false) {
    vnode.dom.remove()
  }
}
```

So in the current algo, when unmounting a component, we remove the DOM node for each subtree when encountering its outermost vnode (regardless of whether its a component or element vnode).

I suspect we could revise this to instead only ever remove the DOM node for element vnodes, but if we're doing that I think this whole function could be simplified (one typeof at the top).

src/util.js
```diff
@@ -14,15 +15,3 @@ export function assign(obj, props) {
...
 	return /** @type {O & P} */ (obj);
 }
-
-/**
- * Remove a child node from its parent if attached. This is a workaround for
- * IE11 which doesn't support `Element.prototype.remove()`. Using this function
- * is smaller than including a dedicated polyfill.
- * @param {import('./index').ContainerNode} node The node to remove
- */
-export function removeNode(node) {
```
JoviDeCroock, line 24: @developit the perf regression still stands after undoing the `assign` change so I am thinking this might be related to `removeNode`

marvinhagemeister (approved):
LGTM

PR #4550, opened 2024-11-10
Recreate unkeyed functional components when they change position.

### LINKED ISSUE #2949: Strange difference in behaviour between preact and react in combination with @material-ui and conditional operator

JoviDeCroock:
With keyed children we can confidently track them as they move throughout an array of components. When this happens with unkeyed functional components we should not risk reusing the state as that could possibly make us re-use state in unrelated components, as seen in #2949.

With skew based diffing we can be sure that we account for offsets in our children arrays so this change should be quite safe.

Prior to this change we could have cases where

```
<Component="x" />
null

Becomes

null
<Component prop="y" />
```

We'd reuse the component state from the first instance for the second one which is dangerous, we should always recreate unkeyed children.

Now there are exceptions to this, imagine we insert a `<div />` in front of this then we want `<Component="x" />` to retain its state, thanks to our skew based diffing that will be the case.

Fixes #2949
Supersedes https://github.com/preactjs/preact/pull/2955

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -397,9 +397,13 @@ function findMatchingIndex(
 	// remainingOldChildren > 1 if the oldVNode is not already used/matched. Else
 	// if the oldVNode was null or matched, then there could needs to be at least
 	// 1 (aka `remainingOldChildren > 0`) children to find and compare against.
+	//
+	// If there is an unkeyed functional VNode, that isn't a built-in like our Fragment,
+	// we should not search as we risk re-using state of an unrelated VNode.
 	let shouldSearch =
+		(typeof type !== 'function' || type === Fragment || key) &&
 		remainingOldChildren >
-		(oldVNode != null && (oldVNode._flags & MATCHED) === 0 ? 1 : 0);
+			(oldVNode != null && (oldVNode._flags & MATCHED) === 0 ? 1 : 0);
 
 	if (
 		oldVNode === null ||
--- a/test/browser/render.test.js
+++ b/test/browser/render.test.js
@@ -1786,6 +1786,37 @@ describe('render()', () => {
 		);
 	});
 
+	// #2949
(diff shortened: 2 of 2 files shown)


rschristian (approved):
Damn, great find!

PR #4553, opened 2024-11-11
Remove _nextDom usage

JoviDeCroock:


andrewiggins (approved):
Nice! Yea the perf change is weird but this is a good change regardless I think. We can investigate that later

PR #4557, opened 2024-11-15
feat: Add `ElementRef` type to compat

### LINKED ISSUE #4481: Module `"preact/compat"` has no exported member `ElementRef`

**Describe the bug**
The `ElementRef` type is not available yet

**To Reproduce**
Check [this](https://stackblitz.com/edit/create-preact-starter-bt5jpo?file=src%2Findex.jsx) Stackblitz demo where i'm importing the `ElementRef` from `preact`. In the same line `render` is also imported. If you hover over `render` you'll see the type definition of it. But that's not the case for `ElementRef`. It doesn't show any type definition (idk why Stackblitz isn't showing any error there. but vscode does)

1. Run `npm init preact elementRefDemo`
2. Check `Typescript` when asked
3. Try importing `ElementRef` in `src/index.ts` from `preact` or `preact/compact`

**Expected behavior**
`ElementRef` should be available in `preact` or `preact/compact`

rschristian:
Closes #4481

For comparison: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b5dc32740d9b45d11cff9b025896dd333c795b39/types/react/index.d.ts#L221-L235

Can confirm #4548 cleared the way for this -- like a total doofus I created a new vite project and was wildy confused to see the same old error re: can't assign `null | undefined` to `VNode<any>` as I had only copied over this particular change. Eventually the issue dawned on me, copied over the new types, and it worked like a treat!

### DIFF
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -286,13 +286,25 @@ declare namespace React {
 		ComponentProps<T>
 	>;
 
-	export type ComponentPropsWithRef<
-		C extends ComponentType<any> | keyof JSXInternal.IntrinsicElements
-	> = C extends new (
+	export type ComponentPropsWithRef<C extends ElementType> = C extends new (
 		props: infer P
 	) => Component<any, any>
 		? PropsWithoutRef<P> & RefAttributes<InstanceType<C>>
 		: ComponentProps<C>;
+
+	export type ElementRef<
+		C extends
+			| ForwardRefExoticComponent<any>
+			| { new (props: any): Component<any, any> }
+			| ((props: any) => ReactNode)
+			| keyof JSXInternal.IntrinsicElements
+	> = 'ref' extends keyof ComponentPropsWithRef<C>
+		? NonNullable<ComponentPropsWithRef<C>['ref']> extends RefAttributes<
+				infer Instance
+			>['ref']
+			? Instance
+			: never
+		: never;
 
 	export function flushSync<R>(fn: () => R): R;
 	export function flushSync<A, R>(fn: (a: A) => R, a: A): R;


compat/src/index.d.ts
```diff
@@ -284,14 +284,26 @@ declare namespace React {
 		ComponentProps<T>
 	>;
 
-	export type ComponentPropsWithRef<
-		C extends ComponentType<any> | keyof JSXInternal.IntrinsicElements
-	> = C extends new (
+	export type ComponentPropsWithRef<C extends ElementType> = C extends new (
```
rschristian, line 287: We didn't originally have `ElementType` when I added `ComponentPropsWithRef` but someone added it a few months ago so it's worth switching this over, helps ensure things are kept in-sync.

For comparison: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b5dc32740d9b45d11cff9b025896dd333c795b39/types/react/index.d.ts#L1687-L1689

rschristian:
That's a great question. Will investigate further

rschristian:
Ignoring our resolution issues for the moment, this seems to work as we need? It does produce a type error in the compat declaration file, but in 5.0 & 5.1, the user is presented with a seemingly functional & identical type. This would require `skipLibCheck`, but most of our users are probably doing so anyhow (our Vite templates all do this by default).

Copy/pasting the shadcn examples given in the issue, there doesn't seem to be a user-facing problem, it's just that the generic doesn't satisfy. I'm not well enough informed to say whether this is a problem or not though.

jakebailey:
I'm currently trying to chase down why so much of DT is broken right now due to preact; I can definitely say that you do not want to end up a situation where the types do not check without `skipLibCheck`. DefinitelyTyped is by definition declaration files which can only be checked without `skipLibCheck` so it's critical that all packages there and their dependencies contain no errors.

jakebailey:
If you don't want to support TS 5.0, then theoretically we can bump all of the packages on DT up to a minimum of TS 5.1, though that may of course break someone.

rschristian:
> you do not want to end up a situation where the types do not check without `skipLibCheck`

This has always been the case in Preact due to the limited aliasing mechanisms we have in TS; we don't provide every single React type, but do need consumers to use our types over `@types/react`, so `skipLibCheck` seems like it'll be necessary forever?

As for issues on DefinitelyTyped we'd love to know more -- last release we did a fair few type changes (longstanding issues) and while we were a bit worried about doing it, we found no issues ourselves. Is it perhaps the added strictness in per-element types or related to `FunctionComponent` now returning a `ComponentChild` in TS 5.1?

rschristian:
Re:`skipLibCheck`, I think I misunderstood you.

We've held out on merging this due to that type error, as it seems TS won't use `typeVersions` or the equivalent in our export map when resolving submodules like in this case. We need `preact/compat` to resolve either the TS 5.0 `preact` types or the 5.1+ types, though haven't been able to figure it out. Using the module specifier (i.e., `preact`) doesn't seem to work. Would we need to make TS 5.0 & 5.1+ splits for everything that consumes the split types too?

jakebailey:
> As for issues on DefinitelyTyped we'd love to know more -- last release we did a fair few type changes (longstanding issues) and while we were a bit worried about doing it, we found no issues ourselves.

See: https://github.com/DefinitelyTyped/DefinitelyTyped/actions/runs/12009667461

Most failures here are due to those changes, it seems.

jakebailey:
> We've held out on merging this due to that type error, as it seems TS won't use `typeVersions` or the equivalent in our export map when resolving submodules like in this case. We need `preact/compat` to resolve either the TS 5.0 `preact` types or the 5.1+ types, though haven't been able to figure it out. Using the module specifier (i.e., `preact`) doesn't seem to work. Would we need to make TS 5.0 & 5.1+ splits for everything that consumes the split types too?

typesVersions is a little annoying to use since it's largely an external config, not an internal one. The node types have full checking but it's nontrivial.

It may be easier to just drop older versions

rschristian:
That's unfortunate. Last time we bumped TS versions a lot of external projects had considerable issues, especially within the Angular ecosystem which I believe locks TS versions per-release. Preact, being a rather great option for widgets, sees a fair number of wrappers created around Preact components for use elsewhere.

As such, I'm not a fan of touching that minimum, especially to raise it to something so new.

jakebailey:
The minimum version supported by DT is 5.0; 5.1 is not that far off I wouldn't think.

But, I haven't dug more deeply into the problems here (but plan to do DT is unblocked)

rschristian:
As for the DefinitelyTyped errors, I'm not sure what we're doing wrong here? We would expect an error on TS 5.0 and older with the new types, but they should be using the forked types, see: https://github.com/preactjs/preact/blob/bd1443795c912239ea3561fdde5496c8d9e865f1/package.json#L12-L16 https://github.com/preactjs/preact/blob/bd1443795c912239ea3561fdde5496c8d9e865f1/package.json#L18-L21

jakebailey:
I was busy fixing other DT issues and didn't get to looking at this one; I plan to look harder at this come tomorrow.

> Dropping support for TS < 5.1 is a no-go, we have had grievances with this in the past

By nature of DT itself dropping 5.0, it's very possible that 5.0 will break for someone in the ecosystem anyway (via other types packages), but I'm sure we'll be able to find something that works anyway. Two years of types support is pretty generous when the alternative for downstream projects is to... Not update their deps? (Surely if they can update preact, they could update a devDep like TypeScript, but, I digress)

jakebailey:
I overcounted; the preact related failures are just:

```
Error in mdx
Error: 
/home/runner/work/DefinitelyTyped/DefinitelyTyped/types/mdx/preact-tests.tsx
   57:2   error  TypeScript@5.0 tsconfig.preact.json, 5.1 tsconfig.preact.json, 5.2 tsconfig.preact.json, 5.3 tsconfig.preact.json, 5.4 tsconfig.preact.json, 5.5 tsconfig.preact.json, 5.6 tsconfig.preact.json, 5.7 tsconfig.preact.json, 5.8 tsconfig.preact.json compile error: 
Type '{}' is not assignable to type 'IntrinsicAttributes & SVGAttributes<SVGSymbolElement> & ObjectHTMLAttributes<HTMLObjectElement> & ... 158 more ... & SemanticsMathMLAttributes<...>'.
  Property 'datetime' is missing in type '{}' but required in type 'DelHTMLAttributes<HTMLModElement>'  @definitelytyped/expect
   57:2   error  TypeScript@5.0 tsconfig.preact.json compile error: 
'Div' cannot be used as a JSX component.
  Its element type 'ComponentChildren' is not a valid JSX element.
    Type 'undefined' is not assignable to type 'Element | ElementClass | null'                                                                                                                                                                                                                                                                                                                         @definitelytyped/expect
   73:13  error  TypeScript@5.0 tsconfig.preact.json, 5.1 tsconfig.preact.json, 5.2 tsconfig.preact.json, 5.3 tsconfig.preact.json, 5.4 tsconfig.preact.json, 5.5 tsconfig.preact.json, 5.6 tsconfig.preact.json, 5.7 tsconfig.preact.json, 5.8 tsconfig.preact.json expected type to be:
  HTMLAttributes<HTMLAnchorElement>
got:
  AnchorHTMLAttributes<HTMLAnchorElement>     
```

```
Error in preact-custom-element
Error: 
/home/runner/work/DefinitelyTyped/DefinitelyTyped/types/preact-custom-element/preact-custom-element-tests.tsx
  4:7  error  TypeScript@5.0 compile error: 
Type '(props: any) => Element' is not assignable to type 'FunctionalComponent<any>'.
  Call signature return types 'Element' and 'VNode<{}> | null' are incompatible.
    The types of 'type' are incompatible between these types.
      Type 'string | ComponentType<any>' is not assignable to type 'string | ComponentType<{}>'.
        Type 'ComponentClass<any, {}>' is not assignable to type 'string | ComponentType<{}>'.
          Type 'ComponentClass<any, {}>' is not assignable to type 'ComponentClass<{}, {}>'.
            Types of property 'contextType' are incompatible.
              Type 'preact.Context<any> | undefined' is not assignable to type 'import("/home/runner/work/DefinitelyTyped/DefinitelyTyped/types/preact-custom-element/node_modules/.pnpm/preact@10.25.0/node_modules/preact/src/index-5").Context<any> | undefined'.
                Type 'preact.Context<any>' is not assignable to type 'import("/home/runner/work/DefinitelyTyped/DefinitelyTyped/types/preact-custom-element/node_modules/.pnpm/preact@10.25.0/node_modules/preact/src/index-5").Context<any>'.
                  The types returned by 'Consumer(...)' are incompatible between these types.
                    Type 'ComponentChildren' is not assignable to type 'VNode<{}> | null'.
                      Type 'undefined' is not assignable to type 'VNode<{}> | null'  @definitelytyped/
```

```
Error in preact-virtual-list
Error: 
/home/runner/work/DefinitelyTyped/DefinitelyTyped/types/preact-virtual-list/preact-virtual-list-tests.tsx
  19:7  error  TypeScript@5.0 compile error: 
Type '(props: any) => h.JSX.Element' is not assignable to type 'FunctionalComponent<any>'.
  Call signature return types 'Element' and 'VNode<{}> | null' are incompatible.
    The types of 'type' are incompatible between these types.
      Type 'string | ComponentType<any>' is not assignable to type 'string | ComponentType<{}>'.
        Type 'ComponentClass<any, {}>' is not assignable to type 'string | ComponentType<{}>'.
          Type 'ComponentClass<any, {}>' is not assignable to type 'ComponentClass<{}, {}>'.
            Types of property 'contextType' are incompatible.
              Type 'preact.Context<any> | undefined' is not assignable to type 'import("/home/runner/work/DefinitelyTyped/DefinitelyTyped/types/preact-virtual-list/node_modules/.pnpm/preact@10.25.0/node_modules/preact/src/index-5").Context<any> | undefined'.
                Type 'preact.Context<any>' is not assignable to type 'import("/home/runner/work/DefinitelyTyped/DefinitelyTyped/types/preact-virtual-list/node_modules/.pnpm/preact@10.25.0/node_modules/preact/src/index-5").Context<any>'.
                  The types returned by 'Consumer(...)' are incompatible between these types.
                    Type 'ComponentChildren' is not assignable to type 'VNode<{}> | null'.
                      Type 'undefined' is not assignable to type 'VNode<{}> | null'  @definitelytyped/expect
```

You can see that the latter two ones are resolving `index-5`, but what isn't good is that it seems to be cross comparing the `<=5.0` types and the `>=5.1` types. `--explainFiles` says:

```
node_modules/.pnpm/preact@10.25.0/node_modules/preact/src/index.d.ts
  Imported via './index' from file 'node_modules/.pnpm/preact@10.25.0/node_modules/preact/src/jsx.d.ts' with packageId 'preact/src/index.d.ts@10.25.0'
  File is CommonJS module because 'node_modules/.pnpm/preact@10.25.0/node_modules/preact/package.json' does not have field "type"
```

And indeed `jsx.d.ts` imports `index` directly. Theoretically the way to fix this would be to not use a relative import there, but instead import from the public name `preact` such that it goes through typesVersions resolution. Alternatively, you can define two separate, different files that declare _just_ the differing type, then import _that_ from `index.d.ts` instead. That's sort of the problem with `typesVersions`, it really makes local references tricky.

The `mdx` issue seems to be a different one entirely, breaking on all versions.

I'm not sure if this PR is the right place to chat about this, happy to talk wherever.

rschristian:
The `datetime`/`mdx` issue appears to be (in part) the bug fixed in #4570, so some of it can be ignored at least.

Will need to look into the rest yet

rschristian:
Okay so _I think_ #4576 should correct our types here. As you suggested, it goes through and swaps out all references to the now-split types with a bare `preact` specifier. Seems to work correctly, fixing the issue we had in this PR (where the default, 5.1+ types were pulled in when they weren't supposed to be).

jakebailey:
#4576 does seem to work to fix things. mdx is also fixed after the datetime thing, though needing a test change to account for `HTMLAttributes<HTMLAnchorElement> -> AnchorHTMLAttributes<HTMLAnchorElement>`.

I've commented on that PR, though, since it seems like the absolute import is not working out.

rschristian:
> though needing a test change to account for `HTMLAttributes<HTMLAnchorElement> -> AnchorHTMLAttributes<HTMLAnchorElement>`.

Ah yeah, I think that should be the only intentional change now that we're offering stricter types. `HTMLAttributes` used to be a mega interface of all possible attributes -- convenient for us to maintain but not so type safe.

Glad that PR otherwise fixes things though!

rschristian:
Now looks good. Both 5.0 & 5.1+ show no errors in `/compat/index.d.ts` now that the type resolution is set up for all consumers of `/src/index.d.ts` (`/src/index-5.d.ts`). Gonna merge this in then.

PR #4560, opened 2024-11-21
Ensure state updates work in signals

JoviDeCroock:
I encountered a case where a component with both a state-hook and a signal would update both at once, the state-hook would settle to the same value meaning that the signal state update would get fully ignored.

This makes sense as we don't account for multiple sources of state in our hooks implementation, with the new implementation when we'd bail before we now check against a possibly existing scu implementation first.


rschristian (approved):
Bug fix AND byte reduction? You're spoiling us :)

marvinhagemeister (approved):

PR #4564, opened 2024-11-22
Allow rendering in document

### LINKED ISSUE #3794: Uncaught DOMException when having `<!DOCTYPE html>` and setting the `parentDom` to `document` when rendering

**Describe the bug**
If we have `<!DOCTYPE html>` in our HTML document where we use Preact, and try to render or hydrate a component in the `document`, we get an Uncaught DOMException error, and the component does not render/hydrate. Although, it'll be fine if we remove `<!DOCTYPE html>`. This is very a strange behavior.

**To Reproduce**

```html
<!DOCTYPE html>
<html>
  <body>
  </body>
</html>
<script type="module">
  import { h, Component, render } from 'https://unpkg.com/preact?module';

  const app = h('html', null, h('body', null, h('h1', null, 'Hello World!')));

  render(app, document);
</script>
```

1. Open the above HTML in your browser.
2. You will face the error.
3. Try removing the first line, `<!DOCTYPE html>`.
4. See how the error is gone.

**Expected behavior**
Should act just fine :)

JoviDeCroock:
Resolves #3794 

When we want to render straight into `document` we have to alias it to `.documentElement` as this is used in Remix we could consider supporting it, or moving it to compat.

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -12,6 +12,11 @@ import { slice } from './util';
  * existing DOM tree rooted at `replaceNode`
  */
 export function render(vnode, parentDom, replaceNode) {
+	// https://github.com/preactjs/preact/issues/3794
+	if (parentDom === document) {
+		parentDom = document.documentElement;
+	}
+
 	if (options._root) options._root(vnode, parentDom);
 
 	// We abuse the `replaceNode` parameter in `hydrate()` to signal if we are in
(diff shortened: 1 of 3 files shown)


marvinhagemeister (commented):
LGTM, lets add a test if possible

marvinhagemeister (approved):

rschristian (approved):
I don't suppose Remix is willing to switch to using `document.documentElement`? Seems a bit odd to use plain `document`.

PR #4568, opened 2024-11-23
fix: `max` and `min` input element types

### LINKED ISSUE #4566: BUG: REACT HOOK FORM register type max input is a number ERROR
- [x ] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
React hook form method register have a type max input like number and input preact is Signalish<string | undefined> .
```tsx
import { useForm } from "react-hook-form";

export function App() {
    const { register } = useForm<{ name: string }>();
    return (
        <>
            <input type="text" {...register("name")} />
        </>
    );
}

```

**To Reproduce**
https://codesandbox.io/p/sandbox/error-input-react-hook-form-y8rprh

1. example but no show error, but in vite js it happen
2. npm init vite@latest app
3. choose preact typescript and install dependencies.
4. install react hook form  

SOLUTION
**Expected behavior**
the solution is that  min and max add number 
		min?: Signalish<string | undefined|number>;
	MAX?: Signalish<string | undefined|number>;

rschristian:
Fixes #4566

We had the correct types in the old `HTMLAttributes` (now `AllHTMLAttributes`) but in the per-element types I copied from compat seemed to be incorrect.

Will try to write up a script to compare all of our old props to the new, spotting any other discrepancies.

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -1781,10 +1781,10 @@ export namespace JSXInternal {
 		height?: Signalish<number | string | undefined>;
 		indeterminate?: Signalish<boolean | undefined>;
 		list?: Signalish<string | undefined>;
-		max?: Signalish<string | undefined>;
+		max?: Signalish<number | string | undefined>;
 		maxlength?: Signalish<number | undefined>;
 		maxLength?: Signalish<number | undefined>;
-		min?: Signalish<string | undefined>;
+		min?: Signalish<number | string | undefined>;
 		minlength?: Signalish<number | undefined>;
 		minLength?: Signalish<number | undefined>;
 		multiple?: Signalish<boolean | undefined>;


JoviDeCroock (approved):

VigilioYonatan:
when will update version preact package. i need this for my production web. thanks

rschristian:
We have some other type issues to fix, so in the meantime, please just use `// @ts-ignore`.

VigilioYonatan:
mi web project is biggest, i use react hook form always. lol

JoviDeCroock:
@VigilioYonatan there are many people with big projects, downgrade your version then... This is really impolite, we as maintainers sacrifice our time alongside our day job to help y'all but comments like these are just annoying

VigilioYonatan:
Dont worry bro, i will hope. i am going to try to create a d.ts file to fix for a moment

devlux:
I think MeterHTMLAttributes and ProgressHTMLAttributes were forgotten

rschristian:
Not so much forgotten as missed; there's unfortunately no nice list of all standardized attributes. PRs welcome though.

devlux:
please see https://github.com/preactjs/preact/pull/4598

Issue #4572, opened 2024-11-26
createPortal renders children out of order when state updates asynchronously


superlucky84:
- [v] Check if updating to the latest Preact version resolves the issue

## Description:

I encountered an issue where the rendering order of elements inside a createPortal changes unexpectedly when the state updates asynchronously in a Preact component.

## Steps to Reproduce:

1. Use createPortal to render a component (Body) into a target DOM node.
2. Inside the Body component:
    * Render elements conditionally based on state (e.g., using state === 1 to conditionally render).
    * Include other static elements to observe the rendering order.
3. Change the state (e.g., from 0 to 1) after a delay using setTimeout.

Here is a minimal code example to reproduce the issue:

```javascript
import { h, Fragment, render } from 'preact';
import { useState, useEffect } from 'preact/hooks';
import { createPortal } from 'preact/compat';

function App() {
  const [state, setState] = useState(0);

  useEffect(() => {
    setTimeout(() => setState(1), 1000); // Change state after 1 second
  }, []);

  const resultElement = document.getElementById('portal-target');

  return createPortal(
    <Body state={state} />,
    resultElement
  );
}

function Body({ state }) {
  return (
    <Fragment>
      {state === 1 && <div>top</div>}
      <div>middle</div>
      <div>bottom</div>
    </Fragment>
  );
}

// Attach target for portal
document.body.innerHTML = '<div id="portal-target"></div>';
render(<App />, document.body);

```

#### Observed Behavior:
When the state transitions from 0 to 1, the rendered output becomes:

```javascript
middle
bottom
top
```

#### Expected Behavior:
The output should respect the JSX order:

```javascript
top
middle
bottom
```

#### Environment:
Preact version: 10.15.0 or 10.10.0
Browser: chrome 131.0.6778.86

JoviDeCroock:
This has been fixed in #4573 and will be released in the 10.25.1/10.26.0 release of Preact.

superlucky84:
@JoviDeCroock Thank you so much for the incredibly quick fix. You're my hero! I can't wait to try version 10.25.1 :)

PR #4573, opened 2024-11-26
Fix portal children ordering

JoviDeCroock:
Fixes https://github.com/preactjs/preact/issues/4572

Technically `appendChild` isn't needed anymore as our diffing algorithm only uses `insertBefore` - also facepalm at this bug 😅


rschristian (approved):
Ha! How have none of use spotted that?

PR #4576, opened 2024-11-27
fix: Ensure resolution of core types goes through typesVersions resolution

rschristian:
Hopefully fixes the DefinitelyTyped issues & sets us up to land #4557 w/out issue. Looks correct now that I've replaced the relative path with `preact` everywhere.

### DIFF
--- a/compat/client.d.ts
+++ b/compat/client.d.ts
@@ -1,4 +1,6 @@
-import * as preact from '../src';
+// Intentionally not using a relative path to take advantage of
+// the TS version resolution mechanism
+import * as preact from 'preact';
 
 export function createRoot(container: preact.ContainerNode): {
 	render(children: preact.ComponentChild): void;
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -1,5 +1,7 @@
 import * as _hooks from '../../hooks';
-import * as preact from '../../src';
+// Intentionally not using a relative path to take advantage of
+// the TS version resolution mechanism
+import * as preact from 'preact';
 import { JSXInternal } from '../../src/jsx';
 import * as _Suspense from './suspense';
 import * as _SuspenseList from './suspense-list';
--- a/compat/src/suspense-list.d.ts
+++ b/compat/src/suspense-list.d.ts
@@ -1,4 +1,6 @@
-import { Component, ComponentChild, ComponentChildren } from '../../src';
+// Intentionally not using a relative path to take advantage of
+// the TS version resolution mechanism
+import { Component, ComponentChild, ComponentChildren } from 'preact';
 
 //
 // SuspenseList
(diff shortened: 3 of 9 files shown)


compat/client.d.ts
```diff
@@ -1,4 +1,6 @@
-import * as preact from '../src';
+// Intentionally not using a relative path to take advantage of
+// the TS version resolution mechanism
+import * as preact from 'preact';
```
JoviDeCroock, line 3: You might need to configure paths in the tsconfig.json for this

rschristian: Working through the issues right now, I think the problem is that we're not actually exporting some of the types (like `ContainerNode`).

rschristian: Alright I have no idea how to fix this. TS seems to refuse to use the paths set in our jsconfig no matter the configuration, and also we have a problem in that rts drags in a copy of preact (as it's a peerDep) which we also have to fight the resolution on.

rschristian: This isn't really working correctly, `preact` resolves to `node_modules/preact` and there's seemingly nothing I can do to override it. However, that's just a maintainer issue so we might be able to ship this if it unblocks users downstream.

jakebailey: DT resolves this internally by making every package dev dep on itself, but maybe that's not so good for your case.

An alternative here is to keep using relative imports, and duplicate the types into a second dir (which is what we do on DT to keep things simple). It comes at the cost of needing to update it, though. Since these are handwritten, you could also store a patch that is the diff, and then make the copy during the build.

Though, I'm trying to find the issue that you're having; what is breaking locally? CI and build seem to succeed.

rschristian: Sorry, didn't word that well. The issue is that we install `preact-render-to-string`, our server renderer, to run some tests. This has a peer dependency against `preact`, meaning NPM will oh-so-helpfully install it. `import ... from 'preact'` in these types files will therefore resolve against `node_modules/preact`, not the project root. Using `"paths"` in our jsconfig doesn't seem to solve this either.

There's probably some config combination that addresses it, but I'm not sure what. The CI passing is therefore incidental and likely to break in the future, as `import ... from 'preact'` is actually importing from an installed copy of Preact, not the latest code/types. If we ever need to add a type and import it elsewhere, that won't work.

jakebailey: I wouldn't expect `paths` to work; it bypasses all resolution rules and defeats the purpose of the change (I would never recommend the option for anything these days).

You could technically set an npm override to point `"preact": "link:."` or something, which would ensure that the other package got the local preact installed.

If you want, I can attempt to test stuff locally, it's just a challenge given I'm walking into an unfamiliar codebase trying to find solutions 😄

jakebailey: IIRC `file:.` will make package managers copy the package contents into `node_modules` instead of linking them, so that may break things during local editing. But, different package managers behave differently so maybe npm is doing something different here...

JoviDeCroock (approved):

PR #4578, opened 2024-11-29
fix: `createElement` & `h` types

### LINKED ISSUE #4577: TS error for h() function overload match when element properties do not include global attribute

**Describe the bug**
The issue is related to type updates in [PR 4546](https://github.com/preactjs/preact/pull/4546). It is not reproducible with Preact 10.24.3.

The h() function definitions still use HTMLAttribute instead of AllHTMLAttributes. Since the HTMLAttribute interface is now limited to global attributes/properties, the h() function overload match fails when given properties do not contain global attributes.

Screenshot 2024-11-27 at 3 08 18 PM

**To Reproduce**
You can use this [example in Typescript playground](https://www.typescriptlang.org/play/?ts=5.4.5#code/JYWwDg9gTgLgBAbzgCzgXzgMyhEcBEYUApgIYDGM+A3AFC3kQB2AzvDMWwMrEA2xlOAF44ACgCUwgHyJacOI1bwIYGCwCCUKKQCewuAG058xHABupXgFdiALgIXrxAIz4ANHF6kARn3v4ANUsbV3Q3Y3kkRxt-aOIAJndPHz8CIKdEsOMAXTp5RTY4FRhgZhZ9Yo0tXQA6EFIwUVFi0qZJIRkECLgSGCsoJhRRbvl8FuZ3EdM4+3GmGriskxM5mq9fXm7xPPRt4wL4IhVykS7lplIQOwIQHQA5S+J8YzQd3v7B5FF8Fj4BKg8RzALA8cxYe1etGIAA9ILBTBxuH9BJDaEA) to see the issue.

1. Use the following code in Typescript playground 
```const testSelect = () => {
  const optsArray = [
    { value: "value1", label: "Value1" },
    { value: "value2", label: "Value2" },
  ];
  const options = optsArray.map((option) => {
    return h(
      "option",
      { value: option.value },
      option.label
    );
  });
  const props = {
    name: "myName"
  };
  return h("select", props, options);
};
export { testSelect };
```
2. See errors for 'option' and 'select':
```
  The last overload gave the following error.
    Argument of type 'string' is not assignable to parameter of type 'ComponentType<{ value: string; }>'.(2769)
index.d.ts(278, 17): The last overload is declared here.
```

**Expected behavior**
The correct overload for HTMLElement should be found.

rschristian:
Closes #4577 

Seems like a bit of a shame, essentially the type falls through to just match against a string and offer no real type checking, but it's seemingly [what React does](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a6e9e491ff5d0fd7f438fb77a70b54b28e356ced/types/react/index.d.ts#L555).

### DIFF
--- a/src/index-5.d.ts
+++ b/src/index-5.d.ts
@@ -226,7 +226,7 @@ export function createElement<T extends HTMLElement>(
 	ClassAttributes<T> & JSXInternal.HTMLAttributes & JSXInternal.SVGAttributes
 >;
 export function createElement<P>(
-	type: ComponentType<P>,
+	type: ComponentType<P> | string,
 	props: (Attributes & P) | null,
 	...children: ComponentChildren[]
 ): VNode<P>;
@@ -276,7 +276,7 @@ export function h<T extends HTMLElement>(
 	| null
 >;
 export function h<P>(
-	type: ComponentType<P>,
+	type: ComponentType<P> | string,
 	props: (Attributes & P) | null,
 	...children: ComponentChildren[]
 ): VNode<Attributes & P>;


src/index.d.ts
```diff
@@ -210,7 +210,7 @@ export function createElement<
 	P extends JSXInternal.SVGAttributes<T>,
 	T extends HTMLElement
 >(
-	type: keyof JSXInternal.IntrinsicElements,
+	type: keyof JSXInternal.IntrinsicSVGElements,
```
rschristian, line 213: Sorry, drive-by change.

I couldn't find a situation in which this actually made a difference, as `createElement`/`h` seemingly always falls back to a very loose overload that allows anything, but we extracted out `IntrinsicSVGElements` a bit more than a year ago and so we should be using it.

JoviDeCroock (approved):

JoviDeCroock:
Let's add a ts-test

JoviDeCroock:
The TS-playground test from the issue replicated for me in our tests

rschristian:
Yeah that one does for me too, but seemingly not other equivalent tests? Which is why I'm confused.

I can copy/paste that and move on, just don't understand what exactly is diverging & what we're testing. Something seems off.

JoviDeCroock:
Hmm, that's odd, the type-checking used to work very well and I think with the examples I gave still does. Not sure what changed though, to your point

rschristian:
OH nvm, it's falling into the `HTMLAttributes & SVGAttributes` overload. Somehow didn't notice that

PR #4579, opened 2024-11-29
Move some frequently used things to constants

JoviDeCroock:
This checks the performance of moving some longer strings to constants to avoid re-allocating in the critical path. Some of these could also optimise easier as the function gets shorter


rschristian (approved):

JoviDeCroock:
Memory bump seems worth the bytes

Issue #2913, opened 2021-01-05
useReducer: stale value in closure or lost update?


mischnic:
### Reproduction

https://codesandbox.io/s/nostalgic-wilson-3vs0s

```jsx
import React, {
  createElement,
  createContext as createContextOrig,
  useContext as useContextOrig,
  useLayoutEffect,
  useReducer,
  useRef,
  useState,
} from "preact/compat";

function createContext() {
  const context = createContextOrig();

  const ProviderOrig = context.Provider;
  context.Provider = ({ value, children }) => {
    const valueRef = useRef(value);
    const contextValue = useRef();

    if (!contextValue.current) {
      contextValue.current = {
        value: valueRef,
        listener: null,
      };
    }

    useLayoutEffect(() => {
      valueRef.current = value;
      if (contextValue.current.listener) {
        contextValue.current.listener([value]);
      }
    }, [value]);
    return <ProviderOrig value={contextValue.current}>{children}</ProviderOrig>;
  };

  return context;
}

function useContextSelector(context) {
  const contextValue = useContextOrig(context);
  const {
    value: { current: value },
  } = contextValue;
  const [state, dispatch] = useReducer(
    () => {
      return {
        value,
      };
    },
    {
      value,
    }
  );
  useLayoutEffect(() => {
    contextValue.listener = dispatch;
  }, []);
  return state.value;
}

const context = createContext(null);

function Child() {
  const [count, setState] = useContextSelector(context);
  const [c, setC] = useState(0);
  return (
    <div>
      <div>Context count: {count}</div>
      <div>Local count: {c}</div>
      <button
        onClick={() => {
          setC((s) => s + 1);
          setState((s) => s + 1);
        }}
      >
        +1
      </button>
    </div>
  );
}

// Render this
export function App() {
  const [state, setState] = useState(0);
  return (
    <context.Provider value={[state, setState]}>
      <Child />
    </context.Provider>
  );
}
```

### Steps to reproduce

Click on the "+1" button multiple times.

### Expected Behavior

What React does: both counters increment simultaneously.

### Actual Behavior

With Preact, the first update is ignored and the "local counter" is one bigger than the "global counter".

cc @marvinhagemeister @dai-shi

developit:
Curious what the reason is for deferring the assignment of `valueRef.current` into the layout effect callback - why not just assign it unconditionally in `Provider()`?

mischnic:
[...]

PR #4580, opened 2024-11-29
Delay executing state updates

JoviDeCroock:
Resolves https://github.com/preactjs/preact/issues/2913

This puts our implementation more in line with React which could resolve a few issues as well as reduce the byte size impact of our code. Another benefit this brings would be that we stop mutating the `Component` instance and hence can't mess with other implementations like signals.

What React does is when you call `reduce` is that it enqueues an update but does not execute the action yet. It first renders the component so the outer scope to the state setter can be recaptured. This way it ensures that it deals with up-to-date props/...

Essentially what this comes down to is that we store every reducer invocation and only execute it as the component is rendered. This way it has access to the outer component scope, .... This needs us to add a new `SKIP_CHILDREN` bit-flag on the `VNode` so we can bail out when the state settles to equal.

For this to properly work we'd need to introduce `options._afterRender` which would check all of the hooks and set `SKIP_CHILDREN` when all hooks are equal.

To get the last test to pass we'd need to find a way to deterministically see whether this component started the diff as we should not bail if the parent instructed the change.

I think this should generally be a performance improvement as well as we aren't touching the `Component` instance anymore on every pass.


src/constants.js
```diff
@@ -7,6 +7,8 @@ export const INSERT_VNODE = 1 << 2;
 /** Indicates a VNode has been matched with another VNode in the diff */
 export const MATCHED = 1 << 1;
 
+export const SKIP_CHILDREN = 1 << 18;
```
JoviDeCroock, line 10:
```suggestion
export const SKIP_CHILDREN = 1 << 3;
```

JoviDeCroock:
I think this could cause issues for signals, when we combine signals with hooks we won't be able to rerender if the hooks have executed but ended up at the same value.

PR #4583, opened 2024-11-29
refactor: Remove `declare global` from internal types

rschristian:
#4173 introduced `declare global` to skip having to import them in `src/`. However, this blocks usage in other repos (like [devtools](https://github.com/preactjs/preact-devtools/blob/e74468faf45deaca5d7051e508bb2f6567d733af/src/adapter/10/bindings.ts#L6), which needed to patch this) and we're really abusing the language server to support it. From a TS file you would not be able to import interfaces declared within `declare global {}`, it blocks all use. While it "works" for our JS usage (every subpath imports from `src/internal`) at the moment, we have a bunch of TS issues that I'm trying to track down and I don't think this is helping matters one bit.

It's also just really ambiguous. If we're going to have multiple `VNode`, `Component`, etc. definitions sitting around, let's at least point to which one we're referencing, no?


hooks/src/internal.d.ts
```diff
@@ -1,8 +1,14 @@
+import {
+	Options as PreactOptions,
+	Component as PreactComponent,
+	VNode as PreactVNode,
+	Context as PreactContext,
+} from '../../src/internal';
```
rschristian, line 6: This is what we use in [debug](https://github.com/preactjs/preact/blob/cfd2f41035e79d4d07ed173f0e7aa0f83ac6a219/debug/src/internal.d.ts#L1) & [compat](https://github.com/preactjs/preact/blob/cfd2f41035e79d4d07ed173f0e7aa0f83ac6a219/compat/src/internal.d.ts#L1-L6), makes everything a bit more consistent

marvinhagemeister (approved):

marvinhagemeister:
Nice, the `declare global` stuff seemed always more of a hack to me.

Issue #4581, opened 2024-11-29
types: `EventHandler` returning (seemingly) incorrect type


rschristian:
- [X] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
Found this trying to help [Hypothesis upgrade their frontend-shared repo](https://github.com/hypothesis/frontend-shared/pull/1803).

#4538 seems to be the cause of a regression in the following type, resulting in the following error:

```
Property 'call' does not exist on type 'GenericEventHandler<HTMLInputElement>'.
  Property 'call' does not exist on type 'EventHandlerObject<TargetedEvent<HTMLInputElement, Event>>'.
```

Not really sure what that type change is even doing, need more time to investigate but wanted to get this opened.

**To Reproduce**

[TS Playground](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgCwDRwFIGUAacC+cAZlBCHAERhQCmAhgMYwUDcAUG0QK4B2TwEHnADCyGgwDWAIwgAPABRJBoujwDmNAgC5MuAHQBJHmC4wAEgBUAsgBkAgjBhRgU0zQDOAHku2jJmACiADY0IDQ8MAB8AJSIHHDEvPyCKKoAJiEq6jTysQhsCQnKyKoaAPx6DHRBQbnsCfjxcLQwXFBCnsDGpnDFpTQAvAglPBk0WRqEAPSRbI1sQA)

```ts
import { JSX } from "preact";

function Checkbox({ onChange }: JSX.HTMLAttributes<HTMLInputElement>) {

  function handleChange(this: void, event: JSX.TargetedEvent<HTMLInputElement>) {
    onChange?.call(this, event);
           // ^ Error here
  }

  return <input onChange={handleChange} />
}
```

**Expected behavior**

Shouldn't be any error

JoviDeCroock:
Hmm, the whole bivariancehack situation is concerning already. I wonder whether we can get away with a similar fix to https://github.com/preactjs/preact/pull/4547

lilnasy:
I think the error really is expected. Starting with the latest version, Checkbox can be passed something other than a function.

For example, with this usage:
```tsx
class Form {
    handleEvent(event: Event) {
        ....
    }
    
    render() {
        return <Checkbox onChange={this}/>
    }
}
```

The userland fix would be something like this:
```tsx
import type { JSX } from "preact";

[...]

PR #4584, opened 2024-11-29
Revert handle event

JoviDeCroock:
Resolves https://github.com/preactjs/preact/issues/4581
Reverts https://github.com/preactjs/preact/pull/4538

After trying to fix this for a while this looks like it's just incompatible with our current ecosystem and would warrant a breaking change, my sincerest apologies for backtracking on this. I've added a test in the TS tests to avoid this in the future.

There's an inherent incompatibility with using the `handleEvent` interface as these types are also used by our consumers to type their events and components.


test/ts/preact.tsx
```diff
@@ -381,3 +382,15 @@ h('form', { onSubmit: onSubmit });
...
 h('option', { value: 'foo' });
 createElement('option', { value: 'foo' });
+
+function Checkbox({ onChange }: JSX.HTMLAttributes<HTMLInputElement>) {
+	function handleChange(
+		this: void,
+		event: JSX.TargetedEvent<HTMLInputElement>
+	) {
+		onChange?.call(this, event);
+		// ^ Error here
```
JoviDeCroock, line 392:
```suggestion
```

rschristian (approved):

PR #4585, opened 2024-11-30
fix: Correct `defaultValue` property types for inputs, selects, & textareas

rschristian:


src/jsx.d.ts
```diff
@@ -1755,7 +1755,7 @@ export namespace JSXInternal {
 		capture?: Signalish<'user' | 'environment' | undefined>; // https://www.w3.org/TR/html-media-capture/#the-capture-attribute
 		checked?: Signalish<boolean | undefined>;
 		defaultChecked?: Signalish<boolean | undefined>;
-		defaultValue?: Signalish<string | undefined>;
+		defaultValue?: Signalish<string | number | undefined>;
```
rschristian, line 1758: Not supported in our old types, but I think it should be? Nothing wrong with  `input.defaultValue = 2`.

JoviDeCroock (approved):

PR #4586, opened 2024-11-30
Test whether reducing bits leads to memory improvements

JoviDeCroock:
Initial assumption was this would reduce memory consumption but... the byte size is also pretty great 😅


rschristian (commented):
Excuse me as it's late & my brain has become mush from chasing down TS errors all day, but would this not impact say [preact-iso](https://github.com/preactjs/preact-iso/blob/57ce361e1d0463e5bf98830baa2daf92c77e769e/src/router.js#L256-L257) or anyone else who's copy/pasted these values in the past to add behavior?

JoviDeCroock:
@rschristian That's a good point, I've switched back SUSPENDED and HYDRATE the others are only present in our internal diff and should not be used externally.

rschristian (approved):

PR #4588, opened 2024-11-30
chore: Correct some internal test types

rschristian:
In a (perhaps futile) desire to reduce the number of files that light up my editor like a Christmas tree, this PR tries to correct a couple hundred or so of our type errors. Switching [this line](https://github.com/preactjs/preact/blob/42524db0382b398da63a22c8a07ed487f9ca3137/package.json#L145) over to just `jsconfig.json` produces 1501 TS errors, which this PR brings down to 1292. Only corrects a small fraction of the type errors, and mostly the quick ones at that, but it's a chunk I suppose.

The only non-comment/type changes should be in adding an initial param to `createContext()` calls & passing `{}`/`null` to `h()`/`createElement()`.

### DIFF
--- a/mangle.json
+++ b/mangle.json
@@ -37,7 +37,6 @@
       "$_args": "__H",
       "$_factory": "__h",
       "$_depth": "__b",
-      "$_nextDom": "__d",
       "$_dirty": "__d",
       "$_mask": "__m",
       "$_detachOnNextRender": "__b",
--- a/test/_util/helpers.js
+++ b/test/_util/helpers.js
@@ -210,14 +210,6 @@ export function clearOptions() {
 export function teardown(scratch) {
 	if (!document.contains(scratch)) return;
 
-	if (
-		scratch &&
-		('__k' in scratch || '_children' in scratch) &&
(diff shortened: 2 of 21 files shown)


mangle.json
```diff
@@ -37,7 +37,6 @@
       "$_args": "__H",
       "$_factory": "__h",
       "$_depth": "__b",
-      "$_nextDom": "__d",
```
rschristian, line 40: We should be able to remove post https://github.com/preactjs/preact/pull/4553, no?

test/_util/helpers.js
```diff
@@ -236,21 +228,6 @@ export function teardown(scratch) {
...
-
-	if (vnode._children) {
-		for (let child of vnode._children) {
-			if (child) {
-				verifyVNodeTree(child);
-			}
-		}
-	}
-}
-
```
rschristian, line 253: Same here

JoviDeCroock (approved):

JoviDeCroock:
This is awesome! Also good spot on nextDom

PR #4599, opened 2024-12-07
Use == where possible and precompute .length

JoviDeCroock:
Just  did a pass of where we can use == instead of === which should save some bytes. I've also hoisted the `.length` accessor from children which should be cheaper perf wise.


rschristian (approved):
Good finds

PR #4600, opened 2024-12-08
Replace npm-run-all and lint-staged

JoviDeCroock:
Replaces some deps with suggestions from es-tooling


rschristian (approved):
Ha, was just browsing that list and about to test out these alts myself!

PR #4604, opened 2024-12-11
Revert unkeyed no-search

JoviDeCroock:
Reverts https://github.com/preactjs/preact/pull/4550

This revert isn't necessarily needed but after a lot of investigation we do increase components remounting a lot, to avoid people having to deal with a breaking change like that I'd like to propose a revert here.

If people fall into this issue, please use `key` to avoid elements being reused when you don't want to.

### DIFF
--- a/test/browser/render.test.js
+++ b/test/browser/render.test.js
@@ -1800,7 +1800,7 @@ describe('render()', () => {
 	});
 
 	// #2949
-	it('should not swap unkeyed chlildren', () => {
+	it.skip('should not swap unkeyed chlildren', () => {
 		class X extends Component {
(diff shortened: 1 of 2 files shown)


test/browser/render.test.js
```diff
@@ -1830,6 +1830,44 @@ describe('render()', () => {
 		expect(scratch.textContent).to.equal('A');
 	});
 
+	// #2949
+	it.skip('should not swap unkeyed chlildren', () => {
```
JoviDeCroock, line 1834: Added a test highlighting the issue

rschristian (approved):

PR #4606, opened 2024-12-12
refactor: Add default event target types for HTML attribute interfaces

rschristian:
Neither [React](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d7ca55e97f79a85f8357f51ef2420b932917a6de/types/react/index.d.ts#L2899) nor [Solid](https://github.com/ryansolid/dom-expressions/blob/828f9e9cef5e0482d904ae4f6ec1069dba378a44/packages/dom-expressions/src/jsx.d.ts#L939) (if I've understood their types right, had to go looking for that and it's seemingly not in their org? weird) do this so I'm a tad apprehensive as otherwise seems like a pretty obvious DX improvement. Take the following example:

```ts
import type { JSX } from 'preact';

// Before
interface MyInputProps extends JSX.InputHTMLAttributes<HTMLInputElement> { ... }

// After
interface MyInputProps extends JSX.InputHTMLAttributes { ... }
interface MyWhackyInputProps extends JSX.InputHTMLAttributes<HTMLButtonElement> { ... }
```
 
Users can still supply the generic parameter if they wish but 99% of the time they're going to want the element associated with the attributes which we can simply assume as the default.

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -1607,7 +1607,8 @@ export namespace JSXInternal {
 		width?: Signalish<number | string | undefined>;
 	}
 
-	interface ColHTMLAttributes<T extends EventTarget = HTMLTableColElement> extends HTMLAttributes<T> {
+	interface ColHTMLAttributes<T extends EventTarget = HTMLTableColElement>
+		extends HTMLAttributes<T> {
 		span?: Signalish<number | undefined>;
 		width?: Signalish<number | string | undefined>;
 	}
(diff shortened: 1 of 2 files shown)


src/jsx.d.ts
```diff
@@ -1533,7 +1533,7 @@ export namespace JSXInternal {
 		| '_top'
 		| (string & {});
 
-	interface AnchorHTMLAttributes<T extends EventTarget>
+	interface AnchorHTMLAttributes<T extends EventTarget = HTMLAnchorElement>
```
JoviDeCroock, line 1536: Can we just add TypeScript tests for `createElement` and `jsx` rendering so that we ensure we cover all bases here.

JoviDeCroock (approved):
This looks like a no-brainer honestly, great DX win, just the tests and we are good

src/jsx.d.ts
rschristian: Good call on the tests, forgot to push mine, but I'm not quite sure I follow re:JSX rendering in particular? JSX will always go through `IntrinsicElements` which always specifies the event target type:

```ts
interface IntrinsicElements { 
    button: ButtonHTMLAttributes<HTMLButtonElement>;
}
```

This PR wouldn't affect the behavior here at all, as there's no way you should be able to have a different element other than `HTMLButtonElement` when using `<button>`.

Let me know if I'm misunderstanding though, certainly possible

rschristian: Fair enough! Added a matching test case for JSX too then

PR #4607, opened 2024-12-13
refactor: Restrict aria roles by element type

rschristian:
Starting to take a look at #3853, seems role restriction should be simple enough for most elements. [Spec Doc](https://www.w3.org/TR/html-aria/#docconformance)

Edit: Unfortunately there's some rare situations in which the spec should be ignored which presents an issue here: adding stricter types might dissuade people against using the correct role on some elements because the spec (and therefore our types) don't allow it. I'd like to guess that this occurs far less often than a user simply reaching for the wrong roles, however, hence this PR.

---

There are some roles that we can't limit via types, mainly due to descendant restrictions ("If not a descendant of `<article>`, `<aside>`, `<main>`, ..., etc.) or specific attribute values ("If `size` attribute is greater than 1"), but this does get most elements/roles.

With this I think I'm all typed out for the foreseeable future 😅


JoviDeCroock:
@rschristian Let's target this one at `v11-2` as well

rschristian:
Worried it might be breaking? I have no strong opinion either way, though with how verbose this is, it might make v11 look a bit messier.

JoviDeCroock (approved):
This should allow us to close out the #3853 issue I reckon

rschristian:
Hm, I don't think so.

While this does take a big first step, it's limited to aria roles; it does not attempt to provide (restricted) types for any of the myriad of other aria attributes that are available.

rschristian:
Going to merge, everything seems good!

PR #4615, opened 2024-12-19
fix: Incorrect onToggle event type

### LINKED ISSUE #4614: `onToggle` types do not use `ToggleEvent`
The types for `onToggle`, for example in `<div popover onToggle={fn} />`, are typed as `JSX.GenericEventHandler<HTMLDivElement> | undefined` rather than `JSX.ToggleEventHandler<HTMLDivElement> | undefined`. This is important, because `ToggleEvent` has a `newState` and `oldState` field that `Event` does not.

rschristian:
Closes #4614

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -508,6 +508,10 @@ export namespace JSXInternal {
 		Target,
 		TouchEvent
 	>;
+	export type TargetedToggleEvent<Target extends EventTarget> = TargetedEvent<
+		Target,
+		ToggleEvent
+	>;
 	export type TargetedTransitionEvent<Target extends EventTarget> =
 		TargetedEvent<Target, TransitionEvent>;
 	export type TargetedUIEvent<Target extends EventTarget> = TargetedEvent<
@@ -535,6 +539,9 @@ export namespace JSXInternal {
 		EventHandler<TargetedCompositionEvent<Target>>;
 	export type DragEventHandler<Target extends EventTarget> = EventHandler<
 		TargetedDragEvent<Target>
+	>;
+	export type ToggleEventHandler<Target extends EventTarget> = EventHandler<
+		TargetedToggleEvent<Target>
 	>;
 	export type FocusEventHandler<Target extends EventTarget> = EventHandler<
 		TargetedFocusEvent<Target>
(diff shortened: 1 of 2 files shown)


test/ts/preact.tsx
```diff
@@ -375,13 +379,17 @@ const onBeforeInput = (e: h.JSX.TargetedInputEvent<HTMLInputElement>) => {};
...
 h('input', { onBeforeInput: onBeforeInput });
 
+// Should accept onSubmit
 const onSubmit = (e: h.JSX.TargetedSubmitEvent<HTMLFormElement>) => {};
 <form onSubmit={e => e.currentTarget.elements} />;
 createElement('form', { onSubmit: onSubmit });
 h('form', { onSubmit: onSubmit });
 
-h('option', { value: 'foo' });
-createElement('option', { value: 'foo' });
```
rschristian, line 384: Sorry for the diff, placed this rather wonkily a few weeks ago. Moved it out so it wasn't sat in the middle of the event tests.

rschristian:
Damn, `ToggleEvent` is newer than our TS version. I'd rather not raise the type floor

src/jsx.d.ts
```diff
@@ -20,6 +20,28 @@ type Defaultize<Props, Defaults> =
...
+    prototype: ToggleEvent;
+    new(type: string, eventInitDict?: ToggleEventInit): ToggleEvent;
+};
+
+interface ToggleEventInit extends EventInit {
+    newState?: string;
+    oldState?: string;
+}
+
+// End TS >5.2
```
rschristian, line 43: We've never done this before (to my knowledge), but maybe it's a viable strategy? It's not too gnarly, and hopefully with the blocks, it can be removed in the future pretty easily.

marvinhagemeister: looks good to me. Admittedly I'm not even sure if we've settled on a particular TS version to support.

marvinhagemeister (approved):

src/jsx.d.ts
rschristian: We haven't, though we do have split types for supporting >5.1 & <5.1 at the moment.

Issue is that every time we've bumped up the minimum people in other communities have had problems -- unfortunate side effect of being popular for widgets & component libs. IIUC, Angular still restricts TS versions per (Angular) major, for example.

JoviDeCroock (approved):

PR #4618, opened 2024-12-23
Allow for Context as JSX

JoviDeCroock:
This is a WIP, currently the types aren't quite collaborating at the moment. Just wanted to highlight that this one is pretty easy to solve for us.

### DIFF
--- a/src/create-element.js
+++ b/src/create-element.js
@@ -8,7 +8,7 @@ let vnodeId = 0;
  * Create an virtual node (used for JSX)
  * @param {import('./internal').VNode["type"]} type The node name or Component constructor for this
  * virtual node
- * @param {object | null | undefined} [props] The properties of the virtual node
+ * @param {object | null | undefined | import('.').Context} [props] The properties of the virtual node
  * @param {Array<import('.').ComponentChildren>} [children] The children of the
  * virtual node
  * @returns {import('./internal').VNode}
@@ -18,6 +18,11 @@ export function createElement(type, props, children) {
 		key,
 		ref,
 		i;
+
+	if (typeof type === 'object' && type != null && 'Provider' in type) {
+		type = type.Provider;
+	}
+
 	for (i in props) {
 		if (i == 'key') key = props[i];
 		else if (i == 'ref') ref = props[i];
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -388,7 +388,7 @@ export type ContextType<C extends Context<any>> = C extends Context<infer T>
 	? T
 	: never;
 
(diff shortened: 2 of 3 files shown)


src/create-element.js
```diff
@@ -18,6 +18,13 @@ export function createElement(type, props, children) {
 		key,
 		ref,
 		i;
+
+	// @ts-expect-error type can't be null, however TS is confused
+	if (type != null && typeof type === 'object' && 'Provider' in type) {
```
JoviDeCroock, line 23: I wonder whether we should move this to compat instead, this looks like some overhead in a very hot path

rschristian: Given the pretty minimal utility, I think I'd vote for/prefer moving to compat myself.

rschristian (approved):

developit:
I'm good with this, though I do wonder if we should just implement it in core.

marvinhagemeister:
+1 for core if the snippet @developit provided works

JoviDeCroock:
@developit @marvinhagemeister added the suggestion

mangle.json
```diff
@@ -60,8 +60,8 @@
       "$_processingException": "__",
       "$_globalContext": "__n",
       "$_context": "c",
-      "$_defaultValue": "__",
```
JoviDeCroock, line 63: One potential issue here is https://github.com/preactjs/prefresh/blob/main/packages/babel/src/index.mjs#L404 where `_defaultValue` might fail to update as it's not on `__` anymore. Instead we'd update the `_parent` of `Context` which sounds... problematic

mangle.json
```diff
@@ -60,8 +60,8 @@
...
-      "$_id": "__c",
+      "$_defaultValue": "__d",
+      "$_id": "__l",
```
JoviDeCroock, line 64:
```suggestion
      "$_defaultValue": "__",
      "$_id": "__l",
```

mangle.json
andrewiggins: OH. `"_defaultValue": "__"` conflicts with `"_contextRef": "__"` (see line 65 in mangle.json and line 54 in create-context.js). One of them will need to change :/

mangle.json
```diff
@@ -61,8 +61,8 @@
       "$_globalContext": "__n",
       "$_context": "c",
       "$_defaultValue": "__",
-      "$_id": "__c",
-      "$_contextRef": "__",
+      "$_id": "__l",
```
JoviDeCroock, line 64:
```suggestion
      "$_id": "__c",
```

PR #4619, opened 2024-12-26
fix: support multiple synchronous timings for act

rock-57blocks:
### What this PR do?

As we talked [here](https://github.com/preactjs/signals/issues/633#issuecomment-2561724827)
There is a issue:
it just looks like the signal-effect update never gets flushed when there's a child depending on the signal, this might be because we first initiate a [rerender in act](https://github.com/preactjs/preact/blob/main/test-utils/src/index.js#L67) which will queue up the useEffect and potentially override the scheduled one from the signal-effect.

### DIFF
--- a/test-utils/test/shared/act.test.js
+++ b/test-utils/test/shared/act.test.js
@@ -518,9 +518,7 @@ describe('act', () => {
 	});
 
 	describe('act function with finish implementations', () => {
-		let actDepth = 0;
 		beforeEach(function () {
-			actDepth = 0;
 			options.requestAnimationFrame = null;
 		});
 
--- a/test-utils/test/shared/act.test.js
+++ b/test-utils/test/shared/act.test.js
@@ -517,26 +517,26 @@ describe('act', () => {
 		});
 	});
 
-	describe('act function with finish implementations', function () {
+	describe('act function with finish implementations', () => {
 		let actDepth = 0;
 		beforeEach(function () {
-			actDepth = 0; // Reset actDepth before each test
-			options.requestAnimationFrame = null; // Reset before each test
+			actDepth = 0;
+			options.requestAnimationFrame = null;
 		});
 
-		it('should execute the flush callback using single flush', function () {
+		it('should execute the flush callback using single flush', () => {
 			let called = false;
 
 			act(() => {
 				options.requestAnimationFrame(() => {
-					called = true; // This should be called
+					called = true;
 				});
 			});
 
-			expect(called).to.be.true; // Verify the callback was executed
+			expect(called).to.be.true;
 		});
 
-		it.only('should execute all callbacks using array flush', function () {
+		it.only('should execute all callbacks using array flush', () => {
 			let callCount = 0;
 
 			act(() => {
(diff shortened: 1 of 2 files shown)


rschristian (commented):
Adding a test case would be appreciated if you can

rock-57blocks:
> Adding a test case would be appreciated if you can

Sure, will do~!

rock-57blocks:
@rschristian 
I've added some tests for the finish function but I also want to add tests with rendering component. Howerver, this issue only happened when using `useSignalEffect`, It seems that we should not introduce `useSignalEffect` in Preact tests
 what do you suggest? 

```
it.only("signals should not stop context from propagating", () => {
	let update: any;

	const MuteButton = ({
		isMuted,
			}) => {
		useEffect(() => {
				console.log('isMuted', isMuted);
		}, [isMuted]);

		return null;
			};

	const muteVideo = signal(false);
		const App = () => {
		const videoRef = useRef<HTMLVideoElement>(null);
		useSignalEffect(() => {
				console.log('running effect', muteVideo.value)
				videoRef.current!.muted = muteVideo.value;
		});
		console.log('rendering', muteVideo.value)

		return (
			<div>
				<video
				data-testid='video-player'
				ref={videoRef}>
				</video>
				<MuteButton
					isMuted={muteVideo.value}
				/>
			</div>
		);
			};

	act(()=>{
		render(<App />, scratch);
	})

	expect(scratch.innerHTML).to.equal('<div><video data-testid="video-player"></video></div>');
	expect(scratch.firstChild?.firstChild.muted).to.eql(false)
	act(() => {
		muteVideo.value = true;
	});
	expect(scratch.innerHTML).to.equal('<div><video data-testid="video-player"></video></div>');
	expect(scratch.firstChild?.firstChild.muted).to.eql(true)
});
```

test-utils/test/shared/act.test.js
```diff
@@ -516,4 +516,56 @@ describe('act', () => {
 			});
 		});
 	});
+
+	describe('act function with finish implementations', function () {
```
rschristian, line 520: Nit: should generally be using arrow functions, as the other examples in this file show.

test-utils/test/shared/act.test.js
```diff
@@ -516,4 +516,56 @@ describe('act', () => {
...
+		let actDepth = 0;
+		beforeEach(function () {
+			actDepth = 0; // Reset actDepth before each test
```
rschristian, line 523: Nit: all of the comments you added in this file can be removed. It's very straight-forward code, it speaks for itself.

rschristian (commented):
Indeed, we shouldn't attempt to directly use `useSignalsEffect` in our tests here but we should be able to emulate it, (roughly) recreating the case in which `act` works incorrectly.

It's unfortunately pretty late here & I'm doing this phone, but your tests might cover this well enough already. I'll try to carve out some time in the next few days (holiday season, might be next week) to look into it more (if Jovi can't) as I'm not super familiar here.

test-utils/test/shared/act.test.js
```diff
@@ -516,4 +516,56 @@ describe('act', () => {
...
+			act(() => {
+				options.requestAnimationFrame(() => {
+					called = true;
+				});
+			});
+
+			expect(called).to.be.true;
+		});
+
+		it.only('should execute all callbacks using array flush', () => {
```
JoviDeCroock, line 539:
```suggestion
		it('should execute all callbacks using array flush', () => {
```

JoviDeCroock (approved):
The gist of what happens is

- update signal
- effect notifies (call options.raf)
- rerender happens 
- useEffect deps change (call raf)

The tests sufficiently cover that

PR #4629, opened 2025-01-10
chore: Fix mocha in Node v22+

rschristian:
Just quick fix (that we don't need quite yet) as I noticed it.

[Mocha seemingly requires an explicit `"type": "commonjs"` as of Node 22+](https://github.com/mochajs/mocha-examples/issues/103), perhaps because of the new `require(esm)` that landed? Not quite sure.

This fixes it, in lieu of Mocha fixing something upstream (which they very well might do yet).

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "version": "10.26.0",
   "private": false,
   "description": "Fast 3kb React-compatible Virtual DOM library.",
+  "type": "commonjs",
   "main": "dist/preact.js",
   "module": "dist/preact.module.js",
   "umd:main": "dist/preact.umd.js",


JoviDeCroock (approved):

marvinhagemeister (approved):

rschristian:
As best as I can tell, this is a change in how Node tries to determine module types.

Whereas in the past CJS was assumed here, Node now begins to read the file and guesses that this is meant to be ESM once it sees those `import` statements. As it's not CJS, the require hook doesn't get called and Node then fails to parse the module due to the JSX we have within.

As this is meant to be the assumed/default value, I'm really hoping nothing breaks here.

PR #4630, opened 2025-01-11
Use .pop() to flush rerenderQueue instead of .shift()

jviide:
This pull request slightly modifies how `rerenderQueue` is flushed.

The idea is to use `.pop()` to flush the rerender queue instead of using `.shift()`. This also requires inverting the depth sorting comparison function to keep the queue in reverse order.

The rerender queue is an array that was previously emptied by calling the array's `.shift()` method repeatedly. In v8 the `.shift()` operation appears to O(n) time cost, based on looking at the [relevant parts of v8's source code](https://github.com/v8/v8/blob/main/src/builtins/array-shift.tq) and [a relevant open issue](https://issues.chromium.org/issues/42202676). Therefore, when `.shift()` is called repeatedly for a large array the aggregate time spent seems to grow faster than linearly.

In contrast the `.pop()` method of arrays seems to behave in O(1) way, at least when amortized over many operations. Therefor calling it repeatedly tends to take a linear-ish time in aggregate.

The overall performance effect of these changes is likely very small in the common case, as these changes address only the overhead from the rerender queue management. Usually everything else, like the queued rerenders themselves, take _much_ more time. Still, these changes aim to avoid the queue management becoming a bottleneck in extreme cases by avoiding O(n^2) behavior when hundreds or thousands rerenders are getting queued simultaneously.

The performance effect seems to be a wash when the queue stays relatively small, with the exception when the queue length is 1. There is now an optimization that skips sorting queues of length < 2.

### DIFF
--- a/compat/test/browser/render.test.js
+++ b/compat/test/browser/render.test.js
@@ -545,6 +545,40 @@ describe('compat render', () => {
 		expect(scratch.textContent).to.equal('foo');
 	});
 
+	it('should allow context as a component', () => {
+		const Context = createContext(null);
+		const CONTEXT = { a: 'a' };
+
+		let receivedContext;
+
+		class Inner extends Component {
+			render(props) {
+				return <div>{props.a}</div>;
+			}
+		}
+
+		sinon.spy(Inner.prototype, 'render');
+
+		render(
+			<Context value={CONTEXT}>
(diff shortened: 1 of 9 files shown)


JoviDeCroock (approved):
Genius!

PR #4636, opened 2025-01-18
Comment denoted hydration

JoviDeCroock:
Relates to https://github.com/preactjs/preact/issues/4442
Implementation from https://github.com/preactjs/preact-render-to-string/pull/376
Supersedes https://github.com/preactjs/preact/pull/4444

This implements a hydration approach with the comment markers, one thing that's noteworthy is that `oldDom` in our `diffChildren` algorithm becomes stale.

We diff the wrapping element of the Suspense and non-suspenseful child, we get to an oldDom which is the first child-node of said series. When we suspend we don't move oldDom, we can try and move it inside of `diff` but this won't affect the continued `diffChildren` so we are stuck inserting that node in front of our oldDom.

This has been solved in https://github.com/preactjs/preact/pull/4444/commits/647d13865b3c9730f17f8ce68644145925a226d7 through correctly skipping comment nodes while placing children.

I think the last part of this research would be to set up a comprehensive test suite for both `renderToStringAsync` as well as `renderToStream` so we can have fixtures for this behaviour, what do you all think?

### DIFF
--- a/compat/test/browser/component.test.js
+++ b/compat/test/browser/component.test.js
@@ -1,6 +1,6 @@
 import { setupRerender } from 'preact/test-utils';
 import { setupScratch, teardown } from '../../../test/_util/helpers';
-import React, { createElement, Component } from 'preact/compat';
+import React, { createElement } from 'preact/compat';
 
 describe('components', () => {
 	/** @type {HTMLDivElement} */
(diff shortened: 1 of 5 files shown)


rschristian (approved):
Sorry this took so long, wanted to find time to go back through your RFC & the RTS pull as my memory had become rusty. Looks great!

Re:comprehensive test suite, certainly would be a big benefit though I imagine it'd be a fair bit of work to cover all the various situations. Wonder how comprehensive React's suite is, might be able to repurpose some tests with less effort than writing from scratch.

marvinhagemeister (approved):

PR #4639, opened 2025-01-19
Support the template tag

### LINKED ISSUE #4635: Bug: `<template>` elements rendered incorrectly

**Describe the bug**

The children of template elements are not attached to the inner document fragment, but to the template element itself which is wrong.

**To Reproduce**

```tsx
import { useEffect, useRef } from 'preact/hooks';

export function App() {
  const ref = useRef<HTMLTemplateElement>(null)
  const refContainer = useRef<HTMLDivElement>(null)

  useEffect(() => {
    const clone = ref.current.content.cloneNode(true)
    refContainer.current.appendChild(clone)
  }, [])

  return (
    <div>
      <h1>Template bug</h1>
      <div ref={refContainer}></div>
      <template ref={ref}>
        <h2>this should show up</h2>
      </template>
    </div>
  );
}

render(<App />, document.getElementById('app'));
```

**Actual**

Image

**Expected behavior**

Image

JoviDeCroock:
Resolves #4635 

This builds out support for the template tag, the test asserts that the document fragment contains the tags we've rendered into it.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -525,7 +525,7 @@ function diffElementNodes(
 			if (oldHtml) dom.innerHTML = '';
 
 			diffChildren(
-				dom,
+				newVNode.type === 'template' && dom.content ? dom.content : dom,
 				isArray(newChildren) ? newChildren : [newChildren],
 				newVNode,
 				oldVNode,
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
Sweet, thanks for making a PR! Removed the `'content' in dom` check as template tags are guaranteed to have a fragment per spec.

PR #4644, opened 2025-01-25
ci: Fix path filter in PR Reporter action

rschristian:
Follow up to #4643

[The action is running again](https://github.com/preactjs/preact/actions/runs/12963641125) but it falls over as it cannot use the branch name from the fork as a reference to compare against -- that'll only work for local branches. To fix this, we simply switch over to the SHA, should do the trick for both forks & local branches all the same.

[Tests here](https://github.com/rschristian/debug__workflow_run-fork/actions), if anyone cares to take a look. Top 3 are a local branch (real workflow + 2 `workflow_run`-triggered events), 3 after that are from a fork.

### DIFF
--- a/.github/workflows/pr-reporter.yml
+++ b/.github/workflows/pr-reporter.yml
@@ -22,7 +22,7 @@ jobs:
       - uses: dorny/paths-filter@v3
         id: filter
         with:
-          ref: ${{ github.event.workflow_run.head_branch }}
+          ref: ${{ github.event.workflow_run.head_sha }}
           # Should be kept in sync with the filter in the CI workflow
           filters: |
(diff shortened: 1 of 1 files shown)


JoviDeCroock (approved):

rschristian:
Posting this here for myself in the future: whilst it seems like an obvious improvement to copy this to the CI workflow, ensuring it filters based on the latest commit only (thereby skipping running the benches for commits that haven't changed the source JS themselves), unfortunately `dorny/paths-filter` does not allow configuring `base` & `ref` parameters for PRs for some reason.

That might (& hopefully will) change, but for now, this is the best we can do short of splitting off the benches too, which I don't find particularly compelling.

PR #4649, opened 2025-02-08
Ensure it's tested with working imports

JoviDeCroock:
This ensures that `react/server`, `react/client` and many others keep working when they use the react-19 alias point. This also adds tests so that we won't regress in the future. There are still some issues with the import typings, will take another look soon.


compat/react-19.mjs
```diff
@@ -1,12 +1,15 @@
-import 'preact/compat';
+import all from 'preact/compat';
 
 import { options } from 'preact';
 
 const oldVNode = options.vnode;
 options.vnode = vnode => {
-	if (typeof vnode.type === 'function') {
-		vnode.type._forwarded = true;
+	if (typeof vnode.type === 'function' && vnode.ref) {
```
rschristian, line 7: This change probably needs to be mirrored to the CJS copy (in the off chance you didn't already have that change local & just not yet pushed)

rschristian (approved):

compat/react-19.mjs
```diff
@@ -1,12 +1,15 @@
...
+		vnode.type.__f = true;
 	}
 
 	if (oldVNode) oldVNode(vnode);
 };
+
+export default all;
```
rschristian, line 14: Not that it really matters but I suppose you could do this instead:

```suggestion
export { default } from 'preact/compat';
```

compat/package.json
```diff
@@ -34,19 +34,44 @@
...
+    "./jsx-dev-runtime": {
+      "import": "./jsx-dev-runtime.mjs",
+      "require": "./jsx-dev-runtime.js"
+    },
+    "./scheduler": {
+      "import": "./scheduler.mjs",
+      "require": "./scheduler.js"
+    },
     "./react-19": {
+      "types": "./src/index.d.ts",
```
rschristian, line 50: Oh this should be compat

```suggestion
      "types": "./compat/src/index.d.ts",
```

compat/react-19.mjs
```diff
@@ -1,12 +1,15 @@
-import 'preact/compat';
+import all from 'preact/compat';
```
JoviDeCroock, line 1:
```suggestion
import 'preact/compat';
```

compat/package.json
```diff
@@ -34,19 +34,44 @@
...
       "import": "./react-19.mjs",
       "require": "./react-19.js"
     },
-    "./jsx-runtime": {
+    "./react-19/jsx-runtime": {
+      "types": "./jsx-runtime/src/index.d.ts",
```
rschristian, line 55:
```suggestion
      "types": "../jsx-runtime/src/index.d.ts",
```

compat/package.json
```diff
@@ -34,19 +34,44 @@
...
       "import": "./jsx-runtime.mjs",
       "require": "./jsx-runtime.js"
     },
-    "./jsx-dev-runtime": {
+    "./react-19/jsx-dev-runtime": {
+      "types": "./jsx-runtime/src/index.d.ts",
```
rschristian, line 60:
```suggestion
      "types": "../jsx-runtime/src/index.d.ts",
```

rschristian (commented):
Missed this, just a copy/paste bug in the paths

Edit: Though do these need types? The non-React 19 entries don't

PR #4650, opened 2025-02-10
IE11: Fix incorrect node order with conditionals and text nodes

deadem:
There is an issue with using Element.contains in IE11. IE11 only partially supports contains: it works for Element nodes but does not support Text nodes. This can cause incorrect node ordering when conditionally rendering elements that include text nodes.

Minimal Repro:

```js
import { useEffect, useState } from 'preact/hooks';

export function App() {
  const [ showFirst, setShowFirst ] = useState(false);

  useEffect(function() {
    setTimeout(() => setShowFirst(true), 100);
  }, []);

  return (
    <div>
      {showFirst && <div>FIRST LINE</div>}
      plain text
      <div>LAST LINE</div>
    </div>
  );
}
```

Expected output:

```
FIRST LINE
plain text
LAST LINE
```

Actual output in IE11:

```
LAST LINE
FIRST LINE
plain text
```

### DIFF
--- a/compat/src/portals.js
+++ b/compat/src/portals.js
@@ -39,7 +39,7 @@ function Portal(props) {
 			nodeType: 1,
 			parentNode: container,
 			childNodes: [],
-			contains: () => true,
+			isPortal: true,
 			// Technically this isn't needed
 			appendChild(child) {
 				this.childNodes.push(child);
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -7,7 +7,7 @@ import {
 	MATCHED,
 	UNDEFINED
 } from '../constants';
-import { isArray } from '../util';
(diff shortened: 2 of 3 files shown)


compat/src/portals.js
```diff
@@ -39,7 +39,7 @@ function Portal(props) {
 			nodeType: 1,
 			parentNode: container,
 			childNodes: [],
-			contains: () => true,
+			isPortal: true,
```
JoviDeCroock, line 42: This is an unsafe change as other faux roots could be using this, also `contains` is not a hack, it's a DOM built-in

JoviDeCroock:
Closing: IE11-specific fix; v11 dropped IE11 support and the targeted `parentDom.contains()` path no longer exists in the rewritten children diff.

PR #4652, opened 2025-02-11
refactor: Breaking changes to outputs & pkg.json

rschristian:
Lands a few breaking changes at once, happy to revert any part or all of this if we feel anything's a problem.

1. Drop `.module.js` for `.mjs`

Obviously this is where the ecosystem is going and it already has pretty good support: Webpack started to adopt back in late 2018 and Node has supported it since v12.16 onward. This moves our effective back compat date up to about 5-6 years ago.

Perhaps too soon, but thought I'd at least suggest it.

2. Switch `package.json#exports.browser` to `package.json#exports.module`

This comes up fairly often, usually due to test runners trying to load modules meant for the browser in Node and then crashing as our browser build does not adhere to Node's ESM semantics. See #4406, https://github.com/preactjs/jest-preset-preact/pull/13#issuecomment-1404316519, https://github.com/preactjs/preact/pull/3634. While I'd tend to call that ecosystem error, it's not going away any time soon and we don't even use this output; it's just a standard ESM build, nothing browser-specific.

Instead of dropping this entirely though we can switch to `module` which some bundlers use to guarantee the package isn't instantiated twice ([Webpack's docs on the subject](https://webpack.js.org/guides/package-exports/#providing-commonjs-and-esm-version-stateful)). We had this change in the [v11 branch](https://github.com/preactjs/preact/commit/14d314d578d044e261cbb50b90b38c558df828c0) and I think it's good to carry over. Might also reduce some of the `__H` reports we have to deal with.

3. Drop the `.min` builds

These largely appear to be broken as-is (`preact.min.module.js` is not ESM, there's no CJS build as the IIFE output overwrites it immediately, `preact.min.umd.js` doesn't appear to be valid UMD either); the only thing maybe of value is the IIFE output `preact.min.js`. It's 30b lighter than `preact.umd.js`, but I don't know who's actually using it (if anyone at all).

Users are almost certainly better off with the plain UMD build and we'd be better off dropping it to reduce some confusion as it's unclear when this is meant to be used.


marvinhagemeister (approved):
LGTM, sounds reasonable to me. I'm kinda wondering if it's worth it to move to `type: "module"` at some point or if that would be too much of a breaking change.

rschristian:
Not sure why coveralls is breaking, weird.

I don't think we'd gain much from `"type": "module"` and it does move up the minimum a fair bit, `.mjs` saw way earlier support than `.cjs` after all.

marvinhagemeister:
fair point!

JoviDeCroock (approved):

karma.conf.js
```diff
@@ -181,7 +181,7 @@ function createEsbuildPlugin() {
 							coverage && [
 								'istanbul',
 								{
-									include: minify ? '**/dist/**/*.js' : '**/src/**/*.js'
+									include: minify ? '**/dist/**/*.{js,mjs}' : '**/src/**/*.js'
```
JoviDeCroock, line 184: Can we add `mjs` to the size tracker

rschristian: Compressed Size? Will do

rschristian: That actually needs to be addressed in the action itself.

It's _meant to_ support `.mjs` as-is, but we're using a [JS fallback](https://github.com/preactjs/compressed-size-action/blob/6fa0e7ca017120c754863b31123c5ee2860fd434/src/index.js#L45) when we define a default in the [action config](https://github.com/preactjs/compressed-size-action/blob/6fa0e7ca017120c754863b31123c5ee2860fd434/action.yml#L39). I'll write up a PR fixing that shortly, up to you on whether we want to hold off on this until that gets in though.

rschristian: All good now

PR #4657, opened 2025-02-12
Move `defaultProps` into `preact/compat`

JoviDeCroock:
This will be handled in `options.vnode` for function/class components. This hook gets called for every invocation of `jsx`/`createElement` and `cloneElement`.

Not sure yet on how to fix the types for `Context`


rschristian:
Still playing around with it, but _I think_ we want something starting like this:

```ts
declare module 'preact' {
	export interface FunctionComponent<P = {}> {
		(props: preact.RenderableProps<P>, context?: any): preact.ComponentChildren;
		displayName?: string;
		defaultProps?: Partial<P> | undefined;
	}

	export interface ComponentClass<P = {}, S = {}> {
		new (props: P, context?: any): preact.Component<P, S>;
		displayName?: string;
		defaultProps?: Partial<P>;
		contextType?: preact.Context<any>;
		getDerivedStateFromProps?(
			props: Readonly<P>,
			state: Readonly<S>
		): Partial<S> | null;
		getDerivedStateFromError?(error: any): Partial<S> | null;
	}
}
```

The problem is the `abstract class Component`, which, as far as I can tell, [cannot be merged](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#disallowed-merges). I don't yet know how to work around that

PR #4659, opened 2025-02-12
Optimize jsx/core bundle size by extracting defaultProps

JoviDeCroock:
Seems like a worthwhile optimization

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -1,3 +1,4 @@
+import { UNDEFINED } from './constants';
 import { _catchError } from './diff/catch-error';
 
 /**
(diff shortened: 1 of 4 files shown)


marvinhagemeister (approved):

rschristian (approved):
Hm, do the benches need a rerun or is that an actual regression? Seems fairly larg, larger than I'd think

src/options.js
```diff
@@ -10,6 +11,18 @@ import { _catchError } from './diff/catch-error';
...
+		// If a Component VNode, check for and apply defaultProps
+		// Note: type may be undefined in development, must never error here.
+		if (typeof type == 'function' && type.defaultProps != null) {
+			for (const i in type.defaultProps) {
+				if (props[i] === UNDEFINED) {
+					props[i] = type.defaultProps[i];
+				}
+			}
+		}
+	},
```
marvinhagemeister, line 25: wondering if this is the reason for the perf regression. This means we always must call `options.vnode` which we didn't before.

JoviDeCroock: It probably is, I did an attempt at making it cheaper by not destructuring but might not help. Was really excited about the byte savings 😅

developit:
regression is definitely from the unconditional hook call yeah.

might be a good time to try a PR that swaps us to getter/setter pairs for each hook that define locals?

PR #4665, opened 2025-02-14
Remove automatic px suffix

JoviDeCroock:


jsx-runtime/package.json
```diff
@@ -3,6 +3,7 @@
   "amdName": "jsxRuntime",
   "version": "1.0.0",
   "private": true,
+  "sideEffects": false,
```
JoviDeCroock, line 6: Just noticed we were missing this, this might make the `jsx` stuff tree shake out and same for the `IS_NON_DIMENSIONAL`

jsx-runtime/src/index.js
```diff
@@ -1,6 +1,5 @@
 import { options, Fragment } from 'preact';
-import { encodeEntities } from './utils';
-import { IS_NON_DIMENSIONAL } from '../../src/constants';
+import { encodeEntities, IS_NON_DIMENSIONAL } from './utils';
```
rschristian, line 2: Would it be too problematic to make this compat-only? Moving it out of core is good for marketing and whatnot but I'd prefer to go a step further if at all possible

rschristian (approved):

jsx-runtime/src/index.js
JoviDeCroock: I will defer this to @marvinhagemeister as Deno uses the JSX template transform exclusively atm

marvinhagemeister (approved):

PR #4672, opened 2025-02-16
types: Add auxClick event types

rschristian:
From the project referenced here: https://github.com/preactjs/preact-www/issues/1240#issuecomment-2661172956

One thing worth calling out here is that the spec recently (-ish, 4 years ago) changed this event from extending a `MouseEvent` to a `PointerEvent` and [only Chrome/Chromium-based browsers support this](https://developer.mozilla.org/en-US/docs/Web/API/Element/auxclick_event#browser_compatibility) yet. Do we want to make this a `MouseEvent` with a note to change this in the future, or use `PointerEvent` now (as this PR does)? I have no opinions one way or the other.


JoviDeCroock (approved):
Love it when specs are in flux, todo sounds fine

PR #4680, opened 2025-02-17
Fix memory leak

### LINKED ISSUE #4679: Fragments cause memory leak

**Describe the bug**
Wrapping component expressions in a Fragment causes a memory leak. https://github.com/preactjs/preact/issues/3668 also concerns a possible memory leak, but it may be unrelated.

**To Reproduce**

```js
<!-- no memory leak -->>
<div>{state}</div>

<>
  <!-- memory leak -->
  <div>{state}</div>
</>
```

I reproduced the issue [here](https://github.com/builds-dev/preact-memory-leak-repro) by initializing the vite preact starter template and wrapping the top level component in `<></>`. To make the memory leak obvious, I added some fluff (a bunch of divs) so that the memory piles up quickly, but it should very slowly leak memory without it. I originally discovered this issue in Electron 22 and Electron 32 (Chrome), and then tested my reproduction repo on Firefox and Chromium (on Ubuntu 22), and observed the memory leak.

I tested Preact v10.6.0, the current v10.26.0 and the upcoming v11. The memory leak occurs in all versions.

JoviDeCroock:
Fixes #4679 

Basically what happens is that when we encounter a fragment we unwrap it to return is children, this way when elements get wrapped with fragments we don't risk losing state. This has a consequence when it is used in hooks environments.

As can be seen in the reproduction, if an effect closure takes a snapshot of the outer closure it will take the `Fragment` we unwrap into memory, this means that we'll effectively start mutating its children.

Every time we rerender we'll take a new snapshot of the closure which retains a copy of `Fragment` with pointers to all of its populated children.

With this measure we counter-act this by removing the reference to `children` which removes the risk of tracking this.

Fun fact, when we implement #4679's replay with class components there is no memory leak.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -259,6 +259,11 @@ export function diff(
 				tmp != null && tmp.type === Fragment && tmp.key == null;
 			let renderResult = isTopLevelFragment ? tmp.props.children : tmp;
 
+			if (isTopLevelFragment) {
+				tmp.props.children = null;
+				console.log('isTopLevelFragment', tmp);
+			}
+
 			oldDom = diffChildren(
 				parentDom,
 				isArray(renderResult) ? renderResult : [renderResult],
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -261,7 +261,6 @@ export function diff(
 
 			if (isTopLevelFragment) {
 				tmp.props.children = null;
-				console.log('isTopLevelFragment', tmp);
 			}
 
 			oldDom = diffChildren(


src/diff/index.js
```diff
@@ -259,6 +259,11 @@ export function diff(
 				tmp != null && tmp.type === Fragment && tmp.key == null;
 			let renderResult = isTopLevelFragment ? tmp.props.children : tmp;
 
+			if (isTopLevelFragment) {
+				tmp.props.children = null;
+				console.log('isTopLevelFragment', tmp);
```
JoviDeCroock, line 264:
```suggestion
```

marvinhagemeister (approved):
LGTM, really good work on debugging and finding this!!

rschristian (approved):
Awesome find!

developit (approved):

vpzomtrrfrt:
This appears to have broken rendering for me in some cases

haven't narrowed down exactly when yet

JoviDeCroock:
The only possibility that I see for this to fail would be something like `const x = (<>Foo</>)` and reusing that across your app

PR #4681, opened 2025-02-18
perf: convert null to constant

Ben-Brady:


Ben-Brady:
Meant to do it on a local branch to test, didn't mean to do a PR upstream

PR #4683, opened 2025-02-18
types: Require initial value in `useRef`

rschristian:
I quite like the [change React made to their types in v19](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#useref-requires-argument) to require an initial argument be passed to `useRef`, it fixes some of the fiddly details of `useRef` that are really easy to stumble into. This is a breaking type change so v11 it goes.

### DIFF
--- a/hooks/src/index.d.ts
+++ b/hooks/src/index.d.ts
@@ -52,13 +52,6 @@ export function useReducer<S, A, I>(
 	init: (arg: I) => S
 ): [S, Dispatch<A>];
 
-/** @deprecated Use the `Ref` type instead. */
-type PropRef<T> = MutableRef<T>;
-
-interface MutableRef<T> {
-	current: T;
-}
-
 /**
  * `useRef` returns a mutable ref object whose `.current` property is initialized to the passed argument
  * (`initialValue`). The returned object will persist for the full lifetime of the component.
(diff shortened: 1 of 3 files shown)


src/index.d.ts
```diff
@@ -37,9 +37,9 @@ export interface VNode<P = {}> {
 
 export type Key = string | number | any;
 
-export type RefObject<T> = { current: T | null };
+export type RefObject<T> = { current: T };
 export type RefCallback<T> = (instance: T | null) => void;
-export type Ref<T> = RefObject<T> | RefCallback<T> | null;
+export type Ref<T> = RefCallback<T> | RefObject<T | null> | null;
```
rschristian, line 42: [React for reference](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/9efe05446271637b014cc445c892923a4f4510ef/types/react/index.d.ts#L192)

hooks/src/index.d.ts
```diff
@@ -68,9 +61,9 @@ interface MutableRef<T> {
  *
  * @param initialValue the initial value to store in the ref object
  */
-export function useRef<T>(initialValue: T): MutableRef<T>;
-export function useRef<T>(initialValue: T | null): RefObject<T>;
-export function useRef<T = undefined>(): MutableRef<T | undefined>;
+export function useRef<T>(initialValue: T): RefObject<T>;
+export function useRef<T>(initialValue: T | null): RefObject<T | null>;
+export function useRef<T>(initialValue: T | undefined): RefObject<T | undefined>;
```
rschristian, line 66: [React for reference](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/9efe05446271637b014cc445c892923a4f4510ef/types/react/index.d.ts#L1722-L1746)

JoviDeCroock (approved):

PR #4687, opened 2025-02-18
Vitest switch

JoviDeCroock:
## Requirements before merge

- [x] Remove `(done)` tests
- [x] Ensure we have coverage setup
- [x] Check how we differ from the existing karma setup
- [x] Look into potential duplicate bundle issue (options.hook not called or mangle not applied)
- [x] Transform `options` property names in tests with regards to mangle
- [x] Fix size and performance diff
- [x] Figure out why tests are running against dist rather than src
- [x] Fix types in test files
- [x] Fix TODO comments in tests
- [x] Get coverage working, looks like it's not reported
- [x] Make `pnpm test:vitest:min` work, this somehow makes two tests fail
- [x] Get `tsc` to work again... Broke with bump of types/node, this is part of `pnpm lint`
- [x] Fix `npm run test:mocha`
- [x] Use `playwright`
- [x] Restore CI to use `npm run lint && npm run test:unit`

## Follow-ups

- Get rid of `sinon`
- Get rid of `sinon-chai`


vitest.config.mjs
```diff
@@ -0,0 +1,131 @@
...
+								loader: 'jsx',
+								contents: await fs.readFile(args.path, 'utf8')
+							};
+						});
+					}
+				}
+			]
+		}
+	},
+	test: {
```
rschristian, line 112: I think we're missing an `include` filter here, probably want to match karma's?

vitest.config.mjs
```diff
@@ -0,0 +1,156 @@
...
+	),
+	'^preact/hooks$': path.join(
+		root,
+		MINIFY ? 'hooks/dist/hooks.js' : 'hooks/src/index.js'
+	),
+	'^preact/test-utils$': path.join(
+		root,
+		MINIFY ? 'test-utils/dist/test-utils.js' : 'test-utils/src/index.js'
+	)
+};
```
rschristian, line 50: Correct form for rollup I believe, a quick check seems like we can drop the alias from esbuild? Playing around with this myself, feel free to ignore me ofc

```suggestion
const alias = [
	{ find: /^react$/, replacement: path.join(root, 'compat/src/index.js') },
	{ find: /^react-dom$/, replacement: path.join(root, 'compat/src/index.js') },
	{ find: /^preact$/, replacement: path.join(root, 'src/index.js') },
	{ find: /^preact\/compat$/, replacement: path.join(root, 'compat/src/index.js') },
	{ find: /^preact\/jsx-runtime$/, replacement: path.join(root, 'jsx-runtime/src/index.js') },
	{ find: /^preact\/jsx-runtime\/src$/, replacement: path.join(root, 'jsx-runtime/src') },
	{ find: /^preact\/jsx-dev-runtime$/, replacement: path.join(root, 'jsx-dev-runtime/src/index.js') },
	{ find: /^preact\/debug$/, replacement: path.join(root, 'debug/src/index.js') },
	{ find: /^preact\/hooks$/, replacement: path.join(root, 'hooks/src/index.js') },
	{ find: /^preact\/test-utils$/, replacement: path.join(root, 'test-utils/src/index.js') }
];
```

test/browser/components.test.js
```diff
@@ -1163,21 +1163,22 @@ describe('Components', () => {
...
-				.and.to.have.returned(
-					sinon.match({
-						props: {
-							j: 2,
-							i: 2,
-							foo: 'bar'
-						}
-					})
-				);
+			// TODO: fix this
```
JoviDeCroock, line 1170: Some of these todos still need fixing

rschristian (approved):
Lot of work went into this, great job!

---

However, I'm not in love with merging this as-is, entirely due to the state of Vitest, not any of the work here.

- It seems slower across the board
- It's an extra 100MB+ of `node_modules` & an 800MB in `~/.cache/ms-playright` (tried using local Chrome install but I think that'd require different config/setup, it doesn't seem to function properly out-of-the-box)
- It's a hell of a lot more dependencies
- We've had to switch to some Vitest-specific APIs. While the "lock-in" doesn't look to be too rough, we are switching from runner agnostic (or meant to be runner agnostic) tools to something that's specific & will need to be changed if we need to migrate again in the future
- We've encountered a fair number of Vitest bugs in the process of this PR

I suppose most importantly, I don't see a clear benefit to switching at the moment? It's something newer and that has some value, and we do shave a bit of config, but browser testing seems pretty well established, as are our tests, so I don't see any obvious possibilities this opens up. Karma, while deprecated, is super stable & doesn't seem to be holding us back at all.

I know Vitest's maintainers have been hard at work improving things, and with a bit of luck some of this will be addressed soon-ish, but at this particular moment it seems like a rather large regression instead of a clear improvement.

---

That being said, I don't want to hold this back at all; if people want this then let's go forward with it.

JoviDeCroock:
> It seems slower across the board and It's an extra 100MB+ of node_modules & an 800MB in ~/.cache/ms-playright (tried using local Chrome install but I think that'd require different config/setup, it doesn't seem to function properly out-of-the-box)

I agree, merging this feels like a regression, all though I am convinced that we should eventually move to a new tool because of karma being deprecated and it blocking upgrading our dependencies. I am going to explore the landscape a bit more, maybe there's a middle ground here.

Ideally `webdriver` would be as fast as `playwright` but... I am unsure where this is headed in terms of performance.

> We've had to switch to some Vitest-specific APIs. While the "lock-in" doesn't look to be to rough, we are switching from runner agnostic (or meant to be runner agnostic) tools to something that's specific & will need to be changed if we need to migrate again in the future

This is more because `sinon` in more recent versions is very broken, when we use `sinon.spyOn` it will break the `prototype` of the underlying object 😅 

> We've encountered a fair number of Vitest bugs in the process of this PR

This was part of my intention, I wanted to uncover bugs so we can help the vitest people with testing browser mode as I am really looking for a community project to step into the shoes of karma.

> The --skipLibCheck flags can probably be removed now that it was added to js-lint.json, right?

Well, sadly, the setting doesn't work 😅 

----

I agree with you, let's hold off on merging this, I will surface this with the vitest people so there's information sharing.

rschristian:
> This is more because sinon in more recent versions is very broken, when we use sinon.spyOn it will break the prototype of the underlying object 😅

Oh fair enough, wasn't quite sure if it was sinon or vitest that was the problem there.

> Well, sadly, the setting doesn't work 😅

Huh. I thought the flag would act the exact same there. Admittedly I can't even run `test:ts:core` at the moment, some Node error... `test:ts:compat` is a-ok though. Fair enough on that too then.

.github/workflows/build-test.yml
```diff
@@ -35,9 +35,7 @@ jobs:
           COVERAGE: true
           FLAKEY: false
         # Not using `npm test` since it rebuilds source which npm ci has already done
-        run: |
-          npm run lint
-          npm run test:unit
+        run: npm run lint && npm run test:unit
```
JoviDeCroock, line 38:
```suggestion
        run: npm run test:vitest
```

rschristian (approved):
Looks great!

marvinhagemeister (approved):
LGTM, great work!

PR #4692, opened 2025-02-18
feat: Add command invoker types

lukewarlow:
Reference:

- [command](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#command)
- [commandfor](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#commandfor)

------------------------------------

See chrome [Intent to ship](https://groups.google.com/a/chromium.org/g/blink-dev/c/ctNbl4gWLuk)

Leaving as Draft till the spec is merged / Chrome i2S is approved.

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -1623,8 +1623,8 @@ export namespace JSXInternal {
 	interface ButtonHTMLAttributes<T extends EventTarget = HTMLButtonElement>
 		extends HTMLAttributes<T> {
 		command?: Signalish<string | undefined>;
-		commandfor?: Signalist<string | undefined>;
-		commandFor?: Signalist<string | undefined>;
+		commandfor?: Signalish<string | undefined>;
+		commandFor?: Signalish<string | undefined>;
 		disabled?: Signalish<boolean | undefined>;
 		form?: Signalish<string | undefined>;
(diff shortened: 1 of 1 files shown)


src/jsx.d.ts
```diff
@@ -1613,6 +1613,9 @@ export namespace JSXInternal {
 
 	interface ButtonHTMLAttributes<T extends EventTarget = HTMLButtonElement>
 		extends HTMLAttributes<T> {
+		command?: Signalish<string | undefined>;
+		commandfor?: Signalist<string | undefined>;
+		commandFor?: Signalist<string | undefined>;
```
lukewarlow, line 1618: There's a .commandForElement property that you can set to an element, it would be good to support this in Preact but popoverTargetElement also doesn't seem to be handled so will omit that for now.

lukewarlow:
TODO: Add onCommand IDL event handler

src/jsx.d.ts
rschristian, line 1618: Not sure why our `.d.ts` files aren't type checked, we have something that seems to poison the ts lang server

```suggestion
		commandfor?: Signalish<string | undefined>;
		commandFor?: Signalish<string | undefined>;
```

rschristian (approved):
Thanks!

PR #4693, opened 2025-02-18
feat: Add dialog closedby types

lukewarlow:
See https://html.spec.whatwg.org/multipage/interactive-elements.html#attr-dialog-closedby


marvinhagemeister (approved):
Sweet, thanks!

PR #4698, opened 2025-02-19
chore: Fix TS lang server support in our `.d.ts` files

rschristian:
Long been a bit of a headache, we don't get an TS support in our `.d.ts` files which allows typos, usused values, etc. to run a bit rampant. Turns out it's a simple issue: `jsconfig.json` has `skipLibCheck` set to `false` by default, and this is seemingly undocumented.

Easy fix and has surfaced a few (internal) errors already.

### DIFF
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -6,14 +6,6 @@ import { JSXInternal } from '../../src/jsx';
 import * as _Suspense from './suspense';
 import * as _SuspenseList from './suspense-list';
 
-interface SignalLike<T> {
-	value: T;
-	peek(): T;
-	subscribe(fn: (value: T) => void): () => void;
-}
-
-type Signalish<T> = T | SignalLike<T>;
-
 // export default React;
 export = React;
 export as namespace React;
@@ -272,7 +264,7 @@ declare namespace React {
 	}
 
 	export function forwardRef<R, P = {}>(
-		fn: ForwardFn<P, R>
+		fn: ForwardRefRenderFunction<R, P>
 	): preact.FunctionalComponent<PropsWithoutRef<P> & { ref?: preact.Ref<R> }>;
 
 	export type PropsWithoutRef<P> = Omit<P, 'ref'>;
--- a/hooks/src/internal.d.ts
+++ b/hooks/src/internal.d.ts
@@ -2,7 +2,9 @@ import {
 	Options as PreactOptions,
 	Component as PreactComponent,
 	VNode as PreactVNode,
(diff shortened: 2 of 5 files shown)


hooks/src/internal.d.ts
```diff
@@ -30,14 +32,14 @@ export interface ComponentHooks {
 	_pendingEffects: EffectHookState[];
 }
 
-export interface Component extends PreactComponent<any, any> {
+export interface Component extends Omit<PreactComponent<any, any>, '_renderCallbacks'> {
```
rschristian, line 35: This PR adds a healthy sprinkling of `Omit` which I think is correct given our use cases -- generally, our internal types intentionally clash with the base/user-exposed types.

hooks/src/internal.d.ts
```diff
@@ -52,12 +54,12 @@ export type HookState =
...
-	_pendingArgs?: undefined;
-	_component?: undefined;
-	_cleanup?: undefined;
+	_nextValue?: unknown;
+	_pendingValue?: unknown;
+	_args?: unknown;
+	_pendingArgs?: unknown;
+	_component?: unknown;
+	_cleanup?: unknown;
 }
```
rschristian, line 63: `undefined` is a more specific type than `unknown` which disallows some interfaces from extending these properties as they need to.

jsconfig-lint.json
```diff
@@ -1,4 +1,9 @@
 {
   "extends": "./jsconfig.json",
-  "include": ["src/**/*", "hooks/src/**/*"]
-}
+  "include": [
+    "src/**/*",
+    "hooks/src/**/*",
+    "compat/**/*.d.ts",
+    "jsx-runtime/**/*.d.ts"
```
rschristian, line 7: compat & jsx-runtime source & test files throw up some errors yet but I think it's worth linting their types at least as these are user-exposed. Will help guard against typos and what not

JoviDeCroock (approved):

compat/src/index.d.ts
```diff
@@ -272,7 +264,7 @@ declare namespace React {
 	}
 
 	export function forwardRef<R, P = {}>(
-		fn: ForwardFn<P, R>
+		fn: ForwardRefRenderFunction<R, P>
```
rschristian, line 267: This is from #4675 -- `ForwardFn` is not the name in React & notably it has the generics reversed. It was marked as deprecated in that PR so all this is doing is replacing it with the correct, non-deprecated version (the internals of both types match).

Issue #4699, opened 2025-02-19
Preact diffing incorrectly when there is a `null` or `false` in the rendered VNode's children and one of its sibling is remounted


navahg:
- [x] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
Say, I am rendering the following structure

> div
>> Comp A {key="foo"}
>> null
>> Comp B

here, if I remount the Comp A by changing its key, both Comp A and Comp B are remounted. And if the div has more siblings, then the diffing gets weird.

This only happens when the remounting node has a `null` right after it. Any other place, no issue is happening.

**To Reproduce**

https://codesandbox.io/p/sandbox/unexpected-unmounting-preact-glx9zw

1. Go to codesanbox
2. Click on the button - "Remount Comp:A"
3. Check the console logs to see other siblings getting mounted/unmounted

**Expected behavior**
The expected behavior is only the Comp A should be unmounted and mounted. React works as expected - https://codesandbox.io/p/sandbox/44vv8d

rschristian:
Keys are meant to be stable between renders -- intentionally changing them is putting yourself at the mercy of the diff which may result in sub-optimal guesses for how to reconcile. This is why keys exist after all, to fix the situations where frameworks have to guess what app developers intend. Add stable keys and the issue will go away.

There might be a path to matching React but the better solution I think is to correct your usage. We have [a portion of our tutorial that goes through keys](https://preactjs.com/tutorial/08-keys) which you might want to take a look at.

JoviDeCroock:
I see why this is annoying and I'd have to look deeper into the heuristic that React uses to determine the offsets here. It comes down to the following, I believe

- We can't find a matching vnode because the key changed in https://github.com/preactjs/preact/blob/main/src/diff/children.js#L383
- We decrement the skew to deal with offsets, i.e. you insert a node somewhere https://github.com/preactjs/preact/blob/main/src/diff/children.js#L256
- There is an explanation about the current optimisations our algorithm accounts for https://github.com/preactjs/preact/blob/main/src/diff/children.js#L264-L279 (these are chosen very intentionally)
- When we get to the next node our skew will effectively point us at an offset of -1
- Which means that your next `<Comp />` will point at the empty spot
- This means that we'll consider this a component that disappeared https://github.com/preactjs/preact/blob/main/src/diff/children.js#L410

[...]

PR #4700, opened 2025-02-20
Keyed nullish placeholders cause re-mounts

JoviDeCroock:
Fixes https://github.com/preactjs/preact/issues/4699 and a whole lot more

There's a few cases that get solved in this PR, let's go over them!

First and foremost, the issue points us at an issue where having a null'ish middle node can lead to false positives. In the example we have one keyed node like the following

```
<CompA key=1 />
null
<CompB />
<CompB />
```

When we create this initially we'll see 3 creations for A, B and B. This is expected, when we swap out CompA by changing its key to i.e. 2 we expect that the only new creation we'll see is from `<CompA />`, however this is not the case we'll also remount one of the two `<CompB />` nodes and reset their state if they are using state which is unexpected.

This is caused by the following:

- We can't find a matching vnode because the key changed
- We decrement the skew to deal with offsets, we assume that the amount of children grew in the new variant hence we increment skew. (when we increment skew we'll look at a smaller index in the oldChildren i.e. [1, 2, 3] becomes [0, 1, 2, 3] --> when we diff 0 that's technically -1 in oldChildren and we want to compare 1 with 1)
- When we get to the next skew in newChildren we'll compare `<CompB />` with `null`, this means that we'll erroneously assume we swapped out `<CompB />` with null which makes us unmount it.

We fix the above by adding a check in the mount case that does the following

- Decrement the skew if the list of newChildren is larger than the list of oldChildren (we added a node 0 becomes -1)
- Increment the skew if the list of newChildren is smaller than the list of oldChildren (we removed a node 0 becomes 1)
- If the list stays the same, like our issue case, we don't do anything!

This made me start looking into non-keyed cases and indeed, this issue existed in the unkeyed case as well which luckily is also fixed with the above measure. I've added tests for all of these cases. When digging even deeper I found some [general issues](https://github.com/preactjs/preact/pull/4700/files#diff-bbe8c651c89dccfac2b57fc4ac13928d4bef226b6163dd6975ee6caf53797c41R1016-R1126) with how we handle placeholders (aka holes in our array). Generally when we replace a keyed node, append or prepend with a placeholder we see re-mounting issues. I've counter-acted that by replacing our `null` case in the child-matching with a `&& !key` condition, this means that for keyed nodes we'll always search the children as this is kind of the goal of keys, we can move them around without fear.

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -253,6 +253,20 @@ function constructNewChildrenArray(
 
 		if (isMounting) {
 			if (matchingIndex == -1) {
+				// When the array of children is growing we need to decrease the skew
+				// as we are adding a new element to the array.
+				// Example:
+				// [1, 2, 3] --> [0, 1, 2, 3]
+				// oldChildren   newChildren
+				//
+				// The new element is at index 0, so our skew is 0,
+				// we need to decrease the skew as we are adding a new element.
+				// The decrease will cause us to compare the element at position 1
+				// with value 1 with the element at position 0 with value 0.
+				//
+				// A linear concept is applied when the array is shrinking,
+				// if the length is unchanged we can assume that no skew
+				// changes are needed.
 				if (newChildrenLength > oldChildrenLength) {
 					skew--;
 				} else if (newChildrenLength < oldChildrenLength) {
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -252,8 +252,12 @@ function constructNewChildrenArray(
 		const isMounting = oldVNode == NULL || oldVNode._original === NULL;
 
 		if (isMounting) {
-			if (matchingIndex == -1 && oldChildrenLength !== newChildrenLength) {
-				skew--;
+			if (matchingIndex == -1) {
+				if (newChildrenLength > oldChildrenLength) {
+					skew--;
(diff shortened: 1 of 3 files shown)


src/diff/children.js
```diff
@@ -407,7 +412,7 @@ function findMatchingIndex(
 		(oldVNode != NULL && (oldVNode._flags & MATCHED) == 0 ? 1 : 0);
 
 	if (
-		oldVNode === NULL ||
+		(oldVNode === NULL && !childVNode.key) ||
```
JoviDeCroock, line 415:
```suggestion
		(oldVNode === NULL && childVNode.key == null) ||
```

src/diff/children.js
```diff
@@ -253,7 +253,12 @@ function constructNewChildrenArray(
 
 		if (isMounting) {
 			if (matchingIndex == -1) {
-				skew--;
+				//skew--;
+				if (newChildrenLength > oldChildrenLength) {
```
JoviDeCroock, line 257: This can probably be golfed but my brain isn't there today

marvinhagemeister (approved):
LGTM, surprised how little changes were needed. Great job!

rschristian (approved):

PR #4701, opened 2025-02-20
chore: Fix mocha in Node v22+

rschristian:
Alternative to #4629, should be much safer.

Basic gist is that `require(esm)` seemingly begins to parse modules to determine whether it's CJS or ESM, and in the case of the latter, require hooks are not yet supported against ESM modules so Babel doesn't transform the JSX within.

---

From Joyee's blog:

> So that led to [this PR](https://github.com/nodejs/node/pull/51977). The major difference between this and the PR in 2019 was that this tried to keep the scope of require(esm) small and only supported loading synchronous ESM. As it turned out, this is not a controversial idea among collaborators/TSC at all and landed without much pushbacks (it derailed a tiny bit due to the discussions about loader hooks, but at least we agreed that hooks support could be left for follow-ups and require(esm) itself should be back on track).

https://joyeecheung.github.io/blog/2024/03/18/require-esm-in-node-js/


marvinhagemeister (approved):
LGTM, would d o a quick format because newlines are missing at the end.

JoviDeCroock (approved):

rschristian:
Do we want newlines? Our editor config disables those for `.json` files:

https://github.com/preactjs/preact/blob/2a788d638ecfe97fd88c52609c7d426bc50aa1c0/.editorconfig#L10-L13

Will check with biome shortly, not sure what it enforces by default

PR #4705, opened 2025-02-24
types: Flesh out HTMLMediaElement types

rschristian:
Adds some more types & corrects a few for `HTMLMediaElement` specifically

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -1263,19 +1263,24 @@ export namespace JSXInternal {
 		/** @deprecated See https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contextmenu */
 		contextmenu?: Signalish<string | undefined>;
 		controls?: Signalish<boolean | undefined>;
-		controlsList?: Signalish<string | undefined>;
+		controlslist?: Signalish<'nodownload' | 'nofullscreen' | 'noremoteplayback' | undefined>;
+		controlsList?: Signalish<'nodownload' | 'nofullscreen' | 'noremoteplayback' | undefined>;
 		coords?: Signalish<string | undefined>;
 		crossOrigin?: Signalish<string | undefined>;
 		crossorigin?: Signalish<string | undefined>;
+		currentTime?: Signalish<number | undefined>;
 		data?: Signalish<string | undefined>;
 		dateTime?: Signalish<string | undefined>;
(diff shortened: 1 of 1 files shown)


src/jsx.d.ts
```diff
@@ -1263,19 +1263,24 @@ export namespace JSXInternal {
 		/** @deprecated See https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contextmenu */
 		contextmenu?: Signalish<string | undefined>;
 		controls?: Signalish<boolean | undefined>;
-		controlsList?: Signalish<string | undefined>;
+		controlslist?: Signalish<'nodownload' | 'nofullscreen' | 'noremoteplayback' | undefined>;
+		controlsList?: Signalish<'nodownload' | 'nofullscreen' | 'noremoteplayback' | undefined>;
```
rschristian, line 1267: Might be worth altering `AllHTMLElements` to get all of its properties from `IntrinsicElements`, eliminating the need to copy from the more per-element interfaces. Will give that a go separately

src/jsx.d.ts
```diff
@@ -1917,16 +1925,23 @@ export namespace JSXInternal {
...
+		currentTime?: Signalish<number | undefined>;
+		defaultMuted?: Signalish<boolean | undefined>;
+		defaultPlaybackRate?: Signalish<number | undefined>;
+		disableremoteplayback?: Signalish<boolean | undefined>;
+		disableRemotePlayback?: Signalish<boolean | undefined>;
 		loop?: Signalish<boolean | undefined>;
 		mediaGroup?: Signalish<string | undefined>;
 		muted?: Signalish<boolean | undefined>;
-		playsinline?: Signalish<boolean | undefined>;
-		playsInline?: Signalish<boolean | undefined>;
```
rschristian, line 1927: These are specific to `HTMLVideoElement`, they were moved down to that interface.

marvinhagemeister (approved):
LGTM

piotr-cz:
PR: #4744

Indeed, the MDN docs for [Video > controlslist](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#controlslist) are misleading, but quite clear on [HTMLMediaElement.controlsList](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/controlsList)

PR #4706, opened 2025-02-24
types: Refactor 'AllHTMLAttributes' interface

rschristian:
In the interest of reducing the maintenance burden of having per-element type interfaces AND the legacy `AllHTMLAttributes` interface, which should theoretically be kept in sync, this converts the latter into an extension of every per-element interface we have.

Doing so has caught a few errors already, though there's a limited few properties that have to become `any` when used from this interface (`HTMLInputElement.type` vs `HTMLAnchorElement.type`, for instance).

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -1290,9 +1290,9 @@ export namespace JSXInternal {
 			TrackHTMLAttributes<RefType>,
 			VideoHTMLAttributes<RefType> {
 		/* Properties that are incompatible between elements */
+		loop?: Signalish<any | undefined>;
 		target?: Signalish<any | undefined>;
 		type?: Signalish<any | undefined>;
-		loop?: Signalish<any | undefined>;
 	}
 
 	export interface HTMLAttributes<RefType extends EventTarget = EventTarget>
@@ -1714,7 +1714,8 @@ export namespace JSXInternal {
 		as?: Signalish<string | undefined>;
 		crossorigin?: Signalish<HTMLAttributeCrossOrigin>;
 		crossOrigin?: Signalish<HTMLAttributeCrossOrigin>;
-		fetchPriority?: Signalish<'high' | 'low' | 'auto'>;
+		fetchpriority?: Signalish<'high' | 'low' | 'auto' | undefined>;
+		fetchPriority?: Signalish<'high' | 'low' | 'auto' | undefined>;
 		href?: Signalish<string | undefined>;
 		hreflang?: Signalish<string | undefined>;
 		hrefLang?: Signalish<string | undefined>;
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -4,7 +4,6 @@
 import {
 	ClassAttributes,
 	Component,
-	ComponentChild,
 	ComponentType,
 	FunctionComponent,
 	PreactDOMAttributes,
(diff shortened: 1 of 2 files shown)


src/jsx.d.ts
```diff
@@ -24,20 +23,20 @@ type Booleanish = boolean | 'true' | 'false';
...
 /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/ToggleEvent) */
 interface ToggleEvent extends Event {
-    /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/ToggleEvent/newState) */
-    readonly newState: string;
-    /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/ToggleEvent/oldState) */
-    readonly oldState: string;
+	/** [MDN Reference](https://developer.mozilla.org/docs/Web/API/ToggleEvent/newState) */
+	readonly newState: string;
+	/** [MDN Reference](https://developer.mozilla.org/docs/Web/API/ToggleEvent/oldState) */
+	readonly oldState: string;
```
rschristian, line 29: I've reverted this in a few commits now, figure these formatting corrections are worth getting in just to get out of the way

src/jsx.d.ts
```diff
@@ -1675,7 +1477,6 @@ export namespace JSXInternal {
 	interface DetailsHTMLAttributes<T extends EventTarget = HTMLDetailsElement>
 		extends HTMLAttributes<T> {
 		open?: Signalish<boolean | undefined>;
-		onToggle?: GenericEventHandler<T> | undefined;
```
rschristian, line 1678: Luke saw this in #4694 -- essentially this type on a `GenericEventHandler` predates the better type we backported in #4615. Toggle events do trigger from popover targets now, so it's better as a global event type rather than a per-element type

src/jsx.d.ts
```diff
@@ -1216,257 +1215,60 @@ export namespace JSXInternal {
...
+			TdHTMLAttributes<RefType>,
+			TextareaHTMLAttributes<RefType>,
+			ThHTMLAttributes<RefType>,
+			TimeHTMLAttributes<RefType>,
+			TrackHTMLAttributes<RefType>,
+			VideoHTMLAttributes<RefType> {
+		/* Properties that are incompatible between elements */
+		target?: Signalish<any | undefined>;
+		type?: Signalish<any | undefined>;
+		loop?: Signalish<any | undefined>;
```
rschristian, line 1271:
```suggestion
		loop?: Signalish<any | undefined>;
		target?: Signalish<any | undefined>;
		type?: Signalish<any | undefined>;
```

JoviDeCroock (approved):

test/ts/preact.tsx
```diff
@@ -387,13 +399,15 @@ h('form', { onSubmit: onSubmit });
 
 // Should accept onToggle
 const onToggle = (e: h.JSX.TargetedToggleEvent<HTMLDetailsElement>) => {};
```
lukewarlow, line 401: Shouldn't this be HTMLDialogElement?

rschristian: Er yes, yes it should. I think that was a typo.

On the other hand, it does show that our `h` and `createElement` types aren't quite correct as they'll accept handlers with generics for other elements than their own. IIUC, the way to fix this is add a `h` & `createElement` [type definition for every single element](https://github.com/preactjs/preact/blob/daba00255753b6c7764ca2649dd183163458a5f5/src/index.d.ts#L190-L200).

PR #4711, opened 2025-02-26
Prevent memo'd components that unsuspend from bailing

### LINKED ISSUE #4631: useEffect didn't run in some rare case (lazy + memo)

**Describe the bug**

I have a simple app with the following structure:

- App (suspense)
  - Foobar (lazy)
    - Foo (memo)
    - Bar (lazy)

I found that in this case, Foo's useEffect() will NOT be executed. (But it does in React 18)

**To Reproduce**

https://codesandbox.io/p/sandbox/preact-memo-gpmrxk

**Expected behavior**

Show foobar

**Actual behavior**

Show bar

CaptainWang98:
Set `Memo` Component to when unsuspending `_force` so when the subsequent render arrives there is no possibility of bailing out of the render pass.

fixes #4631

### DIFF
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -57,6 +57,9 @@ function detachedClone(vnode, detachedParent, parentDom) {
 			if (vnode._component._parentDom === parentDom) {
 				vnode._component._parentDom = detachedParent;
 			}
+
+			vnode._component._force = true;
+
 			vnode._component = null;
 		}
 
--- a/compat/test/browser/suspense.test.js
+++ b/compat/test/browser/suspense.test.js
@@ -1,4 +1,4 @@
-import { setupRerender } from 'preact/test-utils';
+import { act, setupRerender } from 'preact/test-utils';
 import React, {
 	createElement,
 	render,
@@ -9,7 +9,8 @@ import React, {
 	createContext,
 	useState,
 	useEffect,
-	useLayoutEffect
+	useLayoutEffect,
+	memo
 } from 'preact/compat';
 import { setupScratch, teardown } from '../../../test/_util/helpers';
 import { createLazy, createSuspender } from './suspense-utils';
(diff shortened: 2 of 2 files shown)


compat/src/suspense.js
```diff
@@ -57,6 +57,10 @@ function detachedClone(vnode, detachedParent, parentDom) {
 			if (vnode._component._parentDom === parentDom) {
 				vnode._component._parentDom = detachedParent;
 			}
+
+			// issue #4631
+			if (vnode.type.name === 'Memoed') vnode._component._force = true;
```
JoviDeCroock, line 62:
```suggestion
			node._component._force = true;
```

JoviDeCroock (commented):
This does not fix https://github.com/preactjs/preact/issues/3449 yet, for that you would need a different mechanism in error-boundaries. That being said this also needs tests

CaptainWang98:
Thanks
I'll check #3449 later and create a new PR.

JoviDeCroock (changes requested):
There are no tests asserting that the behaviour is corrected, also see my comment, there is no need for the comment nor for making this exclusive to memoed as this would also trigger for normal sCU components

compat/test/browser/suspense.test.js
```diff
@@ -2159,4 +2160,38 @@ describe('suspense', () => {
 			expect(scratch.innerHTML).to.equal('<div><p>hello world</p></div>');
 		});
 	});
+
+	it('should correctly render memod child component', () => {
```
JoviDeCroock, line 2164: This test does not test the behaviour, it even passes on main. You might want to look at the reproduction to more closely reproduce the failure. Currently this component renders successfully and retains said state, what we want to assert is that it can attain new state

CaptainWang98:
@JoviDeCroock Hi there, could you plz check this pr? Thanks for your time~

JoviDeCroock (approved):

PR #4713, opened 2025-02-26
Add fetchPriority

### LINKED ISSUE #4712: Missing `fetchPriority` type for HTMLImageAttribute

**Describe the bug**
Typescript definition bug: Missing type
The `fetchPriority` attribute is not recognized on the image element. 

Reference to attribute on MDN
https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/fetchPriority

**To Reproduce**
The following snippet will not compile with typescript.
```jsx
<img fetchPriority="high" .../>
```

**Expected behavior**
`fetchPriority` and `fetchpriority` should be a part of the relevant type definition here:

https://github.com/preactjs/preact/blob/1effaa680969d77fe0134fb2f3476e6efacc701c/src/jsx.d.ts#L1779-L1796

React has the following entry:
```ts
fetchPriority?: "high" | "low" | "auto";
```

JoviDeCroock:
Resolves #4712

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -1782,6 +1782,7 @@ export namespace JSXInternal {
 		crossorigin?: Signalish<HTMLAttributeCrossOrigin>;
 		crossOrigin?: Signalish<HTMLAttributeCrossOrigin>;
 		decoding?: Signalish<'async' | 'auto' | 'sync' | undefined>;
+		fetchpriority?: Signalish<'high' | 'auto' | 'low' | undefined>;
 		fetchPriority?: Signalish<'high' | 'auto' | 'low' | undefined>;
 		height?: Signalish<number | string | undefined>;
 		loading?: Signalish<'eager' | 'lazy' | undefined>;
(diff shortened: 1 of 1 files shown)


src/jsx.d.ts
```diff
@@ -1782,6 +1782,7 @@ export namespace JSXInternal {
 		crossorigin?: Signalish<HTMLAttributeCrossOrigin>;
 		crossOrigin?: Signalish<HTMLAttributeCrossOrigin>;
 		decoding?: Signalish<'async' | 'auto' | 'sync' | undefined>;
+		fetchPriority?: Signalish<'high' | 'auto' | 'low' | undefined>;
```
rschristian, line 1785: Should probably also add the attribute casing while we're at it:

```suggestion
		fetchpriority?: Signalish<'high' | 'auto' | 'low' | undefined>;
		fetchPriority?: Signalish<'high' | 'auto' | 'low' | undefined>;
```

rschristian (approved):

PR #4715, opened 2025-02-27
types: Remove 'AllHTMLAttributes' from core

rschristian:
Targetting v11 as it's breaking.

I'd like to kill off this interface from core, only offering it as `any` through compat going forward. It's fundamentally a bad interface and no one should be using it:

1. Using it allows for nonsensical props to be set. This is most often utilized by component libs that do something like `props: AllHTMLAttributes<HTMLDivElement>`, I'm guessing expecting the `HTMLDivElement` generic to somehow narrow the type but it doesn't work like that. This allows the user to set (say) `referrerPolicy` despite that doing nothing on the target `<div>` element.

2. There's obvious & unavoidable conflicts in the interface members, such as `InputHTMLAttributes.type` vs `AnchorHTMLAttributes.type` -- which `type` do we expose here? Common denominator `string`? `any`?

3. It's a pain to maintain. Theoretically every single prop/attr needs to be added to that interface and we can't simply extend our per-element interfaces as they're not static (or won't be w/ #4607). Every type alteration would need to be added in two separate places, else, the types start getting a bit flakey for users.

We have per-element types now, and while they're probably not perfect, users should be reaching for them instead as they're much, much more useful.


JoviDeCroock (commented):
I'm atm very hesitant on this one due to people that might be relying on this

rschristian:
Upon more thought, let's not. The value is unclear & it's a pain to maintain correctly, but letting it sit isn't much of an issue. I'll revert part of #4706 instead.

PR #4717, opened 2025-02-27
Address context fragments issue

### LINKED ISSUE #4716: Fragment unmounts when it shouldn't

**Describe the bug**
After updating to 10.26.2, some Fragments are disappearing during a second render. I think this was caused by #4680.

**To Reproduce**

https://git.sr.ht/~vpzom/fragment-gone-issue

1. Open the page
2. Click either of the links
3. Both links disappear

**Expected behavior**
Nothing

JoviDeCroock:
Fixes #4716 

When we do strict equality bailouts we rely on `props.children` as we aren't invoking render, this rather than setting the top-level fragment children to null clones them.

I've confirmed that the memory leak of https://github.com/preactjs/preact/issues/4679 is still addressed

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -261,13 +261,7 @@ export function diff(
 			let renderResult = tmp;
 
 			if (isTopLevelFragment) {
-				renderResult = isArray(tmp.props.children)
-					? tmp.props.children.map(v =>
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
LGTM

PR #4720, opened 2025-03-06
refactor: Skip dom check for inferring lower-cased event names

rschristian:
To my knowledge, all event types are lower cased with the possible exception of user-created events, as they can be any arbitrary name. We could simplify our check here & reduce a few bytes with the assumption that any user-created event will match the platform in naming/casing which, to me, sounds pretty reasonable? Using any other form of naming would be quite odd & something we could reasonably discourage, IMO.

Breaking change though as `onMyClickEvent` will now listen for `myclickevent` instead of `MyClickEvent`.

~~Need to fiddle with the check yet, though I think that's the smallest & fastest form. Beats regex or charCodeAt for sure.~~ Think this is the best option, lmk if anyone has other ideas!


JoviDeCroock (approved):
This is clever!

PR #4723, opened 2025-03-08
refactor: Golf down unnecessary strict equality checks

rschristian:
This always feels hideously dirty but most of these cases are an obvious & safe byte improvement

~~Makes sense that there'd be some slowdown, will go through the cases to try to find the culprit(s)~~


marvinhagemeister (approved):

JoviDeCroock (approved):
Nice find!

PR #4726, opened 2025-03-10
Object.is fn for dependency comparison

darshan-rajadhyaksha:
To match the dependency comparison behaviour with React.

**Preact:** runs on every re-render
`useEffect(() => {
  console.log("deps changed");
 },  [NaN]);`
Example - https://codepen.io/dsr/pen/RNwZqKP
 
**React:** only once
`useEffect(() => {
  console.log("deps changed");
 },  [NaN]);`
 Example - https://codepen.io/dsr/pen/MWMWEbG
 
 Similar for: useLayoutEffect, useMemo.

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -537,7 +537,7 @@ function argsChanged(oldArgs, newArgs) {
 	return (
 		!oldArgs ||
 		oldArgs.length !== newArgs.length ||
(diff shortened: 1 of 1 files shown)


rschristian:
This is intentional: Preact X supports IE11 where this is not available.

Thanks for the PR though.

PR #4733, opened 2025-03-17
Avoid cloning reused nodes

JoviDeCroock:
Fixes https://github.com/preactjs/signals/issues/659

Avoid cloning nodes that we are re-using, the diffChildren algorithm will already ensure we clone these and we do it in a more correct way there https://github.com/preactjs/preact/blob/main/src/diff/children.js#L210. In signals we rely on this cloning with https://github.com/preactjs/signals/blob/main/packages/preact/src/index.ts#L164

I have yet to test this fix, but this looks like the right approach 😅 

EDIT: does seem to work https://stackblitz.com/edit/vitejs-vite-jdff8rn3?file=src%2Fsignals.js,src%2Fapp.tsx,src%2Fpreact.js,src%2Fhooks.js,src%2Fmain.tsx

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -172,7 +172,7 @@ export function diff(
 							c._nextState,
 							componentContext
 						) === false) ||
-						newVNode._original == oldVNode._original
+					newVNode._original == oldVNode._original
 				) {
 					// More info about this here: https://gist.github.com/JoviDeCroock/bec5f2ce93544d2e6070ef8e0036e4e8
 					if (newVNode._original != oldVNode._original) {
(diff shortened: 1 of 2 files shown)


src/diff/index.js
```diff
@@ -369,14 +369,22 @@ export function commitRoot(commitQueue, root, refQueue) {
...
 	}
 
 	if (isArray(node)) {
 		return node.map(cloneNode);
 	}
 
+	if (node._depth && node._depth > 0) {
+		return node;
+	}
+
```
rschristian, line 387: This should be removed or replace the conditional above, no?

rschristian (approved):

src/diff/index.js
JoviDeCroock, line 387:
```suggestion
```

PR #4734, opened 2025-03-20
10.26.5

JoviDeCroock:
## Types

- Return 'AllHTMLAttributes' to manually including all attrs (#4728, thanks @rschristian)

## Fixes

- Avoid cloning reused nodes (#4733, thanks @JoviDeCroock)
- Bring back strict-equality bailout for children even w/ context updates (#4724, thanks @rschristian)
- Effect in memoed boundary should be re-executed when the lazy boundary resolves (#4711, thanks @CaptainWang98)

## Maintenance

- Add case for context w/ strictly-equal children (#4725, thanks @rschristian)
- Golf down unnecessary strict equality checks (#4723, thanks @rschristian)

### DIFF
--- a/devtools/src/devtools.js
+++ b/devtools/src/devtools.js
@@ -13,7 +13,7 @@ export function initDevTools() {
 		globalVar !== undefined &&
 		globalVar.__PREACT_DEVTOOLS__
 	) {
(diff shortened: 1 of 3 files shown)


rschristian (approved):
Are you good w/ #4724? We didn't really get back to that Slack convo

marvinhagemeister (approved):

PR #4735, opened 2025-03-20
correct hydrateRoot return type

plesiecki:
Aligning type with implementation https://github.com/preactjs/preact/blob/29b0021af790ee6a112387298d84d92639981274/compat/client.js#L18-L21


rschristian (approved):
Thanks

PR #4738, opened 2025-03-26
Fix bailouts breaking select value setting

### LINKED ISSUE #4737: `select` value update fails when set to a newly added `option` simultaneously

**Describe the bug**

Imagine you have a `select` element and its `option` elements in different Preact components, both subscribed to the same signal (or reading from the same context).

```jsx
const Options = () =>
  items.value.map((item) => (
    <option key={item.id} value={item.id}>
      {item.id}
    </option>
  ));

const Select = ({ children }) => {
  const selectedItem = items.value.find((item) => item.selected).id;
  return <select value={selectedItem}>{children}</select>;
};

<Select>
  <Options />
</Select>
```

If you try to add a new option and immediately select it, Preact first updates the `value` attribute of the `select` element in the DOM before adding the new `option`, and therefore the change of the `value` attribute fails.

**To Reproduce**

- Using signals: https://stackblitz.com/edit/vitejs-vite-cd6mlnbv
- Using context: https://stackblitz.com/edit/vitejs-vite-4tf6mbza

Interestingly, the same code in React works fine: https://stackblitz.com/edit/vitejs-vite-m6klnuay

Steps to reproduce the behavior: just click on the button.

**Expected behavior**

After clicking the button, the select value should be the third, newly added option, but it's not. Actually, it is correct in the virtual DOM, but not in the DOM.

The same code works if you refactor it so that both elements are in a single element.

JoviDeCroock:
Resolves #4737 

The problem as seen in the test and reproduction is that we are doing this diff in two runs, the first runbeing the one where we change the select value, we descend one level due to a bail (memo) and we bubble back up to the `setProperty` call. This `setProperty` call relies on the `selected` option being mounted, which is not the case this will happen in the second run.

To counteract this edge case which is increasingly likely to happen with the bail we implement in signals we defer the property setting to happen post-diff.


src/diff/index.js
```diff
@@ -174,6 +174,7 @@ export function diff(
 						) === false) ||
 					newVNode._original == oldVNode._original
 				) {
+					console.log('bailout', newVNode.type.name);
```
JoviDeCroock, line 177:
```suggestion
```

marvinhagemeister (approved):
LGTM, seems like a reasonable fix, though the `defer` is unfortunate.

rschristian (approved):
Reasonable fix but pretty rough...

I'm almost wondering if we can suggest that be handled in user land instead. It's not a super niche scenario but it should be pretty easy to handle in app code

JoviDeCroock:
Trying to figure out what is causing this perf bump 😅 first assertion was that the function is now too big to inline but not sure

PR #4740, opened 2025-03-31
Update `CommandEventHandler` type to use `EventHandler`

robertknight:
Update the `CommandEventHandler` type declaration to match other `*EventHandler` types by using the `EventHandler` utility. This fixes a downstream typing error reported in
https://github.com/hypothesis/frontend-shared/pull/1899#issuecomment-2765368697.

Some unrelated formatting changes were applied when the commit was made. Is this expected?

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -51,7 +51,7 @@ interface CommandEvent extends Event {
 
 declare var CommandEvent: {
 	prototype: CommandEvent;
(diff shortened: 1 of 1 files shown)


rschristian (approved):
Good spot, I suppose #4692 mis-copied the existing types.

Looks like that PR (and a couple others perhaps) skipped the formatter on commit, maybe done by the web interface. Would have to get that correct at some point or another, may as well be now.

Thanks!

PR #4753, opened 2025-05-04
Bump vitest

JoviDeCroock:


vitest.config.mjs
```diff
@@ -265,7 +265,7 @@ export default defineConfig({
 		},
 		setupFiles: ['./vitest.setup.js'],
 		globals: true,
-		dangerouslyIgnoreUnhandledErrors: true,
+		// dangerouslyIgnoreUnhandledErrors: true,
```
JoviDeCroock, line 268: This was part of enabling `parallel` runs but we shouldn't enable this if we're not using it

vitest.config.mjs
```diff
@@ -265,7 +265,7 @@ export default defineConfig({
 		},
 		setupFiles: ['./vitest.setup.js'],
 		globals: true,
```
JoviDeCroock, line 267: One day I hope we can get rid of globals as well 😅

marvinhagemeister: +1

marvinhagemeister (approved):
LGTM

PR #4760, opened 2025-05-08
Implement deferring state updates

JoviDeCroock:
Copy of https://github.com/preactjs/preact/pull/4580

> [!NOTE]
> I realized that we can make `signals` work by having a generic sCU in there, this enables us to say that we need to run through the update because of the state hooks. One concern that I still have is where hooks and state are mixed in the same component, if the state hooks end up settling but the signal changed we risk skipping an update we need. We could fix that by doing an `options._afterRender` check in signals as well though.

This puts our implementation more in line with React which could resolve a few issues as well as reduce the byte size impact of our code. Another benefit this brings would be that we stop mutating the `Component` instance and hence can't mess with other implementations like signals.

What React does is when you call `reduce` is that it enqueues an update but does not execute the action yet. It first renders the component so the outer scope to the state setter can be recaptured. This way it ensures that it deals with up-to-date props/...

Essentially what this comes down to is that we store every reducer invocation and only execute it as the component is rendered. This way it has access to the outer component scope, .... This needs us to add a new `SKIP_CHILDREN` bit-flag on the `VNode` so we can bail out when the state settles to equal.

For this to properly work we'd need to introduce `options._afterRender` which would check all of the hooks and set `SKIP_CHILDREN` when all hooks are equal.

To get the last test to pass we'd need to find a way to deterministically see whether this component started the diff as we should not bail if the parent instructed the change.

I think this should generally be a performance improvement as well as we aren't touching the `Component` instance anymore on every pass.

There was a failing test for the NaN case, this is a top-down update which is considered a bail so it will render 25 times and bail afterwards.
For clarity, this loop is also present in React https://stackblitz.com/edit/vitejs-vite-mqaupwag?file=src%2Fmain.tsx,src%2FApp.tsx&terminal=dev


marvinhagemeister (approved):
LGTM

rschristian (approved):

JoviDeCroock:
Thinking about this, a case I should test for having a bug is where it's the initial render and we trigger a render that goes to a new state and then back to the initial state. We should not allow that to bail out of render 😅

PR #4766, opened 2025-05-16
fix(types): add `name` to `<details>`

marvinhagemeister:
We were missing types for the `name` attribute for `<detail>`-elements, see https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/details#name


rschristian (approved):
TIL about grouping `<details>`, nice!

PR #4768, opened 2025-05-24
Remove lifecycles from core

JoviDeCroock:
Checking out the byte size impact of removing these lifecycles, we can remove `componentWillUpdate` as well due to #4760. 53b is a lower impact than I had expected, what do we think about this one?

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -190,6 +190,10 @@ export function diff(
 					break outer;
 				}
 
+				if (c.componentWillUpdate != NULL) {
+					c.componentWillUpdate(newProps, c._nextState, componentContext);
+				}
+
 				if (isClassComponent && c.componentDidUpdate != NULL) {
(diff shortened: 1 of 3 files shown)


rschristian:
Super late here, but I think we should go for it or revert dropping support for the `UNSAFE_` prefix.

React 16.3 allowed & encouraged users to start using these prefixes and React 17.0 required used of them; as such, now that we've dropped support for the prefix in #4656, we only support 16.x components (and older) that never underwent the optional migration. This feels like a somewhat awkward middle ground for support and might be surprising to some users, if any are still using these methods, that is. If they aren't, well, `-53b` ain't bad.

JoviDeCroock:
@rschristian I support that, let's do it when we decide whether we revert deferred state settling though as the previous iteration does use componentWillUpdate 😅

rschristian:
Ah, that's why there's that revert commit, forgot we used to use that!

PR #2608, opened 2020-07-08
Fix accuracy of the non-dimensional property regex

kripod:
Resolves #2607

PR #4772, opened 2025-05-27
refactor (v11): Improve accuracy of non-dimensional regex

rschristian:
Copy of #2608

Uncommented the failing cases and added a few extra that were listed in #2607 but didn't make it into our test files. It's certainly a bit bigger and likely a bit slower but thought I'd at least put this up as a suggestion as I happened to remember it.


compat/src/util.js
```diff
@@ -13,4 +13,4 @@ export function shallowDiffers(a, b) {
 }
 
 export const IS_NON_DIMENSIONAL =
-	/acit|ex(?:s|g|n|p|$)|rph|grid|ows|mnc|ntw|ine[ch]|zoo|^ord|itera/i;
+	/^(-|f[lo].*[^se]$|g.{5,}[^ps]$|z|o[pr]|(W.{5})?[lL]i.*(t|mp)$|an|(bo|s).{4}Im|sca|m.{6}[ds]|ta|c.*[st]$|wido|ini)/;
```
JoviDeCroock, line 16: My main concern here is that this is a significant byte size increase for something that I'm not sure is worth it

PR #4780, opened 2025-05-29
Performance: compat `shallowDiffers`

romgrk:
This util is used to compare props. Props are more likely (I'm guessing) to have the same shape from one render to the next. Therefore, it's better to check first if the values differ than if the keys differ.

This benchmark compares 20-keys objects with the same shape for the same-values and different-values cases.

image

Benchmark: https://github.com/romgrk/js-benchmark/blob/master/benchmarks/object-shallow-compare.js


romgrk:
Ping, and you can close the PR if it's not going to be merged.

rschristian:
Thanks for the ping, sorry this PR slipped our notice.

I believe this is only used for `memo()` & `PureComponent`, both of which are pretty uncommon nowadays. Did you run into any detectable issue with one of them to end up looking at this? Just curious; I think it's a fine change & I agree changing values will occur more often.

Just needs a rebase, and also, `main` is targeting the upcoming v11, `v10.x` is the branch for the v10 release line. You can mirror this to v10 if you want or not

rschristian (approved):

romgrk:
I didn't notice any issue, I was just reading the codebase to understand the differences with React and noticed it. It's more of a lint. I'm happy with just targetting `main`.

PR #4787, opened 2025-06-05
fix(types): ensure `forwardRef` overrides the default ref type

aleclarson:
This fixes an issue with `forwardRef` that was leading to a type conflict, which was basically…

```ts
(Ref<HTMLDivElement> | undefined) …from forwardRef's return type
  & (Ref<any> | undefined) // …from the RenderableProps type
```

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -86,12 +86,12 @@ export type ComponentProps<
 		? JSXInternal.IntrinsicElements[C]
 		: {};
 
-export interface FunctionComponent<P = {}> {
-	(props: RenderableProps<P>, context?: any): ComponentChildren;
+export interface FunctionComponent<P = {}, RefType = any> {
+	(props: RenderableProps<P, RefType>, context?: any): ComponentChildren;
 	displayName?: string;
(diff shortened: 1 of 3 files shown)


src/index.d.ts
```diff
@@ -86,12 +86,12 @@ export type ComponentProps<
 		? JSXInternal.IntrinsicElements[C]
 		: {};
 
-export interface FunctionComponent<P = {}> {
-	(props: RenderableProps<P>, context?: any): ComponentChildren;
+export interface FunctionComponent<P = {}, RefType = any> {
+	(props: RenderableProps<P, RefType>, context?: any): ComponentChildren;
```
rschristian, line 90: I don't think this is the correct solution, [React's `FunctionComponent` type does not take a second generic param](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/9cf64a4f21790bc306c4937a41efc6658fb2ba19/types/react/index.d.ts#L1043) so we shouldn't either.

It looks like the return type of `forwardRef` should be altered instead.

Edit: We do also have TS tests, in both core & compat, adding this case to one or both of those tests would be appreciated.

PR #4791, opened 2025-06-07
chore: Bump terser

rschristian:
Terser was locked in #4687 as we saw significant perf degradation from bumping the version further. Upon further investigation, it turns out this was due to a lack of hoisting:

In our CJS & ESM builds, `NULL` & `UNDEFINED` are not hoisted as constants but inlined throughout the bundle. For reasons that I still don't quite understand, hoisting these two, as newer versions of Terser will do, is a huge perf issue for us, causing double-digit regressions across the board. The byte impact is completely overshadowed by the perf hit, so, I've removed the `NULL` & `UNDEFINED` constants. We'll go back to using `null` & `undefined`/`void 0` throughout the source.

Namespaces are a similar story; right now they're inlined, newer versions of Terser will hoist, but this is a very tiny byte reduction (we reference Math only once and XHTML & SVG twice each) with a 1-4% perf degradation in a few tests. Inlining them too seems better.

Edit: The perf regression is much larger w/ JIT enabled, disable it & hoisting is a slight improvement in some, still a regression in others. Not that it matters much ofc, just an interesting tidbit.

---

Therefore, the only real changes of this PR are as follows:

- Switches us to patterns that match our desired output/perf characteristics w/ modern Terser
- Matches the UMD build to our CJS & ESM builds
  - It's probably safe to assume it largely matches the performance characteristics of the CJS & ESM builds, and if so, it's likely slower than it should be due to the hoisting that only it is receiving at the moment.
- Compat's `version` is now hoisted too, instead of inlined into the CJS exports. Harder to fix (or at least, I'm not sure how to) as we don't easily control how the exports are generated.
  - i.e., `exports.version = '18.3.1'` has become `exports.version = P`

No worries if we want to just stick w/ older Terser, I mostly did this as an exploration as I was curious. There's no definite advantage here (I could go bench UMD builds I suppose though), just refactoring us so new Terser lines up with our old Terser output.


src/render.js
```diff
@@ -41,11 +41,12 @@ export function render(vnode, parentDom) {
...
 		oldVNode
-			? NULL
+			? null
 			: parentDom.firstChild
 				? slice.call(parentDom.childNodes)
-				: NULL,
+				: null,
 		commitQueue,
+		// @ts-expect-error TODO: Legitimate type mismatch that our use of `NULL` hid
 		oldVNode ? oldVNode._dom : parentDom.firstChild,
```
rschristian, line 50: This is a weird one. Above, we have this:

```js
let oldVNode = isHydrating ? NULL : parentDom._children;
```

For whatever reason, type inference has failed for `NULL`, so `oldVNode` was actually `any`. As such, this is a legitimate (internal) type mismatch in what `diff` expects as a parameter and what we provide. Will fix separately.

PR #4794, opened 2025-06-11
fix: escape style object value in precompile transform

marvinhagemeister:
We didn't escape the value when passing an `object` to `style` when using Deno's `precompile` transform. This does not affect other JSX transforms.


JoviDeCroock:
@marvinhagemeister does this need to be released as part of X, else you would need to PR to 10.x

I haven't properly fixed the actions yet to run there though, sorry about that.

JoviDeCroock (approved):

PR #4798, opened 2025-06-11
fix: signal attribute values not working with precompile transform

marvinhagemeister:


JoviDeCroock (approved):

jsx-runtime/src/index.js
```diff
@@ -145,7 +160,7 @@ function jsxAttr(name, value) {
 		return '';
 	} else if (value === true) return name;
 
-	return name + '="' + encodeEntities(value) + '"';
+	return name + '="' + encodeEntities('' + value) + '"';
```
marvinhagemeister, line 163: Added the type conversion because `encodeEntities` expects a string, but we could also pass a number here

Issue #4732, opened 2025-03-17
Bug: `<template>` tag content rendered twice


marvinhagemeister:
- [x] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
Template tag content is rendered twice. My guess is that the `oldDom` pointer is wrong somewhere when recursing into `<template>` tags.

**To Reproduce**

If possible, please provide a link to a StackBlitz/CodeSandbox/Codepen project or a GitHub repository that demonstrates the issue. You can use the following template on StackBlitz to get started: https://stackblitz.com/edit/create-preact-starter

```ts
it('should hydrate <template> tags ', () => {
  function App() {
    return (
        <div>
          <template>
            <h1>it works</h1>
          </template>
          <template>
            <h1>it works 2</h1>
          </template>
        </div>
      );
  }

  scratch.innerHTML = `<div><template><h1>it works</h1></template><template><h1>it works 2</h1></template></div>`

  render(<App />, scratch);

  expect(scratch.innerHTML).to.equal(`<div><template><h1>it works</h1></template><template><h1>it works 2</h1></template></div>`)
});
```

3. See error

**Expected behavior**
No duplicate content.

PR #4803, opened 2025-06-18
fix: reuse DOM nodes inside <template> during hydration

frontman-git:
fix [3732](https://github.com/preactjs/preact/issues/4732)

When hydrating a `<template>` element, Preact fails to reuse existing DOM nodes inside `template.content`.  
This is because `<template>` elements encapsulate their children within a `DocumentFragment`, but the `diff`/`hydrate` logic doesn't switch the context to `template.content`.

As a result, hydration skips the existing fragment content and instead appends new nodes — leading to duplicated content in the DOM.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
+> [!NOTE]
+> This is the branch for the upcoming release, for patches to v10 you need the [v10.x branch](https://github.com/preactjs/preact/tree/v10.x)
+
 <p align="center">
 <a href="https://preactjs.com" target="_blank">
 


test/browser/render.test.js
```diff
@@ -2013,4 +2013,11 @@ describe('render()', () => {
 		render(<App />, scratch);
 		expect(scratch.innerHTML).to.equal('hello world');
 	});
+	it('should hydrate <template> tags ', () => {
+		const App = () => <template><h1>it works</h1></template>		
+		scratch.innerHTML = `<template><h1>it works</h1></template>`
+
+		render(<App />, scratch);
+		expect(scratch.innerHTML).to.equal(`<template><h1>it works</h1></template>`)
```
rschristian: Not sure if you intentionally skipped the precommit hook, but this would pop up the next time anyone else tried to edit the file w/ hooks enabled.

```suggestion
		const App = () => <template><h1>it works</h1></template>;
		scratch.innerHTML = `<template><h1>it works</h1></template>`;

		render(<App />, scratch);
		expect(scratch.innerHTML).to.equal(`<template><h1>it works</h1></template>`);
```

src/diff/children.js
```diff
@@ -53,6 +53,13 @@ export function diffChildren(
 	isHydrating,
 	refQueue
 ) {
+	if (
+		parentDom instanceof DocumentFragment &&
+		newParentVNode.type === 'template' &&
+		excessDomChildren?.length === 0
+	) {
+		excessDomChildren = Array.from(parentDom.childNodes);
+	}
```
rschristian: We'd want to move the cheaper tag type check first, `instanceof` is considerably more expensive, especially as a first check in a chain when most will bail (`<template>` is pretty uncommonly used).

```suggestion
	if (
		newParentVNode.type === 'template' &&
		excessDomChildren?.length === 0 &&
		parentDom instanceof DocumentFragment
	) {
		excessDomChildren = slice.call(parentDom.childNodes);
	}
```

Will also need to add an import for `slice`.

test/browser/render.test.js
frontman-git: Thanks for pointing that out! I didn’t intend to skip the pre-commit hook — it looks like I accidentally committed with some local changes that I forgot to revert. I’ll commit suggestion.

src/diff/children.js
frontman-git: Thanks for the suggestion! I overlooked the performance cost of instanceof and agree it’s better to reorder the checks. I’ll update the condition accordingly.

test/browser/render.test.js
rschristian: No worries! I do that myself _a lot_

rschristian:
**Size Change:** +120 B (+0.26%) 

**Total Size:** 46.8 kB

| Filename | Size | Change |
| :--- | :---: | :---: |
| `dist/preact.js` | 4.68 kB | +37 B (+0.8%) |
| `dist/preact.mjs` | 4.71 kB | +43 B (+0.92%) |
| `dist/preact.umd.js` | 4.76 kB | +40 B (+0.85%) |

<details><summary>ℹ️ <strong>View Unchanged</strong></summary>

| Filename | Size |
| :--- | :---: |
| `compat/dist/compat.js` | 3.78 kB |
| `compat/dist/compat.mjs` | 3.69 kB |
| `compat/dist/compat.umd.js` | 3.83 kB |
| `debug/dist/debug.js` | 3.85 kB |
| `debug/dist/debug.mjs` | 3.85 kB |
| `debug/dist/debug.umd.js` | 3.94 kB |
| `devtools/dist/devtools.js` | 260 B |
| `devtools/dist/devtools.mjs` | 271 B |
| `devtools/dist/devtools.umd.js` | 346 B |
| `hooks/dist/hooks.js` | 1.52 kB |
| `hooks/dist/hooks.mjs` | 1.54 kB |
| `hooks/dist/hooks.umd.js` | 1.58 kB |
| `jsx-runtime/dist/jsxRuntime.js` | 892 B |
| `jsx-runtime/dist/jsxRuntime.mjs` | 861 B |
| `jsx-runtime/dist/jsxRuntime.umd.js` | 966 B |
| `test-utils/dist/testUtils.js` | 473 B |
| `test-utils/dist/testUtils.mjs` | 473 B |
| `test-utils/dist/testUtils.umd.js` | 555 B |

</details>



frontman-git:
Thanks for the review — commits have been squashed.

src/diff/children.js
```diff
@@ -53,6 +54,13 @@ export function diffChildren(
 	isHydrating,
 	refQueue
 ) {
+	if (
+		newParentVNode.type === 'template' &&
+		excessDomChildren?.length === 0 &&
+		parentDom instanceof DocumentFragment
+	) {
```
JoviDeCroock, line 61: Let's make this fast by adding the `isHydrating` check, we don't need to check `excessDomChildren` length as we expect this to be 0 as per your assumptions in this PR. I would assume that we are in a DocumentFragment when we are dealing with a template.

```suggestion
	if (isHydrating && newParentVNode.type === 'template') {
```

I assume you could save even more bytes by moving this to https://github.com/preactjs/preact/blob/1dae4898efd3c4a4e596917ee9426ed617efe50b/src/diff/index.js#L596-L611 and changing this condition to use `newVNode.type === 'template'` as well as hoisting the duplicated check

PR #4806, opened 2025-06-24
chore: migrate from mocha to vitest

43081j:
Removes mocha and runs all tests in vitest.

Notable changes:

- TypeScript tests are still run via `tsc`, but the associated test
  suites are now run in vitest _node mode_
- Shared tests are now run in both _browser mode_ and _node mode_
- Node tests are now run in _node mode_

The type tests have also been renamed to `*.test-d.tsx`, to make them
easier to exclude and in case we ever switch to vitest's type testing
feature.


marvinhagemeister (approved):
Sweet!

PR #4807, opened 2025-06-24
chore: use v8 coverage

43081j:
Switches to using vitest's v8 coverage provider which should be faster
and more accurate.

Depends on #4806 and #4808

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "@types/node": "^18.19.87",
         "@types/sinon": "^17.0.3",
         "@vitest/browser": "^3.2.1",
-        "@vitest/coverage-istanbul": "^3.2.1",
+        "@vitest/coverage-v8": "^3.2.4",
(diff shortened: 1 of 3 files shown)


JoviDeCroock (approved):

JoviDeCroock:
Thinking more about this, how does this work with coveralls? https://github.com/preactjs/preact/blob/7fbc10d1e8718dc83db1cb8c5bb180bcc91b4483/.github/workflows/build-test.yml#L39

I can see

```
🔍 Detected coverage file: coverage/lcov.info
🚨 Nothing to report
```

43081j:
That's the same output before this change. If you look at another branch that doesn't include this, it is the same

I think it just means there's no coverage errors, i.e thresholds were met

When I'm at a laptop again I can have a look at the report it outputs to be sure it is right still

43081j:
@JoviDeCroock you were right, it is fixed now 👍

PR #4808, opened 2025-06-24
chore: switch to playwright for browser tests

43081j:
Switches to playwright when running browser tests in vitest.

This will allow us to use v8 coverage in a follow-up change.

Depends on #4806

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -29,6 +29,8 @@ jobs:
           cache: 'npm'
           cache-dependency-path: '**/package-lock.json'
       - run: npm ci
+      - name: Install Playwright
+        run: npx playwright install
       - name: test
         env:
           CI: true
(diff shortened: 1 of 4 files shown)


.github/workflows/build-test.yml
```diff
@@ -29,6 +29,8 @@ jobs:
           cache: 'npm'
           cache-dependency-path: '**/package-lock.json'
       - run: npm ci
+      - name: Install Playwright
```
JoviDeCroock: I don't disagree with doing this, however it does slow down the CI with this installation step, I kind of dislike that part. The tests themselves do look to be running faster in playwright so it might be a win in the end

43081j: we do also have the option of only installing chromium since we don't currently use any others

`npx playwright install chromium`

43081j:
@JoviDeCroock i've switched this to only install chromium now and caught up from main 👍

.github/workflows/build-test.yml
```diff
@@ -29,6 +29,8 @@ jobs:
...
+        run: npx playwright install chromium
```
rschristian: This should probably be moved to our `prepare` script, all contributors will need this ran too.

PR #4814, opened 2025-06-25
perf: reduce some repeated logic

43081j:
this reduces the bundle size a little (few hundred bytes i think)

do we have a bot or a tool to diff the sizes?

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -392,14 +392,6 @@ export function toChildArray(children, out) {
 }
 
 /**
- * @param {VNode} node
- * @returns {boolean}
- */
-function nodeMatches(node) {
-	return node != NULL && (node._flags & MATCHED) == 0;
-}
-
-/**
  * @param {VNode} childVNode
  * @param {VNode[]} oldChildren
  * @param {number} skewedIndex
@@ -415,7 +407,7 @@ function findMatchingIndex(
 	const key = childVNode.key;
 	const type = childVNode.type;
 	let oldVNode = oldChildren[skewedIndex];
-	const matched = nodeMatches(oldVNode);
+	const matched = oldVNode != NULL && (oldVNode._flags & MATCHED) == 0;
 
 	// We only need to perform a search if there are more children
 	// (remainingOldChildren) to search. However, if the oldVNode we just looked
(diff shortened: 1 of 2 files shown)


JoviDeCroock:
This increases byte size as you can see in https://github.com/preactjs/preact/actions/runs/15876304094/job/44764641936?pr=4814 the bot sadly can't post messages for forks 😅 The action can be clicked and inspected though (compressed size)

43081j:
I'll look into it as it's smaller in the bundle locally 🤔 so must be minification differences somehow

43081j:
its actually the cloneElement change i think, because it was literally the same duplicated code before, it'd compress better!

but the actual JS build output _is_ smaller 😬

src/diff/children.js
```diff
@@ -391,6 +391,14 @@ export function toChildArray(children, out) {
 	return out;
 }
 
+/**
+ * @param {VNode} node
+ * @returns {boolean}
+ */
+function nodeMatches(node) {
```
JoviDeCroock: Adding additional functions like this is most likely either going ot be inlined or when not inlined make the bundle bigger

43081j: it isn't this causing the size difference mostly, its the cloneNode stuff

this increases performance, and the reduced repetition further down decreases the bundle size. but the compressed size overall is more it seems

JoviDeCroock: When you run `npm run build:core` you'll get compressed size that is identical to the GitHub action unless some of your deps are different from the lockfile

43081j: i put it back as the algorithm refactor is the important bit, not this function

the bundle size increase is larger when compressed because of the cloneElement stuff it seems

43081j:
@JoviDeCroock what i mean is the files on disk are smaller, and the logic should be faster (but micro optimisation i guess)

the _compressed_ files are larger. presumably because `cloneElement` was a duplicate of `createElement` before. so it meant it compressed well but was bad for maintenance

ill see if there's any way i can reduce it

JoviDeCroock:
I reckon that reverting the `cloneElement` part is what fits our philosophy most, we want to reduce size of these utils as much as possible. We don't really care about the disk size, we are optimising for compressed size. A util that isn't used often should be as small as possible, a micro-optimisation in something like child reconciliation is much preferred

43081j:
| Filename | Size | Change |
| :--- | :---: | :---: |
| `dist/preact.js` | 4.63 kB | -13 B (-0.28%) |
| `dist/preact.mjs` | 4.64 kB | -19 B (-0.41%) |
| `dist/preact.umd.js` | 4.71 kB | -15 B (-0.32%) |

JoviDeCroock (approved):
This is really clever, it basically flattens the loop

PR #4820, opened 2025-06-27
chore: remove sinon

43081j:
Now that we've switched entirely to vitest's spies, we can drop the
following:

- `sinon` (since we use `vi.fn` and `vi.spyOn`)
- `chai` (since vitest pulls chai in itself)
- `sinon-chai` (since we use vitest's own assertions now)
- associated types packages


JoviDeCroock (approved):
AWESOME!

PR #4821, opened 2025-06-28
perf: reduce some repeated logic (#4814)

JoviDeCroock:
* perf: use createElement when cloning

To avoid duplicating logic, we can use `createElement` with pre-populated props.

* chore: reduce search logic

* perf: reuse matched flag

* chore: put the function inline again

* perf: simplify to a ternary

* perf: revert clone function

---------


marvinhagemeister (approved):
LGTM

PR #4822, opened 2025-06-28
Fix empty children

JoviDeCroock:
In https://github.com/preactjs/prefresh/issues/568 the point was rightfully made that we lose our assurance that `_children` is always an array. When a component throws on first-render we risk setting an `undefined` value and then re-using it for recovered renders.


JoviDeCroock:
I am well and thoroughly confused by our tests erroring out right now :joy:

EDIT: ah a rebase was all it took

src/diff/index.js
```diff
@@ -312,7 +312,7 @@ export function diff(
 				}
 			} else {
 				newVNode._dom = oldVNode._dom;
-				newVNode._children = oldVNode._children;
+				newVNode._children = oldVNode._children || [];
```
rschristian, line 315:
```suggestion
				newVNode._children = oldVNode._children || EMPTY_ARR;
```

Will need to add it as an import tho, wish GH allowed users to suggest outside the diff.

Weird that the CI is failing. Edit: Ah, looks like we'll need to manually upload something to get the 10.x line working again.

rschristian (approved):

src/diff/index.js
marvinhagemeister: Would add a small comment here too on why we set it to a default value (e.g. point to prefresh issue)

PR #4830, opened 2025-07-07
fix: export helper functions from `preact/debug`

colinaaa:
Lynx utilizes Preact as its rendering library, and we're aiming to enhance diagnostic messages by incorporating the component stack.

It would be highly beneficial if `preact/debug` could provide exports for these helper functions.

We are now forking `component-stack.js` in our codebase. See: https://github.com/lynx-family/lynx-stack/pull/1238

### DIFF
--- a/debug/src/index.js
+++ b/debug/src/index.js
@@ -4,3 +4,9 @@ import 'preact/devtools';
 initDebug();
 
 export { resetPropWarnings } from './check-props';
+
+export {
+	getCurrentVNode,
+	getDisplayName,
+	getOwnerStack
+} from './component-stack';


debug/src/index.js
```diff
@@ -4,3 +4,9 @@ import 'preact/devtools';
 initDebug();
 
 export { resetPropWarnings } from './check-props';
+
+export {
+	getCurrentVNode,
+	getDisplayName,
+	getOwnerStack
+} from './component-stack';
```
rschristian, line 12: Mind updating the [types](https://github.com/preactjs/preact/blob/main/debug/src/index.d.ts) to match?

colinaaa: Thank you! Types has been added!

JoviDeCroock (approved):

PR #4842, opened 2025-07-15
test: Add simple test for `unstable_batchedUpdate` to improve coverage

rschristian:


jsconfig.json
```diff
@@ -14,7 +14,9 @@
     },
     "target": "es5",
     "noEmit": true,
-    "skipLibCheck": false
+    "skipLibCheck": false,
+    "typeRoots": ["./node_modules"],
+    "types": ["vitest/globals"]
```
rschristian, line 19: Similar was done in #4806 for `test/ts`, but that only applied to that specific directory. All of our other test files would result in type errors as `describe` & `expect` weren't pulled in anywhere.

PR #4843, opened 2025-07-15
refactor: DOM event name casing

rschristian:
Noticed we were calling `.toLowerCase()` and then `.slice(2)`, which, conceptually & (somewhat) measurably, should be less than optimal. Then may as well call `.slice()` early, doesn't hinder our casing check & reduces some bytes.

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -67,8 +67,8 @@ export function setProperty(dom, name, value, oldValue, namespace) {
 		useCapture = name != (name = name.replace(CAPTURE_REGEX, '$1'));
 
 		// Infer correct casing for DOM built-in events:
-		if (name[2].toLowerCase() != name[2]) name = name.slice(2).toLowerCase();
-		else name = name.slice(2);
+		name = name.slice(2);
+		if (name[0].toLowerCase() != name[0]) name = name.toLowerCase();
 
 		if (!dom._listeners) dom._listeners = {};
 		dom._listeners[name + useCapture] = value;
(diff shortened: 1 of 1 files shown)


src/diff/props.js
```diff
@@ -67,8 +67,8 @@ export function setProperty(dom, name, value, oldValue, namespace) {
 		useCapture = name != (name = name.replace(CAPTURE_REGEX, '$1'));
 
 		// Infer correct casing for DOM built-in events:
-		if (name[2].toLowerCase() != name[2]) name = name.toLowerCase().slice(2);
-		else name = name.slice(2);
+		name = name.slice(2);
+		if (name[0].toLowerCase() != name[0]) name = name.toLowerCase();
```
rschristian, line 71: My mind is telling me there's a better way to write this but I'm running on fumes here and can't think of it 😅

Lmk if there is indeed!

JoviDeCroock (approved):

src/diff/props.js
JoviDeCroock: Maybe

```
name = name[0] === name[0].toLowerCase() ? name : name.toLowerCase();
```

I guess yours is more performant though

rschristian: That'd be a couple bytes bigger which makes sense as the first branch of that ternary (`? name`) is redundant.

Edit: Terser already turns this into the `name = (name = name.slice(2) ...` form anyhow, should've checked & confirmed that first. Oh well, good enough I guess.

PR #4844, opened 2025-07-16
chore: Update 'replaceNode' deprecation comment to point at new shim

rschristian:
Links to the new shim package rather than the gist.

A [comment on the gist](https://gist.github.com/developit/f4c67a2ede71dc2fab7f357f39cff28c?permalink_comment_id=4729975#gistcomment-4729975) stated they were seeing the comment on the deprecated overload even when not using the third parameter, and whilst I can't reproduce (it only pops up when I actually use the third param in my editor), I figured expanding the comment to explicitly mentioning `replaceNode` would be a good idea.

### DIFF
--- a/src/index-5.d.ts
+++ b/src/index-5.d.ts
@@ -301,9 +301,10 @@ interface ContainerNode {
 
 export function render(vnode: ComponentChild, parent: ContainerNode): void;
 /**
- * @deprecated Will be removed in v11.
+ * @deprecated The `replaceNode` parameter will be removed in v11.
  *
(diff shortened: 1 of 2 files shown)


rschristian:
Hrm, not related but it seems the PR reporter is now comparing the branch against `main`, rather than the target. Will look into it

Edit: Oh, it might be that the PR reporter action is out of date? Still weird though as the PR `base` is set to `main`, which doesn't make sense to me. ~~Will make a separate PR to at least align `v10.x` w/ `main` there~~

JoviDeCroock (approved):

PR #4849, opened 2025-07-24
Add parameter `oldVNode` when calling `options._diff`

hzy:
`oldVnode` is not exposed in `options._diff`, which makes it difficult to get the component instance in this hook (because `newVNode` usually does not have a `_component` property). This PR provides a new way to get the component instance in the update process, which can still work when, for example, parent and child components `setState` at the same time.


JoviDeCroock:
Closing: niche draft adding `oldVNode` to `options._diff`; conflicting, no reviews, low demand.

Issue #3668, opened 2022-08-13
Posible memory leak  


Z2Y:
**Describe the bug**
When i use preact with a virtualized list, I found memroy not gabage collected. this may cased by a chrome issue:
The MouseEventManager keeps removed DOM nodes alive until subsequent click   https://bugs.chromium.org/p/chromium/issues/detail?id=1177010
if there is no click event for a while , (such as scrolling infinite card list ), the memory usage grows  until next click.

marvinhagemeister:
Please share more information about your setup. We cannot see your screen and don't know what your code looks like, so it's pretty much impossible to help with the issue you're seeing. Specifically the virtual list implementation that's used in your project and how you're using it might be of interest.

Z2Y:
<img width="772" alt="1660469565(1)" src="https://user-images.githubusercontent.com/68933742/184531182-9afd68db-075f-47ba-9ba0-812a64829c24.png">

the event handler hold the ref of a unmountd vnode, ant the vnode hold the parent vnode. if the detached dom not gabage collected , it will cause memory leak.

JoviDeCroock:
From the linked issue this is a chrome bug and not a Preact one, there's not much we can do about the implementation of a browser event handler as this would occur in any library. The screenshot attached also does not seem to relate to the bug as this is a VDOM structure.

The event-handler in question is a native one as we aren't creating our own events so the only things interacting here should be your callback and the browser event-handler.

Z2Y:
I managed to create a demo to reproduce it  https://codesandbox.io/s/preact-vlist-demo-ehphz9
steps: Add some items and then scroll up and down, As the chrome devetool show , the number of detached HtmlButtonElement grows up.
<img width="1380" alt="WeChat0a55e3857971618997c10743f53c67a5" src="https://user-images.githubusercontent.com/68933742/184571304-f1fcff54-6331-4456-9dcf-92bfb9bdea21.png">

developit:
@Z2Y this happens because `SomeButtonItem` closes over `props.children`, which retains it. The issue is already fixed in Preact 11, where `props.children` is a pure JSX element with no tree structure:
https://codesandbox.io/s/preact-vlist-demo-forked-o26dg7?file=/src/index.js

In the meantime, here's a workaround that adds some additional unmount cleanup that likely addresses the issue:

```js
import { options } from 'preact';

let oldUnmount = options.unmount;
options.unmount = (vnode) => {
[...]

PR #4853, opened 2025-07-26
Provide solution to long standing memleak

JoviDeCroock:
We can't safely remove `_listeners` as that would mean we are risking suspense-hydration breaking 😅 basically during suspense we store the vnodes which means that old and new vnode will always be equal, if we want to remove `_listeners` as well we'll have to probably stop storing our virtual DOM tree when it's a mount and store the previous variant when it's an update. This would however mean that our mutative approach is a bad thought.

EDIT: I solved the `_listeners` issue

Resolves https://github.com/preactjs/preact/issues/3668

### DIFF
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -1,5 +1,5 @@
 import { Component, createElement, options, Fragment } from 'preact';
-import { MODE_HYDRATE, COMPONENT_FORCE } from '../../src/constants';
+import { MODE_HYDRATE, FORCE_PROPS_REVALIDATE, COMPONENT_FORCE } from '../../src/constants';
 import { assign } from './util';
 
 const oldCatchError = options._catchError;
@@ -67,6 +67,10 @@ function detachedClone(vnode, detachedParent, parentDom) {
 
 function removeOriginal(vnode, detachedParent, originalParent) {
 	if (vnode && originalParent) {
+		if (typeof vnode.type == 'string') {
+			vnode._flags |= FORCE_PROPS_REVALIDATE;
+		}
+
 		vnode._original = null;
 		vnode._children =
 			vnode._children &&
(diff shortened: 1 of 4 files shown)


DracoRunner:
Facing the same issue 

https://github.com/preactjs/preact-iso/issues/104

compat/src/suspense.js
```diff
@@ -67,6 +67,10 @@ function detachedClone(vnode, detachedParent, parentDom) {
 
 function removeOriginal(vnode, detachedParent, originalParent) {
 	if (vnode && originalParent) {
+		if (typeof vnode.type == 'string') {
```
JoviDeCroock, line 74: For DOM nodes we need to revalidate the props so that we re-attach event-listeners which are deleted by `unmount` as we do `dom._listeners = undefined`

compat/src/suspense.js
```diff
@@ -135,11 +139,21 @@ Suspense.prototype._childDidSuspend = function (promise, suspendingVNode) {
 			// suspended children into the _children array
 			if (c.state._suspended) {
 				const suspendedVNode = c.state._suspended;
-				c._vnode._children[0] = removeOriginal(
-					suspendedVNode,
-					suspendedVNode._component._parentDom,
-					suspendedVNode._component._originalParentDom
-				);
+				if (!c._vnode._children) {
```
JoviDeCroock: It's now possible in non hydration suspense that we've cleared _children

src/diff/index.js
```diff
@@ -719,11 +721,13 @@ export function unmount(vnode, parentVNode, skipRemove) {
...
-	if (!skipRemove) {
-		removeNode(vnode._dom);
+	if (!skipRemove) removeNode(vnode._dom);
+
+	if (!vnode._component && vnode._dom && vnode._dom._listeners) {
+		vnode._dom._listeners = NULL;
 	}
 
-	vnode._component = vnode._parent = vnode._dom = UNDEFINED;
+	vnode._dom = vnode._component = vnode._parent = NULL;
```
JoviDeCroock, line 732: I stopped removing the reference to children because it has perf issues in the micro benchmarks, going to verify this further

lucas-bortoli:
Great job Jovi, thank you

PR #4857, opened 2025-07-28
Ensure we properly re-render bailing errored children

JoviDeCroock:
Frontport of #4856 (description is there)

Found the issue in https://github.com/preactjs/preact/pull/4854

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -352,10 +352,12 @@ export function diff(
 					for (let i = excessDomChildren.length; i--; ) {
 						removeNode(excessDomChildren[i]);
 					}
+					markAsForce(newVNode);
 				}
 			} else {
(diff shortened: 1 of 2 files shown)


src/diff/index.js
```diff
@@ -352,10 +352,12 @@ export function diff(
 					for (let i = excessDomChildren.length; i--; ) {
 						removeNode(excessDomChildren[i]);
 					}
+					markAsForce(newVNode);
 				}
 			} else {
 				newVNode._dom = oldVNode._dom;
 				newVNode._children = oldVNode._children;
+				if (e.then) markAsForce(newVNode);
```
```suggestion
				if (!e.then) markAsForce(newVNode);
```

rschristian (approved):

PR #4858, opened 2025-07-28
Bump testing/transpilation/formatting deps

JoviDeCroock:
This bumps vitest, babel, ...

We also have a new major of biome which slightly changes formatting as well as shows that we have a bug in the node packages package.json as we enforce commonjs while it needs  to be module as we use .js files for export/import.


package.json
```diff
@@ -203,36 +203,36 @@
...
     "oxlint": "^1.8.0",
-    "playwright": "^1.53.1",
+    "playwright": "^1.54.1",
     "preact-render-to-string": "^6.5.0",
     "prop-types": "^15.8.1",
     "sade": "^1.8.1",
     "terser": "5.16.0",
     "typescript": "5.1.6",
     "undici": "^4.12.0",
     "vite": "^6.2.0",
```
JoviDeCroock, line 234: Vite has a 7.0.0 but it looks like it currently crashes due to `crypto.hash` not being a function locally. Volta does not seem to apply the node v20.19 for me 😅 going to leave the bump in the middle for now

test/shared/package.json
```diff
@@ -1,3 +1,3 @@
 {
-  "type": "commonjs"
+  "type": "module"
 }
```
rschristian: This file might be redudant now, I think I added it for mocha w/ Node's require(esm)? Might be worth a check, anyhow

Edit: Same with the other one(s)

JoviDeCroock: Yes, as we use vitest now they are not needed

rschristian: Awesome! Fewer files to support tooling oddities 🎉

PR #4863, opened 2025-07-30
fix(types) make React.SVGAttributes generic

shyguy1412:
React.SVGAttributes should be generic.

<https://github.com/DefinitelyTyped/DefinitelyTyped/blob/7662e054f506bf695a4e352017d4dafe6926791d/types/react/index.d.ts#L3492>

React itself seems to demand the generic type argument but since I wasn't sure about the implications of adding mandatory generic type arg I made it default to the same type its was before.

### DIFF
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -87,7 +87,8 @@ declare namespace React {
 		extends JSXInternal.SVGAttributes<T>,
 			preact.ClassAttributes<T> {}
 
-	interface SVGAttributes extends JSXInternal.SVGAttributes {}
+	interface SVGAttributes<T extends EventTarget = SVGElement>
+		extends JSXInternal.SVGAttributes<T> {}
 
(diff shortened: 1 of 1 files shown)


rschristian (approved):
Thanks! I'll add it to my to-dos to check whether we can make the definition from core generic too.

Just FYI, `main` is for the upcoming v11 release, you may want to instead target `v10.x`?

shyguy1412:
it seems i'll need to do some git magic in order to retarget cleanly but yeah, targeting v10.x sounds like a good idea

shyguy1412:
that was quite painful but should be good now

rschristian:
Sorry about that, not quite sure what you did, but for the future, using something like this makes it pretty quick & painless:

```sh
git rebase -i @~1 --onto v10.x
```

The `1` is however many commits you want to rebase, the `v10.x` the branch. It is destructive so be a bit careful, but for situations like this, really convenient.

PR #4867, opened 2025-07-31
Add flags for component type

JoviDeCroock:
Kind of want to add more things onto flags while maintaining backwards compatability, type will always be present so this one does not break anything special. We could consider `_dirty` and others as well.

Follow ups

- https://github.com/preactjs/preact-render-to-string/blob/d424e9c46b8df425e6c988936b3be0ba316925d4/src/index.js#L178
- https://github.com/preactjs/prefresh/blob/main/packages/core/src/runtime/catchError.js#L4
- https://github.com/preactjs/preact-ssr-prepass/blob/2f63c40fb8c152f29df5ae255fc0e9d4cb05501a/src/index.js#L59

### DIFF
--- a/src/internal.d.ts
+++ b/src/internal.d.ts
@@ -166,7 +166,6 @@ export interface Component<P = {}, S = {}>
 	state: S; // Override Component["state"] to not be readonly for internal use, specifically Hooks
 
 	_excess?: PreactElement[];
-	_force?: boolean;
 	_renderCallbacks: Array<() => void>; // Only class components
 	_stateCallbacks: Array<() => void>; // Only class components
 	_globalContext?: any;
--- a/src/internal.d.ts
+++ b/src/internal.d.ts
@@ -166,7 +166,6 @@ export interface Component<P = {}, S = {}>
 	state: S; // Override Component["state"] to not be readonly for internal use, specifically Hooks
 
 	_excess?: PreactElement[];
-	_dirty: boolean;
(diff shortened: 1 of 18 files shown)


src/internal.d.ts
```diff
@@ -165,9 +165,8 @@ export interface Component<P = {}, S = {}>
 	constructor: ComponentType<P>;
 	state: S; // Override Component["state"] to not be readonly for internal use, specifically Hooks
 
+	_flags: number;
```
JoviDeCroock: CC @marvinhagemeister it looks like _prevState from devtools uses __u as well can we swap that?

EDIT: I cut out component flags for now as they behave weirdly, would have to dig deeper

marvinhagemeister (approved):
LGTM

src/internal.d.ts
```diff
@@ -180,10 +179,7 @@ export interface Component<P = {}, S = {}>
 	 * components or array returns.
 	 */
 	_parentDom?: PreactElement | null;
-	// Always read, set only when handling error
-	_processingException?: Component<any, any> | null;
-	// Always read, set only when handling error. This is used to indicate at diffTime to set _processingException
-	_pendingError?: Component<any, any> | null;
+	_bits: number;
```
JoviDeCroock, line 181: Named it bits because the _flags mangle would conflict

rschristian (approved):

PR #4870, opened 2025-07-31
fix(compat): add .render property referencing original render function

upupming:
Follows #4793, port the `.render` of forwarded component to preact v10.x


marvinhagemeister (approved):
Thanks!

rschristian:
Just wanted to say good job on catching this! We merged our v11 mega branch/PR into `main` around the time of #4793 and must've missed backporting it ourselves, probably wouldn't have noticed now that it's a month+ old.

Thanks again!

PR #4875, opened 2025-08-05
feat(debug): add `captureOwnerStack`

colinaaa:
Export a new `captureOwnerStack` API from `preact/debug`.

> [`captureOwnerStack`](https://react.dev/reference/react/captureOwnerStack) is the new API added in React [v19.1.0](https://github.com/facebook/react/releases/tag/v19.1.0).

### DIFF
--- a/debug/src/index.js
+++ b/debug/src/index.js
@@ -5,8 +5,4 @@ initDebug();
 
 export { resetPropWarnings } from './check-props';
 
-export {
-	getCurrentVNode,
-	getDisplayName,
-	getOwnerStack
-} from './component-stack';
+export { captureOwnerStack } from './component-stack';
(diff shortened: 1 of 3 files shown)


debug/src/index.js
```diff
@@ -5,8 +5,4 @@ initDebug();
 
 export { resetPropWarnings } from './check-props';
 
-export {
-	getCurrentVNode,
-	getDisplayName,
-	getOwnerStack
-} from './component-stack';
+export { captureOwnerStack } from './component-stack';
```
rschristian: Unfortunately we did backport this in #4860, and I don't see anything that demands removal, so can we make this additive instead? If those APIs don't get used much then that's fine, `/debug` is pretty isolated & non-probelmatic.

colinaaa: The removed APIs have been added back. Sorry for missing the backport.

rschristian: No problem, we backported in the off chance it might've been useful to y'all.

Thanks!

PR #4878, opened 2025-08-05
refactor: Move most interfaces out of the JSX namespace

### LINKED ISSUE #2706: Remove irrelevant types from JSX namespace

rschristian:
Closes #2706

This PR cuts down the `JSX` namespace to only include the actual JSX interfaces, i.e., `Element`, `IntrinsicElements`, `IntrinsicAttributes`, etc. We've been using it to hold most of our DOM-ish typings forever which is a) surprising for users looking for these types, if they can find them at all and b) probably not ideal given the magic handling of `JSX` by TS. Best to keep it pretty minimal and only load it up with the interfaces expected of it.

This will let users import types directly from Preact, like so:

```ts
// Before
import { JSX } from 'preact';

let MyButtonProps: JSX.ButtonHTMLAttributes & {
  //...
}

// After
import { ButtonHTMLAttributes } from 'preact';

let MyButtonProps: ButtonHTMLAttributes & {
  //...
}
```

---

Moving these out to a separate declaration file seemed easiest, as we need to re-export all of these types except for the `JSXInternal` namespace and I found that to be quite fiddly; I frequently broke `IntrinsicElements` when working on that and so just moved the unrelated types out. Seems less error-prone like this, and does somewhat nicely isolate the JSX-specific types. Might be easier to maintain.

Big ol' breaking change of course, but for the better I think. Should be pretty quick for consumers to fix up.


src/index.d.ts
```diff
@@ -284,7 +286,7 @@ export namespace h {
 //
 // Preact render
 // -----------------------------------
-interface ContainerNode {
+export interface ContainerNode {
```
rschristian, line 289: Fairly odd: we reference this in `/compat/index.d.ts`, and that's always been fine, but now it's throwing an error, stating `ContainerNode` cannot be found. This seems pretty legit to me, seeing as how we weren't exporting it.

test/ts/custom-elements.tsx
```diff
@@ -1,4 +1,4 @@
-import { createElement, Component, createContext } from '../../';
+import { createElement, Component, createContext, HTMLAttributes, MouseEventHandler } from '../../';
```
rschristian, line 1: `MouseEventHandler`, like `ContainerNode` above, seems to be a weird instance of global type access that shouldn't exist. I don't know why that worked, or what has since broke it, but I'd say this is correct?

src/jsx.d.ts
```diff
@@ -2,12 +2,117 @@
...
 	Component,
 	ComponentType,
 	FunctionComponent,
-	PreactDOMAttributes,
-	VNode
+	VNode,
+	SVGAttributes,
+	HTMLAttributes,
+	AnnotationMathMLAttributes,
+	AnnotationXmlMathMLAttributes,
```
rschristian: This mega import statement is kinda gross but I'm not sure an `import * as domTypes` is all that much better?

JoviDeCroock (approved):

PR #4881, opened 2025-08-08
Call tolowercase only when needed

JoviDeCroock:
- Defer toLowerCase() to only when needed (event props and translate), avoiding per-prop lowercasing on the hot path.
-  Remove ON_ANI regex and rely on core preact event normalization.
-  Preserve React-compat remaps:
   - onDoubleClick -> ondblclick
   - onChange -> oninput for input/textarea (except file/checkbox/radio)
   - onFocus -> onfocusin, onBlur -> onfocusout
   - Handle oninput duplication via oninputCapture

Reduces unnecessary string ops and regex checks.
Small bundle size reduction and improved render perf in compat.


compat/src/render.js
```diff
@@ -165,9 +163,11 @@ function handleDomVNode(vnode) {
 			// value will be used as the file name and the file will be called
 			// "true" upon downloading it.
 			value = '';
-		} else if (lowerCased === 'translate' && value === 'no') {
+		} else if (i === 'translate' && value === 'no') {
+			// Lowercase only for this specific, rare prop
 			value = false;
```
rschristian: This comment doesn't make sense to me, might be a leftover from an earlier iteration? Or am I being daft (also very possible)?

rschristian (approved):

compat/src/render.js
rschristian: And good spot BTW! Dropping the regex is great

JoviDeCroock: Whups, you're right, I first had `toLowerCase` for translate and then realised we don't need that 😅

PR #4882, opened 2025-08-08
refactor: Avoid breaking change for ARIA element types

rschristian:
Revert of #4876

All of the per-element types (`AnchorHTMLAttributes`, `AreaHTMLAttributes`, etc) via core & compat are now interfaces again (as they are in React), with the new accessible types exported & hooked up to `IntrinsicElements` so that users still get correct ARIA type checking. Users will now need to augment the `Accessible`-prefixed types for a few elements, which may be a surprise & needs documentation for sure, but I think that's a better tradeoff; more people will import element types to extend or consume than will need to augment (say) all anchor tags globally.

I spent so much time working on those aria roles & reading the spec that I suppose I overestimated/forgot how many elements in total need discriminated unions and can be represented properly, which turns out to be only `<a>`, `<area>`, ``, `<input>`, and `<select>`. In my head there were a lot more 😅

### DIFF
--- a/src/dom.d.ts
+++ b/src/dom.d.ts
@@ -33,15 +33,13 @@ interface CommandEventInit extends EventInit {
 	command: string;
 }
 
-
-
 type Booleanish = boolean | 'true' | 'false';
 
 export interface SignalLike<T> {
-		value: T;
-		peek(): T;
-		subscribe(fn: (value: T) => void): () => void;
-	}
+	value: T;
+	peek(): T;
+	subscribe(fn: (value: T) => void): () => void;
+}
 
 export type Signalish<T> = T | SignalLike<T>;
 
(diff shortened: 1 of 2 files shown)


src/dom.d.ts
```diff
@@ -1593,12 +1593,17 @@ export type AnchorAriaRoles =
...
+export type AccessibleAnchorHTMLAttributes<T extends EventTarget = HTMLAnchorElement> = Omit<
 	PartialAnchorHTMLAttributes<T>,
 	'role'
 > &
 	AnchorAriaRoles;
 
+export interface AnchorHTMLAttributes<T extends EventTarget = HTMLAnchorElement> extends PartialAnchorHTMLAttributes<T> {
+	href?: Signalish<string | undefined>;
+	role?: Signalish<AriaRole | undefined>;
+}
```
rschristian: This is all that the PR does: makes plain & wide interfaces to compliment the specific accessible types.

JoviDeCroock (approved):

PR #4883, opened 2025-08-09
Try new build

JoviDeCroock:
## General changes

- Drops umd output
- CJS uses destructured require rather than namespace (ESBuild limitation, we could transform but... not sure if it's worth it)
- Lifts syntax to ES2020
- Removes microbundle (a lot less deps)

## TODO

- Chase down size diff in `compat.mjs`
- Refactor the script
- Undo src change

### DIFF
--- a/scripts/build-packages.cjs
+++ b/scripts/build-packages.cjs
@@ -10,7 +10,6 @@ const { minify } = require('terser');
 const zlib = require('node:zlib');
 const { init: initEsmLexer, parse } = require('es-module-lexer');
 const MagicString = require('magic-string');
-const { platform } = require('node:os');
 
 /**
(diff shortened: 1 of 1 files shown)


rschristian:
> Lifts syntax to ES2020

~~I was under the impression that the ES2020 setting was just a temporary thing whilst you were working on this, is that not the case? I feel that's a bit too new to place our floor at, and I'm not sure it's necessary -- do we/the transpiler even use anything that new? ES2017 or so, IMO, should be as new as it gets if absolutely required.~~

Moving past ES5 also likely calls for investigation into ES6 classes, they'll certainly be smaller but perf could go either way. Not necessarily related to this PR of course, just opens up a few new options that could be good for us.

PR #4888, opened 2025-08-16
Do not re-insert memoized vnodes that keep their relative order after swap

vasylenkoval:
## Description 

I was looking at [js-framework-benchmark](https://krausest.github.io/js-framework-benchmark/2025/table_chrome_138.0.7204.50.html) and found something that looked a bit strange. All of the Preact benchmarks that use memoized rows have `swap rows` performance similar to React, while all other benchmarks that don't use memoized rows have faster swaps but incur the cost of not being able to skip processing unchanged rows in `select row`. 

After taking a closer look at your codebase, I was pleasantly surprised that Preact is flagging all moved nodes granularly during reconciliation as opposed to React's more optimistic handling (deopt backwards swaps). So I wanted to track down where the cost was coming from.

Screenshot 2025-08-13 at 9 03 59 PM

## Debugging
I modified the `preact-hooks` benchmark to have memoized rows and made an unminified build to debug. 

The insert condition seems to be triggered for these 3 cases:
- vnode was explicitly marked for insertion
- vnode has memoized children
- Combination of the two

```js
if (
   childVNode._flags & INSERT_VNODE ||
   oldVNode._children === childVNode._children
) {
   oldDom = insert(....)
}
```

For most cases, if the `insert` call is made for a memoized vnode without an insert flag, it will match the old dom and skip the insert. Unfortunately, when a swap is made, it's not guaranteed to match, resulting in unnecessary calls to `insertBefore` even though the order of vnodes between swapped items did not change.

Notice how it starts re-inserting rows between the swapped ones while pushing the second (swapped) row forward, instead of skipping over nodes in between and swapping the second row at the end when it encounters it during the diff. It's happening because the second row is still in the DOM at its old position and children in between are failing to skip the dom equality check that would bail out the `insertBefore` call.

https://github.com/user-attachments/assets/6895a197-be4c-4d71-b131-678c70706639

## Proposed change

I figured that a quick and dirty fix with the least amount of code would be to just pass a flag to `insert` to disable the DOM insertion in cases when it's not required, while still keeping the logic around sCU bailout and incrementing the next oldDom pointer in tact. Ideally it would be a separate code path but I wanted to get your feedback first. Let me know your thoughts!

## Tests

I added a test that verifies extra `insertBefore` calls are not happening in this case. I can also add more tests around simply testing re-ordering with memoized children, however, you seemed to have a pretty decent coverage there already.

## Results

Here's the comparison between unmodified `preact-classes` bench with and without the above commit. I chose `react-classes` because it's using memoized rows and I did not want to compare against a benchmark that I refactored to keep it fair.

Please disregard the `v10.25.2` at the top, both benchmarks are using the latest build from main at this commit 68b30fef5571d55eb44572586a2bdcbe3a22cb63, not npm. I just swapped the preact package in node modules, the version in the bench is pulled from package.json.

Other Preact benchmarks that do not use memoized rows would see even bigger shifts in overall scores if refactored.

Screenshot 2025-08-15 at 8 27 34 PM

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -128,16 +128,9 @@ export function diffChildren(
 			firstChildDom = newDom;
 		}
 
-		if (
-			childVNode._flags & INSERT_VNODE ||
-			oldVNode._children === childVNode._children
-		) {
-			oldDom = insert(
-				childVNode,
-				oldDom,
-				parentDom,
-				!(childVNode._flags & INSERT_VNODE) /* shouldSkipDomUpdate */
-			);
+		let shouldPlace = !!(childVNode._flags & INSERT_VNODE);
+		if (shouldPlace || oldVNode._children === childVNode._children) {
+			oldDom = insert(childVNode, oldDom, parentDom, shouldPlace);
 		} else if (typeof childVNode.type == 'function' && result !== UNDEFINED) {
 			oldDom = result;
 		} else if (newDom) {
@@ -347,10 +340,10 @@ function constructNewChildrenArray(
  * @param {VNode} parentVNode
  * @param {PreactElement} oldDom
  * @param {PreactElement} parentDom
- * @param {boolean} shouldSkipDomUpdate
+ * @param {boolean} shouldPlace
  * @returns {PreactElement}
  */
-function insert(parentVNode, oldDom, parentDom, shouldSkipDomUpdate) {
+function insert(parentVNode, oldDom, parentDom, shouldPlace) {
 	// Note: VNodes in nested suspended trees may be missing _children.
 	if (typeof parentVNode.type == 'function') {
 		let children = parentVNode._children;
@@ -361,13 +354,13 @@ function insert(parentVNode, oldDom, parentDom, shouldSkipDomUpdate) {
 				// children's _parent pointer to point to the newVNode (parentVNode
 				// here).
 				children[i]._parent = parentVNode;
-				oldDom = insert(children[i], oldDom, parentDom, shouldSkipDomUpdate);
+				oldDom = insert(children[i], oldDom, parentDom, shouldPlace);
 			}
 		}
 
 		return oldDom;
 	} else if (parentVNode._dom != oldDom) {
-		if (!shouldSkipDomUpdate) {
+		if (shouldPlace) {
 			if (oldDom && parentVNode.type && !oldDom.parentNode) {
 				oldDom = getDomSibling(parentVNode);
 			}


rschristian:
~~Smaller even? Compression works in mysterious ways...~~

Edit: Bit bigger now

---


**Size Change:** +13 B (+0.03%) 

**Total Size:** 47.5 kB

| Filename | Size | Change |
| :--- | :---: | :---: |
| `dist/preact.js` | 4.71 kB | +3 B (+0.06%) |
| `dist/preact.mjs` | 4.72 kB | +3 B (+0.06%) |
| `dist/preact.umd.js` | 4.78 kB | +7 B (+0.15%) |

<details><summary>ℹ️ <strong>View Unchanged</strong></summary>

| Filename | Size |
| :--- | :---: |
| `compat/dist/compat.js` | 3.9 kB |
| `compat/dist/compat.mjs` | 3.82 kB |
| `compat/dist/compat.umd.js` | 3.95 kB |
| `debug/dist/debug.js` | 3.9 kB |
| `debug/dist/debug.mjs` | 3.9 kB |
| `debug/dist/debug.umd.js` | 3.98 kB |
| `devtools/dist/devtools.js` | 260 B |
| `devtools/dist/devtools.mjs` | 271 B |
| `devtools/dist/devtools.umd.js` | 346 B |
| `hooks/dist/hooks.js` | 1.55 kB |
| `hooks/dist/hooks.mjs` | 1.58 kB |
| `hooks/dist/hooks.umd.js` | 1.62 kB |
| `jsx-runtime/dist/jsxRuntime.js` | 892 B |
| `jsx-runtime/dist/jsxRuntime.mjs` | 861 B |
| `jsx-runtime/dist/jsxRuntime.umd.js` | 966 B |
| `test-utils/dist/testUtils.js` | 473 B |
| `test-utils/dist/testUtils.mjs` | 473 B |
| `test-utils/dist/testUtils.umd.js` | 555 B |

</details>



---

Thanks for the PR & thorough investigation! I'll need to take a better look tomorrow (or someone else may jump in first), but looks like you did quite a bit of research & testing on this and we appreciate it!

JoviDeCroock (approved):
This is an awesome find, when I re-wrote the heuristics for skew based diffing I never checked the js-framework-benchmark and I even explicitly pointed out better swap performance 😅 

Thank you for the thorough research, this is a _really_ great find!

>  Ideally it would be a separate code path but I wanted to get your feedback first. Let me know your thoughts!

I agree with this generally, in Preact though we have a tendency to balance bytes and performance. Separating code paths often leads to relatively large byte size increases. We could prototype it and see what that balance is and I'm happy to discuss it but wanted to be clear about our philosophy.

src/diff/children.js
```diff
@@ -132,7 +132,12 @@ export function diffChildren(
 			childVNode._flags & INSERT_VNODE ||
 			oldVNode._children === childVNode._children
 		) {
-			oldDom = insert(childVNode, oldDom, parentDom);
+			oldDom = insert(
+				childVNode,
+				oldDom,
+				parentDom,
+				!(childVNode._flags & INSERT_VNODE) /* shouldSkipDomUpdate */
```
JoviDeCroock: We could probably hoist this check into a variable so we don't do it twice but don't feel strongly about that as it's just a bit check

marvinhagemeister (approved):
What a great find! This is a fascinating PR and explanation how you found this. This is great!!

rschristian (approved):
Just to add: the `main` branch here is our upcoming v11 release, `v10.x` is the current release line. This probably colored your results a bit more than intended as v11 but by all means this looks to be a really solid improvement (and w/ a byte reduction no less).

This change could likely also be ported to v10 w/ similar (I'm guessing/hoping) improvements there; it should be more or less a direct copy/paste, I don't think these code paths have changed in much v11.

Looks great though, and thanks again for the investigation!

vasylenkoval:
> Just as an aside: the `main` branch here is our upcoming v11 release, `v10.x` is the current release line. This probably colored your results a bit as v11 has landed some other improvements too, but by all means, this looks to be a really solid improvement (and w/ a byte reduction no less).
> 
> This change could likely also be ported to v10 w/ similar (I'm guessing/hoping) improvements there; it should be more or less a direct copy/paste, I don't think these code paths have changed in much v11.
> 
> Looks great though, and thanks again for the investigation!

Thanks for heads up! Will address comments and put up a PR to the v10 branch in a couple of hours. 

Also, amazing job on the V11 @rschristian @marvinhagemeister @JoviDeCroock!

I did in fact see a good boost in majority of the scores relative to V10. It was big enough jump for me to realize that I need to make sure I bench against latest main otherwise it wouldn’t be a fair benchmark lol. So the scores you see are technically both V11. 

Exciting stuff! Would be fun to get even closer to inferno :))

src/diff/children.js
vasylenkoval: @JoviDeCroock Thanks for your feedback! Pushed a commit that hoists the check into a variable.

After sleeping on it, I decided to also get rid of double negation to make it simpler to read, so I renamed `shouldSkipDomUpdate` to `shouldPlace`. Please let me know if you have any preferences on the flag name or anything else.

vasylenkoval: Re-requested approvals, will cherry-pick in another PR to 10.x right after review.

PR #4890, opened 2025-08-17
Port #4888 to v10

vasylenkoval:
## Description

This PR ports commits from https://github.com/preactjs/preact/pull/4888 to the v10 branch.

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -124,16 +124,9 @@ export function diffChildren(
 			firstChildDom = newDom;
 		}
 
-		if (
-			childVNode._flags & INSERT_VNODE ||
-			oldVNode._children === childVNode._children
-		) {
-			oldDom = insert(
-				childVNode,
-				oldDom,
-				parentDom,
-				!(childVNode._flags & INSERT_VNODE) /* shouldSkipDomUpdate */
-			);
+		let shouldPlace = !!(childVNode._flags & INSERT_VNODE);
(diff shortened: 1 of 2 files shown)


src/diff/children.js
```diff
@@ -352,16 +351,18 @@ function insert(parentVNode, oldDom, parentDom) {
...
 				// here).
 				children[i]._parent = parentVNode;
-				oldDom = insert(children[i], oldDom, parentDom);
+				oldDom = insert(children[i], oldDom, parentDom, shouldPlace);
 			}
 		}
 
 		return oldDom;
 	} else if (parentVNode._dom != oldDom) {
-		if (oldDom && parentVNode.type && !parentDom.contains(oldDom)) {
```
vasylenkoval, line 361: The ported code diff also happens to contain one more improvement that replaces the `parentDom.contains(...)` check with `!oldDom.parentNode`. Original PR: https://github.com/preactjs/preact/pull/4666. Please let me know if you don't want to include it here.

rschristian: Should be fine I think, test suite passes anyhow. And thanks for tracking down the blame there, saves me from doing it myself

vasylenkoval:
Thanks @rschristian! Will update the `js-framework-benchmark` Preact implementations as well once a new release is cut.

JoviDeCroock:
Published @vasylenkoval Thank you again for all of this work, you're a hero!

PR #4891, opened 2025-08-18
ci: Fix benchmarks for v10.x

rschristian:
Our [benchmarks action is failing](https://github.com/preactjs/preact/actions/runs/17023195272/job/48255153404) on PRs opened against v10.x as we're not running the "CI" workflow on push. The "CI" workflow triggers "Build & Test" which uploads assets that future commits will compare against; the merge commits we're making at the moment aren't associated with a workflow run, ergo, the action will fail to download the artifact for comparison.

Edit: Or at least that's what I think is happening. Fingers crossed?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,8 +8,7 @@ on:
   push:
     branches:
       - main
-      - restructure
-      - v11
+      - v10.x
 
 jobs:
   filter_jobs:


rschristian:
On a semi-related note, the Tachometer comment will keep appearing for PRs against `v10.x`, even when it shouldn't, as there's no way at this time to get a reference to the target branch for `workflow_run` events when the workflow is triggered due to a PR from a fork. To get the reporter at least semi-working, it uses the SHA of the location in which the reporter is running, aka, `main`, to use as the base ref. This, however, has the effect of always reporting the JS changed as `main` and `v10.x` are now divergent so the diff between them is meaningless.

So it'll be a bit noisy and there will be some empty comments like above, but we'll still get reporting when the benches run, which seems preferable to disabling the thing on branches other than `main`.

I don't know what else to do here besides write a script to start querying GH's API manually which I'm not exactly thrilled about nor enthusiastic to do; feels like there's a really weird gap in the event payload data here. I opened a couple threads on their site-wide discussions repo, we'll see if that goes anywhere I guess.

PR #4894, opened 2025-08-20
Save bytes on hydration 2.0

JoviDeCroock:
More ideas are 100% welcomed

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -80,7 +80,7 @@ export function diff(
 		(isHydrating = !!(oldVNode._flags & MODE_HYDRATE))
 	) {
 		excessDomChildren = oldVNode._component._excess;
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
LGTM

src/diff/index.js
```diff
@@ -75,13 +75,14 @@ export function diff(
 	if (newVNode.constructor != UNDEFINED) return NULL;
 
 	// If the previous diff bailed out, resume creating/hydrating.
-	if (oldVNode._flags & MODE_SUSPENDED) {
-		isHydrating = !!(oldVNode._flags & MODE_HYDRATE);
-		if (oldVNode._component._excess) {
```
rschristian, line 80: I have yet to dig into an attempt to reproduce but the removal of this check is causing CI failures in iso. Was it intentionally removed? I'd think we'd still need it.

PR #4895, opened 2025-08-20
Proposal to update ecma in terser

JoviDeCroock:
Size changes are quite nice

### DIFF
--- a/mangle.json
+++ b/mangle.json
@@ -19,7 +19,8 @@
     },
     "compress": {
       "hoist_vars": true,
-      "reduce_funcs": false
+      "reduce_funcs": false,
+      "ecma": 2020
     }
   },
   "props": {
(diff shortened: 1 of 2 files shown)


rschristian:
Size benefits would likely be much greater if we made more adjustments to the source too, depending on perf characteristics. ES2020 means ES6 classes, which might save some bytes for `BaseComponent`, optional chaining in a few places, optional catch bindings (though perhaps terser already does this), etc. There'd be more opportunities besides the -14b I think, though that might mean dealing with Microbundle.

A 7 year jump in minimum browser versions probably isn't too unreasonable given the 6 year lifespan of X, though that is a big jump.

JoviDeCroock:
Did a lot of the work in https://github.com/preactjs/preact/pull/4883 however it's really hard to make ESBuild work well with CJS/UMD, maybe I should try rolldown 😅

rschristian:
I mean, we could just add a more minimal Rollup config? Speed is great, but if the newer tools can't produce output worth a damn, we can keep using older tools? That being said, Vitest is still pretty much unusable on my system so Microbundle might only seem acceptably slow to me.

I wouldn't be surprised if Microbundle is considerably slower than a more modern & purpose-built Rollup/Babel config too, maybe I'll take a look after work.

PR #4901, opened 2025-08-25
Add TrustedHTML (from @types/trusted-types) to dangerouslySetInnerHTML

azekeprofit:
If you use [DomPurify with TrustedTypes](https://github.com/cure53/DOMPurify?tab=readme-ov-file#what-about-dompurify-and-trusted-types), setting dangerouslySetInnerHTML requires faking the type:

```jsx
<div dangerouslySetInnerHTML={{
   __html: parsed as unknown as string // hacking the type since Preact innerHTML doesn't list TrustedHTML as possible type
}} />
```

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -66,7 +66,7 @@ export interface ClassAttributes<T> extends Attributes {
 export interface PreactDOMAttributes {
 	children?: ComponentChildren;
 	dangerouslySetInnerHTML?: {
-		__html: string;
+		__html: string | TrustedHTML;
 	};
 }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,6 +18,7 @@
         "@babel/register": "^7.27.1",
         "@biomejs/biome": "^2.1.2",
         "@types/node": "^18.19.87",
+        "@types/trusted-types": "^2.0.7",
         "@vitest/browser": "^3.2.4",
         "@vitest/coverage-v8": "^3.2.4",
         "babel-plugin-transform-rename-properties": "0.1.0",
@@ -3468,6 +3469,13 @@
       "dev": true,
       "optional": true,
       "peer": true
+    },
+    "node_modules/@types/trusted-types": {
+      "version": "2.0.7",
+      "resolved": "https://registry.npmjs.org/@types/trusted-types/-/trusted-types-2.0.7.tgz",
+      "integrity": "sha512-ScaPdn1dQczgbl0QFTeTOmVHFULt394XJgOQNoyVhZ6r2vLnMLJfBPd53SB52T/3G36VI1/g2MZaX0cwDuXsfw==",
+      "dev": true,
+      "license": "MIT"
     },
     "node_modules/@types/which": {
       "version": "2.0.2",
--- a/package.json
+++ b/package.json
@@ -212,6 +212,7 @@
     "@babel/register": "^7.27.1",
     "@biomejs/biome": "^2.1.2",
     "@types/node": "^18.19.87",
+    "@types/trusted-types": "^2.0.7",
     "@vitest/browser": "^3.2.4",
     "@vitest/coverage-v8": "^3.2.4",
     "babel-plugin-transform-rename-properties": "0.1.0",
--- a/src/dom.d.ts
+++ b/src/dom.d.ts
@@ -2,6 +2,8 @@
 
 import { ClassAttributes, PreactDOMAttributes } from 'preact';
 
+export interface TrustedHTML {}
+
 // Implementations of some DOM events that are not available in TS 5.1
 interface ToggleEvent extends Event {
 	readonly newState: string;


src/index.d.ts
```diff
@@ -66,7 +66,7 @@ export interface ClassAttributes<T> extends Attributes {
 export interface PreactDOMAttributes {
 	children?: ComponentChildren;
 	dangerouslySetInnerHTML?: {
-		__html: string;
+		__html: string | TrustedHTML;
```
rschristian, line 69: Packages listed in `devDependencies` won't be available at runtime, and we won't be adding deps for the sake of types. Can you vendor this type by chance?

Edit: Does this just define a `.toString()` that gets called? I might prefer users calling that manually & passing the result in rather than the `TrustedHTML` object as it'll simplify the types situation.

azekeprofit: Trusted Types are browser DOM API so they will be available at runtime, and one of the few ways they will get it is something like DOMPurify which also uses same @types/trusted-types (which is planned to be integrated into Typescript's lib.dom.ts anyway).

Calling .toString() manually won't work because we need object to be Trusted Type at runtime on sites that are marked as [`Content-Security-Policy: trusted-types`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/trusted-types)

azekeprofit: I can fake TrustedHTML like React does [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d65f24c95e732e4b0eadaebf63f5699801a05e59/types/react/global.d.ts#L161), but i don't know where would that go in Preact

rschristian: Trusted Types are available at runtime, the `TrustedHTML` type, which you used above, will not be. That'll be a type error for every single consumer that doesn't happen to have `@types/trusted-types` installed at the moment, which I imagine is most.

> which is planned to be integrated into Typescript's lib.dom.ts anyway

Our TS support is locked at v5.1 for the upcoming Preact 11 release line so we can't rely on this eventually making its way into `lib.dom.ts`. This will have to be vendored.

> Calling `.toString()` manually won't work because we need object to be Trusted Type at runtime on sites that are marked as [Content-Security-Policy: trusted-types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/trusted-types)

Gotcha, didn't realize that. Thanks!

rschristian: > I can fake TrustedHTML like React does [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d65f24c95e732e4b0eadaebf63f5699801a05e59/types/react/global.d.ts#L161), but i don't know where would that go in Preact

We have some vendored implementations here: https://github.com/preactjs/preact/blob/f23272c3d7c3e45e6d99f807080ccd802e795045/src/dom.d.ts#L5

Before or after, either works.

azekeprofit: I added empty `TrustedHTML` interface, will this work?

rschristian: An empty interface that isn't imported, so I doubt it? Is the result of DOMPurify assignable to that? It seems like it wouldn't be.

azekeprofit: I was looking into ways to make it work, but this has gone way beyond my Typescript skills.

The simpler/lazier solution would be to simply wait until it gets merged into [lib.dom.ts](https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1246) and Preact adopts that version.

rschristian: No worries, and thanks for trying! I'll take a look when I can, maybe I can come up with a hack.

> wait until it gets merged into lib.dom.ts and Preact adopts that version.

Unfortunately the minimum supported TS version should only change in a major version, and we've spent almost 6 years on v10... whilst I can't say I know what the future holds, we probably want to find a solution besides telling users to wait for v12 as that might be years away.

rschristian:
Looked into this further tonight & unfortunately you're going to need to cast or patch Preact's types, we can't accommodate this.

We can't vendor the type, as it's a class so they'll never align, and you apparently can't override existing interface properties with module augmentation, so even users can't override it nicely. I don't think there's any other options available.

Going to close this on out, but if you do find a solution, we'd love to have it.

lukewarlow:
`__html: Parameters<DOMParser['parseFromString']>[0]` - it's not pretty but this will at least future proof Preact types, once dom types update you'll get string | TrustedHTML on newer TS.

Provided `Parameters` exists far enough back in TS this should work?

rschristian:
`Parameters` is within our support range, yes. Back to TS 3.1 if their docs are correct

lukewarlow:
This has been partially addressed by #4930 fwiw. Once TS updates (which it should do soon given this is shipped in 2 browsers now) it will take TrustedHTML when you use new TS versions.

PR #4904, opened 2025-08-27
refactor (v10): Mirror non-JSX types to the 'preact' namespace

### LINKED ISSUE #4903: Preact 11 backwards compatibility concern
## Context

I am a web component library developer. I provide Preact JSX typings for my web components.
Preact 11 changes the JSX namespace. That will cause my JSX typings to no longer work - I can update them for Preact 11 but I don't think I can support both 10 and 11 with the same file.

## Request

Could you please re-export the HTMLAttributes interface from the JSX namespace (as deprecated if you wish) for one more major release? That will give time for users to update to Preact 11

For an example typings file see dist/types/preact.d.ts in https://www.npmjs.com/package/@esri/calcite-components?activeTab=code
You could use it for compatibility testing if it helps.

rschristian:
Closes #4903
Non-breaking form of #4878

Duplicates our DOM types into a new `dom.d.ts` file (matching v11), exporting them under the Preact namespace (so you can do `import type { HTMLAttributes } from 'preact';`), but maintains back-compat . This will allow library authors to switch to the Preact namespace and keep compatibility with v10 & v11 going forward.

There's obviously some duplication here but unfortunately it's necessary -- turns out you can't label namespace aliases (i.e., `export import ...`) as deprecated. We need to keep the current types where they are so that users are warned they're using deprecated APIs.

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -1635,6 +1635,7 @@ export namespace JSXInternal {
 		itemType?: Signalish<string | undefined>;
 	}
 
+	/** @deprecated Please import from the Preact namespace instead */
 	type HTMLAttributeReferrerPolicy =
 		| ''
 		| 'no-referrer'
@@ -1646,6 +1647,7 @@ export namespace JSXInternal {
 		| 'strict-origin-when-cross-origin'
 		| 'unsafe-url';
 
+	/** @deprecated Please import from the Preact namespace instead */
 	type HTMLAttributeAnchorTarget =
 		| '_self'
 		| '_blank'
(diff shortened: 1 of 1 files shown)


src/jsx.d.ts
```diff
@@ -88,2621 +41,468 @@ export namespace JSXInternal {
...
-			| 'getPropertyPriority'
-		>]?: string | number | null | undefined;
-	};
-	export type AllCSSProperties = {
-		[key: string]: string | number | null | undefined;
-	};
-	export interface CSSProperties extends AllCSSProperties, DOMCSSProperties {
-		cssText?: string | null;
-	}
+	export import DOMCSSProperties = preact.DOMCSSProperties;
```
rschristian: Whilst I'd like to add `@deprecated` jsdoc comments here to steer folks to the Preact namespace, turns out you can't: `export import` forwards on the comments of the type, it itself cannot be directly annotated.

rschristian: Update: This might be a bug/missing feature in TS, `export import` specifically is missing this functionality at the moment

maxpatiiuk:
Hi
Do you know when this change will be available on npm for preact@10?

rschristian:
Next time we cut a release, but when that happens is TBD. We don't schedule things.

PR #4907, opened 2025-08-27
Prevent frequently updated components from retaining memory

### LINKED ISSUE #4905: VNode leak in Context
**Title:** Memory leak in Context providers - old component instances not disposed during re-renders

Tested with preact11 beta, same issue.

**Describe the bug**
Context provider components are not properly disposed of when their state updates, causing a memory leak where multiple component instances accumulate in memory. Each state update should properly clean up the previous Context provider component instance, but instead old instances persist with their associated hooks, vnodes, and DOM references. This occurs with regular `useState` updates, not just signal-driven updates.

**To Reproduce**

https://stackblitz.com/edit/vitejs-vite-21mvd37v?file=src%2Fmain.tsx,index.html

Minimal reproduction code:
```javascript
import { render, createContext } from 'preact';
import { useState, useContext } from 'preact/hooks';

const RouterContext = createContext(null);

let setState; // Expose setState so child can update parent state

function RouterProvider({ children }) {
  const [routerState, setRouterState] = useState({ params: 'a' });
  
  // Expose setState globally so button can call it directly
  setState = setRouterState;

  return (
    <RouterContext.Provider value={{ params: routerState.params }}>
      {children}
    </RouterContext.Provider>
  );
}

function RouterContent() {
  const { params } = useContext(RouterContext);
  return (
    <button onClick={() => setState({ params: params + "!" })}>
      {`No account found for username "${params}"`}
    </button>
  );
}

render(
  <RouterProvider>
    <RouterContent />
  </RouterProvider>, 
  document.getElementById('app')
);
```

1. Go to the app with the code above
2. Click on the button multiple times (e.g., 5 times) 
3. Take a memory snapshot in browser dev tools and analyze retained objects
4. See error: Multiple RouterProvider component instances are retained in memory, each with their own hook state objects and component trees containing old parameter values ("a", "a!", "a!!", etc.)

**Expected behavior**
After each state update in the Context provider, only the current component instance should remain in memory. Previous RouterProvider instances and their associated hooks/vnodes should be properly disposed of and garbage collected. Memory should only contain references to the current state value, not historical values.

JoviDeCroock:
Resolves #4905 
Duplicate of #4906

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -6,8 +6,7 @@ import {
 	COMPONENT_FORCE,
 	COMPONENT_DIRTY,
 	MODE_HYDRATE,
-	NULL,
-	UNDEFINED
+	NULL
 } from './constants';
 
 /**
--- a/src/component.js
+++ b/src/component.js
@@ -6,7 +6,8 @@ import {
 	COMPONENT_FORCE,
 	COMPONENT_DIRTY,
 	MODE_HYDRATE,
-	NULL
+	NULL,
+	UNDEFINED
 } from './constants';
 
 /**
@@ -154,6 +155,7 @@ function renderComponent(component) {
 		newVNode._original = oldVNode._original;
 		newVNode._parent._children[newVNode._index] = newVNode;
 		commitRoot(commitQueue, newVNode, refQueue);
+		oldVNode._parent = oldVNode._dom = NULL;
 
 		if (newVNode._dom != oldDom) {
 			updateParentDomPointers(newVNode);


src/component.js
```diff
@@ -6,7 +6,8 @@ import {
 	COMPONENT_FORCE,
 	COMPONENT_DIRTY,
 	MODE_HYDRATE,
-	NULL
+	NULL,
+	UNDEFINED
```
JoviDeCroock:
```suggestion
	NULL
```

rschristian (approved):

korDen:
FYI this is fixing partial leak but RouterContent and RouterProvider objects are still leaking (_original: -2 and _original: -1 if that helps)

PR #4912, opened 2025-09-01
ci: Revert #4911, run benches on v10 again

rschristian:


rschristian:
Should be all good now! I reverted the commit to trigger the benches but it ran successfully, as you can see from the comment above.

PR #4925, opened 2025-09-28
test: Switch tests to use `.jsx` file extension

rschristian:
Seems to be the way the ecosystem has moved over the last 4-5 years, therefore it's likely that the only thing holding us back from doing this was the rather boring work of renaming all these files? Lmk though, not sure if anyone else has opinions here.


jsconfig.json
```diff
@@ -14,7 +14,8 @@
     },
     "target": "es5",
     "noEmit": true,
-    "skipLibCheck": false
+    "skipLibCheck": false,
+    "types": ["vitest/globals"]
```
rschristian, line 18: Was this not an issue for everyone else btw? In every single file (beyond our TS tests, which also set this in their `tsconfig.json` files) I get hundreds of `Cannot fine name 'describe'/'expect'. Do you need to install type definitions...`

I'd almost have to assume folks are using other editor setups than me as it makes mine light up like a christmas tree without this

vitest.config.mjs
```diff
@@ -137,11 +136,11 @@ export default defineConfig({
 	},
 	esbuild: {
 		loader: 'jsx',
-		include: /.*\.js$/,
+		include: /.*\.jsx$/,
 		exclude: ['node_nodules'],
 		jsx: 'transform',
-		jsxImportSource: 'preact',
-		jsxDev: true
```
rschristian, line 144: _I believe_ these settings did nothing and were actively incorrect as ESBuild seems to prefer our settings in the `jsconfig.json` instead, which uses the classic transform? Hence why we have the default `React` imports in compat & the `/** @jsx createElement */` comments everywhere else?

Switched it over to use the classic transform options, i.e., `jsxFactory` & `jsxFragment` though, as mentioned, it seems entirely redundant as these are overridden.

marvinhagemeister (approved):

PR #4930, opened 2025-10-23
Updates dangerouslySetInnerHTML type so future TS will accept Trusted…

lukewarlow:
…HTML automatically

Need to see if there's tests for this, and how to update them.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -66,7 +66,7 @@ export interface ClassAttributes<T> extends Attributes {
 export interface PreactDOMAttributes {
 	children?: ComponentChildren;
 	dangerouslySetInnerHTML?: {
-		__html: string;
+		__html: Parameters<DOMParser['parseFromString']>[0];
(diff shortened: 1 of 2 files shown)


rschristian (approved):
Thanks!

Just to expound: `parseFromString` should take [A TrustedHTML or string instance defining HTML to be parsed.](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString), though TS has yet to update `lib/dom.d.ts` to account for this. The method here should in theory be future proof though and collect the correct types for `.innerHTML` when/if TS makes them available.

Quite crafty

lukewarlow:
To provide a bit of extra detail we can't do something using innerHTML itself as that is currently a property but might become a getter setter pair and there's not a TS way to handle both those cases, nor does there seem to be a way to get the setter type, so this is a hack to get string | TrustedHTML in a workable way.

PR #4935, opened 2025-10-24
fix: Avoid Trusted Types violation when dangerouslySetInnerHTML is re…

### LINKED ISSUE #4934: Conditional dangerouslySetInnerHTML triggers Trusted Types violation

**Describe the bug**
If you conditionally apply dangerouslySetInnerHTML, Preact internal code leads to a TrustedTypes violation.

**To Reproduce**

If possible, please provide a link to a StackBlitz/CodeSandbox/Codepen project or a GitHub repository that demonstrates the issue. You can use the following template on StackBlitz to get started: https://stackblitz.com/edit/create-preact-starter

1. Go to https://codepen.io/lukewarlow/pen/wBMjmJM
2. Click on "Toggle HTML" *twice*
3. Open browser dev tools and see the Trusted Types violation (will need Chromium or Safari 26+)

**Expected behavior**
What should have happened when following the steps above?

The text should toggle between the two values.

lukewarlow:
…moved

Setting innerHTML to an empty string violates Trusted Types. This changes to using textContent tp empty element children.

Fixes #4934


lukewarlow:
No new test because tests such as 'should apply proper mutation for VNodes with dangerouslySetInnerHTML attr' already cover this (if you change the empty string to 'foo' for example it fails.

src/diff/index.js
```diff
@@ -607,7 +607,7 @@ function diffElementNodes(
 
 			newVNode._children = [];
 		} else {
-			if (oldHtml) dom.innerHTML = '';
+			if (oldHtml) dom.textContent = '';
```
JoviDeCroock, line 616: I'd prefer to have a test for the trusted HTML case if that's possible at all

JoviDeCroock (approved):

PR #4938, opened 2025-10-24
Experiment: add moveBefore rather than insertBefor when available

JoviDeCroock:
One issue is that the selection-range is reset when moveBefore is used, heh looks like we found a browser bug https://stackblitz.com/edit/vitejs-vite-k5ibmktp?file=index.html,src%2Fmain.js,src%2Fstyle.css&terminal=dev

### DIFF
--- a/src/diff/children.js
+++ b/src/diff/children.js
@@ -88,11 +88,8 @@ export function diffChildren(
 
 		// At this point, constructNewChildrenArray has assigned _index to be the
 		// matchingIndex for this VNode's oldVNode (or -1 if there is no oldVNode).
-		if (childVNode._index == -1) {
-			oldVNode = EMPTY_OBJ;
-		} else {
-			oldVNode = oldChildren[childVNode._index] || EMPTY_OBJ;
-		}
+		oldVNode =
+			(childVNode._index != -1 && oldChildren[childVNode._index]) || EMPTY_OBJ;
 
 		// Update childVNode._index to its final index
 		childVNode._index = i;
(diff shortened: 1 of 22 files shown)


src/diff/children.js
```diff
@@ -362,7 +365,13 @@ function insert(parentVNode, oldDom, parentDom, shouldPlace) {
 			if (oldDom && parentVNode.type && !oldDom.parentNode) {
 				oldDom = getDomSibling(parentVNode);
 			}
-			parentDom.insertBefore(parentVNode._dom, oldDom || NULL);
+
+			if (HAS_MOVE_BEFORE_SUPPORT && !isMounting) {
+				// @ts-expect-error This isn't added to TypeScript lib.d.ts yet
+				parentDom.moveBefore(parentVNode._dom, oldDom);
```
rschristian: I believe `.moveBefore` can throw, _I think_ it need to be in a `try/catch` falling back to `.insertBefore` (or at least that used to be the case when I was testing it)

JoviDeCroock: It doesn't throw as long as we check whether we're not mounting, because that seems to be the throw condition. Re-parenting would also be iffy so we might need to add `parentVNode._dom.parentNode === parentDom `

rschristian: Ah yeah, try/catch might be a poor way around handling the parenting stuff properly. Sorry for the silly comment!

JoviDeCroock: nah it's a good shout, created the chrome bug and we can try to break it after

PR #4946, opened 2025-10-29
Adds `scrollsnapchange` and `scrollsnapchanging` event support

argyleink:
1. https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollsnapchange_event
1. https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollsnapchanging_event

### DIFF
--- a/src/dom.d.ts
+++ b/src/dom.d.ts
@@ -33,6 +33,22 @@ interface CommandEventInit extends EventInit {
 	command: string;
 }
 
+/** [MDN Reference](https://developer.mozilla.org/en-US/docs/Web/API/SnapEvent) */
+interface SnapEvent extends Event {
+  readonly snapTargetBlock: Element | null;
+  readonly snapTargetInline: Element | null;
+}
+
+declare var SnapEvent: {
+  prototype: SnapEvent;
+  new (type: string, eventInitDict?: SnapEventInit): SnapEvent;
+}
+
+interface SnapEventInit extends EventInit {
+  snapTargetBlock?: Element | null;
+  snapTargetInline?: Element | null;
+}
+
 type Booleanish = boolean | 'true' | 'false';
 
 export interface SignalLike<T> {
@@ -494,6 +510,10 @@ export type TargetedPointerEvent<Target extends EventTarget> = TargetedEvent<
 	Target,
 	PointerEvent
 >;
+export type TargetedSnapEvent<Target extends EventTarget> = TargetedEvent<
+    Target,
+    SnapEvent
+>;
 export type TargetedSubmitEvent<Target extends EventTarget> = TargetedEvent<
 	Target,
 	SubmitEvent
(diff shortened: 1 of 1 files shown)


test/browser/events.test.jsx
```diff
@@ -291,4 +291,29 @@ describe('event handling', () => {
...
+			'scrollsnapchange',
+			expect.any(Function),
+			false
+		);
+		expect(proto.addEventListener).toHaveBeenCalledWith(
+			'scrollsnapchanging',
+			expect.any(Function),
+			false
+		);
+	});
```
rschristian: As this is a types-only change, we don't really need a new test case here but in `/test/ts`, if necessary. There's nothing particularly interesting about these two as far as I can see (just above there's `focusin`/`focusout`, which have been a bit iffy historically, and `pointercapture` which caused a hiccup for our regex)

src/dom.d.ts
```diff
@@ -775,6 +775,8 @@ export interface DOMAttributes<Target extends EventTarget>
 	// UI Events
 	onScroll?: UIEventHandler<Target> | undefined;
 	onScrollEnd?: UIEventHandler<Target> | undefined;
+	onScrollSnapChange?: UIEventHandler<Target> | undefined;
+	onScrollSnapChanging?: UIEventHandler<Target> | undefined;
```
rschristian: Actually, are these the correct handlers? Looks like this is meant to be a [`SnapEvent`](https://developer.mozilla.org/en-US/docs/Web/API/SnapEvent), not a `UIEvent`

argyleink: i modeled the types after the other scroll events like `onScroll` and `onScrollEnd` which are `UIEvent`

test/browser/events.test.jsx
argyleink: cool np, tests removed

argyleink:
> [The `main` branch is targetting v11](https://github.com/preactjs/preact/blob/main/CONTRIBUTING.md#important-branches), which is still in beta. If you're willing to mirror this to the `v10.x` branch, targetting the current v10 release line, we'd appreciate it, else one of us will do it when we can.

mirror into `v10.x` https://github.com/preactjs/preact/pull/4947

src/dom.d.ts
rschristian: Er, but that's not correct? The MDN links show that, despite the similarity in name, they are different events.

[scrollEnd](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollend_event#event_type)
  - A generic [Event](https://developer.mozilla.org/en-US/docs/Web/API/Event)

[scrollSnapChange](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollsnapchange_event#event_type) 
  - A [SnapEvent](https://developer.mozilla.org/en-US/docs/Web/API/SnapEvent), which inherits from the generic [Event](https://developer.mozilla.org/en-US/docs/Web/API/Event) type.

With this type, you would not be able to access any of the `SnapEvent`-specific properties like `.snapTargetBlock` or `.snapInlineTarget`, they'd surface a type error instead.

---

~~I'm not sure why `UIEvent` is used in our types for `scrollEnd`, doesn't look to be correct to me, but that's not an issue to solve here.~~ Edit: Fixed that, they're now both generic events.

PR #4951, opened 2025-10-31
refactor(children): remove redundant assignment to childVNode

ali-garajian:


JoviDeCroock (approved):
Great find! Looks like it saves 2 bytes, thank you for cleaning this up and contributing! If you want you can also backport this to v10.x which I intend to draft a release for this weekend

PR #4956, opened 2025-11-09
refactor(children): remove redundant assignment to oldVNode

ali-garajian:


rschristian:
This one I'm less sure of, looks a little suspect looking at line 253/254 & this commit: https://github.com/preactjs/preact/commit/b4a1cc206b3519d80b931134933c917b0f08178c

Also, rebasing before opening PRs is generally appreciated. This has 6 commits associated with it because you haven't.

ali-garajian:
@rschristian sorry about the extra commits, won't happen again

I can see your concern, but I'm going through every branch of code after that assignment and it seems everything would be fine without it. 
besides, the tests are passing

rschristian:
> sorry about the extra commits, won't happen again

No worries, and you can rebase this yet. Just makes it a bit easier for us is all.

> but I'm going to every branch of code after that assignment and it seems everything would be fine without it.

Looks a bit suspect to me yet, would need to look into it more later this week

> besides, the tests are passing

In a perfect world that'd be enough, but in reality, test suites can have gaps. That's certainly a good sign though.

AhmedEjaz-Frontend (approved):

PR #4962, opened 2025-11-21
Delete redundant code of conduct

JoviDeCroock:
Looks like our global one can then take presedence once https://github.com/preactjs/.github/pull/3 is merged


rschristian (approved):
Could also delete [`.github/FUNDING.yml`](https://github.com/preactjs/preact/blob/main/.github/FUNDING.yml), if you open it GitHub provides a banner stating it's overriding the global setting which we also don't really need.

PR #4970, opened 2025-12-02
feat: support use hook

### LINKED ISSUE #4756: React 19 compatibility: Add support for `use(promise)` to `preact/compat`
React 19 added a [use()](https://react.dev/reference/react/use) function that can be called from components to do async data fetching and suspend the component in the meantime. It would be great to have this in `preact/compat`.

This was asked about in https://github.com/preactjs/preact/issues/4613#issuecomment-2657751573, but I think it warrants its own issue, so here it is.

gaoachao:
close #4756

### DIFF
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -125,6 +125,7 @@ declare namespace React {
 	export import useRef = _hooks.useRef;
 	export import useState = _hooks.useState;
 	// React 18 hooks
+	export import use = _hooks.use;
 	export import useInsertionEffect = _hooks.useLayoutEffect;
 	export function useTransition(): [false, typeof startTransition];
 	export function useDeferredValue<T = any>(val: T): T;
--- a/compat/src/index.d.ts
+++ b/compat/src/index.d.ts
@@ -125,7 +125,6 @@ declare namespace React {
 	export import useRef = _hooks.useRef;
 	export import useState = _hooks.useState;
 	// React 18 hooks
-	export import use = _hooks.use;
 	export import useInsertionEffect = _hooks.useLayoutEffect;
 	export function useTransition(): [false, typeof startTransition];
 	export function useDeferredValue<T = any>(val: T): T;
@@ -133,6 +132,8 @@ declare namespace React {
 		subscribe: (flush: () => void) => () => void,
 		getSnapshot: () => T
 	): T;
+	// React 19 hooks
+	export import use = _hooks.use;
 
 	// Preact Defaults
 	export import Context = preact1.Context;
--- a/hooks/src/index.d.ts
+++ b/hooks/src/index.d.ts
@@ -129,7 +129,7 @@ export function useContext<T>(context: PreactContext<T>): T;
  * Preact implementation of React's use hook
  * Supports Promise and Context consumption
  */
-export function use<T>(usable: Promise<T> | PreactContext<T>): T;
+export function use<T>(resource: Promise<T> | PreactContext<T>): T;
 
 /**
  * Customize the displayed value in the devtools panel.
--- a/hooks/src/index.d.ts
+++ b/hooks/src/index.d.ts
@@ -126,6 +126,12 @@ export function useMemo<T>(factory: () => T, inputs: Inputs | undefined): T;
 export function useContext<T>(context: PreactContext<T>): T;
 
 /**
+ * Preact implementation of React's use hook
+ * Supports Promise and Context consumption
+ */
+export function use<T>(usable: Promise<T> | PreactContext<T>): T;
+
+/**
  * Customize the displayed value in the devtools panel.
  *
  * @param value Custom hook name or object that is passed to formatter
--- a/hooks/src/index.d.ts
+++ b/hooks/src/index.d.ts
@@ -126,7 +126,6 @@ export function useMemo<T>(factory: () => T, inputs: Inputs | undefined): T;
 export function useContext<T>(context: PreactContext<T>): T;
 
 /**
- * Preact implementation of React's use hook
  * Supports Promise and Context consumption
  */
 export function use<T>(resource: Promise<T> | PreactContext<T>): T;
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -164,7 +164,6 @@ function getHookState(index, type) {
 	return hooks._list[index];
 }
 
-const PROMISE_CACHE = new WeakMap();
 /**
  * Supports Promise and Context consumption
  *
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -191,25 +191,25 @@ export function use(resource) {
  */
 function usePromise(promise) {
 	const [promiseState, update] = useState({
-			value: undefined,
-			reason: undefined,
+		value: undefined,
+		reason: undefined
 	});
 
 	if (!promiseState.value && !promiseState.reason) {
 		promise.then(
 			value => {
-				update({ status: 'fulfilled', value })
+				update({ status: 'fulfilled', value });
 			},
 			reason => {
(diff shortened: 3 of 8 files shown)


src/create-context.js
```diff
@@ -42,6 +42,7 @@ export function createContext(defaultValue) {
 		return props.children;
 	}
 
+	Context.$$typeof = CONTEXT_TYPE;
```
JoviDeCroock: Let's definitely not do this, we are leaking react implementation details into the core implementation 😅

hooks/src/index.js
```diff
@@ -164,6 +164,101 @@ function getHookState(index, type) {
 	return hooks._list[index];
 }
 
+const PROMISE_CACHE = new WeakMap();
```
JoviDeCroock: We should not be responsible for ensuring that a Promise is cached, nor does react do that. The user is responsible for the referential integrity of the Promise and we just use `.then`/`.catch` to trigger re-render.

hooks/src/index.js
```diff
@@ -164,6 +164,101 @@ function getHookState(index, type) {
...
+	const [, forceUpdate] = useState(/** @type {object} */ ({}));
+
+	let promiseState = PROMISE_CACHE.get(promise);
+
+	if (!promiseState) {
+		promiseState = {
+			status: 'pending',
+			value: undefined,
+			reason: undefined,
+			subscribers: new Set()
```
JoviDeCroock: This feels like bundle size bloat imho, we can get rid of subscribers

compat/src/index.d.ts
```diff
@@ -125,6 +125,7 @@ declare namespace React {
 	export import useRef = _hooks.useRef;
 	export import useState = _hooks.useState;
 	// React 18 hooks
+	export import use = _hooks.use;
```
rschristian: Nit: `use` is from React 19

hooks/src/index.d.ts
```diff
@@ -125,6 +125,12 @@ export function useMemo<T>(factory: () => T, inputs: Inputs | undefined): T;
  */
 export function useContext<T>(context: PreactContext<T>): T;
 
+/**
+ * Preact implementation of React's use hook
```
rschristian: Nit: This is self-evident & unneeded

hooks/src/index.js
```diff
@@ -164,6 +164,101 @@ function getHookState(index, type) {
 	return hooks._list[index];
 }
 
+const PROMISE_CACHE = new WeakMap();
+/**
+ * Preact implementation of React's use hook
```
rschristian: Same here

hooks/src/index.d.ts
```diff
@@ -125,6 +125,12 @@ export function useMemo<T>(factory: () => T, inputs: Inputs | undefined): T;
...
+ * Supports Promise and Context consumption
+ */
+export function use<T>(usable: Promise<T> | PreactContext<T>): T;
```
rschristian: Nit: Matches React:

```suggestion
export function use<T>(resource: Promise<T> | PreactContext<T>): T;
```

hooks/src/index.js
```diff
@@ -164,6 +164,101 @@ function getHookState(index, type) {
...
+ * Supports Promise and Context consumption
+ *
+ * @template T
+ * @param {Promise<T> | import('preact').PreactContext<T>} usable
+ * @returns {T}
+ */
+export function use(usable) {
```
rschristian: Same as above w/ the param


```suggestion
export function use(resource) {
```

hooks/src/index.js
```diff
@@ -164,6 +164,101 @@ function getHookState(index, type) {
...
+ * @returns {T}
+ */
+function usePromise(promise) {
+	const [, forceUpdate] = useState(/** @type {object} */ ({}));
+
+	let promiseState = PROMISE_CACHE.get(promise);
+
+	if (!promiseState) {
+		promiseState = {
+			status: 'pending',
```
rschristian: `status` should be redundant here to my knowledge, you can instead omit `value` & `reason` to use `value in promiseState` or `reason in promiseState` to determine if the promise resolved or rejected, reducing bytes.

hooks/src/index.js
```diff
@@ -164,6 +164,101 @@ function getHookState(index, type) {
...
+		}
+
+		if ('$$typeof' in usable && usable.$$typeof === CONTEXT_TYPE) {
+			return useContext(
+				/** @type {import('./internal').PreactContext} */ (
+					/** @type {unknown} */ (usable)
+				)
+			);
+		}
+	}
```
rschristian: Not sure we need to be so defensive here with our conditionals, `if (usable.then) { /* handle promise */ } else { /* handle context */ }` should pretty much suffice.

rschristian (commented):
Thanks for taking a swing at implementing this! Few comments though, mostly minor (in addition to what Jovi's already said):

hooks/src/index.js
```diff
@@ -164,6 +164,101 @@ function getHookState(index, type) {
...
+		case 'fulfilled':
+			return promiseState.value;
+		case 'rejected':
+			throw promiseState.reason;
+		case 'pending':
+			throw promise;
+		default:
+			throw promise;
+	}
+}
```
JoviDeCroock:
```suggestion
	const [promiseState, update] = useState({
			value: undefined,
			reason: undefined,
	});

	if (!promiseState.value && !promiseState.reason) {
		promise.then(
			value => {
				update({ status: 'fulfilled', value })
			},
			reason => {
				update({ status: 'rejected', reason })
			}
		);
	}

    if (promiseState.value) return promiseState.value;
    if (promiseState.reason) throw promiseState.reason;
    
    throw promise;
}
```

hooks/src/index.js
JoviDeCroock:
```suggestion
```

hooks/src/index.js
```diff
@@ -164,6 +164,101 @@ function getHookState(index, type) {
...
+		if ('$$typeof' in usable && usable.$$typeof === CONTEXT_TYPE) {
+			return useContext(
+				/** @type {import('./internal').PreactContext} */ (
+					/** @type {unknown} */ (usable)
+				)
+			);
+		}
+	}
+
+	throw new Error(`An unsupported type was passed to use(): ${usable}`);
```
JoviDeCroock:
```suggestion
  if (usable && typeof usable.then === 'function') {
	  return usePromise(usable);
  }

	return useContext(
		/** @type {import('./internal').PreactContext} */ (
			/** @type {unknown} */ (usable)
		)
	);
```

JoviDeCroock:
As I tried to make clear in my suggestions/comments, the philosophy of preact is to optimize for performance and size. We should also not leak implementation details into core for the sake of it.

EDIT: thinking more about the `PROMISE_CACHE`, I think I know what you were going for with persisting across discarded renders in relation to the missing lifecycle problem. As we assume the Promise identity to be stable we could hang the value off of that and re-hydrate it though so I think we are safe there. It would have been easier to realise what you were going for with a description for the PR.

hooks/src/index.js
rschristian: Late here but we don't want `if (promiseState.value) return promiseState.value;`, promises should be able to resolve to falsy values

rschristian (commented):
I'm not sure we can get away with using `useContext` or `useState`, both fundamentally break the promise of `use` in that it can be conditionally called.

Huxpro:
Second to @rschristian. `use` is intentionally designed to be semantically order-insensitive (it's not a real "Hook" in the sense that it doesn't allocate a hook node/slot in the Fiber), so we shouldn't implement it on top of existing order-sensitive hooks.

PR #4979, opened 2025-12-20
feat: implement getServerSnapshot for useSyncExternalStore

### LINKED ISSUE #4972: Support for the third argument of useSyncExternalStore
A clear and concise description of what you'd love to see added to Preact.

By supporting `getServerSnapshot`, we improve compatibility with React, enabling detection of hydration steps like:
https://github.com/TanStack/router/blob/425626822255604efc9a0598c58baeebba5fb592/packages/react-router/src/ClientOnly.tsx#L44-L72

* https://react.dev/reference/react/useSyncExternalStore#reference

ssssota:
close #4972


compat/src/hooks.js
```diff
@@ -1,12 +1,35 @@
...
  */
-export function useSyncExternalStore(subscribe, getSnapshot) {
-	const value = getSnapshot();
+export function useSyncExternalStore(
+	subscribe,
+	getSnapshot,
+	getServerSnapshot
+) {
+	const value =
+		typeof window === 'undefined' || hydrating
```
ssssota: Is there better logic to detect SSR?

compat/test/browser/useSyncExternalStore.test.jsx
```diff
@@ -781,6 +782,46 @@ describe('useSyncExternalStore', () => {
 			expect(container.textContent).to.equal('NaN');
 		});
 
+		it('basic server hydration', async () => {
```
ssssota, line 810: test from: https://github.com/facebook/react/blob/3e09c27b880e1fecdb1eca5db510ecce37ea6be2/packages/use-sync-external-store/src/__tests__/useSyncExternalStoreShared-test.js#L732-L782

compat/src/hooks.js
rschristian: React uses this:

https://github.com/facebook/react/blob/65eec428c40d542d4d5a9c1af5c3f406aecf3440/packages/shared/ExecutionEnvironment.js#L10-L14

...but I think what you have is absolutely fine & what we'd want.

compat/src/hooks.js
```diff
@@ -52,6 +75,10 @@ function didSnapshotChange(inst) {
 	}
 }
 
+function missingGetServerSnapshot() {
+	throw new Error('Missing getServerSnapshot for server rendering');
+}
```
rschristian: Throwing is okay for Preact v11 (the `main` branch, which this PR is targeting), but for v10, we'll probably want to use `console.warn` to avoid breakages.

I'd also change the error message to something like this, for clarity:

```suggestion
function missingGetServerSnapshot() {
	throw new Error('Missing "getServerSnapshot" parameter for "useSyncExternalStore", this is required for server rendering & hydration of server-rendered content`);
}
```

compat/src/hooks.js
```diff
@@ -1,12 +1,35 @@
 import { useState, useLayoutEffect, useEffect } from 'preact/hooks';
+import { MODE_HYDRATE } from '../../src/constants';
```
rschristian: I don't think this will work outside our test suite, you'd need to copy/paste `MODE_HYDRATE` into compat somewhere as it won't be available at runtime.

I understand copy/pasting magic values is kinda gross, but it's what we do elsewhere when needed. For example, [in `preact-iso`](https://github.com/preactjs/preact-iso/blob/195d5a7d5536d9f74ed7deb22050cd8879988fd2/src/router.js#L288-L289).

compat/src/hooks.js
ssssota: fixed here!
f1f096a

compat/src/hooks.js
ssssota: fixed!
7c07af8

compat/src/hooks.js
```diff
@@ -1,12 +1,36 @@
...
  */
-export function useSyncExternalStore(subscribe, getSnapshot) {
-	const value = getSnapshot();
+export function useSyncExternalStore(
+	subscribe,
+	getSnapshot,
+	getServerSnapshot
+) {
+	const value =
+		typeof window === 'undefined' || hydrating
```
JoviDeCroock: This or we use `options._skipEffects`

compat/src/hooks.js
```diff
@@ -1,12 +1,36 @@
...
+			? getServerSnapshot
+				? getServerSnapshot()
+				: missingGetServerSnapshot()
```
JoviDeCroock: Let's not add oversized error strings, this seems a bit against our bundle philosophy

compat/src/hooks.js
```diff
@@ -1,12 +1,36 @@
...
+const MODE_HYDRATE = 1 << 5;
+
+/** @type {boolean} */
+let hydrating;
+// Cast to use internal Options type
+const options = /** @type {import('../../src/internal').Options} */ (_options);
+let oldBeforeRender = options._render;
+
+/** @type {(vnode: import('./internal').VNode) => void} */
+options._render = _vnode => {
```
JoviDeCroock: If we move this to the `render.js` file you can benefit from `options._render` there and save more bytes

JoviDeCroock (commented):
The general trend of this looks good, left some comments on how to reduce bundle size. Let's get this into v11

compat/src/hooks.js
rschristian: That (the oversized string) was my suggestion FWIW, my bad

ssssota:
@JoviDeCroock Thank you for the review! I've made the changes, so please take a look.

JoviDeCroock:
I meant move the whole hook over 😅 this made the byte size regress

compat/src/render.js
```diff
@@ -18,21 +18,75 @@ import {
...
+ * @typedef {{ _value: any; _getSnapshot: () => any }} Store
+ */
+export function useSyncExternalStore(
+	subscribe,
+	getSnapshot,
+	getServerSnapshot
+) {
+	const serverRendering = options._skipEffects || hydrationRoot;
+	if (serverRendering && !getServerSnapshot) {
+		throw new Error('Missing getServerSnapshot');
```
JoviDeCroock: Remove the error all together imho, we can just fall back to regular snapshot.

JoviDeCroock (commented):
Let's add tests in the suspense-hydration case whether this works correctly. When a resumed boundary continues hydration we expect components there to correctly follow the server snapshot

compat/src/hooks.js
rschristian: Having `hooks.js` only contain some hooks, not all, is kinda weird... what was the byte difference there, if you remember? Should we even keep this module around, or should it be inlined into `render`?

JoviDeCroock: When I tried it locally it was 32b gzip

rschristian: Roughly a third then, hefty.

compat/src/render.js
ssssota: I will make that change as suggested!
Please note that React throws an error in this case.

> If you omit this argument, rendering the component on the server will throw an error.

_[useSyncExternalStore – React](https://react.dev/reference/react/useSyncExternalStore#:~:text=If%20you%20omit%20this%20argument%2C%20rendering%20the%20component%20on%20the%20server%20will%20throw%20an%20error.)_

JoviDeCroock (approved):
Thank you for implementing this!

PR #4981, opened 2025-12-27
types: Correct outdated imports in `jsx-runtime`

### LINKED ISSUE #4980: stale JSXInternal.HTMLAttributes and JSXInternal.SVGAttributes references in 11.0.0-beta.0

**Describe the bug**
It looks like the index.d.ts file within preact/jsx-runtime/src has stale references to JSXInternal.HTMLAttributes and JSXInternal.SVGAttributes in 11.0.0-beta.0 causing TS to throw errors during build.

Those two types should be imported directly from 'preact' instead.

**To Reproduce**

Load the index.d.ts file in a TS environment and see the errors.

**Expected behavior**

File should be updated to reflect the latest type definitions.

rschristian:
Closes #4980

The actual fix was straightforward, only touching `/jsx-runtime/src/index.d.ts`, but the linting fix was more involved.

Whilst we have a linter that should've caught that problem, turns out we had `skipLibCheck: true` enabled -- this skips checking for source `.d.ts` files as well as those in `node_modules`. Removing this immediately caused issues however as `vite` and `vitest` require more modern TS setups.

As such, the `jsconfig-lint.json` got a bit of a facelift to make it more modern & compatible, whilst also removing the `skipLibCheck` setting so that we now get warned of type errors again. Yay.

Re: `compat/server.d.ts`, no idea what to do about that one. TS is grumpy about the mixed exports (rightfully so) and I spent an hour trying various ways to get around it with no luck. It works at runtime for consumers, so I just silenced the errors. We know that's a problem with our types anyhow (ESM-only anyone??)

### DIFF
--- a/jsx-runtime/src/index.d.ts
+++ b/jsx-runtime/src/index.d.ts
@@ -10,6 +10,7 @@ import {
 	HTMLAttributes,
 	SVGAttributes
 } from 'preact';
+import { JSXInternal } from '../../src/jsx';
 
 export function jsx(
 	type: string,
@@ -59,3 +60,5 @@ export function jsxAttr(name: string, value: any): string | null;
 export function jsxEscape<T>(
 	value: T
 ): string | null | VNode<any> | Array<string | null | VNode>;
+
+export { JSXInternal as JSX };


types/events.d.ts
```diff
@@ -1,58 +0,0 @@
-/**
```
rschristian, line 1: [Was added for Vitest](https://github.com/preactjs/preact/pull/4687) but conflicts with our `@types/node` version and seems unnecessary now. I imagine this was a holdover until we had bumped `@types/node` and just forgot to clean it up.

types/weak-key.d.ts
```diff
@@ -0,0 +1,9 @@
+declare global {
+	export interface WeakKeyTypes {
+		object: object;
+	}
+
+	export type WeakKey = WeakKeyTypes[keyof WeakKeyTypes];
```
rschristian, line 6: Vitest relies on it, copied from [`es5.d.ts`](https://github.com/microsoft/TypeScript/blob/1f5f9f34d325b2c7e3a53179108b0a217e2dae93/src/lib/es5.d.ts#L1669-L1676)

JoviDeCroock (approved):

PR #4987, opened 2026-01-06
10.27 strict equality constructor

JoviDeCroock:


rschristian:
~~Patch to 10.26 & 10.27 doesn't sound good, did you find something particularly problematic?~~ Answered

PR #4997, opened 2026-01-10
types: Separate camelCase SVG attributes to compat only (#4994)

### LINKED ISSUE #4994: JSX types reflect core + compat, when compat may or may not be used

Stemming from a Twitter thread (TODO: go grab link), it's come up a couple times that Preact's JSX types represent a union of core & compat when the latter isn't guaranteed to be used. For example, `strokeWidth`, an SVG prop, is a React invention and we only support it when compat is used, but it's not a type error for those just using core.

This makes it really easy to run into frustrating bugs as the types say properties are supported when they're really not from core alone.

I banged my head on this for an hour and got no where, TS namespace and declaration merging is a land of misery, pain, and abysmal documentation quality. Tossing this here though as it's a valid issue, even if I can't address it.

gayanMatch:
## Summary
  - Move React-style camelCase SVG attributes (e.g., `strokeWidth`, `fillOpacity`) from core to compat only
  - Core Preact now only exposes standard kebab-case SVG attributes (e.g., `stroke-width`, `fill-opacity`)
  - Add `CamelCaseSVGAttributes` interface in compat for React compatibility
  - Create `CompatJSX` namespace that extends core JSX with camelCase SVG support

  ## Test plan
  - [x] Core TypeScript tests verify camelCase SVG props cause type errors
  - [x] Compat TypeScript tests verify camelCase SVG props work correctly
  - [x] `npm run test:ts` passes

  Closes #4994 
---

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=132382032


compat/src/svg-camel-case.d.ts
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * React-compatible camelCase SVG attributes.
+ * These are only supported when using preact/compat, which converts
+ * camelCase attribute names to their kebab-case equivalents at runtime.
+ *
+ * Core Preact does not support these - use the kebab-case versions instead
+ * (e.g., 'stroke-width' instead of 'strokeWidth').
+ */
+export interface CamelCaseSVGAttributes {
```
rschristian: Can we rename this file to `dom.d.ts`, matching what we do in core, and perhaps rename this to `CompatSVGAttributes`? The comment will also need to be adjusted as it's incorrect.

Whilst React does camel case SVG attributes, the comment & interface name aren't quite correct as core _does in fact_ support many camel-cased SVG properties as the spec itself support them. Furthermore, not all of the properties that React camel-cases are kebab-cased in the spec, some, like `hreflang`, are all lowercase. The SVG spec for many years lacked some quality control mechanisms and now the standards are unfortunately rather inconsistent.

compat/src/index.d.ts
```diff
@@ -102,9 +103,40 @@ declare namespace preact {
...
+	export type IntrinsicMathMLElements = JSXInternal.IntrinsicMathMLElements;
+
+	export type IntrinsicHTMLElements = JSXInternal.IntrinsicHTMLElements;
+
+	export interface IntrinsicElements
+		extends IntrinsicSVGElements,
+			IntrinsicMathMLElements,
+			IntrinsicHTMLElements {}
+}
+
```
rschristian: Mind moving this into a `jsx.d.ts` module? Mirrors what we do in core, and just makes it a bit easier to track.

compat/src/index.d.ts
```diff
@@ -175,14 +207,81 @@ declare namespace React {
...
+			preact1.ClassAttributes<T>,
+			CamelCaseSVGAttributes {}
 
 	interface SVGAttributes<T extends EventTarget = SVGElement>
-		extends preact1.SVGAttributes<T> {}
+		extends preact1.SVGAttributes<T>,
+			CamelCaseSVGAttributes {}
+
+	// Compat IntrinsicSVGElements uses SVGAttributes with camelCase properties
+	interface CompatIntrinsicSVGElements {
```
rschristian: Can this also be moved into the new compat JSX namespace, matching what we do in core?

test/ts/preact.tsx
```diff
@@ -419,16 +419,37 @@ createElement<InputHTMLAttributes>('input', {
...
+);
+
+// More camelCase SVG attributes that should fail in core
+// @ts-expect-error fillOpacity is a React-ism
+const fillOpacityTest = <rect fillOpacity="0.5" />;
+// @ts-expect-error stopColor is a React-ism
+const stopColorTest = <stop stopColor="red" />;
+// @ts-expect-error fontFamily is a React-ism
+const fontFamilyTest = <text fontFamily="Arial" />;
+
```
rschristian: nit: Can this move to `test/ts/dom-attributes.test-d.tsx`? Ideally `preact.tsx` would be more general framework-y things, not necessarily standard attributes/properties.

package-lock.json
```diff
@@ -2775,149 +2775,6 @@
       "integrity": "sha512-wwQAWhWSuHaag8c4q/KN/vCoeOJYshAIvMQwD4GpSb3OiZklFfvAgmj0VCBBImRpuF/aFgIRzllXlVX93Jevww==",
       "dev": true
     },
-    "node_modules/@promptbook/utils": {
```
rschristian, line 2778: This whole file needs to be reverted

compat/test/ts/index.tsx
```diff
@@ -40,3 +40,25 @@ React.render(
...
+		<circle fill="blue" stroke-width="2" cx="24" cy="24" r="20" />
+	</svg>
+);
+
+// More camelCase SVG attributes that should work in compat
+const fillOpacityTest = <rect fillOpacity="0.5" />;
+const stopColorTest = <stop stopColor="red" />;
+const fontFamilyTest = <text fontFamily="Arial" />;
+const strokeDasharrayTest = <path strokeDasharray="5,5" />;
+const textAnchorTest = <text textAnchor="middle" />;
```
rschristian: Similar to the other comment, but could this go into new `compat/test/ts/dom-attributes.test-d.tsx` file?

gayanMatch:
Hi @rschristian ,
I updated PR with your feedback. Please review it.
The changes I made are
1. Renamed compat/src/svg-camel-case.d.ts to compat/src/dom.d.ts - matching core's pattern, and renamed the interface from CamelCaseSVGAttributes to CompatSVGAttributes with an updated comment
  2. Created compat/src/jsx.d.ts - moved the CompatJSX namespace here with CompatIntrinsicSVGElements inside it, mirroring what core does with src/jsx.d.ts
  3. Updated compat/src/index.d.ts - imports from the new dom.d.ts and jsx.d.ts files, removed the inline namespace definitions
  4. Moved SVG tests from test/ts/preact.tsx to test/ts/dom-attributes.test-d.tsx - core SVG attribute tests (which verify camelCase attributes are NOT available in core)
  5. Reverted package-lock.json - removed accidental changes
  6. Created compat/test/ts/dom-attributes.test-d.tsx - moved compat SVG tests here from compat/test/ts/index.tsx (which verify camelCase attributes ARE available in compat)

rschristian:
I gave this a test by installing into a separate project and it doesn't seem to be functioning correctly, the JSX namespace is not updated when `preact/compat` is used:

```jsx
// When the following line is uncommented, the type error should disappear
// import 'preact/compat';

const fillOpacityTest = <rect fillOpacity="0.5" />;
```

The `tsconfig.json` within `compat/test/ts` has `"jsx": "react"` set, which is probably hiding this problem (and, IMO, we shouldn't be doing, but that's a separate issue).

gayanMatch:
Hi @rschristian ,
Changeds made:
1. compat/src/index.d.ts - Added global JSX namespace augmentation so import 'preact/compat' updates the JSX types
  2. compat/jsx-runtime.d.ts - New file that exports CompatJSX as JSX for jsxImportSource: "preact/compat"
  3. compat/jsx-dev-runtime.d.ts - Same for dev runtime

  These changes ensure the JSX namespace is properly updated in both scenarios:
  - Using jsxImportSource: "preact/compat"
  - Using import 'preact/compat' with classic JSX transform

rschristian:
Thanks! I'll have to test yet, but can you explain the need for the `jsx-runtime.d.ts` & `jsx-dev-runtime.d.ts` files? Not quite sure I understand why they'd be needed. Maybe if a third-party React library only used the jsx-runtime, so that's the only thing that'd get aliased perhaps?

gayanMatch:
The jsx-runtime.d.ts files would be needed specifically when:

  1. jsxImportSource: "preact/compat" - If someone explicitly sets this in their tsconfig, TypeScript looks for the JSX export from preact/compat/jsx-runtime for type checking (not the global namespace).
  2. Aliasing react/jsx-runtime → preact/compat/jsx-runtime - Your hypothesis is correct. If a bundler aliases the jsx-runtime specifically to compat's version, the types need to come from there.

If you want to test without them first and only add if needed, that's reasonable. The global augmentation might be sufficient if:
  - Most users use jsxImportSource: "preact" (not "preact/compat")
  - The compat import augments the global JSX namespace which TypeScript falls back to.
  
Do you want me to remove them and just keep the global augmentation?

compat/jsx-runtime.d.ts
```diff
@@ -0,0 +1,12 @@
+export {
```
rschristian, line 1: The `"types"` entry for `preact/compat/jsx-runtime` & `preact/compat/jsx-dev-runtime` will need to be updated to match this path:

https://github.com/preactjs/preact/blob/46772a2fb68c54725ae3c0cae350dac23a842b1c/package.json#L91

https://github.com/preactjs/preact/blob/46772a2fb68c54725ae3c0cae350dac23a842b1c/package.json#L96

compat/jsx-dev-runtime.d.ts
```diff
@@ -0,0 +1,12 @@
+export {
```
rschristian: This file is unnecessary, we'll just use the same types file for both.

gayanMatch:
- Removed jsx-dev-runtime.d.ts (using same types file for both runtimes)
- Updated package.json types entries for compat/jsx-runtime and compat/jsx-dev-runtime to point to ./compat/jsx-runtime.d.ts

rschristian:
Did you try using this?

```js
// When the following line is uncommented, the type error should disappear
// import 'preact/compat';

const fillOpacityTest = <rect fillOpacity="0.5" />;
```

Perhaps I'm being daft, but doesn't seem to work as expected, even with the latest changes.

PR #4998, opened 2026-01-19
refactor: ESM-only

rschristian:
Not touching Microbundle here, that can be done separately & later. That'll likely come with other implications that I don't want to delve into.


package-lock.json
```diff
@@ -9793,30 +9792,6 @@
...
-      "resolved": "https://registry.npmjs.org/preact/-/preact-10.10.6.tgz",
-      "integrity": "sha512-w0mCL5vICUAZrh1DuHEdOWBjxdO62lvcO++jbzr8UhhYcTbFkpegLH9XX+7MadjTl/y0feoqwQ/zAnzkc/EGog==",
-      "dev": true,
-      "peer": true,
-      "funding": {
-        "type": "opencollective",
-        "url": "https://opencollective.com/preact"
-      }
-    },
-    "node_modules/preact-render-to-string": {
```
rschristian: TODO: I don't think we're handling this right at the moment & this is just an immediate fix. It really should be an optional peer dep

rschristian:
Hrm, we're apparently using an older version of NPM that has some differences w/ the lockfile. Can't do much about that, I don't have access to any older copies of NPM.

rschristian:
If anyone else wants to work on this, feel free. I'm not going to deal with Vitest as I can't get it to run locally & I don't have the patience to debug our test suite via CI.

Decodetalkers:
This ci failed because the target wrong, not mjs, but esm.js, and the target of preact.mjs becomes preact.esm.esm.js, I do not know why esm repeat twice

rschristian:
Vitest failing doesn't matter too much here, not being able to correctly type `/compat` is the real issue.

We'd probably swap out Microbundle which is the cause of the module naming issues

Decodetalkers:
> Vitest failing doesn't matter too much here, not being able to correctly type `/compat` is the real issue.
> 
> We'd probably swap out Microbundle which is the cause of the module naming issues

I have fixed it , can I send you pr? thanks. but I have rebase this branch, seems it cannot be merged to the esm-only branch

Decodetalkers:
Have you read my comment in the origin pr? I think I have explain clearly, and if you let my pr run the ci, ti will be all green. 

I also have upload the pictures. Can you see it for once a time?

ok, I see, the right way is to set the target name.. I see

but you also need to correct the target in 

https://github.com/preactjs/preact/blob/a7ed5b84a6e31b73daffdd8988f642173fce6679/vitest.config.js#L22

now there is no compat.mjs now

package.json
```diff
@@ -119,17 +85,14 @@
...
     "prepare": "husky && npm run test:install && run-s build",
     "build": "npm-run-all --parallel 'build:*'",
-    "build:core": "microbundle build --raw --no-generateTypes -f cjs,esm,umd",
-    "build:debug": "microbundle build --raw --no-generateTypes -f cjs,esm,umd --cwd debug",
-    "build:devtools": "microbundle build --raw --no-generateTypes -f cjs,esm,umd --cwd devtools",
-    "build:hooks": "microbundle build --raw --no-generateTypes -f cjs,esm,umd --cwd hooks",
-    "build:test-utils": "microbundle build --raw --no-generateTypes -f cjs,esm,umd --cwd test-utils",
-    "build:compat": "microbundle build --raw --no-generateTypes -f cjs,esm,umd --cwd compat --globals 'preact/hooks=preactHooks'",
-    "build:jsx": "microbundle build --raw --no-generateTypes -f cjs,esm,umd --cwd jsx-runtime",
+    "build:core": "microbundle build --raw --no-generateTypes -f esm -o dist/preact.js",
```
Decodetalkers, line 88: if you set the target, then why you set the main to preact.esm.js? it does not exist

Decodetalkers: ok, sorry, I thought I can be so some help.

rschristian: Because it will exist. `-o`, unfortunately, does not fully control anything but CJS output names, Microbundle will still append `.esm`. Long-standing oddity that I attempted to correct years ago: https://github.com/developit/microbundle/pull/896

Decodetalkers: Thank you for explaining it for me. I seems have too much confidence.. sorry for that

rschristian:
Closing, Jovi landed something quite comparable (but, ya know, actually working) in #5159

PR #4999, opened 2026-01-19
Resolve some suspense crashes

JoviDeCroock:
In the first commit we are resolving a suspense crash that happens when we do a state update on a component that is in a suspended state which has been unmounted.

In the second commit we resolve a forced remount of an unmounted component which resolves after unmount.

RE the second commit

Before this when we unmounted a vnode which was in a suspended state it was possible to cause this to render while being unmounted.
One disturbing thing is that I wanted to check `vnode._component._onResolve` in the `options.unmount` hook, this however was not possible because somehow the reference to `_component` isn't being updated on the `vnode` (my assumption is it being related to us cloning the vnode during the suspension) that's actually unmounting. For this reason I opted to use a separate flag.

### DIFF
--- a/compat/test/browser/suspense.test.js
+++ b/compat/test/browser/suspense.test.js
@@ -14,6 +14,7 @@ import React, {
 } from 'preact/compat';
 import { setupScratch, teardown } from '../../../test/_util/helpers';
 import { createLazy, createSuspender } from './suspense-utils';
+import { expect } from 'chai';
 
 const h = React.createElement;
 /* eslint-env browser, mocha */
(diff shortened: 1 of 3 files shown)


compat/test/browser/suspense.test.js
```diff
@@ -2188,6 +2188,174 @@ describe('suspense', () => {
 		});
 	});
 
+	it('should not crash when suspended child updates after unmount', () => {
```
JoviDeCroock, line 2192: Uncommenting line 240 will surface the crash of this

compat/test/browser/suspense.test.js
```diff
@@ -14,6 +14,7 @@ import React, {
 } from 'preact/compat';
 import { setupScratch, teardown } from '../../../test/_util/helpers';
 import { createLazy, createSuspender } from './suspense-utils';
+import { expect } from 'chai';
```
rschristian, line 17: Maybe nit: is this needed?

JoviDeCroock: Whups, darn vscode auto-import

PR #5003, opened 2026-01-23
refactor(props): call toLowerCase on name regardless of its current casing

ali-garajian:
if benchmarks don't get negatively affected, we could save some bytes and perhaps even improve performance a bit


src/diff/props.js
```diff
@@ -67,8 +67,7 @@ export function setProperty(dom, name, value, oldValue, namespace) {
 		useCapture = name != (name = name.replace(CAPTURE_REGEX, '$1'));
 
 		// Infer correct casing for DOM built-in events:
```
rschristian, line 69: Not something that needs to be done here, but we probably should do away with this comment now that we're not inferring anything (and instead just assuming the vent matches DOM event conventions).

PR #5005, opened 2026-01-29
Avoid double scheduling

JoviDeCroock:
People overriding options.debounceRendering risk double scheduling an update when the first update happens before prevDebounce is set

Resolves https://github.com/preactjs/prefresh/issues/597

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -145,7 +145,8 @@ function renderComponent(component) {
 		);
 
 		newVNode._original = oldVNode._original;
-		newVNode._parent._children[newVNode._index] = newVNode;
+		if (newVNode._parent)
+			newVNode._parent._children[newVNode._index] = newVNode;
 		commitRoot(commitQueue, newVNode, refQueue);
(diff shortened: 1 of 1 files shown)


src/component.js
```diff
@@ -240,10 +239,17 @@ function process() {
 		l = rerenderQueue.length;
 
 		if (c._dirty) {
-			renderComponent(c);
+			try {
+				renderComponent(c);
+			} catch (firstError) {
+				// unrecoverable errors in rendering should not prevent
+				// further renders from completing
+				e = firstError;
```
JoviDeCroock: The alternative is that we just do `finally { process._rerenderCount = 0; }` and unset `_dirty` on all of the `rerenderQueue` members (and reset the queue)

marvinhagemeister: Since this is inside a loop, does it mean `e` will always be overwritten?

JoviDeCroock: Good point, let's do `if (!e)`

PR #5010, opened 2026-02-02
refactor: Revert back to normal TS resolution for core types

rschristian:
We had originally done this to support the type version splitting in v10.x:

https://github.com/preactjs/preact/blob/eb1b8c87b6dd71649f4237a33d07180911f387d0/package.json#L19-L21

No longer doing this in v11, so it should be dropped IMO.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,10 +1,10 @@
 # Contributing
 
-This document is intended for developers interest in making contributions to Preact and document our internal processes like releasing a new version.
+This document is intended for developers interested in making contributions to Preact and documents our internal processes like releasing a new version.
 
 ## Getting Started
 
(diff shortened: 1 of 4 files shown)


compat/src/index.d.ts
```diff
@@ -1,25 +1,23 @@
-import * as _hooks from '../../hooks';
-// Intentionally not using a relative path to take advantage of
-// the TS version resolution mechanism
-import * as preact1 from 'preact';
+import * as _preact from '../../src/index';
```
rschristian, line 1: I'm fairly certain this was my doing. `preact1` is dumb as hell, especially sat next to `_hooks` & `_Suspense`. Not sure why in the world I named it like that.

Apologies for the big diff, but I figure it makes sense to correct now & is a very boring find-and-replace change anyhow.

Issue #5002, opened 2026-01-21
Report error when error boundary's `componentDidCatch` do not call `setState`


upupming:
**Describe the feature you'd love to see**

When using `componentDidCatch`, users might forget to run `this.setState` to set a new state. In this case, preact will throw an error at `process` function and fail to set [`process._rerenderCount = 0`](https://github.com/preactjs/preact/blob/43178581442fa0f2428e5bdbca355860b2d12e5d/src/component.js#L246), which will [prevent future `enqueueRender` working properly](https://github.com/preactjs/preact/blob/43178581442fa0f2428e5bdbca355860b2d12e5d/src/component.js#L207).

I have created two example cases: [Preact](https://stackblitz.com/edit/vitejs-vite-ps2fryp2?file=src%2Fapp.tsx), [React](https://stackblitz.com/edit/vitejs-vite-4qhnvukc?file=src%2FApp.tsx). In this example, when we click the counter up to 3, it will throw an error, and the parent error boundary will not `setState` in their `componentDidCatch`.

For React, it will remind user to call `setState` properly, see their impl at [here](https://github.com/facebook/react/blob/d29087523a09d2babff2ce258f3527944a0ecd2e/packages/react-reconciler/src/ReactFiberThrow.js#L186-L196):

<img width="1093" height="270" alt="Image" src="https://github.com/user-attachments/assets/91b985b4-25b6-4af6-bc71-60c6c5db5c61" />

For Preact, it will sliently break and the page will be not interactivable any more.

So I would suggest we can add an error message on this circumstances. I'd be happy to submit a PR to support this; please let me know if you have any questions or concerns.

**Additional context (optional)**
N/A

JoviDeCroock:
In Preact 11 the render queue was reworked with a `try/finally` reset and a `handled` flag, so an error boundary whose `componentDidCatch` does not call `setState` no longer silently bricks rendering — the error is re-thrown instead of being swallowed. The remaining ask (a dev-mode warning when `setState` is not called) is minor; closing this. Happy to take a small PR against `preact/debug` if someone wants the explicit warning.

PR #5017, opened 2026-02-10
Ensure we reset renderCount

JoviDeCroock:
~~Currently due to the condition of `prevDebounce != options.debounceRendering` we leak an implementation detail of `act` in our general implementation. We don't want this so we're just exposing `process` on `options` to avoid us having to account for this in the future. When our consumers override `options.debounceRendering` they might accidentally introduce more renders than intended when is about to be processed.~~

The above did not work out, we should encourage people to override `options.debounceRendering` before scheduling any preact updates which should entirely bypass the possiblity of this

Another issue we had was that if `process()` would throw without any resolution in the component tree that `process._rerenderCount` would never return back to 0 and hence stop any future render from being scheduled. This is handled in the `try finally` block.

I've also added a byte saving opportunity with pushing the `component._dirty` check down a bit.

Resolves https://github.com/preactjs/preact/issues/5002

### DIFF
--- a/test/browser/components.test.js
+++ b/test/browser/components.test.js
@@ -1,4 +1,4 @@
-import { createElement, render, Component, Fragment } from 'preact';
+import { createElement, render, Component, Fragment, options } from 'preact';
 import { setupRerender } from 'preact/test-utils';
 import {
 	setupScratch,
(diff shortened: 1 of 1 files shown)


src/component.js
```diff
@@ -220,30 +219,31 @@ const depthSort = (a, b) => a._vnode._depth - b._vnode._depth;
...
-		c = rerenderQueue.shift();
-		l = rerenderQueue.length;
+			c = rerenderQueue.shift();
+			l = rerenderQueue.length;
 
-		if (c._dirty) {
 			renderComponent(c);
 		}
+	} finally {
+		process._rerenderCount = 0;
```
JoviDeCroock:
```suggestion
		rerenderQueue.length = process._rerenderCount = 0;
```

src/component.js
```diff
@@ -217,28 +217,31 @@ const depthSort = (a, b) => a._vnode._depth - b._vnode._depth;
...
-		c = rerenderQueue.shift();
-		l = rerenderQueue.length;
+			c = rerenderQueue.shift();
+			l = rerenderQueue.length;
 
-		renderComponent(c);
+			renderComponent(c);
+		}
+	} finally {
+		rerenderQueue.length = process._rerenderCount = 0;
```
JoviDeCroock, line 243: This is up for debate, we could also just wrap `renderComponent` in this `try {} catch {}` and throw the first error so that we do process the rest of the queue

PR #5022, opened 2026-02-12
Core size optimizations

JoviDeCroock:


src/diff/index.js
```diff
@@ -227,9 +225,7 @@ export function diff(
 
 				tmp = c.render(c.props, c.state, c.context);
 
-				for (let i = 0; i < c._stateCallbacks.length; i++) {
-					c._renderCallbacks.push(c._stateCallbacks[i]);
-				}
+				c._renderCallbacks.push(...c._stateCallbacks);
```
developit: This seems expensive

rschristian: Ran a couple quick tests, (the transpiled output) seems to actually be marginally faster in Chrome & only slightly slower in FF. Not sure about Safari.

Of course it doesn't stay as a spread, actually becomes `.push.apply()`:

```js
($ = h.__h).push.apply($, h._sb)
```

JoviDeCroock: In all of my benchmarks it was better, the larger the array the faster it is. The size optimisation feels so worth it to me

developit: Oh I didn't realize it would be transpiled in our config, thanks @rschristian   

Yeah I'm that case I'm totally good with this.

developit: One option maybe worth checking the size of given the transpiled output:

```
EMPTY_ARR.push.apply(c._renderCallbacks, c._stateCallbacks);
```

I suspect that may be smaller and faster than the transpiled spread if used everywhere, since it has more repetition and avoids a property read.

Issue #3929, opened 2023-03-09
flushSync implementation needed for new View Transitions API


sschultze:
**Describe the feature you'd love to see**

The [documentation of the View Transitions API](https://developer.chrome.com/docs/web-platform/view-transitions/) available in Chrome 111+ has a section [Working with frameworks](https://developer.chrome.com/docs/web-platform/view-transitions/#working-with-frameworks) which mentions React's `flushSync` function as a key to use it together with the View Transitions API.

Because when using `startViewTransition`, you have to ensure that the DOM is completely updated, which is what `flushSync` does.

Preact's implementation of `flushSync` however is currently just a no-op.

So my suggestion is to implement `flushSync` so that it ensures that the DOM is updated completely [like described here](https://beta.reactjs.org/reference/react-dom/flushSync).

sschultze:
BTW, the comment of the no-op "implementation" in Preact says that

> In React, `flushSync` flushes the entire tree and forces a rerender.

The latter one doesn't seem to be true, at least it is [not documented in the latest reference](https://beta.reactjs.org/reference/react-dom/flushSync) that `flushSync` forces a rerender, and from my point of view it wouldn't make sense to rerender everything to achieve the function's intended effect.

The React documentation however says:

> flushSync lets you force React to flush any updates inside the provided callback synchronously. This ensures that the DOM is updated immediately.

This makes a lot of sense in certain scenarios, like in the case described above (View Transitions API).

sschultze:
As commented in #3928, the flush of the queue actually doesn't have to be synchronous. The callback passed to `startViewTransition` can return a promise, [as described here](https://developer.chrome.com/docs/web-platform/view-transitions/#async-dom-updates-and-waiting-for-content).

But I think that there's a chance that 3rd party libraries will use React's `flushSync` with the View Transition API, so it's probably better to implement this rather than introducing another async function.

foxt:
@sschultze yeah, just tried to implement view transitions, and the problem is that, while the callback passed to SVD can return a promise, you seemingly need to block rendering until that callback is returned. for example, if you try and start a view transition in, for example, componentWillUpdate, and end it in componentDidUpdate, as componentWillUpdate will fire *before* the view transition before snapshot has been created and the callback has been called.

sschultze:
[...]

PR #5025, opened 2026-02-18
Golf down compat

JoviDeCroock:


rschristian:
Need to take a better look after work, but might also address #3929

JoviDeCroock:
A different implementation might be to do

```js
let flush;
options.debounceRendering = (fn) => {
  flush = fn;
  queueMicroTask(fn);
}

export function flushSync(cb) {
  let res = cb();
  flush();
  return res;
}
```

rschristian (commented):
I don't mean to be picky but this PR seems to have a considerable amount of miscellaneous golfing, it might even make up the bulk of the changes. If it's not too much of a hassle, I'd generally prefer that to be extracted out & separate.

As-is, it's bundle fairly different things and muddying up the diff.

JoviDeCroock:
I tried doing our old habit again of compensating for the bytes we added but I'll separate it out.

rschristian (approved):

PR #5026, opened 2026-02-18
Fix crash where a synchronous effect render unmounts the tree

JoviDeCroock:
## Summary

Prevent a crash when an effect unmounts its own vnode tree within an effect, this makes the currentComponent unmount and hence we can't reset the hooks array.


marvinhagemeister (approved):
LGTM

PR #5029, opened 2026-02-19
Maintain elements in head/body during hydration

JoviDeCroock:
Relates to https://github.com/preactjs/preact-render-to-string/pull/446

This is the same thing as what React 19 does to prevent browser extensions and existing styles getting clobbered by full document renders/hdyrates

https://react.dev/blog/2024/12/05/react-19

> In React 19, unexpected tags in the <head> and <body> will be skipped over, avoiding the mismatch errors. If React needs to re-render the entire document due to an unrelated hydration mismatch, it will leave in place stylesheets inserted by third-party scripts and browser extensions.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -557,7 +557,10 @@ function diffElementNodes(
 			// Remove children that are not part of any vnode.
 			if (excessDomChildren != NULL) {
 				for (i = excessDomChildren.length; i--; ) {
-					removeNode(excessDomChildren[i]);
+					const nodeName =
+						excessDomChildren[i] && excessDomChildren[i].localName;
+					if (nodeName != 'script' && nodeName != 'style' && nodeName != 'link')
+						removeNode(excessDomChildren[i]);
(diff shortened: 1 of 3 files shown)


rschristian (commented):
Isn't this a pretty big breaking change? I don't know that blowing away existing nodes is good behavior but to change in a minor seems really iffy as someone very well might be relying upon it.

This also doesn't really align with the title anymore if we're going to keep all `<head>` elements

JoviDeCroock:
That's true, the reason why we implement this is because React actually started doing it as well. When someone hydrates `document.body` or `document` for that matter we risk erasing all styles and extensions like 1Password. This only keeps elements on body/head when we are hydrating.

rschristian:
I get that, just seems risky to do outside of a major, though I can't say I'm a fan of users trying to do full-doc hydration renders anyhow.

Will approve regardless. Up to you

rschristian (approved):

PR #5031, opened 2026-02-21
fix(diff): handle undefined prototype in component type check

### LINKED ISSUE #5011: "TypeError: Cannot read properties of undefined (reading 'render')" when wrapping a functional component with a proxy that has an undefined prototype
- [X] Check if updating to the latest Preact version resolves the issue (tested with V11 beta)

**Describe the bug**

When a functional component is wrapped with a proxy that has an undefined prototype, executing the `render` method throws a TypeError:

`Cannot read properties of undefined (reading 'render')`

**To Reproduce**

* Create a functional component
* Wrap it with something that _has_ the `prototpe` property set to `undefined`
* Render it using `render`

One way to easily achieve this is to wrap the functional component with a Sinon spy:

```ts
import {JSDOM} from "jsdom";
import {spy} from "sinon";

globalThis.document = new JSDOM().window.document;

const spiedComponent = spy(() => {
  return 'foo';
});

render(createElement(spiedComponent , {}), document.body); // TypeError*
```

**Expected behavior**

The functional component is rendered.

**The culprit**

The following piece of code wrongly assumes that, if object that _has_ the `prototype` property, the `prototype` attribute is not `undefined`. It is wrong, per ECMAScript specification:

https://github.com/preactjs/preact/blob/6d29e2c76724ee9b643cfa152fd5b670c22ca51c/src/diff/index.js#L100

I can confirm that, be replacing `'prototype' in newType && newType.prototype.render;` with `'prototype' in newType && newType.prototype?.render;`, the error goes away. Actually, would the source code be written with a typed language like TypeScript, it would probably not compile.

TheAbMehta:
## Summary

Fixes #5011

When a functional component is wrapped with a proxy or object that has `prototype` as an own property but set to `undefined`, the expression `'prototype' in newType && newType.prototype.render` throws a `TypeError: Cannot read properties of undefined (reading 'render')`.

This happens because the `in` operator returns `true` (the property exists), but the value is `undefined`, so accessing `.render` on it crashes.

## Fix

Use optional chaining on the prototype access:

```diff
- 'prototype' in newType && newType.prototype.render
+ 'prototype' in newType && newType.prototype?.render
```

This safely returns `undefined` instead of throwing when `prototype` is `undefined`.

## Reproduction

This occurs in practice when using testing libraries like Sinon that create spies/stubs wrapping functional components, where the spy object has `prototype` set to `undefined`. A test has been added that reproduces this scenario using a `Proxy` with `prototype` returning `undefined`.

## Test

Added a test in `test/browser/components.test.jsx` that creates a functional component wrapped in a Proxy where `get(target, 'prototype')` returns `undefined`, verifying it renders without throwing.

## Disclosure

This PR was co-authored with the help of Claude (LLM) for identifying the fix, writing the test, and structuring the description.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -98,7 +98,7 @@ export function diff(
 				snapshot,
 				newProps = newVNode.props;
 			const isClassComponent =
-				'prototype' in newType && newType.prototype.render;
+				'prototype' in newType && newType.prototype?.render;
 
 			// Necessary for createContext api. Setting this property will pass
 			// the context value as `this.context` just for this component.
--- a/test/browser/components.test.jsx
+++ b/test/browser/components.test.jsx
@@ -82,6 +82,34 @@ describe('Components', () => {
 			expect(scratch.innerHTML).to.equal('<div foo="bar"></div>');
 		});
 
+		it('should render functional components wrapped with a proxy that has prototype set to undefined', () => {
+			function MyComponent(props) {
+				return <div {...props} />;
+			}
+
+			const proxy = Object.create(null, {
+				prototype: { value: undefined },
+				apply: {
+					value: function (_target, _thisArg, args) {
+						return MyComponent(...args);
+					}
+				}
+			});
+
+			const WrappedComponent = new Proxy(MyComponent, {
+				get(target, prop) {
+					if (prop === 'prototype') return undefined;
+					return Reflect.get(target, prop);
+				}
+			});
+
+			expect(() => {
+				render(<WrappedComponent foo="bar" />, scratch);
+			}).not.to.throw();
+
+			expect(scratch.innerHTML).to.equal('<div foo="bar"></div>');
+		});
+
 		it('should render components with props', () => {
 			let constructorProps;
 


rschristian:
> We recognize that LLM's have become an integral tool for many users, however when using LLMs to help write issues or pull requests, disclose this. Transparency preserves the collaborative integrity that makes open-source communities work. It's the difference between "I used AI to help structure this bug report" (honest, helpful) and silently passing off generated content as your own analysis (deceptive, counterproductive). We as maintainers want to engage with humans rather than LLM-generated messages, put time in your issue report and that will be reflected back.

[Preact Code of Conduct](https://github.com/preactjs/.github/blob/master/CODE_OF_CONDUCT.md)

TheAbMehta:
Thanks for pointing that out, I missed the CoC section on LLM disclosure. I used Claude to help me with identifying the fix, writing the test, and structuring the PR description. Happy to answer any questions or make changes.

rschristian:
Failure to disclose until called out, combined with your shotgunning of LLM-driven PRs to other projects (some of which the maintainers have called out as junk & that you obviously didn't test them), means I'm closing this out.

For any other contributor, feel free to open a PR, but I don't want to reward this behavior. We get enough LLM spam as-is.

PR #5032, opened 2026-02-22
refactor(props): use startsWith instead of char check for event handler checks

ali-garajian:
I ran the mentioned benchmark three times and every time `startsWith` gained a higher score than `char check`. Am I missing something?

Screenshot 1404-12-03 at 23 27 01
Screenshot 1404-12-03 at 23 28 00
Screenshot 1404-12-03 at 23 30 12

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -63,7 +63,7 @@ export function setProperty(dom, name, value, oldValue, namespace) {
 		}
 	}
 	// Benchmark for comparison: https://esbench.com/bench/574c954bdb965b9a00965ac6
-	else if (name[0] == 'o' && name[1] == 'n') {
+	else if (name.startsWith('on')) {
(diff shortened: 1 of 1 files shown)


rschristian:
`.startsWith` was newer than our targets allowed for v10, v11 could be viable if it's faster cross-browser. Might very well be

ali-garajian:
I noticed it adds 14B to the bundle size

@rschristian is it a fair tradeoff?

rschristian:
Firefox (Linux)             |  Chromium (Linux)
:-------------------------:|:-------------------------:
Firefox (Linux)  |  Chromium (Linux)

---

Without the benches pointing towards any perf increase in this repo (microbenchmarks like the one above can be useful, but we need to measure holistic perf here), I'd say that unless you can get the byte impact down (by using `.startsWith` elsewhere maybe?), this probably isn't worthwhile.

PR #5039, opened 2026-02-24
help fix the branch of Refactor/esm

Decodetalkers:
I want help fix the branch in Refactor/esm, and since it conflicts with main, so I think I need make pull request to the main 

@rschristian I have made the ci right


package.json
```diff
@@ -5,7 +5,7 @@
   "private": false,
   "description": "Fast 4kb React-compatible Virtual DOM library.",
   "type": "module",
-  "main": "dist/preact.esm.js",
+  "main": "dist/preact.js",
```
rschristian, line 8: `preact.js` won't exist, IIRC. This will be broken

Decodetalkers: It works in this way, it means target

rschristian: Vitest might work, but I'm pretty sure you've broken this entrypoint. Instead, Vitest will need to be corrected to use `preact.js`

The lockfile changes also look incredibly weird

Decodetalkers: I think I did not broke it, in fact, I just rollback it to the original value. And the lock, maybe I just remove the \n in the end..

In fact, the origin one, it also use preact.mjs, not preact.js, and it wrote in this way, the target become preact.mjs. 

https://github.com/preactjs/preact/blob/main/package.json#L7-L11

I see, because you reviewed the commit, then thought it is weird, but I read it from the whole pr, find it nothing changes at all

Maybe I need read the module key? Why you delete that key at the first time? The key of main is used for compiler to generate the target

Decodetalkers: done. https://github.com/developit/microbundle after I read the readme, it may be a mistake for you to remove the module key. main key is for commonjs, and also the entry point for microbundle, so that time you delete module and name the main to preact.esm.js, the esm target become preact.esm.esm.js

Decodetalkers: ok, I drop the commit for package-lock.json. I just run npm install and it changes, I do not know how to handle the changes.. Is it because I am using a different version of npm then the lock file changes?

rschristian: I'm sorry, but this isn't the fix, no.

`module` is a non-standard key used to support legacy ESM bundles. It is entirely superfluous today. Yes, Microbundle's output needs some finagling, but this is not the fix.

I might incorporate parts of 39d634b36dac69d38aae875ca72f9d4dc39e4478, but as a whole, this PR unfortunately isn't helpful. I'm very well aware of what Microbundle is doing (though precisely why Vitest was failing was indeed a mystery as I have to debug that via CI -- it segfaults locally) and know that it needs some altering.

Decodetalkers: but it did not failed. I mean. and you should not name main key to preact.esm.js, 

You should run 
```
npx microbundle build --raw --no-generateTypes -f esm
```
once to see the target under dist

and have you review the whole pr once? I did not change anything just seems like revert the it.

image
main key is for the commonjs, it should not be named that way

Decodetalkers:
image
this is the target when the main is dist/preact.esm.js

image

this is the correct one @rschristian . microbundle use the main key to generate the commonjs target, and by the name of main, it generates the esm target and etc, so it is not right to name main to dist/preact.esm.js

It is very clear that why vitest cannot find the compat, because preact.esm.js does not exist, it become preact.esm.esm.js

and you should also modify the vitest.config.js to the new targets

package.json
rschristian: No, `main` is not for commonjs. `main` refers to the main entrypoint which can be CJS or ESM depending upon the value of `type`. We have `type` set to `module`, therefore, `main` is ESM.

I am a maintainer of Microbundle, I am very well aware of how it works. I don't expect anyone else to fully understand it, as it is rather odd in the way it does things & we will be moving off of it, but please don't quote the docs back to me as if I don't understand what it's doing. What those docs were written for 5+ years ago and what we're doing today are intentionally going to differ.

Decodetalkers:
https://github.com/preactjs/preact/blob/main/package.json#L2-L10

image
and this is the origin target 

hav eyou see? if you do not let it generate commonjs, there is no preact.js, You have never use the `preact.js` once a time

package.json
Decodetalkers: I see. sorry for that.. I think I solve the problem for it. and make the ci green. ok

PR #5040, opened 2026-02-25
Add static/edge entries

JoviDeCroock:
As seen in https://github.com/JoviDeCroock/vinext-preact-test/pull/1

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,6 +91,16 @@
       "import": "./compat/server.mjs",
       "require": "./compat/server.js"
     },
+    "./compat/server.edge": {
+      "browser": "./compat/server.browser.js",
+      "import": "./compat/server.mjs",
+      "require": "./compat/server.js"
+    },
+    "./compat/server.static": {
+      "browser": "./compat/server.browser.js",
+      "import": "./compat/server.mjs",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -91,6 +91,16 @@
       "import": "./compat/server.mjs",
       "require": "./compat/server.js"
     },
+    "./compat/server.edge": {
```
rschristian, line 94: Would this ever be needed for anything outside of toy ports of very, very React tools?

I don't really have an issue with it, but outside of replacing that (very minimal) patch and/or an tooling alias, I have a hard time imagining it'll be used.

Edit: The same could be said for some of the existing server stuff of course, we probably want to avoid creating exports of minimal use is all

rschristian (approved):

package.json
JoviDeCroock: This one isn't currently for anything that I got working in vinext so not pressing but wanted to get it up already

rschristian: Fair enough; like I said, totally supportive of it, expands upon what we are already doing, just wondering how widespread it'll be used or if it'll be specific to folks trying to force Next, Remix, etc., to work with Preact.

PR #5041, opened 2026-02-25
undefined prototype

JoviDeCroock:
This could be faster/save bytes so doing a quick check before adding a test

Resolves https://github.com/preactjs/preact/issues/5011


rschristian (approved):
I hope you're on vacation as you were also pushing a PR 4 hours ago and that'd be staggeringly little sleep otherwise 😅

PR #5054, opened 2026-03-15
fix: retry on Suspense render and preserve hook state

mary-ext:
I was figuring out why [react-freeze](https://github.com/software-mansion/react-freeze) didn't work and it boiled down these issues:

1. if a parent rerenders while a child suspends, Preact doesn't try rerendering the immediate suspended children. react-freeze throws a promise that never resolves and expects the parent Freeze component to unsuspend it
2. when a child suspends, all of the hook state gets destroyed, which doesn't exactly align with React's behavior of preserving the state for already-mounted components (components suspending during initial mount have their state discarded, matching React's WIP fiber discard)
3. ~~when a child suspends, Preact does not differentiate useEffect and useLayoutEffect, only the latter should be cleaned up during suspension~~

### DIFF
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -42,11 +42,21 @@ options.unmount = function (vnode) {
 function detachedClone(vnode, detachedParent, parentDom) {
 	if (vnode) {
 		if (vnode._component && vnode._component.__hooks) {
-			vnode._component.__hooks._list.forEach(effect => {
-				if (typeof effect._cleanup == 'function') effect._cleanup();
-			});
-
-			vnode._component.__hooks = null;
+			if (vnode._dom != null && !vnode._component._excess) {
+				// Already-mounted component: preserve hook state but run
+				// effect cleanups. Effects will rerun when unsuspended.
+				vnode._component.__hooks._list.forEach(effect => {
+					if (typeof effect._cleanup == 'function') {
+						effect._cleanup();
+						effect._cleanup = undefined;
+					}
+				});
+			} else {
+				// Component suspended during initial mount (never committed
+				// to DOM). Discard hook state entirely, matching React's
+				// behavior of discarding the WIP fiber.
+				vnode._component.__hooks = null;
+			}
 		}
 
 		vnode = assign({}, vnode);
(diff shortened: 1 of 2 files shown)


JoviDeCroock:
> when a child suspends, all of the hook state gets destroyed, which doesn't exactly align with React's behavior of preserving the state

This is kind of incorrect, for a mounted component react preserves the hooks-state however for a component that has yet to mount the state is discarded --> WIP Fiber discarded

> when a child suspends, Preact does not differentiate useEffect and useLayoutEffect, only the latter should be cleaned up during suspension

I should check this but in react 17-18 this was that way in React as well - during my research at least. Which is also the reason why `urql` keeps the promise-cache.

I'm going to try and find time in the following weeks to verify this

JoviDeCroock:
I tested this in https://stackblitz.com/edit/vitejs-vite-emrmkojz?file=src%2FApp.tsx&terminal=dev

- Hooks state is reset when suspend happens during mount, not as part of an update
- Layout effects are cleaned up during the suspense bubble, regular effects are only cleaned up when the promise resolves --> I prefer not to make this differentiation

mary-ext:
> This is kind of incorrect, for a mounted component react preserves the hooks-state however for a component that has yet to mount the state is discarded --> WIP Fiber discarded

I inferred this would be the case when I glanced on https://github.com/software-mansion/react-freeze/commit/2e6ced224d72163d1890ce15fe87ebc4a987def9 but somehow ignored it and didn't think of checking in-depth aside from checking whether hook state is preserved at all, definitely worth replicating

mary-ext:
> Layout effects are cleaned up during the suspense bubble, regular effects are only cleaned up when the promise resolves --> I prefer not to make this differentiation

noted, might not make too much sense here I think

compat/test/browser/suspense.test.jsx
```diff
@@ -159,124 +159,6 @@ describe('suspense', () => {
 		resolve().then(assert).catch(assert);
 	});
 
-	it('should reset hooks of components', () => {
```
JoviDeCroock, line 162: Can we re-introduce these tests

compat/src/suspense.js
```diff
@@ -205,7 +216,34 @@ Suspense.prototype.render = function (props, state) {
...
+				suspendedVNode,
+				suspendedVNode._component._parentDom,
+				suspendedVNode._component._originalParentDom
+			);
+		}
+
+		this._suspendedVNode = null;
+		this._pendingSuspensionCount = 0;
+		this._suspenders = null;
+		state._suspended = null;
```
JoviDeCroock: Saving some bytes, I am however weary about mutating state here, what is the reason for this? Also we reset `this._suspendedVNode` does this mean that two subsequent renders will not work?

```suggestion
		this._pendingSuspensionCount = 0;
		state._suspended = this._suspendedVNode = this._suspenders = null;
```

mary-ext: we need `state._suspended` to be `null` immediately as the render code immediately below it checks `state._suspended` to decide whether to render `props.children` or the fallback, using setState mean we'd wait another tick.

I'm not sure which case you mean by two subsequent renders here but I have added tests for them

compat/test/browser/suspense.test.jsx
mary-ext: reintroduced but renamed the test case since the behavior changed

compat/src/suspense.js
```diff
@@ -42,11 +42,21 @@ options.unmount = vnode => {
 function detachedClone(vnode, detachedParent, parentDom) {
 	if (vnode) {
 		if (vnode._component && vnode._component.__hooks) {
-			vnode._component.__hooks._list.forEach(effect => {
-				if (typeof effect._cleanup == 'function') effect._cleanup();
-			});
-
-			vnode._component.__hooks = null;
+			if (vnode._dom != null) {
```
JoviDeCroock: My concern with this is that in Suspenseful scenarios we will set `_dom` maybe we should use `_excess` for everything then to counter-act this becoming a problem https://github.com/preactjs/preact/blob/main/src/diff/index.js#L361 - this does mean that we can't backport this to the 10.x release line safely. It would be part of v11

mary-ext: seems fine as a v11-only change, should I change it to `_excess` then? though it's worth noting that `_detachOnNextRender`, which calls `detachedClone`, is only set when we're not in the middle of hydration

https://github.com/preactjs/preact/blob/21dd6d04c1a9a43e5b60976bb5eb7d856253195b/compat/src/suspense.js#L168-L180

JoviDeCroock: I mainly mean that during the old hydration route the hooks clearing logic will miss behave because it will see suspended vnodes as mounted due to the presence of _dom. I don't think it's a bad thing to, in a separate refactor, move that core logic to fully use `_excess`

mary-ext: ah sorry, should've checked how it actually goes for v10, I've added an `_excess` check as an extra

PR #5055, opened 2026-03-17
Flush subtree effects

JoviDeCroock:
Reproduction at https://stackblitz.com/edit/vitejs-vite-yxpjutvm?file=package.json - notice how the fade-out animation never triggers and instead it just dissapears

## Problem

When libraries like `motion/react` (framer-motion) and `@base-ui/react` are used together with Preact/compat, exit animations break. The popup disappears instantly instead of animating out.

The root cause is a timing issue: when a component re-renders before its `useEffect` callbacks have fired (via `afterPaint`), Preact flushes only that component's stale effects during `options._render`. This causes effects from different components to run in separate render cycles instead of as a single batch.

In the concrete case:

1. Select closes → base-ui store updates → multiple components re-render
2. Base-ui's stale `useEffect` flushes during one render cycle → calls `getAnimations()` → finds 0 animations
3. Motion's stale `useEffect` flushes during a later render cycle → starts WAAPI animation → too late
4. Base-ui has already concluded no animations are running → sets `hidden` → popup disappears

## How React handles this

React calls [`flushPendingEffects()`](https://github.com/facebook/react/blob/main/packages/react-reconciler/src/ReactFiberRootScheduler.js) at the start of `performSyncWorkOnRoot` before beginning any new render work. This flushes **all** pending passive effects (from all components) as a batch, not just the re-rendering component's effects.

From React's `ReactFiberRootScheduler.js`:

```js
// In performSyncWorkOnRoot:
const didFlushPassiveEffects = flushPendingEffects();
if (didFlushPassiveEffects) {
  return null; // Exit to recompute priority
}
// Then proceed with rendering
```

```js
// In performWorkOnRootViaSchedulerTask (concurrent path):
const didFlushPassiveEffects = flushPendingEffectsDelayed();
if (didFlushPassiveEffects) {
  if (root.callbackNode !== originalCallbackNode) {
    return null;
  }
}
// Then proceed with rendering
```

This guarantees that all `useEffect` callbacks from render N complete before render N+1 begins, regardless of which component is re-rendering.

## What Preact did before

In `options._render`, when a component re-renders with unflushed effects (`previousComponent !== currentComponent`):

```js
// Old behavior: flush ONLY this component's stale effects
hooks._pendingEffects.some(invokeCleanup);
hooks._pendingEffects.some(invokeEffect);
hooks._pendingEffects = [];
```

This meant effects from different components could be spread across multiple render cycles when intermediate `setState` calls triggered cascading re-renders.

## The fix

Replace the per-component flush with a call to `flushAfterPaintEffects()`, which drains the entire pending effects queue:

```js
// New behavior: flush ALL pending effects (matching React)
if (hooks._pendingEffects.length) {
  flushAfterPaintEffects();
}
```

This matches React's approach: before starting new render work, ensure all effects from previous renders have completed.

## Behavioral differences from the old code

| Aspect | Old behavior | New behavior (matches React) |
|--------|-------------|------------------------------|
| Which effects flush | Only the re-rendering component's | All components with pending effects |
| Flush order | Re-rendering component first | Tree order (children before parents, as queued by `diffed`) |
| When sibling effects run | Deferred to their own re-render or `afterPaint` | Eagerly, before any re-render proceeds |

- Parent-effects may fire **earlier** than before, but at the correct time per React's contract
- Effect ordering changes from "re-rendering component first" to "tree order" which matches React

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -64,9 +64,9 @@ options._render = vnode => {
 				hookItem._pendingArgs = hookItem._nextValue = undefined;
 			});
 		} else {
-			hooks._pendingEffects.some(invokeCleanup);
-			hooks._pendingEffects.some(invokeEffect);
-			hooks._pendingEffects = [];
+			if (hooks._pendingEffects.length) {
+				flushAfterPaintEffects();
+			}
 			currentIndex = 0;
 		}
 	}
--- a/hooks/test/browser/useEffect.test.jsx
+++ b/hooks/test/browser/useEffect.test.jsx
@@ -1,5 +1,5 @@
 import { act, teardown as teardownAct } from 'preact/test-utils';
-import { createElement, render, Fragment, Component } from 'preact';
+import { createElement, render, Fragment, Component, options } from 'preact';
 import { useEffect, useState, useRef } from 'preact/hooks';
 import { setupScratch, teardown } from '../../../test/_util/helpers';
 import { useEffectAssertions } from './useEffectAssertions';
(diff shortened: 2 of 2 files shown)


marvinhagemeister (approved):
LGTM nice find!

hooks/src/index.js
```diff
@@ -64,9 +64,9 @@ options._render = vnode => {
 				hookItem._pendingArgs = hookItem._nextValue = undefined;
 			});
 		} else {
-			hooks._pendingEffects.some(invokeCleanup);
-			hooks._pendingEffects.some(invokeEffect);
-			hooks._pendingEffects = [];
+			if (hooks._pendingEffects.length) {
```
developit, line 67: Isn't this the wrong branch of this condition? The fix should be in the branch where `currentComponent === previousComponent` since it's for the case where the component is dirtied mid-render?

JoviDeCroock: This is in the right branch, when we immediately re-render due to an in-render setState we will reset the components hooks state. When we however have overlapping renders where we see that we have rendered this component before and it still has pending-effects to flush we will flush it. Now instead of flushing the component only we flush the entire subtree.

JoviDeCroock: The main reason for this is that nothing currently stops the following from happening

- Queue update
- Update runs, our root and an intermediary component enqueues an effect run
- Child queues an update for parent
  - BEFORE: we flush the intermediary effect
  - AFTER: we flush both intermediary as well as root effect
- intermediary renders again

Between 3 and 4 we would flush the parent updates

krskrs:
Hi guys, this PR seems interesting. I was wondering why it didn't land in the last release. Do you still have to work / reason on it a bit more?

PR #5057, opened 2026-03-21
feat: add support for `dangerouslySetInnerHTML` on `Fragments`

marvinhagemeister:
This PR adds support for setting `dangerouslySetInnerHTML` on Fragments. It allows you to inject raw unsafe HTML without a container element. This is useful for styling through CSS like with grids or flexbox which are sensitive to child elements.

Before:

```jsx
<article>
  <h1>{post.title}</h1>
  <div dangerouslySetInnerHTML={...} />
</article>
```
After:
```jsx
<article>
  <h1>{post.title}</h1>
  <Fragment dangerouslySetInnerHTML={...} />
</article>
```

This is the client-side equivalent of https://github.com/preactjs/preact-render-to-string/pull/453/ 

Given the size increase I'm not sure if we want to land this, but seems useful enough. Long-term we could make the byte increase worth it by generalizing the concept of a "virtual element between markers" for suspense, streaming, etc. That would make the size increase more worth it.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -255,6 +255,21 @@ export function diff(
 					? cloneNode(tmp.props.children)
 					: tmp;
 
+			// Teardown old html fragment markers if content changed or transitioning away
+			if (newType === Fragment && oldVNode._dom && oldVNode._dom._endHtml) {
+				const endMarker = oldVNode._dom._endHtml;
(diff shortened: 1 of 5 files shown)


JoviDeCroock:
We have the infrastructure on `main` for ths, I'm hesitant to push out a change this large to X

PR #5058, opened 2026-03-21
chore: rename `.test.js` -> `.test.jsx` when JSX is used

marvinhagemeister:
Future vite versions won't accept JSX inside `.js` files anymore (as they should).

PR #5059, opened 2026-03-22
chore: migrate remaining .js -> .jsx files

marvinhagemeister:
I've missed 2 files and a config change in https://github.com/preactjs/preact/pull/5058


rschristian (approved):

rschristian:
Need to expand here: 

https://github.com/preactjs/preact/blob/71ecde6e70f3b06e5b92feca331af87fae45bc42/babel.config.js#L36

PR #5068, opened 2026-03-30
Create a unique event-clock for each Preact instance on a page.

JoviDeCroock:
If a library bundles its own variant of Preact and relies on event-bubbling to step through its root and into other roots. Think a custom-element component library bubbling up into your regular application. Then the event-clock can be out of sync meaning that events aren't firing on the root application.

Theoretically this can't reproduce the bug of shifting roots as the first touch-point into the new Preact root will re-create with its own root.

We eliminate the bug from deliberate multi-bundle applications but it would now be introduced in ones where the user accidentally does a dual package hazard combination, that however would also have issues with the options hooks so I think that's okay.

It's a bit hard to add tests for this so I decided to leave it be and make huge comments instead.

A lower bundle size variant could be `self._id++`

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -6,9 +6,9 @@ import options from '../options';
 // shared event objects and handler functions cannot collide across instances.
 // ~1 in 60M collision odds - if you have that many praect versions on the page,
 // you deserve some weird bugs.
-let _id = Math.random().toString(8);
-let EVENT_DISPATCHED = '__d' + _id;
-let EVENT_ATTACHED = '__a' + _id;
+let _id = Math.random().toString(8),
+	EVENT_DISPATCHED = '__d' + _id,
+	EVENT_ATTACHED = '__a' + _id;
 
 function setStyle(style, key, value) {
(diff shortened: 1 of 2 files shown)


rschristian (approved):

marvinhagemeister (approved):

src/diff/props.js
```diff
@@ -1,6 +1,15 @@
 import { IS_NON_DIMENSIONAL, NULL, SVG_NAMESPACE } from '../constants';
 import options from '../options';
 
+// Per-instance unique key for event clock stamps. Each Preact copy on the page
+// gets its own random suffix so that `_dispatched` / `_attached` properties on
+// shared event objects and handler functions cannot collide across instances.
+// ~1 in 60M collision odds - if you have that many praect versions on the page,
+// you deserve some weird bugs.
```
JoviDeCroock, line 8:
```suggestion
// you deserve some weird bugs.
// In 11 we can replace this with a
// Symbol
```

developit (approved):

PR #5072, opened 2026-04-03
Trusted publishing

JoviDeCroock:
## Summary

- Adds a `publish` job to the release workflow that publishes to npm with provenance using OIDC
- Derives the npm dist-tag from the git tag (prerelease identifier or `latest`)
- Adds `publishConfig` with `provenance: true` and `access: "public"` to package.json
- Uses `id-token: write` permission for provenance attestation

## Security hardening

In addition to the new publish job, this PR nails down some loose ends in the existing workflow:

- **Trigger scoped to `on: push: tags: ['10.*']`** instead of `on: create`, which fired for both tags and branches. This also restricts to 10.x version tags only.
- **Potential script injection fixed** in the `release` job — `${{ steps.create-release.outputs.result }}` was interpolated directly into a `script:` block. Now passed via `RELEASE_DATA` env var and parsed with `JSON.parse`.
- **All actions pinned to commit SHAs** instead of mutable major version tags, preventing supply-chain attacks via tag mutation.
- **Explicit permissions** on all jobs — `release` declares `contents: write`, `publish` declares `contents: read` + `id-token: write`.

## Open question

- **Environment protection gate**: Should the `publish` job require manual approval via a GitHub Environment (e.g. `npm`) with required reviewers? Currently noted as a TODO — any collaborator with write access can push a tag and trigger a publish without approval. See https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/manage-environments

## Setup required

- On npmjs.com, link the package to the `preactjs/preact` GitHub repo for provenance verification

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,29 +1,70 @@
 name: Release
-on: create
+
+on:
+  push:
+    tags:
+      - '10.*'
 
 jobs:
   build:
-    if: github.ref_type == 'tag'
     uses: preactjs/preact/.github/workflows/build-test.yml@v10.x
 
   release:
     runs-on: ubuntu-latest
     needs: build
+    permissions:
+      contents: write
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/download-artifact@v4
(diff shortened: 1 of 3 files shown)


marvinhagemeister (approved):
LGTM

PR #5078, opened 2026-04-25
fix: vnode clone compatible with frozen intrinsics

kumavis:
Disclaimer: I (a human) identified this issue years ago and have always just worked around it via a local patch. I used an LLM (opus 4.7) to generate the fix. I reviewed the change and wrote most of the PR description.

This PR fixes a compatibility issue when Preact is run in [Hardened JavaScript](https://hardenedjs.org/) environments such as:
- [Nodejs with frozen intrinsics](https://nodejs.org/download/release/v12.22.12/docs/api/cli.html#cli_frozen_intrinsics), as recommended by Nodejs's own [Security Best Practices](https://nodejs.org/learn/getting-started/security-best-practices#monkey-patching-cwe-349).
- the [LavaMoat supplychain security runtime](https://github.com/lavamoat/lavamoat)

The compatibility issue comes from a javascript quirk known as ["the override mistake"](https://github.com/endojs/endo/discussions/1855) related to frozen intrinsics. For this bug, the minimal environment change is just `Object.freeze(Object.prototype)`.

In this environment, assigning the `constructor` property via assignment syntax (`=`) or `Object.assign` causes an error.
```
  TypeError: Cannot assign to read only property 'constructor'
```

Specifically, `Object.assign({}, vnode)` in `renderComponent()` (and `cloneNode` in `src/diff/index.js`) tries to copy a vnode's `constructor: undefined` own property via `[[Set]]`. Under environments with a frozen `Object.prototype` the `Object.prototype.constructor` is non-writable, so the assignment fails.

This PR introduces a `cloneVNode` utility that ensures `VNodes` are cloned safely. It also reorders the `constructor` key for consistent object shape between `VNodes` and their clones.

Note: I erred on the side of excessive documentation in comments, let me know if you would like this reduced.

### DIFF
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -4,7 +4,7 @@ import {
 	FORCE_PROPS_REVALIDATE,
 	MODE_HYDRATE
 } from '../../src/constants';
-import { assign } from './util';
+import { cloneVNode } from '../../src/util';
 
 const oldCatchError = options._catchError;
 options._catchError = (error, newVNode, oldVNode, errorInfo) => {
@@ -49,7 +49,7 @@ function detachedClone(vnode, detachedParent, parentDom) {
 			vnode._component.__hooks = null;
 		}
 
-		vnode = assign({}, vnode);
+		vnode = cloneVNode(vnode);
 		if (vnode._component != null) {
 			if (vnode._component._parentDom === parentDom) {
 				vnode._component._parentDom = detachedParent;
--- a/src/component.js
+++ b/src/component.js
@@ -7,7 +7,7 @@ import {
 import { Fragment } from './create-element';
 import { commitRoot, diff } from './diff/index';
 import options from './options';
-import { assign } from './util';
+import { assign, cloneVNode } from './util';
 
 /**
  * Base Component class. Provides `setState()` and `forceUpdate()`, which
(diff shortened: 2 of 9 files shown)


kumavis:
requesting a review from @JoviDeCroock (bc of recent change to vnode constructor logic (strict equality))

JoviDeCroock:
While I think the fix makes sense, there's a site that has been forgotten by the LLM. `compat/src/suspense.js` also has a clone afaict. That being said there are a lot of libraries in the ecosystem that will still do this through `options` hooks so I am thinking whether there is a more general solution that would prevent us from patching all of them as well

kumavis:
> there's a site that has been forgotten by the LLM

good catch, thanks for the review.

> there are a lot of libraries in the ecosystem that will still do this through options hooks 

can you help me understand this? sounds like 3rd parties are also cloning nodes, but I don't get the "through option hooks" part.

the good news is this change would not break those ecosystem libraries, and users running in HardenedJS environments get a working core `preact`.

to help the ecosystem we could export `cloneVNode` and/or provide a `vnode.clone()`. either way any ecosystem package would need to update its manual cloning. A HardenedJS consumer would need to either patch the ecosystem package or choose a compatible one. I have not personally encountered an issue with ecosystem packages in HardenedJS, but I'm not a heavy user.

JoviDeCroock:
The through options hook part is mostly the `options.vnode` hook but come to think more of it, most of these are mutating the VNode rather than doing immutable copies so it's possible that we're good on this change. The performance looks to be neutral as well so I am gravitating towards accepting the change despite the size changes, going to have a think about whether we can reduce the byte-size (staying true to our philosophy of byte size).

Do note that this is a PR for the main branch which is currently v11 (beta) and we also have the v10.x branch which is the mainline atm.

JoviDeCroock:
Implemented it in https://github.com/preactjs/preact/pull/5195 - gave you credit on the commit!

PR #5081, opened 2026-05-01
Fix hydration when we have defaultValue or value on a textarea

### LINKED ISSUE #5080: Hydrate clears value from <textarea defaultValue='x' />

**Describe the bug**

When using `hydrate` with `<textarea defaultValue='x' />`, the value is cleared instead of retained.

**To Reproduce**

```html
<!doctype html>
<html>
<head>
<script type='module'>
  import { h, hydrate } from 'https://esm.sh/preact'

  function Test() {
    return h('textarea', { defaultValue: 'x' })
  }

  hydrate(h(Test), document.body)
</script>
</head>
<body><textarea>x</textarea></body>
</html>
```

**Expected behavior**

The textarea should retain the value (`x`), but it's cleared when the page loads.

Reproduced in latest Chrome, Firefox, and Safari on macOS and Safari on latest iOS.

JoviDeCroock:
Fixes #5080.

## Summary
- Preserve SSR textarea text during hydration when using `defaultValue`
- Re-apply controlled textarea `value` during hydration
- Add hydration coverage for textarea `defaultValue` and `value`

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -474,7 +474,10 @@ function diffElementNodes(
 		}
 	} else {
 		// If excessDomChildren was not null, repopulate it with the current element's children:
-		excessDomChildren = excessDomChildren && slice.call(dom.childNodes);
+		excessDomChildren =
+			nodeType == 'textarea' && newProps.defaultValue != NULL
+				? NULL
+				: excessDomChildren && slice.call(dom.childNodes);
 
 		// If we are in a situation where we are not hydrating but are using
 		// existing DOM (e.g. replaceNode) we should read the existing DOM
@@ -562,7 +565,7 @@ function diffElementNodes(
 		}
 
 		// As above, don't diff props during hydration
-		if (!isHydrating) {
+		if (!isHydrating || nodeType == 'textarea') {
 			i = 'value';
 			if (nodeType == 'progress' && inputValue == NULL) {
 				dom.removeAttribute('value');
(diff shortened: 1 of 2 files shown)


src/diff/index.js
```diff
@@ -474,7 +474,10 @@ function diffElementNodes(
 		}
 	} else {
 		// If excessDomChildren was not null, repopulate it with the current element's children:
-		excessDomChildren = excessDomChildren && slice.call(dom.childNodes);
+		excessDomChildren =
+			nodeType == 'textarea' && newProps.defaultValue != NULL
```
JoviDeCroock, line 478: Basically makes us ignore children of a `textarea` so that we don't remove them in HTML the children of a textarea are its defaultValue, I know confusing right?

> In HTML, the initial content of a <textarea> is specified between its opening and closing tags, not as a value attribute.

[Source](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/textarea)

src/diff/index.js
```diff
@@ -562,7 +565,7 @@ function diffElementNodes(
 		}
 
 		// As above, don't diff props during hydration
-		if (!isHydrating) {
+		if (!isHydrating || nodeType == 'textarea') {
```
JoviDeCroock, line 568: This ensures that we go into value re-hydration for textareas

src/diff/index.js
marvinhagemeister: TIL!

marvinhagemeister (approved):
LGTM

src/diff/index.js
rschristian: TIL too, I swore `value` was valid

rschristian (approved):

PR #5082, opened 2026-05-06
fix(clone-element): preserve falsy key and ref values in cloneElement

Olexandr88:
Replace || operator with explicit undefined checks for key and ref arguments in cloneElement to correctly handle falsy but valid values like 0, empty string, or null.

### DIFF
--- a/src/clone-element.js
+++ b/src/clone-element.js
@@ -31,8 +31,8 @@ export function cloneElement(vnode, props, children) {
 	return createVNode(
 		vnode.type,
 		normalizedProps,
-		key || vnode.key,
-		ref || vnode.ref,
+		key !== undefined ? key : vnode.key,
+		ref !== undefined ? ref : vnode.ref,
 		NULL
 	);
 }
--- a/src/clone-element.js
+++ b/src/clone-element.js
@@ -1,6 +1,6 @@
 import { assign, slice } from './util';
 import { createVNode } from './create-element';
-import { NULL } from './constants';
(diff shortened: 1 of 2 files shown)


src/clone-element.js
```diff
@@ -31,8 +31,8 @@ export function cloneElement(vnode, props, children) {
 	return createVNode(
 		vnode.type,
 		normalizedProps,
-		key || vnode.key,
-		ref || vnode.ref,
+		key !== undefined ? key : vnode.key,
+		ref !== undefined ? ref : vnode.ref,
```
rschristian: We use [constants ](https://github.com/preactjs/preact/blob/21dd6d04c1a9a43e5b60976bb5eb7d856253195b/src/constants.js#L31) for this as they compress slightly better:

```suggestion
		key !== UNDEFINED ? key : vnode.key,
		ref !== UNDEFINED ? ref : vnode.ref,
```

(Will have to add an import for it too)

rschristian (approved):
Makes sense to me, just one tiny change needed.

Do note that `main` is targeting the upcoming v11, not the current v10. Slightly concerned about introducing this to v10 though given it could break some apps.

Olexandr88:
Replace literal undefined with the UNDEFINED constant in key and ref checks, and add it to the constants import for better minification.

rschristian:
Oh, and could do w/ a test case for each, to ensure the behavior is solid going forward.

rschristian:
Thanks!

PR #5085, opened 2026-05-12
Migrate dependency management to pnpm

andrewiggins:
## Summary
- migrate root, subpackages, test/ts, and demo to a single pnpm workspace and committed pnpm-lock.yaml
- replace npm CI installs with pnpm/action-setup, pnpm cache, and frozen installs
- enable pnpm supply-chain hardening with minimum release age, no-downgrade trust checks, blocked exotic transitive dependencies, strict build-script approval, and explicit allowBuilds policy
- keep npm pack/publish release behavior intact while updating local docs and Husky to use pnpm

## Security decisions
- pin packageManager to pnpm@11.1.0 and move Volta Node to 24.15.0 because pnpm 11 requires Node >=22.13
- use version-specific trustPolicyExclude entries instead of a broad trust bypass
- allow only esbuild install-time builds; explicitly deny @parcel/watcher, fsevents, and styled-components builds
- add explicit demo sass-embedded dependency exposed by pnpm isolation instead of weakening hoisting
- declare nano-staged locally so the pre-commit hook does not rely on npx fetching tools

## Verification

Ran CI with a dummy src change to verify all GH Actions passed

Note: This PR was partially generated using GPT 5.5


demo/package.json
```diff
@@ -10,7 +10,8 @@
   "devDependencies": {
     "@babel/plugin-proposal-class-properties": "^7.0.0-beta.55",
     "@babel/plugin-proposal-decorators": "^7.4.0",
-    "vite": "^5.4.10"
+    "sass-embedded": "^1.99.0",
```
marvinhagemeister, line 13: what's the `sass-embedded` dependency used for?

.github/workflows/run-bench.yml
```diff
@@ -34,29 +34,16 @@ jobs:
...
-      - uses: actions/setup-node@v4
-        with:
-          node-version-file: 'package.json'
-          cache: 'npm'
-          cache-dependency-path: '**/package-lock.json'
-
-      # Setup pnpm
-      - name: Install pnpm
-        uses: pnpm/action-setup@v3
+      - uses: pnpm/action-setup@v6
```
JoviDeCroock: There's some actions missing in this PR that also use npm still

JoviDeCroock (changes requested):

demo/package.json
andrewiggins: The demo project has a sass file that needs compiling. Previously `npm` implicitly installed optional peer deps - pnpm does not.

.github/workflows/run-bench.yml
andrewiggins: I think I got them all. Let me know if you see anything that I missed

JoviDeCroock:
Superseded by #5246.

Issue #5094, opened 2026-05-18
InputHTMLAttributes.type should accept Signalish


rejhgadellaa:
- [x] Check if updating to the latest Preact version resolves the issue

**Describe the bug**

The `InputHTMLAttributes` type in `jsx.d.ts` does not wrap the `type` attribute in `Signalish`:

```
type?: HTMLInputTypeAttribute | undefined;
```

Expected: 

```
type?: Signalish<HTMLInputTypeAttribute | undefined>;
```

**To Reproduce**

```jsx

import { useSignal } from '@preact/signals';

/** @import { HTMLInputTypeAttribute, InputHTMLAttributes, VNode } from 'preact' */
/** @import { Signal } from '@preact/signals' */

/**
 * @param   { InputHTMLAttributes } props
 * @returns { VNode }
 */
export function Input( props ) {
  const { type, ...restProps } = props;
  /** @type { Signal<HTMLInputTypeAttribute | undefined> } */
  const stype = useSignal( type );
  return (
    <input
      type={ stype }
      { ...restProps }
    />
  );
}

```

The snippet above will report a typescript error:

```
Type 'Signal<HTMLInputTypeAttribute | undefined>' is not assignable to type 'HTMLInputTypeAttribute | undefined'.
  Type 'Signal<HTMLInputTypeAttribute | undefined>' is not assignable to type 'string & {}'.
    Type 'Signal<HTMLInputTypeAttribute | undefined>' is not assignable to type 'string'.
```

**Expected behavior**

The `type` attribute should allow `Signalish` values.

PR #5096, opened 2026-05-18
fix: accept Signalish for input type

JoviDeCroock:
- Allow `InputHTMLAttributes.type` to accept `Signalish` on `v10.x`
- Update both namespace typings and the type test
- Refs #5094


rschristian (approved):
Oops, thanks!

PR #5104, opened 2026-05-26
ci: add explicit workflow-level permissions to ci

arpitjain099:
Declares `permissions: contents: read` at the workflow level for `ci.yml`.

The CI workflow is read-only. Setting an explicit minimum scope follows GitHub's [hardening guide](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token) and the OpenSSF Scorecard Token-Permissions check, and limits blast radius if any third-party action used here is ever compromised (the March 2025 tj-actions/changed-files token-leak incident).

YAML parses cleanly. No other changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,9 @@ on:
       - restructure
       - v11
 
+permissions:
+  contents: read
+
 jobs:
   filter_jobs:
     name: Filter jobs


arpitjain099:
Closing this; on a second look the workflow-level `contents: read` here would propagate into the reusable `size.yml` workflow it calls, which uses `preactjs/compressed-size-action` and needs `pull-requests: write` to post the PR size comment. Apologies for the noise. I will rework as per-job permissions on the non-reusable jobs only if you want me to re-open.

PR #5107, opened 2026-05-31
fix(suspense): don't orphan server DOM when a hydration-suspended boundary resolves to non-matching content

sbesh91:
### What

A `<Suspense>` boundary that suspends *during hydration* and then resolves to content that does **not** reuse the server-rendered DOM (a different element, or `null`) leaves the original server DOM orphaned in the document instead of removing it.

### Repro

```jsx
scratch.innerHTML = '<div>Hello</div>';
const [Lazy, resolve] = createLazy();
hydrate(<Suspense><Lazy /></Suspense>, scratch);
rerender();
// resolve to different content:
await resolve(() => <section>Goodbye</section>);
rerender();
// expected: "<section>Goodbye</section>"
// actual:   "<section>Goodbye</section><div>Hello</div>"  <- server <div> orphaned
```

Resolving to `null` leaves `<div>Hello</div>` behind entirely. Same-content resolution is unaffected (it re-adopts the node by matching `localName`, which is why the existing "should leave DOM untouched when suspending while hydrating" test passes).

### Root cause

When a component suspends while hydrating, `diff`'s catch pins the server DOM as the suspended vnode's `_dom` and pulls it out of `excessDomChildren` so hydration won't delete it (flags `MODE_HYDRATE | MODE_SUSPENDED`). On resume, the resume block rebuilds `excessDomChildren = [parkedDom]` and re-enters hydrating mode. If the resolved content doesn't reuse the parked node, nothing removes it: excess-DOM cleanup only runs in `diffElementNodes` (for an element's own children), but the component-resume path calls `diffChildren` directly.

### Fix

After the resumed component's children are diffed, remove any leftover parked excess, advancing `oldDom` past a removed node the same way `diffChildren`'s own removal loop does. ~10 lines in `src/diff/index.js`.

### Testing

Two new cases in `compat/test/browser/suspense-hydration.test.jsx` (beside the existing same-content control). Full vitest suite green (1252 passed). oxfmt / oxlint / tsc clean.

### Note

This also reproduces on `main` (v11). Happy to port it there in a follow-up (or fold in here), whichever you prefer.

### DIFF
--- a/compat/test/browser/suspense-hydration.test.jsx
+++ b/compat/test/browser/suspense-hydration.test.jsx
@@ -145,6 +145,35 @@ describe('suspense hydration', () => {
 		});
 	});
 
+	it('should remove server-rendered DOM but keep later siblings when resolving to different content', () => {
+		scratch.innerHTML = '<p>region</p><div>After</div>';
+		clearLog();
+
+		const [Lazy, resolve] = createLazy();
+		hydrate(
+			<Fragment>
+				<Suspense>
+					<Lazy />
+				</Suspense>
+				<div>After</div>
+			</Fragment>,
+			scratch
+		);
+		rerender(); // Flush rerender queue to mimic what preact will really do
+		expect(scratch.innerHTML).to.equal('<p>region</p><div>After</div>');
+		clearLog();
+
+		// The parked <p> is removed, but the sibling <div>After</div> that
+		// follows the boundary must stay in place. Guards against the leftover
+		// removal disturbing trailing siblings.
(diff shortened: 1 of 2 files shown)


JoviDeCroock:
We have a more scaleable (and expandable) fix for this in https://github.com/preactjs/preact/issues/4442 on v11 - this kind of feels like a placeholder fix.

The reason it's not working is probably because we aren't using the marker.

I do think that possibly we are missing the excessDom cleanup for a lazy node resolving to nothing but that's also an edge case, the ssr and hydrated dom, even when deferred should match. If not you need an isHydrated check that renders the dom post hydration

PR #5108, opened 2026-06-04
Fix useId stability across async Suspense

JoviDeCroock:
## Summary

- stabilize `useId` across async Suspense resumes by assigning each compat `Suspense` boundary its own numeric mask namespace
- keep the Suspense-specific counter in compat instead of hooks
- keep the existing hooks root behavior limited to carrying the previous root mask across render invocations

## Problem

When sibling `Suspense` boundaries that use `useId` resolve out of order, IDs can be assigned from resolution order instead of tree position. That can make the server result differ from the client/hydration result even when the rendered tree is the same.

### DIFF
--- a/compat/src/suspense.js
+++ b/compat/src/suspense.js
@@ -198,14 +198,23 @@ Suspense.prototype.componentWillUnmount = function () {
  * @param {import('./internal').SuspenseState} state
  */
 Suspense.prototype.render = function (props, state) {
+	let vnode = this._vnode;
+	if (!vnode._mask) {
+		let root = vnode;
+		while (root._parent) root = root._parent;
+
+		root = root._mask || (root._mask = [0, 0]);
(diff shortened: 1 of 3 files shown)


marvinhagemeister (approved):

compat/src/suspense.js
```diff
@@ -43,6 +43,8 @@ options.unmount = function (vnode) {
 	if (oldUnmount) oldUnmount(vnode);
 };
 
+let globalSuspenseCounter = 0;
```
JoviDeCroock: I want to experiment with this some more, when I've released RTS but just using this as a counter might be better than using _mask[0]

marvinhagemeister: +1 global suspense counter seems better

marvinhagemeister (approved):

marvinhagemeister (approved):

PR #5115, opened 2026-06-12
perf(hooks): avoid redundant allocations and writes

JoviDeCroock:
- `updateHookState`: drop the `_list.filter()` allocation and merge the `every`/`some` passes into a single traversal
- `options.diffed`: only clear `_pendingArgs` when set, avoiding a write per hook per diff and shape pollution of non-effect hook states

hooks bundle: -11 B gzip / -9 B brotli.


marvinhagemeister (approved):
nice find!

Issue #4909, opened 2025-08-30
VNode leak using Context


korDen:
- [ ] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
This is a followup to another leak: VNode leak in Context · Issue #4905 · preactjs/preact. There is a VNode memory leak occurring when using Context Provider with state updates. After clicking a button that triggers a state change, the previous VNode hierarchy is still retained in memory, which can be observed in Chrome Dev Tools. This results in duplicate VNodes including 2 `<button>`s, 2 `<App>`s, 2 `<Context>`s, 2 `<RouterContent>`s and 2 Text nodes (both the old and new text content).

**To Reproduce**

StackBlitz reproduction link: https://stackblitz.com/edit/vitejs-vite-21mvd37v?file=src%2Fmain.tsx,index.html

Repro:
```ts
import { render, createContext } from 'preact';
import { useState, useContext } from 'preact/hooks';

const RouterContext = createContext(null);

function RouterContent() {
  const { params, setRouterState } = useContext(RouterContext);
  return (
    <button onClick={() => setRouterState({ params: params + "!" })}>
      {`No account found for username "${params}"`}
    </button>
  );
}

function App() {
  const [routerState, setRouterState] = useState({ params: 'a' });

  const contextValue = { 
    params: routerState.params,
    setRouterState
  };

  return (
    <RouterContext.Provider value={contextValue}>
      <RouterContent />
    </RouterContext.Provider>
  );
}

render(<App />, document.getElementById('app'));
```

which compiles to (sometime its easier to check correctness of .js vs .ts):
```js
import { jsx as _jsx } from "preact/jsx-runtime";
import { render, createContext } from 'preact';
import { useState, useContext } from 'preact/hooks';
const RouterContext = createContext(null);
function RouterContent() {
    const { params, setRouterState } = useContext(RouterContext);
    return (_jsx("button", { onClick: () => setRouterState({ params: params + "!" }), children: `No account found for username "${params}"` }));
}
function App() {
    const [routerState, setRouterState] = useState({ params: 'a' });
    const contextValue = {
        params: routerState.params,
        setRouterState
    };
    return (_jsx(RouterContext.Provider, { value: contextValue, children: _jsx(RouterContent, {}) }));
}
render(_jsx(App, {}), document.getElementById('app'));
```

1. Run the web app (preferably with the previous memory leak patch applied -- I tested with the patch applied).
[...]

PR #5116, opened 2026-06-12
Fix memory leak hotspots

JoviDeCroock:
- `render()`: clear the root Fragment's `props.children` after commit so `_children` is the only live tree reference (#4909, #4905)
- `unmount()`: clear `component._globalContext` with the other component references

Local `replace1k` check: `run-final` 10.99ms vs 10.79ms main, overlapping CI.

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -654,7 +654,7 @@ export function unmount(vnode, parentVNode, skipRemove) {
 			}
 		}
 
-		r.base = r._parentDom = NULL;
+		r.base = r._parentDom = r._globalContext = NULL;
 	}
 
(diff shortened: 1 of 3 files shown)


JoviDeCroock:
I feel like I've done this before but... time to analyze why replace is degrading this much... Probably hitting GC a lot more...

EDIT: yes removing the `_unmount` reset of `_children` did the trick https://github.com/preactjs/preact/compare/c40fd133c1cd7dc452ba1594a763c0776807a01f..5670ac8b8b9222783d2527ebbac88c021995b7b8

marvinhagemeister (approved):
nice find!

PR #5117, opened 2026-06-17
fix(compat): avoid mutating user-provided style objects when appending px

JSap0914:
## Bug

`handleDomVNode` in `compat/src/render.js` mutates the style object supplied by the caller when appending `px` to numeric values:

```js
// before
if (i === "style" && typeof value === "object") {
    for (let key in value) {
        if (typeof value[key] === "number" && !IS_NON_DIMENSIONAL.test(key)) {
            value[key] += "px"; // ← mutates props.style in place
        }
    }
}
```

`value` is a direct reference to `props[i]` (the original object passed by the user). After this loop the caller's object is permanently modified—`margin: 10` becomes `margin: "10px"`—which is an unexpected side effect.

## Fix

Shallow-copy the style object before the loop so mutations stay local to `normalizedProps`:

```js
value = assign({}, value);
```

The px-append behaviour and the rendered output are unchanged; only the caller's object is now left untouched.

## Verification

```
npx vitest run compat/test/browser/render.test.jsx
```

New test: **"should not mutate the original style object when appending px"** — this test **failed** before the fix (`style.margin` was mutated to `"10px"`) and **passes** after it.

All 46 existing compat-render tests continue to pass; full suite: **1242 passed, 12 skipped** across 101 test files.

> AI-assisted contribution.


compat/src/render.js
```diff
@@ -136,6 +136,7 @@ function handleDomVNode(vnode) {
 		}
 
 		if (i === 'style' && typeof value === 'object') {
+			value = assign({}, value);
```
marvinhagemeister: The PR description says:
> (...) mutates the style object supplied by the caller when appending `px` to numeric values

which hints that it's only a problem when `px` is appended. The change here though is broader. It also applies when no `px` is appended too. That seems wasteful in terms of performance that we always pay the cost of cloning an object for that too.

compat/test/browser/render.test.jsx
```diff
@@ -886,4 +886,21 @@ describe('compat render', () => {
...
+		render(<div style={style} />, scratch);
+
+		expect(style.margin).to.equal(
+			originalMargin,
+			'style.margin should not be mutated'
+		);
+		expect(style.opacity).to.equal(
+			originalOpacity,
+			'style.opacity should not be mutated'
+		);
```
marvinhagemeister: Let's simplify this:

```suggestion
		const style = { margin: 10, opacity: 0.5 };
		render(<div style={style} />, scratch);

		expect(style).to.equal({ margin: 10, opacity: 0.5 });
```

compat/src/render.js
```diff
@@ -136,8 +136,13 @@ function handleDomVNode(vnode) {
 		}
 
 		if (i === 'style' && typeof value === 'object') {
+			let cloned = false;
 			for (let key in value) {
 				if (typeof value[key] === 'number' && !IS_NON_DIMENSIONAL.test(key)) {
+					if (!cloned) {
+						value = assign({}, value);
```
JSap0914: The clone is now lazy —  only runs the first time a numeric value actually needs  appended, so style objects with no such properties pay zero clone cost.

compat/src/render.js
```diff
@@ -136,8 +136,13 @@ function handleDomVNode(vnode) {
 		}
 
 		if (i === 'style' && typeof value === 'object') {
+			let cloned = false;
```
JoviDeCroock:
```suggestion
			let cloned;
```

compat/src/render.js
```diff
@@ -136,8 +136,13 @@ function handleDomVNode(vnode) {
...
 			for (let key in value) {
 				if (typeof value[key] === 'number' && !IS_NON_DIMENSIONAL.test(key)) {
+					if (!cloned) {
+						value = assign({}, value);
+						cloned = true;
```
JoviDeCroock:
```suggestion
						cloned = value = assign({}, value);
```

JoviDeCroock (approved):
A code golfing suggestion

PR #4851, opened 2025-07-24
Support alternative contentDocument

JoviDeCroock:
Supersedes #4501 
Supersedes #3042
Closes https://github.com/preactjs/preact/issues/4500

Issue #4500, opened 2024-09-16
Add an option to inject the `document` for a `render()`


alshdavid:
# The Problem

I'd like to be able to inject the `document` to use for my `render()` to allow me to use happy-dom or JSDOM to test my Preact components without assigning a global document.

Currently you must assign it to the `globalThis` object like so:

```jsx
import { h, render } from 'preact'

// Imagine this is a mock Window
const browser = new Browser(`<body></body>`)

// Assign itto the global context so Preact can use them
globalThis.document = browser.document

// Run the preact render
render(<div>Hi</div>, browser.document.body)
```

However this prevents me from being able to run the tests concurrently
```jsx
setTimeout(() => {
  const browser = new Browser(`<body></body>`)
  globalThis.document = browser.document
  render(<div>Hi</div>, browser.document.body)
})

setTimeout(() => {
  const browser = new Browser(`<body></body>`)
  globalThis.document = browser.document
  render(<div>Hi</div>, browser.document.body)
})
```

# What I've Tried

I have tried to use Node's `node:vm` module to alter the globalThis context, however this does not work as it doesn't have access to imports.

I have tried bundling the import and using that as a script evaluation in the mock browser, however this is slow and cumbersome, negating the benefits of making tests concurrent.

# Request

From what I can tell, the `globalThis.document` is used in the `diff` function: https://github.com/preactjs/preact/blob/main/src/diff/index.js#L414

1. Use `element.ownerDocument`

This would allow preact to use the `document` object associated with the current `HTMLElement`

2. Add a new function, something like `renderToDom(component, target, globalContext)`.

This would allow me to inject the specific DOM into the render pipeline. This might cause issues with iframes though

```js
const browser = new Browser(`<body></body>`)
renderToDom(
  <div>Hi</div>, 
  browser.document.body,
  browser.document, // <- here
)
```

I can raise a PR for this, is that okay?

JoviDeCroock:
Hey it might be worth revisiting https://github.com/preactjs/preact/pull/3042 for this

alshdavid:
Rebased it: https://github.com/preactjs/preact/pull/4501

JoviDeCroock:
Let's stick with the approach suggested in https://github.com/preactjs/preact/issues/2545 and https://github.com/preactjs/preact/pull/4501. I'll close #2545 to track the progress here, note that #4501 is failing CI

alshdavid:
[...]

PR #5119, opened 2026-06-23
fix: check `nodeType` instead of the global `document`

### LINKED ISSUE #5118: Safely access bare `document` in Preact 11
For an upcoming [bomb.sh](https://bomb.sh) project, we'll be publishing Preact bindings for TUI projects. Because we're running in a non-browser environment, there is no global `document` variable.

To my great pleasure, I saw that **Preact 11** has moved to scoped `doc` references that are dervied from `node.ownerDocument`.

This _almost_ unburdens us from needing a `globalThis.document` shim. Unfortunately, there's still a single [unguarded `document` reference in `render.js`](https://github.com/preactjs/preact/blob/54f628c6242b50c17b9b97fbb319ef6168bd0577/src/render.js#L13-L16).

**Additional context**
On our side, we need to define `globalThis.document = undefined` to avoid the `ReferenceError` (or wire a proper `document` shim).

If Preact 11 added a `typeof document !== 'undefined'` or `'document' in globalThis` before checking if `parentDom == document`, we'd be able to remove our userland shim and Preact would be fully "native-compatible". 

Fairly minimal change, but it would bring me great joy to contribute a fix here.

natemoo-re:
Using `nodeType == 9` aligns `render()` with the existing `parentNode.ownerDocument` lookups (#2569, #4851) and removes the last reference to the global `document` in the codebase (#4500, closes #5118).

It's also more robust. `parentDom == document` tested identity against the global `document`, which meant a document from `contentDocument`/`DOMParser`/`createHTMLDocument()` would be diffed straight into its `Document` node. `nodeType == 9` matches any document and correctly retargets to its own `documentElement`.

Resolves https://github.com/preactjs/preact/issues/5118

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -11,8 +11,9 @@ import { slice } from './util';
  */
 export function render(vnode, parentDom) {
 	// https://github.com/preactjs/preact/issues/3794
-	if (parentDom == document) {
-		parentDom = document.documentElement;
+	// https://github.com/preactjs/preact/issues/5118
+	if (parentDom.nodeType == 9) {
+		parentDom = parentDom.documentElement;
 	}
 
 	if (options._root) options._root(vnode, parentDom);
--- a/src/render.js
+++ b/src/render.js
@@ -10,13 +10,13 @@ import { slice } from './util';
  * @param {import('./internal').PreactElement} parentDom The DOM element to render into
  */
 export function render(vnode, parentDom) {
+	if (options._root) options._root(vnode, parentDom);
+
 	// https://github.com/preactjs/preact/issues/3794
 	// https://github.com/preactjs/preact/issues/5118
 	if (parentDom.nodeType == 9) {
 		parentDom = parentDom.documentElement;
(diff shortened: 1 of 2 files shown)


JoviDeCroock (dismissed):

JoviDeCroock:
If you need this in v10.x there's a branch for this to be PR'ed to there as well, on second thought, this does not support undefined documents which might become problematic and it would expand byte-size to check both 😅

src/render.js
```diff
@@ -10,13 +10,14 @@ import { slice } from './util';
  * @param {import('./internal').PreactElement} parentDom The DOM element to render into
  */
 export function render(vnode, parentDom) {
+	if (options._root) options._root(vnode, parentDom);
```
JoviDeCroock, line 13: This will change the thing people see in options._root but because this is in a major release and pretty internal I think we are fine doing that

JoviDeCroock (approved):

PR #5124, opened 2026-06-26
fix(compat): restore defaultValue fallback for textarea in IE11

### LINKED ISSUE #5123: Regression: defaultValue for <textarea> broken in IE11 (PR #2742)
**Describe the bug**
In Preact 10.5.0 and later, the `defaultValue` prop for `<textarea>` no longer works in Internet Explorer 11 when the `value` prop is not explicitly passed. The textarea appears empty despite having a `defaultValue` set. This worked correctly in earlier versions.

**To Reproduce**
1. Create a Preact component with `<textarea defaultValue="Hello, World!" />`
2. Open the application in Internet Explorer 11
3. Observe that the textarea is empty
4. Add `value={undefined}` to the component: `<textarea defaultValue="Hello, World!" value={undefined} />`
5. Observe that the textarea now displays the text

**Expected behavior**
The textarea should display the value passed to `defaultValue` in all browsers, including Internet Explorer 11, without requiring `value={undefined}` to be explicitly passed.

**Regression Commit:** This issue was introduced in PR [#2742 - `[wip] optimizations for compat`](https://github.com/preactjs/preact/pull/2742).

**Before (working):**
```js
// Old condition: checks if valueProp is null (meaning value hasn't been set yet)
if (i === 'defaultValue' && valueProp == null) {
    if ('value' in props) {
        i = 'value';
    } else {
        valueProp = value; // Preserve defaultValue as fallback
    }
}

// New condition: only works if value prop exists AND is null/undefined
if (i === 'defaultValue' && 'value' in props && props.value == null) {
    i = 'value';
}

```

The Problem:

The old logic handled two cases:
- value is present (even null) → use it
- value is absent → use defaultValue as fallback (valueProp = value)

The new logic handles only one case:
- value={null} or value={undefined} is explicitly passed → transform to value
- value is absent → defaultValue is silently ignored

Why This Is a Regression:
The new condition 'value' in props && props.value == null is fundamentally different from valueProp == null.
By removing the else branch, the refactoring lost the ability to handle the case where value is simply not provided. This was a critical fallback for IE11 compatibility.

ivanq2191-ui:
## Summary

This PR restores the fallback logic for `defaultValue` in `preact/compat`, fixing a regression introduced in PR #2742.

Closes #5123

## Problem

After the refactoring in #2742, `defaultValue` for `<textarea>` stopped working in Internet Explorer 11 when `value` prop is not explicitly passed.


JoviDeCroock:
Main does not support IE11, the v10.x branch does.

PR #5126, opened 2026-06-29
feat(hooks): add use hook

gaoachao:
## Summary
- add React 19-style `use()` support in `preact/hooks` for Promises and Contexts
- keep `use()` order-insensitive by avoiding `useState`/`useContext` internally
- expose `use` through compat default/types and add coverage for Promise, Context, conditional calls, falsy values, rejection, exports, and TS types

## Tests
- `npm test`

### DIFF
--- a/hooks/src/index.d.ts
+++ b/hooks/src/index.d.ts
@@ -123,6 +123,8 @@ export function useMemo<T>(factory: () => T, inputs: Inputs | undefined): T;
  */
 export function useContext<T>(context: PreactContext<T>): T;
 
+export function use<T>(resource: Promise<T> | PreactContext<T>): T;
+
 /**
  * Customize the displayed value in the devtools panel.
  *
--- a/hooks/src/index.d.ts
+++ b/hooks/src/index.d.ts
@@ -123,6 +123,15 @@ export function useMemo<T>(factory: () => T, inputs: Inputs | undefined): T;
  */
 export function useContext<T>(context: PreactContext<T>): T;
 
+/**
+ * Read the value of a resource like a Promise or a Context. Unlike other
+ * hooks, `use` may be called conditionally.
+ *
+ * When passed a pending Promise the component suspends until it settles,
+ * rethrowing the rejection reason if it rejects.
+ *
+ * @param resource The Promise or Context to read
+ */
 export function use<T>(resource: Promise<T> | PreactContext<T>): T;
 
 /**
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -165,6 +165,8 @@ function getHookState(index, type) {
 }
 
 /**
+ * Read the value of a Promise (suspending while pending) or a Context.
+ * Unlike other hooks, `use` may be called conditionally.
  * @template T
  * @param {Promise<T> | import('preact').PreactContext<T>} resource
  * @returns {T}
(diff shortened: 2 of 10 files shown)


gaoachao:
Hi @JoviDeCroock @rschristian , I reworked the `use()` implementation based on your feedback from #4970, especially around keeping it small, avoiding React-specific details in core, and not building it on top of order-sensitive hooks like `useState` or `useContext`.

The new PR implements `use()` for Promises and Contexts, with tests covering conditional calls, falsy promise values, rejection handling, shared promises, Context updates, exports, and TypeScript types. When you have time, could you take another look and let me know if this direction better matches Preact’s design philosophy?

hooks/src/index.d.ts
```diff
@@ -123,6 +123,8 @@ export function useMemo<T>(factory: () => T, inputs: Inputs | undefined): T;
  */
 export function useContext<T>(context: PreactContext<T>): T;
 
+export function use<T>(resource: Promise<T> | PreactContext<T>): T;
```
JoviDeCroock, line 135:
```suggestion
/**
 * Read the value of a resource like a Promise or a Context. Unlike other
 * hooks, `use` may be called conditionally.
 *
 * When passed a pending Promise the component suspends until it settles,
 * rethrowing the rejection reason if it rejects.
 *
 * @param resource The Promise or Context to read
 */
export function use<T>(resource: Promise<T> | PreactContext<T>): T;
```

hooks/src/index.js
```diff
@@ -164,6 +164,65 @@ function getHookState(index, type) {
...
+			reason => {
+				p.status = 'rejected';
+				p.reason = reason;
+			}
+		);
+	}
+
+	throw p;
+}
+
```
JoviDeCroock:
```suggestion
/**
 * Read the value of a Promise (suspending while pending) or a Context.
 * Unlike other hooks, `use` may be called conditionally.
 * @template T
 * @param {Promise<T> | import('preact').PreactContext<T>} resource
 * @returns {T}
 */
export function use(resource) {
	if (options._hook) {
		options._hook(currentComponent, currentIndex, 12);
	}

	if (typeof resource.then == 'function') {
		const thenable =
			/** @type {Promise<T> & { status?: string, value?: T, reason?: any }} */ (
				resource
			);
		if (thenable.status == 'fulfilled') return thenable.value;
		if (thenable.status == 'rejected') throw thenable.reason;
		if (thenable.status != 'pending') {
			thenable.status = 'pending';
			thenable.then(
				value => {
					thenable.status = 'fulfilled';
					thenable.value = value;
				},
				reason => {
					thenable.status = 'rejected';
					thenable.reason = reason;
				}
			);
		}
		throw thenable;
	}

	const context = /** @type {import('./internal').PreactContext} */ (resource);
	const provider = currentComponent.context[context._id];
	if (!provider) return context._defaultValue;
	provider.sub(currentComponent);
	return provider.props.value;
}
```

hooks/src/internal.d.ts
```diff
@@ -36,6 +36,7 @@ export interface ComponentHooks {
 export interface Component
 	extends Omit<PreactComponent<any, any>, '_renderCallbacks'> {
 	__hooks?: ComponentHooks;
+	_contexts?: Record<string, boolean>;
```
JoviDeCroock:
```suggestion
```

gaoachao:
@JoviDeCroock  Thanks for the suggestions! I applied all three changes:
- added the TSDoc for `use()`,
- inlined the Promise handling into `use()` to keep the implementation smaller,
- removed the extra `_contexts` field and simplified the Context path.

I also reran the relevant hooks tests and the full suite, and everything is passing now.

JoviDeCroock:
Merged in https://github.com/preactjs/preact/pull/5186

PR #5128, opened 2026-06-30
Migrate v10.x dependency management to pnpm

andrewiggins:
## Summary
- migrate the v10.x branch from npm package-lock files to a pnpm workspace and committed pnpm-lock.yaml
- update package scripts, Husky, docs, and CI workflows to install/cache/run/pack with pnpm
- keep npm publishing behavior intact while matching PR #5085's pnpm 11 and Node 24 toolchain policy

## Verification
- `pnpm --version` -> `11.1.0`
- `pnpm install --frozen-lockfile`
- `pnpm run lint`
- Used a temporary change to `src` to verify benchmmark and size action still passes

## Notes
- `pnpm run test:unit` was attempted, but this Codespaces host is missing Playwright browser system libraries (`libatk`, `libxcomposite`, `libgbm`, `libasound`, etc.), so the browser portion cannot launch here.

GPT 5.5 was used to generate some parts of this PR


.github/workflows/build-test.yml
```diff
@@ -26,17 +26,22 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           ref: ${{ inputs.ref || '' }}
+      - uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # v6.0.8
```
JoviDeCroock, line 29: You're skipping the release workflow just like the other PR, that would not be good since pnpm stage publish would be installed with npm ci

andrewiggins: Oh I didn't realize that pnpm supports OIDC and staged publishing - I assumed we had to use npm for those. I can switch them to use pnpm instead.

andrewiggins: Updated the release workflow to use pnpm now

JoviDeCroock:
Superseded by #5246.

PR #5132, opened 2026-06-30
fix: setState callback dropped when update is falsy

chatman-media:
setState(() => null, cb) never calls cb. The early-return for a falsy update happens before the callback ever gets pushed onto `_stateCallbacks`, and since that array only flushes during a render pass triggered by `enqueueRender`, the callback just never runs — not deferred, just gone.

This matters because bailing out of an update by returning null/undefined from the updater function is explicitly typed as valid (`index.d.ts` allows `Pick<S,K> | Partial<S> | null`), so it's a real pattern people rely on, not an edge case someone made up. `setState(null, cb)` has the same problem.

Fixed by only skipping the early return when there's a callback to flush — if there's truly nothing to apply and nothing to call, it still bails out the same as before, so the perf shortcut is intact. Added tests for both the updater-returns-null case and passing null directly.

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -51,6 +51,18 @@ BaseComponent.prototype.setState = function (update, callback) {
 	} else if (!callback) {
 		// nothing to apply and no callback to flush, so this is a true no-op
 		return;
+	} else if (this._vnode && !(this._bits & COMPONENT_DIRTY)) {
+		// The update itself was a no-op (e.g. an updater bailing out by
+		// returning null/undefined), and there's no render already queued to
+		// piggyback the callback on. This mirrors React: a no-op update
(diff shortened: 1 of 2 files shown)


JoviDeCroock:
👋 thanks for the PR!

We can't however merge this as-is as it would create a new gap for preact/compat. In react we don't enqueue a render _but_ the callback is called so while this solves half the problem it introduces a redundant render cycle.

chatman-media:
Good catch — pushed a fix. Now it only enqueues a render when one isn't already pending for the component; otherwise the callback just fires right away since there's nothing to commit. Added tests that assert no extra render happens for the lone no-op case, and that it still batches properly when there's a real update alongside it.

PR #5137, opened 2026-07-03
ci: add CVE Lite dependency audit workflow

sonukapoor:
This adds a CVE Lite CLI dependency audit workflow to preact. CVE Lite CLI is an OWASP Lab Project that reads your lockfile directly - no npm install needed - and cross-references every resolved package version against the OSV vulnerability database.

A scan of preact's current lockfile found 41 findings across 699 packages: 5 critical, 22 high, 13 medium, and 1 low. Of those, 4 have direct fixable counterparts with copy-run upgrade commands (vitest, playwright, undici, vite). The workflow gates on high severity so CI will catch new critical and high findings before they reach main.

The workflow runs on push and pull_request to main, plus a weekly Monday morning scheduled scan to catch newly published advisories against the existing lockfile. Results are uploaded to GitHub Code Scanning as SARIF so findings surface in the Security tab alongside any other static analysis the project runs.

All three actions are pinned to immutable commit SHAs rather than floating version tags, which removes the supply chain risk of a tag being moved.

More about CVE Lite CLI and the OWASP Lab Project at owasp.org/cve-lite-cli

### DIFF
--- a/.github/workflows/cve-lite.yml
+++ b/.github/workflows/cve-lite.yml
@@ -1,0 +1,36 @@
+name: CVE Lite dependency audit
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  schedule:
+    - cron: '0 6 * * 1'
+
+permissions:
+  contents: read
+  security-events: write
+
+jobs:
+  scan:
+    name: Scan dependencies
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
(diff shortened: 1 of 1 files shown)


JoviDeCroock:
Hey, this gives the exact same results as npm audit and doesn't seem worthwhile to me.

sonukapoor:
Thanks for the feedback. The key difference is in what you do after detection - `npm audit` surfaces issues but CVE Lite CLI validates that the suggested fix version actually resolves the advisory and gives you the exact command to run per finding (e.g. `npm install vite@8.0.16`), rather than a generic `npm audit fix` that can break peer dependencies. It also classifies each finding as direct vs transitive with the full dependency path, so you know which ones you own vs which require an upstream fix.

That said, totally fair if it doesn't add enough on top of what you already have. Appreciate the look.

PR #5138, opened 2026-07-03
fix: upgrade vulnerable direct dependencies

sonukapoor:
This upgrades three direct dev dependencies that have confirmed high-severity advisories in the OSV database.

A CVE Lite CLI scan (see PR #5137 for the accompanying audit workflow) found 41 total findings across preact's 699 resolved packages. Among the direct dependencies, playwright, undici, and vite each have high-severity advisories with clean upgrade paths.

playwright 1.54.1 to 1.55.1 - resolves a high-severity advisory. The semver range in package.json is updated from ^1.54.1 to ^1.55.1.

undici 4.12.0 to 6.27.0 - resolves multiple high-severity advisories. The semver range is updated from ^4.12.0 to ^6.27.0. undici 6.x is the current stable line.

vite 6.3.3 to 6.4.3 - resolves a high-severity advisory. The vite range stays ^6.2.0 and npm update resolves to 6.4.3 within range.

One additional direct finding (vitest 3.2.4, critical severity) requires upgrading the entire vitest ecosystem from v3 to v4, which also requires bumping @types/node from ^18 to ^20. That is a broader upgrade best handled separately by the preact team.


JoviDeCroock:
From what I can tell you are spamming repos with these commits and the way you are doing that is an infraction upon our CodeOfConduct.

sonukapoor:
Not spam - each repo was individually scanned and the fix commands were validated against the actual lockfile. The reason for contributing both PRs is that the fix PR demonstrates something npm audit doesn't give you: a concrete, validated upgrade command per finding, not just "a vulnerability exists."

Issue #4927, opened 2025-10-14
Support for useEffectEvent


bcomnes:
Any plans to support https://react.dev/reference/react/useEffectEvent in Preact?

marvinhagemeister:
From what I can tell in their docs the `useEffectEvent` hook is mostly used to bypass the hook dependency lint rule. It's basically a function that does nothing, like this:

```ts
function useEffectEvent<T extends (...args: any[]) => any>(fn: T): T {
  return fn
}
```

We might add it at some point to the compat package I'd assume. There is no timeline on that though.

davidje13:
@marvinhagemeister a key difference between `useEffectEvent` and a pass-through is that `useEffectEvent` [always "sees" the latest version of the function](https://react.dev/reference/react/useEffectEvent#reading-the-latest-props-and-state) (and therefore the latest variables), despite returning a stable function.

I believe it is more or less a rebrand of the proposed `useEvent` hook, with additional restrictions on when it should be used (presumably to avoid some of the edge cases which have been noted with `useEvent`). Lots of user-space approximations of `useEvent` exist, but it roughly looks like this:

```js
export const useEvent = (fn) => {
  const latest = useRef(fn);
  useLayoutEffect(() => { latest.current = fn; }, [fn]);
  return useCallback((...args) => latest.current(...args), []);
};
```

(a safer user-space variant uses `useState(() => (...args) => ...)` instead of `useCallback` to guarantee stability, which is required by `useEvent` but not guaranteed by `useCallback` in React)

dhardtke:
I would like to kindly ask if there is a timeline here. We are using https://github.com/mantinedev/mantine and while 8.x works, the recently released 9.x does not due to its use of useEffectEvent.

JoviDeCroock:
Implemented in #5141 - thank you @Mesoptier

bcomnes:
Thank you!

PR #5141, opened 2026-07-05
Add useEffectEvent hook to compat

Mesoptier:
I felt like doing some open source work, so figured I'd try and fix an issue (#4927) for a project I like.

I kept the implementation as minimal as possible, while still matching the main behaviour of the React hook (i.e. callback sees the latest values).

It can use some clean-up, but I figured I'd wait to see if you're actually even interested in adding this hook to compat.


compat/src/hooks.js
```diff
@@ -64,3 +64,15 @@ export function useTransition() {
 // TODO: in theory this should be done after a VNode is diffed as we want to insert
 // styles/... before it attaches
 export const useInsertionEffect = useLayoutEffect;
+
+export function useEffectEvent(cb) {
```
JoviDeCroock: I am not a big fan of this hook, especially because it's solely used to bypass a lint rule and has little semantic meaning. You could just vendor this type of hook and the linter would still conform to this behaviour.

compat/src/hooks.js
```diff
@@ -64,3 +64,15 @@ export function useTransition() {
...
+	const cbRef = useRef(cb);
+	cbRef.current = cb;
+
+	// React docs: "Effect Event functions do not have a stable identity. Their identity intentionally changes on every render."
+	return (...params) => {
+		// Avoid overriding cb's `this` by not calling as a method.
+		const cb = cbRef.current;
+		return cb(...params);
```
JoviDeCroock: From a review perspective this function is a bit wasteful as we optimize for byte-size, we should avoid caching/overriding like that and using rest params

compat/src/hooks.js
Mesoptier: I've personally only ever wanted to use this hook for practical purposes (e.g. using props in a global event handler, without having to remove/add the listener when props change), not for make-the-linter-happy purposes. But the workaround of just using `useRef` works fine too.

In any case, I take it you're not interested in adding this to compat, then? That's fine with me, of course, but maybe it would be good to close the corresponding issue as "not planned"?

compat/src/hooks.js
Mesoptier: I'm not exactly sure what you're referring to with "caching/overriding"?

But as for the rest params, I code-golfed it a bit and it does result in fewer bytes (locally anyway, dunno about gzip etc.) by using rest params:

```js
// using `arguments` keyword instead of rest params:
return function(){return(cb=cbRef.current)(arguments)}
// vs with rest params:
return(...p)=>(cb=cbRef.current)(...p)
// or if you're fine with exposing the ref through `this`:
return(...p)=>cbRef.current(...p)
```

(obviously the `cb` and `cbRef` should still be mangled, but the comparison stands)

rschristian: > But as for the rest params, I code-golfed it a bit and it does result in fewer bytes (locally anyway, dunno about gzip etc.) by using rest params:

Did you run it through our build tooling though? We compile to ES5, the rest params will not survive.

It's going to be much larger in byte cost than the input code suggests.

compat/src/hooks.js
```diff
@@ -64,3 +64,15 @@ export function useTransition() {
...
+	};
+}
```
JoviDeCroock:
```suggestion
**
 * @template {Function} T
 * @param {T} callback
 * @returns {T}
 */
export function useEffectEvent(callback) {
	const ref = useRef(callback);
	ref.current = callback;

	return /** @type {T} */ (
		function () {
			return ref.current.apply(undefined, arguments);
		}
	);
}
```

compat/src/hooks.js
```diff
@@ -64,3 +64,19 @@ export function useTransition() {
 // TODO: in theory this should be done after a VNode is diffed as we want to insert
 // styles/... before it attaches
 export const useInsertionEffect = useLayoutEffect;
+
+**
```
JoviDeCroock:
```suggestion
/**
```

JoviDeCroock (approved):
We've decided to put this hook into v11, thank you for your contribution!

JoviDeCroock:
| Filename | Size | Change |
| :--- | :---: | :---: |
| `compat/dist/compat.mjs` | 3.91 kB | +43 B (+1.11%) |

PR #5142, opened 2026-07-05
[Proposal] remove legacy context

JoviDeCroock:
React 19 removed it and it does look like it saves bytes/improves perf

Measured: -9 B gzipped on core (`dist/preact.mjs`). `createContext` and compat portals now publish context through an internal mangled `_childContext` property; the public `getChildContext` method is no longer called.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


JoviDeCroock:
The impact of this was a lot lower than I had hoped 😅

PR #5157, opened 2026-07-14
Anchor useId to the server boundary id during resumed hydration

JoviDeCroock:
## Summary

- derive the `useId` namespace of a resumed subtree from the boundary id embedded in the server-emitted suspense marker (`<!--$s:0-->`) instead of a client-side counter
- keep the stored start marker when a suspended-while-hydrating boundary re-renders before it resolves, so ancestor updates can no longer drift the boundary onto a nested boundary's marker
- boundaries whose marker carries no id, and client-only boundaries, keep the existing root-counter behavior and cannot collide with a server namespace (`PS<id>-N` vs `P0-N`)

## Problem

`useId` draws ids from a shared counter in render order. With resumed hydration, the content of a suspended boundary renders when its promise resolves, so sibling boundaries that resolve in a different order on the client than on the server consume the counter in a different order and derive ids that don't match the ids already sitting in the server HTML (and referenced by `aria-*` attributes).

We tried deriving a per-boundary namespace from an encounter-order counter in #5108, but had to revert it in #5135: a client-only boundary (`{isHydrated && <Suspense>…}`) rendering in the middle of hydration consumes a namespace the server never allocated and shifts every boundary after it. Any scheme that allocates namespaces by *when* a boundary renders has this failure mode; the namespace has to come from *what* the boundary is.

The streamed-hydration markers already carry exactly that: a stable, document-ordered boundary id assigned by the server. Reading it back from the DOM when the boundary resumes sidesteps re-deriving anything on the client — resolution order becomes irrelevant, and client-only boundaries simply have no marker.

While testing the client-only case this surfaced a second issue: when an ancestor update re-rendered a boundary that was still suspended, the resume path consumed the stored start marker and the re-suspension then stored the first marker found in the rebuilt (inner) range — a nested boundary's marker. The boundary would later resume from the wrong range with the wrong namespace. Re-suspensions now keep the marker they originally started from, and the stored marker is cleared once the boundary renders successfully.

## Server side

`renderToStringAsync` currently emits bare `<!--$s-->` markers; the streamed variant emits `<!--$s:N-->`. The follow-up in preact-render-to-string is to emit the boundary id in both modes and to hand out `PS<id>-N` ids inside a suspended subtree, mirroring the mask this PR restores on the client.

Size: +27 B brotli on core.

### DIFF
--- a/compat/src/internal.d.ts
+++ b/compat/src/internal.d.ts
@@ -44,5 +44,5 @@ export interface SuspenseComponent
 	_pendingSuspensionCount: number;
 	_suspenders: Component[];
 	_detachOnNextRender: null | VNode<any>;
-	_mask?: [number, number];
+	_mask?: [string | number, number];
 }
--- a/compat/test/browser/suspense-hydration.test.jsx
+++ b/compat/test/browser/suspense-hydration.test.jsx
@@ -5,6 +5,7 @@ import React, {
 	Fragment,
 	Suspense,
 	memo,
+	useId,
 	useState
 } from 'preact/compat';
 import { logCall, getLog, clearLog } from '../../../test/_util/logCall';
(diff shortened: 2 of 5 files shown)


JoviDeCroock:
Unsure whether we want to do this over just allowing the user to specify the Suspense name...

PR #5159, opened 2026-07-14
Introduce a modern build setup

JoviDeCroock:
rollup + babel for size equivalence

0c4fed063fb1cf6ea1b870e8efe26cb1b8a04b24 breaking change removes UMD/CJS


package.json
```diff
@@ -1,80 +1,52 @@
...
     ".": {
       "types": "./src/index.d.ts",
-      "module": "./dist/preact.mjs",
-      "umd": "./dist/preact.umd.js",
-      "import": "./dist/preact.mjs",
-      "require": "./dist/preact.js"
+      "default": "./dist/preact.mjs"
     },
     "./compat": {
       "types": "./compat/src/index.d.ts",
```
rschristian, line 20: This might be something to look at separately but wanted to make a comment of it as I remembered it: ESM-only breaks our types in compat pretty badly & fundamentally. This module will always need to be commonjs which then really can't be used to describe our ESM-only output.

Now that might be "fine" for most users (we're already fudging it a bit), and I have no idea how to work around it (spent a lot of time in #4998 and got nowhere good), but wanted to at least make it known.

JoviDeCroock: I ran `arethetypeswrong` and it looks to be okay as long as we don't add `type: "module"` here 😅 when `types/react` goes ESM I assume we can fix these export statements that aren't compatible with ESM

rschristian: We'd still have the "masquerading" issue though, right? I think that becomes slightly worse as it's not "Preact ships CJS & ESM, with ESM reusing the CJS types files" as we'd now be "Preact ships only ESM but still uses CJS types", if that makes sense.

Problem for a different day if it still generally works okay, we can fix that later w/out consequence I think.

JoviDeCroock: It all looked to be working when I was checking it out but I could be missing something 😅 it would indeed masquerade as TS would go for the `.ts` assuming it's  CJS while the runtime would pick `.mjs`

rschristian (approved):
Away from computer so I can't actually run the build, but LGTM!

Thanks for doing this! Long since needed to clear out Microbundle unfortunately. This should be a massive improvement.

PR #5162, opened 2026-07-16
Fix DocumentFragment missing createElementNS

### LINKED ISSUE #4539: Possible bug: phantom VDOM node erroring render when using React Aria `ListBoxItem`

**Describe the bug**
A React Aria [`ComboBox`](https://react-spectrum.adobe.com/react-aria/ComboBox.html) containing [`ListBoxItem`](https://react-spectrum.adobe.com/react-aria/ListBox.html) options fails to render with error `Cannot set property previousSibling of #<Node>`, render works normally for equivalent React application

**To Reproduce**

I created an issue demonstration repository [here](https://github.com/eoncarlyle/react-aira-issue-on-preact) which uses the [latest version](https://github.com/eoncarlyle/react-aira-issue-on-preact/blob/main/preact/package.json#L12) of Preact (10.24.3 at time of writing). The issue demonstration repository is the bare minimum to reproduce the issue, and compares the erroring Preact application with the normally functioning, equivalent React application.

To reproduce the error in Preact and compare against the working React application, run the following after cloning [the repository](https://github.com/eoncarlyle/react-aira-issue-on-preact/tree/main)

```shell
$ npm --prefix preact install
$ npm --prefix preact run dev
```
In another terminal:
```
$ npm --prefix react install
$ npm --prefix react run dev
```

Note that
```shell
$ diff preact/src/app.tsx react/src/App.tsx
1,2c1,2
< import { useState } from "preact/hooks";
< import preactLogo from "./assets/preact.svg";
---
> import { useState } from "react";
> import reactLogo from "./assets/react.svg";
4c4
< import "./app.css";
---
> import "./App.css";
17a18
>
```

Observe the following in the developer console in the Preact application:

```
Uncaught TypeError: Cannot set property previousSibling of #<Node> which has only a getter
    at $681cc3c98f569e39$export$b34a105447964f9f.appendChild (Document.ts:119:13)
    at Object.insertBefore
```

Observe the following output in the React application:

image

**Expected behavior**
The Preact application should produce the same output as the React application. My best guess of what is going on is outlined [here](https://iainschmitt.com/post/october-2024-gdple-development#possible-preact-bug), it may be the case that Preact is attempting to render a layer of the VDOM hierarchy that does not exist, but I'm not certain.

JoviDeCroock:
Fixes #4539

Falls back to `createElement` when `createElementNS` is unavailable. With passing ownerDocument to the portal we save part of the issue, however the faux-DOM passed from the library is missing createElementNS.

Do we want to land this? Hypothetically it's a compat concern

### DIFF
--- a/src/diff/index.js
+++ b/src/diff/index.js
@@ -529,7 +529,9 @@ function diffElementNodes(
 			return doc.createTextNode(newProps);
 		}
 
-		dom = doc.createElementNS(namespace, nodeType, newProps.is && newProps);
+		dom = doc.createElementNS
+			? doc.createElementNS(namespace, nodeType, newProps.is && newProps)
+			: doc.createElement(nodeType);
 
 		// we are creating a new node, so we can assume this is a new subtree (in
 		// case we are hydrating), this deopts the hydrate
--- a/test/browser/render.test.jsx
+++ b/test/browser/render.test.jsx
@@ -2093,5 +2093,24 @@ describe('render()', () => {
 			expect(rootCreateElementSpy).not.toBeCalled();
 			expect(iframeCreateElementSpy).toBeCalled();
 		});
+
+		it('falls back to createElement when createElementNS is unavailable', () => {
+			const iframe = document.createElement('iframe');
+			scratch.appendChild(iframe);
+
+			const iframeDocument = iframe.contentDocument;
+			const iframeCreateElementSpy = vi.spyOn(iframeDocument, 'createElement');
+			Object.defineProperty(iframeDocument, 'createElementNS', {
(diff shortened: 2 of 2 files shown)


rschristian:
> Do we want to land this? Hypothetically it's a compat concern

They're passing in a faux DOM node? IMO, we shouldn't fallback here. I get that it's a compat issue but it should be on the caller to supply a sufficiently robust fake, not on the library to fallback when that fails.

Any chance we/they/some bot can/are willing to fix this downstream instead?

JoviDeCroock:
I'll close out the source issue with that comment, you are right

PR #5163, opened 2026-07-16
Fix effect-flushing for radix and animations

### LINKED ISSUE #3666: [preact/compat] radix-ui Slider component drag points (Thumbs) not working

**Describe the bug**
Thumb (drag points) are not receiving any updates past the initial css applied.

**To Reproduce**

https://stackblitz.com/edit/nextjs-preact-webpack5-pjoxwh

Steps to reproduce the behaviour:

Load the stackblitz example above and drag the edges of the white lines:

Screen Shot 2022-08-08 at 12 43 56 pm

While the line itself is adjustable, the `Thumb` components are not shown. If the nested span's css is modified from `display: none` to `display: block` the `Thumb` element is rendered on left edge with 0, 0 coords.

Screen Shot 2022-08-08 at 12 49 37 pm

**Expected behavior**

For slider thumbs to be attached to end of white line and to move along with them.

Screen Shot 2022-08-08 at 12 54 02 pm

** Additional content **

`Slider` component used to hang the app as did the Dialog in this bug report https://github.com/preactjs/preact/issues/3297. Since version `10.10.1` and this https://github.com/preactjs/preact/pull/3645 fix specifically the `Slider` component now no longer hangs the app but it's not 100% working either with `Thumb` component being broken. 

Same Issue reported in radix-ui repo https://github.com/radix-ui/primitives/issues/1601

JoviDeCroock:
Fixes #3666.

Flushes pending passive effects before scheduled renders. The main difference is that instead of going component by component we flush a commit phase before the next render starts so that component trees are consistent. Technically though I'm not sure whether this is the right approach due to our override of `options.debounceRendering` and us also flushing if the next render is for an entirely different subt-tree so I am leaning on the side of skipping this.

Fixes the Radix Slider reproduction and the Base UI + Motion exit-animation reproduction from #5055.

### DIFF
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -26,6 +26,14 @@ let oldAfterDiff = options.diffed;
 let oldCommit = options._commit;
 let oldBeforeUnmount = options.unmount;
 let oldRoot = options._root;
+let oldDebounce = options.debounceRendering || queueMicrotask;
+
+options.debounceRendering = callback => {
+	oldDebounce(() => {
+		flushAfterPaintEffects();
+		callback();
+	});
+};
 
 // We take the minimum timeout for requestAnimationFrame to ensure that
 // the callback is invoked after the next frame. 35ms is based on a 30hz
--- a/hooks/src/index.js
+++ b/hooks/src/index.js
@@ -48,8 +48,12 @@ options._diff = vnode => {
 };
 
 options._root = (vnode, parentDom) => {
-	if (vnode && parentDom._children && parentDom._children._mask) {
-		vnode._mask = parentDom._children._mask;
+	if (vnode) {
+		flushAfterPaintEffects();
+
+		if (parentDom._children && parentDom._children._mask) {
+			vnode._mask = parentDom._children._mask;
+		}
 	}
 
 	if (oldRoot) oldRoot(vnode, parentDom);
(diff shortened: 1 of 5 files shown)


hooks/src/index.js
```diff
@@ -66,11 +78,6 @@ options._render = vnode => {
 				}
 				hookItem._pendingArgs = hookItem._nextValue = undefined;
 			});
-		} else {
```
JoviDeCroock, line 61: Now because we flush when we start a new render cycle this isn't needed anymore. It was basically a hack for what we're adding. For predictability though I would love to also add this into signals, that would however require a major and it's already behaving like this currently so... it's not pressing for v11 itself

marvinhagemeister (approved):

hooks/src/index.js
```diff
@@ -27,6 +27,14 @@ let oldAfterDiff = options.diffed;
 let oldCommit = options._commit;
 let oldBeforeUnmount = options.unmount;
 let oldRoot = options._root;
+let oldDebounce = options.debounceRendering || queueMicrotask;
+
+options.debounceRendering = callback => {
```
JoviDeCroock, line 31: My main worry with this is that it should be a  breaking change as folks might be overriding this in their components as well when they leverage a custom scheduler

JoviDeCroock:
Closing this — the approach fixes both repros (Radix Slider #3666, Base UI + Motion exit animations from #5055), but wrapping `options.debounceRendering` is too fragile to ship: core reads the option live, so any consumer assigning this documented public option after `preact/hooks` loads silently discards the flush, and the behavior depends on module import order. Our own test-utils needed patching twice in this diff, which is a good indicator third-party schedulers/harnesses would break silently. Flushing the whole effect queue before any scheduled render also effectively turns after-paint effects into pre-paint effects in busy apps.

If we revisit this, a private core hook invoked before draining the rerender queue (rather than piggybacking on the public option) would keep the semantics without the override fragility.

krskrs:
I hope this will be revisited in the future, Preact is a great framework, but the React ecosystem is huge and compatibility is vital for Preact itself. Now this is radix-ui related, but perhaps it could resolve other compatibility issues with other libraries as well.

PR #5170, opened 2026-07-18
fix: don't treat non-function on* props on custom elements as event handlers

### LINKED ISSUE #4085: Uncaught Error: Component's "only-when-overflow" property should be a function, but got [boolean] instead

JoviDeCroock:
Custom elements may expose non-event attributes starting with "on" (e.g. only-when-overflow). Only take the event-listener path when the new or old value is a function, or the element is not a custom element (no dash in its tag name).

Fixes #4085

Co-Authored with GPT5.6-sol for tests

### DIFF
--- a/src/diff/props.js
+++ b/src/diff/props.js
@@ -70,7 +70,17 @@ export function setProperty(dom, name, value, oldValue, namespace) {
 		}
 	}
 	// Benchmark for comparison: https://esbench.com/bench/574c954bdb965b9a00965ac6
-	else if (name[0] == 'o' && name[1] == 'n') {
+	// Custom elements may have non-event properties/attributes starting with
+	// `on` (https://github.com/preactjs/preact/issues/4085), so for tag names
+	// containing a `-` only treat `on*` props as event handlers when the new
+	// or old value is a function.
+	else if (
+		name[0] == 'o' &&
+		name[1] == 'n' &&
+		(typeof value == 'function' ||
+			typeof oldValue == 'function' ||
+			dom.localName.indexOf('-') == -1)
+	) {
 		useCapture = name != (name = name.replace(CAPTURE_REGEX, '$1'));
 
 		// Infer correct casing for DOM built-in events:
--- a/test/browser/events.test.jsx
+++ b/test/browser/events.test.jsx
@@ -276,6 +276,32 @@ describe('event handling', () => {
 		);
 	});
 
+	// see https://github.com/preactjs/preact/issues/4085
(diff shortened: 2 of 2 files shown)


src/diff/props.js
```diff
@@ -70,7 +70,17 @@ export function setProperty(dom, name, value, oldValue, namespace) {
...
 	// Benchmark for comparison: https://esbench.com/bench/574c954bdb965b9a00965ac6
-	else if (name[0] == 'o' && name[1] == 'n') {
+	// Custom elements may have non-event properties/attributes starting with
+	// `on` (https://github.com/preactjs/preact/issues/4085), so for tag names
+	// containing a `-` only treat `on*` props as event handlers when the new
+	// or old value is a function.
+	else if (
+		name[0] == 'o' &&
+		name[1] == 'n' &&
+		(typeof value == 'function' ||
```
JoviDeCroock, line 80: short-circuit so there shouldn't be an extra cost for normal execution)

rschristian:
> Custom elements may expose non-event attributes starting with "on" (e.g. only-when-overflow)

IMO, we shouldn't support this.

`on...` is reserved for event handlers only in the DOM, the spec group will not allow other properties to collide. This is a very explicit rule at this point. CE authors are free to do as they please, but I don't think we should incur cost when they choose to deviate from platform standards... and we already have some restrictions/limitations around CEs in that we expect them to act like standard DOM elements, i.e., we prefer properties over attributes and so if you have a setter property that you don't want us to call, tough luck.

JoviDeCroock:
I mean, onlyOverflow for instance is a valid name for a prop and it runs into this issue. I think it's a non obvious footgun. I guess it has propagated in the ecosystem so we can live with it, tried to see if we had some easy fixes

rschristian:
> `onlyOverflow` for instance is a valid name for a prop

Not on any spec'd DOM element unless it's an event handler, is what I was trying to say earlier. CEs can do as they please though.

I fully agree it's non-obvious but spending bytes to support using property names that should not be used is kinda iffy

PR #5188, opened 2026-07-31
Propose removing useErrorBoundary

JoviDeCroock:
Usage of this hook is quite low and class components will remain part of core so it's just syntax sugar which is against our philosophy


rschristian:
Just to echo myself from Slack so it doesn't get lost: it's syntactic sugar, sure, but sugar that normal users aren't going to ever write or figure out, sugar that users have shown some appreciation towards, and it's only 65b at the end of the day. I don't think anyone will notice the bytes but they will notice the removal.

---

Maybe I'm the odd one out though! If people generally think the bytes outweigh the API, by all means. I'm not blocking this / I don't feel _that_ strongly.

PR #5197, opened 2026-08-06
Reduce bundle size before release candidates go live

JoviDeCroock:
Ran an autoresearch loop with Claude Opus 5 and came to the following result.

Currently looking through perf regressions

### DIFF
--- a/jsx-runtime/src/utils.js
+++ b/jsx-runtime/src/utils.js
@@ -3,34 +3,7 @@ const ENCODED_ENTITIES = /["&<]/;
 /** @param {string} str */
 export function encodeEntities(str) {
 	// Skip all work for strings with no entities needing encoding:
-	if (!str.length || !ENCODED_ENTITIES.test(str)) return str;
-
-	let last = 0,
-		i = 0,
-		out = '',
-		ch = '';
-
-	// Seek forward in str until the next entity char:
-	for (; i < str.length; i++) {
-		switch (str.charCodeAt(i)) {
-			case 34:
-				ch = '&quot;';
-				break;
-			case 38:
-				ch = '&amp;';
-				break;
-			case 60:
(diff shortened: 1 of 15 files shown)


jsx-runtime/src/utils.js
```diff
@@ -3,34 +3,7 @@ const ENCODED_ENTITIES = /["&<]/;
...
-		// Append skipped/buffered characters and the encoded entity:
-		if (i !== last) out += str.slice(last, i);
-		out += ch;
-		// Start the next seek/buffer after the entity's offset:
-		last = i + 1;
-	}
-	if (i !== last) out += str.slice(last, i);
-	return out;
+	return ENCODED_ENTITIES.test(str)
+		? str.replace(/&/g, '&amp;').replace(/"/g, '&quot;').replace(/</g, '&lt;')
```
rschristian: This doesn't look to be functionally equivalent? The old method iterated over the string, potentially replacing multiple times I think? `.replace` is a one-and-done method.

Doing this from my phone but I'm struggling to believe we really had something sitting around that could be replaced by a few `.replace` calls.

JoviDeCroock: This is mainly code used for the jsxTemplate Deno optimization, I still have to validate this one but this is doing `//g` so it should replace all occurrences and is not dissimilar to how we used to do it in RTS.

I hope that this works because this can't be tree-shaken and is paid by all our users atm

rschristian: I did miss `/g`, my bad.

Still, seems.. odd.

JoviDeCroock: Huh, Vite started tree-shaking this nvm what I just said

marvinhagemeister (approved):

PR #5207, opened 2026-08-23
Add missing lowercased html attributes to dom.d.ts

teddybradford:
There were a few missing lowercased versions of html content attributes, so I added them to dom.d.ts


rschristian (approved):
Looks right to me, thanks!

Keep in mind that `main` is the upcoming v11 release, `v10.x` branch is the one you want for the v10 release line.

PR #5208, opened 2026-08-23
Sort lowercase attributes before camelCase ones in dom.d.ts

teddybradford:
This change sorts all the lowercase attributes before camelCase ones. Autocomplete in my code editor was picking the camelCase attributes first for some, so this PR normalizes the sorting across all attributes in dom.d.ts

### DIFF
--- a/src/dom.d.ts
+++ b/src/dom.d.ts
@@ -81,25 +81,10 @@ export interface CSSProperties extends AllCSSProperties, DOMCSSProperties {
 export interface SVGAttributes<
 	Target extends EventTarget = SVGElement
 > extends HTMLAttributes<Target> {
+	'accent-height'?: Signalish<number | string | undefined>;
 	accentHeight?: Signalish<number | string | undefined>;
 	accumulate?: Signalish<'none' | 'sum' | undefined>;
 	additive?: Signalish<'replace' | 'sum' | undefined>;
-	alignmentBaseline?: Signalish<
-		| 'auto'
-		| 'baseline'
-		| 'before-edge'
-		| 'text-before-edge'
-		| 'middle'
-		| 'central'
-		| 'after-edge'
-		| 'text-after-edge'
-		| 'ideographic'
-		| 'alphabetic'
-		| 'hanging'
-		| 'mathematical'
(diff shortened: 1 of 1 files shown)


rschristian:
> Autocomplete in my code editor was picking the camelCase attributes first for some

If this is the case, it should remain so. We prioritize properties over attributes.

teddybradford:
> > Autocomplete in my code editor was picking the camelCase attributes first for some
> 
> If this is the case, it should remain so. We prioritize properties over attributes.

Ah. It seemed that dom.d.ts had more instances of lowercase attributes being sorted first, so the sorting strategy seemed inconsistent.

rschristian:
> It seemed that dom.d.ts had more instances of lowercase attributes being sorted first

That wasn't done with any intention, that's just alphabetic order. Also made sense as some of the camel-cased "props" are non-standard Reactisms that were appended later.

Personally, my editor alphabetizes suggestions itself, apparently source order plays no part in how they're presented. If some editor is prioritizing them in source order though then we probably want to reconsider our setup. 

Edit: Code in question:

1. Check if prop exists on dom element (which is case-sensitive):

https://github.com/preactjs/preact/blob/f53c217b8ae5bb203e3c438b3026a2d796c6ce37/src/diff/props.js#L118

2. If not, fallback to setting the attribute:

https://github.com/preactjs/preact/blob/f53c217b8ae5bb203e3c438b3026a2d796c6ce37/src/diff/props.js#L137

rschristian (approved):
I have looked at 5 different editors at this point and they all present auto complete options in slightly different orders :shrug:

Happy to merge this as it's in better order, even if your editor then presents the options in a less-than-ideal order for autocomplete.

Needs a rebase but then we can merge.

teddybradford:
> I have looked at 5 different editors at this point and they all present auto complete options in slightly different orders 🤷
> 
> Happy to merge this as it's in better order, even if your editor then presents the options in a less-than-ideal order for autocomplete.
> 
> Needs a rebase but then we can merge.

Okay. I resolved the conflict.

rschristian:
Thanks!

PR #5221, opened 2026-08-29
perf(compat): avoid deleting class ref props

JoviDeCroock:
Compat now copies referenced class-component props without `ref` instead of deleting the property, avoiding V8 dictionary-mode objects.

This makes the focused path 32% faster to create and 27% faster to update while retaining 139 fewer bytes per component in Chromium.


JoviDeCroock:
Size impact is too large for a feature that only applies to class-components now

PR #5222, opened 2026-08-29
perf: reuse empty component state and queues

JoviDeCroock:
Function components now reuse shared empty state and callback sentinels, passive-effect queues allocate only when needed, and empty class callback queues persist across commits.

This reduces retained heap by 16–32 bytes per function component and improves class updates by 8% in Chromium.


JoviDeCroock:
The byte size is too heavy for the perf benefit being small

PR #5223, opened 2026-08-29
perf(hooks): share state bailout function

JoviDeCroock:
Stateful function components now share the hooks `shouldComponentUpdate` implementation while retaining any prior component bailout.

This removes one closure per component, reducing retained heap by 36 bytes and improving mounts by 5–8% and updates by 0.7–0.8% in Chromium.


JoviDeCroock:
Hesitant whether this is worth it mainly because 12 bytes for a mount-only win

PR #5239, opened 2026-09-04
Preserve hook state when Suspense re-suspends

JoviDeCroock:
Preserves committed hook state when a Suspense boundary re-suspends while still resetting hooks for an initial suspended mount. Effects are cleaned up and re-run when content is revealed again, with client-rendering and hydration coverage.

### DIFF
--- a/mangle.json
+++ b/mangle.json
@@ -32,6 +32,7 @@
       "$_hydrationMismatch": "__m",
       "$_list": "__",
       "$_pendingEffects": "__h",
+      "$_passive": "__P",
       "$_value": "__",
       "$_nextValue": "__N",
       "$_original": "__v",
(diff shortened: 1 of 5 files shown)


mangle.json
```diff
@@ -32,6 +32,7 @@
       "$_hydrationMismatch": "__m",
       "$_list": "__",
       "$_pendingEffects": "__h",
+      "$_passive": "__P",
```
marvinhagemeister, line 35: Q: Does this clash with `_parentDom`?

marvinhagemeister (approved):

mangle.json
JoviDeCroock: _passive is set on hooks and _parentDom is set on the vnode/component so it should not clash

PR #5246, opened 2026-09-06
Migrate dependency management to a root and demo pnpm workspace

JoviDeCroock:
Supersedes #5085 and #5128.

Migrate the root and demo to pnpm with a shared lockfile, frozen CI installs, supply-chain checks, and pnpm packing and staged publishing. Keep entry-point manifests out of the workspace and retain the independent benchmark workspace.


pnpm-workspace.yaml
```diff
@@ -0,0 +1,31 @@
+# Only independently installed projects belong here. Preact's entry-point
+# manifests are build/publish metadata, and the root is included automatically.
+packages:
+  - demo
+
+minimumReleaseAge: 1440
```
rschristian: This all feels like stuff that's the responsibility of the user to set if they want it, not the project. Nannying users who choose to clone & install the repo feels rather out of scope & not our responsibility.

Our deps rarely get touched anyhow.

marvinhagemeister (approved):

pnpm-workspace.yaml
```diff
@@ -0,0 +1,31 @@
...
+  # Babel requires semver 6. This exact artifact (including SHA-512 integrity)
+  # already exists in main's npm lockfile; 6.3.1 lacks earlier releases' provenance.
+  - semver@6.3.1
+  # @types/node 22 requires this release; its SHA-512 also matches main's
+  # npm lockfile. The release lacks earlier-published versions' provenance.
+  - undici-types@6.21.0
+strictDepBuilds: true
+
+# Check dependency engines against the Node version used by contributors and CI.
+nodeVersion: 22.22.2
```
marvinhagemeister, line 24: Nit: wonder if we should update to 24 now that it's LTS

PR #5254, opened 2026-09-09
Fix compressed size comparison with pnpm

JoviDeCroock:
Since the pnpm migration, an install with unchanged dependencies can skip `prepare`. The compressed-size action therefore reused the PR's bundles for the base revision and reported no size changes. Run `build` explicitly for both revisions.

Keep Husky and `test:install` in `prepare`, and run builds explicitly in CI.

### DIFF
--- a/.github/workflows/size.yml
+++ b/.github/workflows/size.yml
@@ -19,6 +19,6 @@ jobs:
       - uses: preactjs/compressed-size-action@66325aad6443cb7cf89c4bfcd414aea2367cda94 # 2.9.1
         with:
           repo-token: '${{ secrets.GITHUB_TOKEN }}'
-          # Our `prepare` script already builds the app post-install,
-          # building it again would be redundant
-          build-script: '--if-present noop'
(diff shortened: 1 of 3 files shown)


marvinhagemeister (approved):

.github/workflows/size.yml
```diff
@@ -19,6 +19,6 @@ jobs:
       - uses: preactjs/compressed-size-action@66325aad6443cb7cf89c4bfcd414aea2367cda94 # 2.9.1
         with:
           repo-token: '${{ secrets.GITHUB_TOKEN }}'
-          # Our `prepare` script already builds the app post-install,
-          # building it again would be redundant
-          build-script: '--if-present noop'
+          # pnpm can skip prepare when dependencies are unchanged between refs.
```
rschristian, line 22: This comment is entirely redundant and a bit nonsensical now that `prepare` isn't building

rschristian (commented):
Down with `prepare`!

